Issue #347, opened 2022-09-28
Android Automotive: Change the tab order


geekygecko:
When the user has subscribed to podcasts move the "Podcasts" tab to the first position otherwise use the "Discover" tab in this position. 

Taking the user to an empty podcast screen doesn't look good. But when a user is subscribed to podcasts they don't want to see the discover section every time they open the app. 

![Screenshot_20220928_210209](https://user-images.githubusercontent.com/308331/192768412-0c49c0f1-9fb2-4590-905c-a0d0ed643862.png)

PR #361, opened 2022-10-04
Improve Automotive OS tab order

geekygecko:
This change switches the Automotive OS opening tab between "Podcasts" and "Discover". If the user doesn't have any podcasts it starts with the "Discover" so they aren't presented with an empty screen. Otherwise, the user is shown their podcast collection as the first tab.

Fixes https://github.com/Automattic/pocket-casts-android/issues/347

**Test steps**
1. Open the Automotive app

The "Discover" tab should open first.

2. Tap on a podcast
3. Play an episode
4. Leave the app, open the Settings app
5. Open the Apps section
6. If the Pocket Casts app and Force stop the app
7. Open the Bluetooth Audio app
8. Open the Pocket Casts app

The "Podcasts" tab should open first.

Screenshot_20221004_143345
Screenshot_20221004_143306


ashiagr (approved):
LGTM 👍

:shipit:

PR #364, opened 2022-10-04
Enable creating local signed release builds

mchowning:
## Description

The ability to create a locally signed release build no longer works for me after the buildkite updates. This restores the previous behavior where the user can define the relevant information locally in a gradle.properties file (as described in the [README](https://github.com/Automattic/pocket-casts-android/blob/main/README.md#signing-a-release)).

## To Test

1. Set up a signing key and add the relevant properties to one of the gradle.properties files that this project will load
2. Create a release build and ensure that the app runs

# Checklist

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -54,9 +54,21 @@ project.ext {
 
     // App Signing
     storeFile = file("$rootDir/${secretProperties.getProperty("signingKeyStoreFile", null)}")
-    storePassword = secretProperties.getProperty("signingKeyStorePassword", null)
-    keyAlias = secretProperties.getProperty("signingKeyAlias", null)
-    keyPassword = secretProperties.getProperty("signingKeyPassword", null)
+    if (storeFile.exists()) {
+        // Use secret properties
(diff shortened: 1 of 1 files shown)


mchowning:
I also need to have the `sentry.properties` file locally to create a local release build, but it sounds like @geekygecko may not based on [this comment](https://github.com/Automattic/pocket-casts-android/pull/336#discussion_r983153793). 

If that's true and we don't need a `sentry.properties` file to create a local release build, then this PR is probably fine as-is, and I just need to figure out what's going on with my local builds. 

If we cannot create a local release build without the `sentry.properties` file, we should probably update this PR to address that so it is easier for open source contributors to create a release build if they want.

<details>
<summary>Build error I get when I run `:app:assembleRelease` task</summary>
Task :app:uploadSentryProguardMappingsRelease FAILED
compressing mappings
uploading mappings
error: An organization slug is required (provide with --org)
</details>

geekygecko (approved):
This looks good. I have a fix for skipping the Sentry mapping file upload that I will raise in a separate pull request.

PR #368, opened 2022-10-05
Android Grade Plugin upgrade

geekygecko:
This change upgrades the Gradle files for the latest Android Gradle Plugin 7.3.0. I didn't realise this can be done in Android Studio from the menu option 'Tools' -> 'AGP Upgrade Assistant...'.

Running `./gradlew :app:assembleRelease` output the following error.

```
package="au.com.shiftyjelly.pocketcasts.discover" found in source AndroidManifest.xml.
Setting the namespace via a source AndroidManifest.xml's package attribute is deprecated.
Please instead set the namespace (or testNamespace) in the module's build.gradle file, as described here: 
https://developer.android.com/studio/build/configure-app-module#set-namespace
This migration can be done automatically using the AGP Upgrade Assistant, please refer to 
https://developer.android.com/studio/build/agp-upgrade-assistant for more information.
```

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -9,6 +9,8 @@ apply from: "../base.gradle"
 apply plugin: 'com.google.android.gms.oss-licenses-plugin'
 
 android {
+    namespace 'au.com.shiftyjelly.pocketcasts'
+
     defaultConfig {
         applicationId project.applicationId
         multiDexEnabled true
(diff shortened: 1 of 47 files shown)


mchowning (approved):
Changes look good and everything seems to be working in my testing. Thanks for updating this!

Issue #143, opened 2022-07-20
Unable to Select Text in Show Notes From Episode Details View


mchowning:
Reported by @mdrockwell 

Steps to reproduce:

1. Navigate to an individual episode's show notes (not the currently playing show). Any episode in the Listening History or the Discover tab, for example.
2. Attempt to select text from the show notes.

This can be done from Now Playing's Notes view, but not from other episode details views on Android. Although, it can be done all of these locations from iOS.

karacca:
I see that the WebView's `LongClickListener` has been turned of by @geekygecko on `EpisodeFragment:494`. When I removed these lines It works as you expected, but I don't completely understand why It turned off it the first place. 

@geekygecko If you recall the problem that made you to turned off this setting, maybe we can focus on solving that 🤔 

```kotlin
setOnLongClickListener { true }
isLongClickable = false
isHapticFeedbackEnabled = false
```

geekygecko:
@karacca, I wonder if we had issues with long show notes and scrolling the page. But I have just tried this, and it doesn't affect it. Thanks so much for taking a look into this. I have raised a pull request https://github.com/Automattic/pocket-casts-android/pull/372.

PR #372, opened 2022-10-05
Allow select text in show notes from episode details view.

geekygecko:
This change allows the user to select text from the show notes on the full-screen episode details view.

Thanks to @karacca for looking into this issue.

Fixes https://github.com/Automattic/pocket-casts-android/issues/143

**Test steps**
1. Tap on the "Podcasts" tab
2. Tap on a podcast
3. Tap on an episode
4. Long press on the show notes to select text

You should be allowed to copy text from the show notes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,10 @@
 -----
 
 *   Bug Fixes:
-    *   Allow Discover feed collection titles to wrap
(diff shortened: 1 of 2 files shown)


ashiagr (approved):
Nice fix, @geekygecko, @karacca 🙌 

:shipit:

PR #375, opened 2022-10-06
Fix crash when swiping to open Up Next queue

### LINKED ISSUE #370: Crash when swiping Up to open Up Next
### Description

We have had several reports that swiping up to open the Up Next queue is crashing the app in the 7.24-rc-1 release candidate.

I can intermittently reproduce this locally using my Pixel 3 running Android 12.

The stacktrace I get when I reproduce this locally matches [this Sentry report](https://sentry.io/share/issue/e176d50efd9a475da30ff9642e6dcdc2/). Interestingly though, that crash report is 100% from Android 13 users, but my test device sees the issue on Android 12.

1. Open app
2. Open full screen player
3. Swipe up to open the Up Next queue
4. 💥 Crash (this crash does not occur consistently for me though. It may happen more frequently when I swipe quickly, but I'm not sure about that)

https://user-images.githubusercontent.com/4656348/194063928-f91e56b7-cfef-439b-abcf-59954e86fd31.mp4

Sentry: 100% Android 13, 7.24-rc-1
My test device: Pixel 3 running Android 12

mchowning:
## Description

Fixes #370 by adding a top level view to capture the down events that for some reason don't seem to always make it down to the `NestedScrollView` (which results in a crash). 

_h/t to @geekygecko for coming up with this solution. 🙇_

> **Warning**
> I am targeting the `release/7.24` branch because I think we should do a betafix for this. It is a regression in the 7.24 release (probably because of [bumping our target SDK to 33](https://github.com/Automattic/pocket-casts-android/pull/312)) and has a pretty high number of occurrences in Sentry considering it only just started with the 7.24-rc-1 release we put out earlier this week. Let me know if you disagree (or if you think that targeting the release branch like this isn't the right way to proceed with a beta release).

## Test steps

This crash is very hard to reproduce, and I don't have any clear test steps to cause the crash. For some reason I seem to be able to get it pretty consistently on my Pixel 3 test device running Android 12, but it seems like all the crashes in Sentry are with Android 13. 🤷  TLDR: You may not be able to reproduce this crash locally.

1. Open app
2. Open full screen player
3. Swipe up to open the Up Next queue
4. ✅  Verfiy there is no crash

# Checklist

### DIFF
--- a/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/PlayerHeaderFragment.kt
+++ b/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/PlayerHeaderFragment.kt
@@ -61,7 +61,7 @@ import au.com.shiftyjelly.pocketcasts.localization.R as LR
 
 private const val UPNEXT_DRAG_DISTANCE_MULTIPLIER = 1.85f // Open up next at a different rate than we are dragging
 private const val UPNEXT_HEIGHT_OPEN_THRESHOLD = 0.15f // We only have an open threshold because we only control swipe up, swipe down is the standard bottom sheet behaviour
-private const val UPNEXT_OUTLIER_THRESHOLD = 150.0f // Sometimes we get a random large delta, it seems better to filter them out or else you get random jumps
+private const val UPNEXT_OUTLIER_THRESHOLD = 400.0f // Sometimes we get a random large delta, it seems better to filter them out or else you get random jumps
 
 @AndroidEntryPoint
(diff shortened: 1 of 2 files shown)


ashiagr (approved):
I was able to reproduce the issue on `Android 12` and `Android 13` for `v7.24`. 

Looks like it is `API33` bug and reproducible only on `7.24` after we targetted the app to `Android 13`:
https://issuetracker.google.com/issues/248212163
https://issuetracker.google.com/issues/206855618

I tested the solution, it is working well for me. Hopefully, we'll have an official fix soon when we can remove this workaround.

This indeed is a beta fix candidate. 👍

PR #382, opened 2022-10-07
Add BasePreferenceFragment and Implement in PodcastSettingsFragment

### LINKED ISSUE #73: Podcast Settings: Switches animate every time you open the page.
On the podcast settings page, any switch that is toggled on animates when you open it.

https://user-images.githubusercontent.com/308331/177689986-7f74638f-05d3-413c-957c-89455ef52cc2.mov

karacca:
# Description

Add `BasePreferenceFragment` and Implemented in `PodcastSettingsFragment` to handle loading state changes. I created a base Fragment to handle this logic because I saw other problems that implements Preferences like this one.

Also when I disabled the RecyclerView animation that handles ListPreference's childeren animations, It still has a small flicker. (not the switches only the child items) Which implementation is better in your opinion @geekygecko 

https://user-images.githubusercontent.com/47274419/194526328-74bbb803-426b-4179-b9ec-67a68f941ea9.mp4 

https://user-images.githubusercontent.com/47274419/194526340-1645ef59-81b5-42b0-9b54-e17e09f91d6f.mp4

Fixes #73 

# Checklist


geekygecko:
> Which implementation is better in your opinion

They both look so much better than before! Leaving the animation on for the child list might be cleaner in the code, so maybe that's a better option.

Thanks for looking into this issue.

geekygecko (approved):

Issue #261, opened 2022-08-23
Project Issue: Tracks Implementation


mchowning:
## Description
Implement analytics using Tracks in the app. See https://github.com/Automattic/pocket-casts-ios/issues/154 for the iOS work on this.

### Setup

- [x] Create a base implementation of an analytics tracker in the app @ashiagr (https://github.com/Automattic/pocket-casts-android/pull/265)
- [x] Create a way to support multiple types of analytics trackers @ashiagr (https://github.com/Automattic/pocket-casts-android/pull/265)
- [x] ~Add switching between anonymous and authenticated user tracking when the user signs in/out~ Adding logic to regenerate anonymous IDs to protect our users privacy (see p4H3ND-1i9-p2) (https://github.com/Automattic/pocket-casts-android/pull/268)
- ~Add analytics checkbox to the PR template~ (see [comment](https://github.com/Automattic/pocket-casts-android/issues/261#issuecomment-1228626746))
- [x] Disable Tracks in release builds until we're ready to release (https://github.com/Automattic/pocket-casts-android/pull/291)
- [x] Turn Tracks on for release builds and off for debug builds once this is ready to be released
- [ ] Print event logs using LogBuffer so all the tracks events show up in the user's logs if they contact support.
- [ ] Enable Tracks for Android Automotive builds

#### Predefined Events @ashiagr (https://github.com/Automattic/pocket-casts-android/pull/273)
- [x] `plus_has_subscription`: Whether the user has a plus subscription
- [x] `plus_has_lifetime`: if they're a lifetime plus member
- [x] `plus_subscription_type`: Their subscription type: none, plus, or supporter
- [x] `plus_subscription_platform`: The platform they subscribed to plus on: none, ios, android, web, gift
- [x] `plus_subscription_frequency`:  Their renewal frequency: none, monthly, yearly

`is_rtl_language` is not required to be added as it is already added by the `TracksClient` as `device_info_is_rtl_language`.

#### App Lifecycle Events @ashiagr (https://github.com/Automattic/pocket-casts-android/pull/270)
- [x] application_installed: When a user launch the app for the very first time
- [x] application_opened: When a user launches or foregrounds the app after the first open
- [x] application_updated: When the app launches for the first time after installing a new version
- [x] application_closed: When the the app is put into the background

#### User lifecycle events @mchowning (https://github.com/Automattic/pocket-casts-android/pull/276)
- [x] user_signed_in: When the user successfully signs into the app using existing credentials
[...]

PR #383, opened 2022-10-07
Track playback effect events 

ashiagr:
| 📘 Project: #261 | 🛫 Depends on: #381 |
|:---:|:---:|

Corresponding iOS PR: https://github.com/Automattic/pocket-casts-ios/pull/291

This adds events for the playback effect settings
- `playback_effect_speed_changed`:  When the user changes the speed the podcast plays
- `playback_effect_trim_silence_toggled`:  When the user toggles the trim silence option
- `playback_effect_trim_silence_amount_changed`: When the user changes the trim select amount
- `playback_effect_volume_boost_toggled`: When the user toggles the volume boost option

## To test

1. Play an episode if you aren't already
3. Tap the dial icon to open the playback effects view
4. Tap the + button on the Speed option, then tap outside the playback effects view
5. ✅ Verify you see the following in console: `🔵 Tracked: playback_effect_speed_changed ["source": "player_playback_effects", "speed": 1.1]`
6. Tap the dial icon again to open the playback effects view, tap the - button, then tap outside the playback effects view
7. ✅ `🔵 Tracked: playback_effect_speed_changed ["source": "player_playback_effects", "speed": 1.0]`
8. Tap the dial icon again to open the playback effects view, then tap the middle option to cycle through predefined options,  then tap outside the playback effects view
9. ✅ Verify the `speed` property in console changes accordingly (speed changes are tracked when the playback effects view is destroyed to [match the behavior on iOS](https://github.com/Automattic/pocket-casts-ios/pull/291#discussion_r981633702))
10. Enable the Trim Silence option
11. ✅ `🔵 Tracked: playback_effect_trim_silence_toggled ["source": "player_playback_effects", "enabled": true]`
12. Change your trim silence setting to Medium
13. ✅ `🔵 Tracked: playback_effect_trim_silence_amount_changed ["amount": "medium", "source": "player_playback_effects"]`
14. Change to mad max
15. ✅ `🔵 Tracked: playback_effect_trim_silence_amount_changed ["source": "player_playback_effects", "amount": "mad_max"]`
16. Change back to mild
17. ✅ `🔵 Tracked: playback_effect_trim_silence_amount_changed ["source": "player_playback_effects", "amount": "mild"]`
18. Enable the Volume boost option
19. ✅ `🔵 Tracked: playback_effect_volume_boost_toggled ["source": "player_playback_effects", "enabled": true]`
20. Repeat above tests for subscribed podcast's playback effects settings which you can reach as follows:
     - Select a subscribed podcast
     - Tap on the gear icon on podcast screen 
     - Tap on "Playback effects" on podcasts settings
20. ✅ Verify that `"source": "podcast_settings"` is shown for these events

# Checklist
N/A 

### DIFF
--- a/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastEffectsViewModel.kt
+++ b/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastEffectsViewModel.kt
@@ -10,7 +10,6 @@ import au.com.shiftyjelly.pocketcasts.models.entity.Podcast
 import au.com.shiftyjelly.pocketcasts.models.type.TrimMode
 import au.com.shiftyjelly.pocketcasts.preferences.Settings
 import au.com.shiftyjelly.pocketcasts.repositories.playback.PlaybackManager
-import au.com.shiftyjelly.pocketcasts.repositories.playback.PlaybackManager.PlaybackSource
 import au.com.shiftyjelly.pocketcasts.repositories.podcast.PodcastManager
 import au.com.shiftyjelly.pocketcasts.utils.extensions.clipToRange
 import dagger.hilt.android.lifecycle.HiltViewModel
@@ -57,7 +56,7 @@ class PodcastEffectsViewModel
         val podcast = this.podcast.value ?: return
         launch {
             podcastManager.updateTrimMode(podcast, trimMode)
-            trackPlaybackEffectsEvent(AnalyticsEvent.PLAYBACK_EFFECT_TRIM_SILENCE_AMOUNT_CHANGED, mapOf(AMOUNT_KEY to trimMode.analyticsVale))
+            trackPlaybackEffectsEvent(AnalyticsEvent.PLAYBACK_EFFECT_TRIM_SILENCE_AMOUNT_CHANGED, mapOf(PlaybackManager.AMOUNT_KEY to trimMode.analyticsVale))
             if (shouldUpdatePlaybackManager()) {
                 playbackManager.updatePlayerEffects(podcast.playbackEffects)
             }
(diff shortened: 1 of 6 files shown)


modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastEffectsViewModel.kt
```diff
@@ -51,6 +56,7 @@ class PodcastEffectsViewModel
         val podcast = this.podcast.value ?: return
         launch {
             podcastManager.updateTrimMode(podcast, trimMode)
+            trackPlaybackEffectsEvent(AnalyticsEvent.PLAYBACK_EFFECT_TRIM_SILENCE_AMOUNT_CHANGED, mapOf(PlaybackManager.AMOUNT_KEY to trimMode.analyticsVale))
```
mchowning, line 59: This event gets fired with an amount of "mild" when you toggle on trim silence from the podcast settings here, but if you turn on trim silence from the player there is no change event fired until you change from the default of mild. Can we make it behave consistently. 

I would probably lean toward not firing the event since it seems like it always defaults to mild, but I don't feel strongly at all.

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastEffectsViewModel.kt
```diff
@@ -24,12 +26,15 @@ class PodcastEffectsViewModel
 @Inject constructor(
     private val podcastManager: PodcastManager,
     private val playbackManager: PlaybackManager,
-    private val settings: Settings
+    private val settings: Settings,
+    private val analyticsTracker: AnalyticsTrackerWrapper,
```
mchowning, line 30: It looks like this `analyticsTracker` isn't used.

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastEffectsViewModel.kt
```diff
@@ -24,12 +26,15 @@ class PodcastEffectsViewModel
...
 ) : ViewModel(), CoroutineScope {
 
     override val coroutineContext: CoroutineContext
         get() = Dispatchers.Default
 
+    private var isSpeedChanged: Boolean = false
+    private var finalSpeed: Double = 1.0
```
mchowning, line 37: Both here and in the `EffectsFragment`, instead of having a default `finalSpeed`, I think we could default it to `null` and use a `null` `finalSpeed` as indicating the speed had not been changed and a non-`null` `finalSpeed` as indicating that the speed had been changed (renaming it to something like `updatedSpeed` might fit better in that case).

This would have the benefit of dropping the `isSpeedChanged` field. Also, it would avoid us defaulting `finalSpeed` to 1.0 even when that isn't the actual starting speed (i.e., when the speed had been changed previously).

I don't feel strongly about this change, so this is more just food for thought. Definitely feel free to leave it as-is if you prefer.

mchowning (commented):
Looks good. Great work! 

I think the only issue that needs to be fixed is to make it consistent whether a change event is fired when turning on trim silence.

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastEffectsViewModel.kt
ashiagr: Makes sense to me 👍 
Updated in 7ee4fac

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastEffectsViewModel.kt
ashiagr: Yep, it'll be an improved version this way, I applied it in 74626ec.

mchowning (approved):
This is working well!

PR #385, opened 2022-10-08
Tracks for Filters part 2(filter_updated)

mchowning:
| 📘 Project: #261 |
|:---:|

iOS PRs: https://github.com/Automattic/pocket-casts-ios/pull/226 and https://github.com/Automattic/pocket-casts-ios/pull/304

Adds the `filter_updated` event

### 1. Creating a filter

1. Go through the creation flow and create a filter, making changes to the various possible settings.
2. ✅  Confirm that at no point during the flow was a `filter_updated` event fired.

### 2. Filter Chips
1. Tap on a filter from the filters list
2. Tap the little down arrow by the filter's name to open the filter chips
3. ✅  Tap on each chip in turn, make a change, and verify that the appropriate `filter_updated` event is fired. Note that even if the end result is no change to the filter, the event will fire if the user made any selections (i.e., if the user made a change and then undid that change without exiting the screen, an update event would fire when they exited the screen).
     - Podcasts -> `filter_updated {"group": "podcasts", "source": "filters",...}
     - Episode Status -> `filter_updated {"group": "episode_status", "source": "filters",...}
     - Release Date -> `filter_updated {"group": "release_date", "source": "filters",...}
     - Duration -> `filter_updated {"group": "episode_duration", "source": "filters",...}
     - Download Status -> `filter_updated {"group": "download_status", "source": "filters",...}
     - Media Type -> `filter_updated {"group": "media_type", "source": "filters",...}
     - Starred -> `filter_updated {"group": "starred", "source": "filters",...}

### 3. Podcast Settings
1. Open a podcast's screen
2. Tap the gear icone to open settings
3. Tap the Filters row
4. Change which filters this podcast is included in
5. ✅ Observe the event `filter_updated {"group": "podcasts", "source": "podcast_settings", ...}`

# Checklist

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/PlaylistManagerImpl.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/PlaylistManagerImpl.kt
@@ -2,6 +2,8 @@ package au.com.shiftyjelly.pocketcasts.repositories.podcast
 
 import android.content.Context
 import android.os.Build
+import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsEvent
+import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsTrackerWrapper
 import au.com.shiftyjelly.pocketcasts.models.db.AppDatabase
 import au.com.shiftyjelly.pocketcasts.models.entity.Episode
 import au.com.shiftyjelly.pocketcasts.models.entity.Playlist
@@ -34,6 +36,7 @@ private const val CREATED_DEFAULT_PLAYLISTS = "createdDefaultPlaylists"
 class PlaylistManagerImpl @Inject constructor(
     private val settings: Settings,
     private val downloadManager: DownloadManager,
+    private val analyticsTracker: AnalyticsTrackerWrapper,
     @ApplicationContext private val context: Context,
     appDatabase: AppDatabase
 ) : PlaylistManager, CoroutineScope {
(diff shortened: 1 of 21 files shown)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/PlaylistManagerImpl.kt
```diff
@@ -309,6 +319,35 @@ class PlaylistManagerImpl @Inject constructor(
...
+        // Don't send a filter updated event if the playlist is being created or if
+        // the user did not initiate the update
+        if (!playlist.draft && userPlaylistUpdate != null) {
+            userPlaylistUpdate.properties.map { playlistProperty ->
+                when (playlistProperty) {
+
+                    is FilterUpdatedEvent -> {
+                        val properties = mapOf(
+                            "group" to playlistProperty.groupValue,
+                            "source" to userPlaylistUpdate.source.analyticsValue
```
mchowning, line 335: "group" and "source" are extracted to constants in the follow-up PR: https://github.com/Automattic/pocket-casts-android/pull/386

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/UserPlaylistUpdate.kt
```diff
@@ -0,0 +1,12 @@
+package au.com.shiftyjelly.pocketcasts.repositories.podcast
+
+data class UserPlaylistUpdate(
+    val properties: List<PlaylistProperty>,
+    val source: PlaylistUpdateSource
+) {
```
ashiagr, line 6: I like the approach you took with passing `userPlaylistUpdate` to `PlaylistManagerImpl.update(...)` and understand [your earlier comment](PlaylistManagerImpl) better now. As you mentioned, this approach has a type-checking advantage.

For filters, it is working very well. I know, we need to resolve unknown sources for playback events and maybe this approach can help. I'll discuss it in a separate PR once I have a better idea about the complexity.

ashiagr (approved):

PR #386, opened 2022-10-08
Tracks for Filters part 3 (auto_download_*)

mchowning:
| 📘 Project: #261 | 🛫 Depends on: #385 |
| --- | --- |

Adds the `filter_auto_download_updated` and `filter_auto_download_limit_updated` events.

### 1. Filter Options
1. Open the filters tab
2. Tap on a filter
3. From the overflow menu select "Filter Settings"
4. Toggle Auto Download  on
5. ✅  Observe the event `filter_auto_download_updated, {"source": "filters", "enabled": "true", ...}`
6. Change the number of episodes to Auto Download
7. ✅  Observe the event `filter_auto_download_limit_updated, {"limit": [SELECTED LIMIT], ...}`
8. Toggle Auto Download off
5. ✅  Observe the event `filter_auto_download_updated, {"source": "filters", "enabled": "false", ...}`

### 2. Auto Download Settings
1. Open the profile tab
2. Tap on the Settings cog
3. Tap on Auto download
4. Tap on the Filters setting
5. ✅  Tap the checkboxes for some of the filters and observe a`filter_auto_download_updated` event is fired with the appropriate "enabled" property and a "source" property of "auto_download_settings"
6. Tap on Select All/None
7. ✅  Observe that an appropriate `filter_auto_download_updated` event is fired for each filter that is updated. For example if you have 5 filters, but 2 of them are already set up to auto download, then tapping Select All should result in 3 events.

# Checklist

- [X] Should this change be included in the release notes? If yes, please add a line in CHANGELOG.md
- [X] Have you tested in landscape?
- [X] Have you tested accessibility with TalkBack?
- [X] Have you tested in different themes?
- [X] Does the change work with a large display font?
- [X] Are all the strings localized?
- [X] Could you have written any new tests?
- [X] Did you include Compose previews with any components?

PR #387, opened 2022-10-08
Tracks for Filters part 4 (filter_sort_by_changed)

mchowning:
| 📘 Project: #261 | 🛫 Depends on: #386 |
| --- | --- |

Adding the `filter_sort_by_changed` event.

## To Test

1. Open the filters tab
2. Tap on a filter
3. From the overflow menu, select "Sort By"
4. Dismiss the sort modal without selecting anything
5. ✅  Observe that there is no `filter_sort_by_changed` event fired
6. Reopen the "Sort By" modal
7. Tap on any sorting option
8. ✅  Observe that a `filter_sort_by_changed` event is fired with an appropriate "sort_order" property of either "newest_to_oldest", "oldest_to_newest", "shortest_to_longest", or "longest_to_shortest"

# Checklist

### DIFF
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/Playlist.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/Playlist.kt
@@ -33,7 +33,7 @@ data class Playlist(
     @ColumnInfo(name = "autoDownload") var autoDownload: Boolean = false,
     @ColumnInfo(name = "autoDownloadWifiOnly") var autoDownloadUnmeteredOnly: Boolean = false,
     @ColumnInfo(name = "autoDownloadPowerOnly") var autoDownloadPowerOnly: Boolean = false,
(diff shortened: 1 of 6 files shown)


modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/Playlist.kt
```diff
@@ -156,4 +150,19 @@ data class Playlist(
 
     val isAllEpisodes: Boolean
         get() = unplayed && partiallyPlayed && finished && notDownloaded && downloaded && audioVideo == AUDIO_VIDEO_FILTER_ALL && allPodcasts && !filterDuration && filterHours == 0 && !starred
+
+    fun sortOrder() = SortOrder.fromInt(sortId)
+
+    enum class SortOrder(val value: Int) {
```
mchowning, line 156: I created this enum to minimize how much we have to pass around and match against plain `Int`s.

ashiagr (approved):
Works as described 👍

PR #388, opened 2022-10-08
Tracks for filters part 5 (filter_edit_dismissed)

mchowning:
| 📘 Project: #261 | 🛫 Depends on: #387  |
| --- | --- |

Adding the `filter_edit_dismissed` event.

## To Test

For each of these tests, start by:
1. From the filters tab, tap on a filter
2. From the overflow menu, select "Filter Settings"

### 1. No changes
1. Immediately exit from the Filter Settings screen without making any changes by tapping on the navigation arrow
2. ✅  Observe that a `filter_edit_dismissed` event is fired, but all of the setting properties are false (did_change_auto_download, did_change_episode_count, did_change_icon, did_change_name, did_change_color)
3. Reopen the Filter Settings screen
4. Exit from the Filter Settings screen without making any changes by using a back gesture
5. 2. ✅  Observe that a `filter_edit_dismissed` event is fired, but all of the setting properties are false

### 2. Change all Settings
1. Change the Filter's name, icon, color, auto download and auto download limit settings (note that the auto download and auto download limit change will fire their own events immediately when those settings are changed)
2. Exit the Filter Settings screen
3. ✅  Observe that a `filter_edit_dismissed` event is fired, and all of the setting properties are true (did_change_auto_download, did_change_episode_count, did_change_icon, did_change_name, did_change_color)
4. Reopen the Filter Settings screen
5. Again, change all of the filter's settings
6. Rotate the device so that a configuration change occurs
7. Exit the Filter settings screen
8. ✅  Again, observe that a `filter_edit_dismissed` event is fired, and all of the setting properties are true even though the changes occurred before the configuration change (did_change_auto_download, did_change_episode_count, did_change_icon, did_change_name, did_change_color)

# Checklist

### DIFF
--- a/modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/filters/CreateFilterViewModel.kt
+++ b/modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/filters/CreateFilterViewModel.kt
@@ -4,6 +4,8 @@ import androidx.lifecycle.LiveData
 import androidx.lifecycle.LiveDataReactiveStreams
 import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.ViewModel
+import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsEvent
+import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsTrackerWrapper
 import au.com.shiftyjelly.pocketcasts.models.entity.Episode
 import au.com.shiftyjelly.pocketcasts.models.entity.Playlist
 import au.com.shiftyjelly.pocketcasts.repositories.extensions.calculateCombinedIconId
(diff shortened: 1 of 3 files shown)


modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/filters/CreateFilterViewModel.kt
```diff
@@ -24,9 +26,24 @@ import javax.inject.Inject
...
     CoroutineScope {
+
+    companion object {
+        private object AnalyticsPropKey {
+            const val DID_CHANGE_AUTO_DOWNLOAD_EPISODE_COUNT = "did_change_episode_count"
+            const val DID_CHANGE_ICON = "did_change_icon"
+            const val DID_CHANGE_NAME = "did_change_name"
+            const val DID_CHANGE_AUTO_DOWNLOAD = "did_change_auto_download"
+            const val DID_CHANGE_COLOR = "did_change_color"
+        }
```
mchowning, line 44: I kind of liked how this grouped the related strings together and kind of gave them their own namespace. I thought this could be really nice in classes that had other unrelated strings in the companion object. wdyt?

ashiagr: I was thinking about it lately and I like this idea. 👍 
I'll start grouping properties like this in future PRs.

ashiagr (approved):
I found everything working in my testing. Code changes look good as well. 

Great job!

Issue #379, opened 2022-10-07
Android Automotive: Allow users to view played episodes


geekygecko:
### What problem does this address?

In Automotive OS currently, the played episodes are hidden. It would be good to give the user the option to view these. 

### What is your proposed solution?

As we can't have extra buttons in the UI the option will need to be added to the settings page. Something like:

**Show played episodes** 
Episodes that you have finished playing will be shown.

![Screenshot_20221007_130011](https://user-images.githubusercontent.com/308331/194454179-37048b3f-275d-4cde-b0d3-773d398491b1.png)

The episode rows could use the new content styles to show if an episode is Play, In Progress, or Played.

https://developer.android.com/training/cars/media#per-item-styles
<img width="686" alt="Screen Shot 2022-10-07 at 1 04 36 pm" src="https://user-images.githubusercontent.com/308331/194455969-bd882296-7185-4f44-9b6a-bb094ae8ef2c.png">

### Did you search for existing feature suggestions?

- [X] I have searched for existing feature suggestions.

PR #389, opened 2022-10-10
Added Automotive setting to show played.

geekygecko:
This change adds a setting to allow Automotive OS users to show played episodes.

Fixes https://github.com/Automattic/pocket-casts-android/issues/379

Other changes are:

- Moved the Automotive library dependency versions from the build.gradle to dependencies.gradle.
- Upgraded AboutLibraries from 10.4.0 to 10.5.0
  https://github.com/mikepenz/AboutLibraries/releases/tag/v10.5.0

**Test Steps**

1. Install the Automotive app on an emulator
2. Start playing an episode
3. On the full screen play mark the episode as played
4. ✅ Verify you don't see the episode in the list
5. Open the settings page and turn on "Show played episodes"
6. ✅ Verify you do see the episode in the list

Screenshot_20221010_132421

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
         ([#335](https://github.com/Automattic/pocket-casts-android/pull/335)).
     *   Allow select text in show notes from episode details view.
         ([#372](https://github.com/Automattic/pocket-casts-android/pull/372)).
+    *   Added Automotive OS setting to show played episodes.
+        ([#389](https://github.com/Automattic/pocket-casts-android/pull/389)).
 
 7.24
 -----
--- a/automotive/build.gradle
+++ b/automotive/build.gradle
@@ -49,17 +49,17 @@ dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
     // Commented out the Automotive library as it clashes with the Material library and we don't use it. Duplicate value for resource attr/navigationIconTint.
     //implementation 'androidx.car:car:1.0.0-alpha7'
-    implementation 'androidx.appcompat:appcompat:1.4.1'
-    implementation 'androidx.core:core-ktx:1.7.0'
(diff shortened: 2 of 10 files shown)


mchowning (approved):
This is working pretty well although I am seeing an issue where I have to reload the episode list screen to get the completed episodes "removed".

This seems like a pre-existing issue (based on the second set of steps to reproduce which do not involve this setting), but it seems like more of a concern now because in the first set of steps to reproduce it makes it seem like the "Show Played" toggle doesn't work. I'll go ahead and approve this in case you'd like to address this issue outside of this PR.

#### 1. Updated Settings not reflected in episode list

1. Mark an episode of a given podcast as complete
2. Go to the episode list for that podcast
3. Observe that the played episode is (or is not) shown correctly based on the "Show Played" toggle
4. Tap on the gear icon
5. Update the "Show Played" toggle
6. Tap back
7. Observe that the episode list has not been updated based on the new state of the "Show Played" toggle
8. Back out of the episode list and reload that list
9. Observe that the episode list has now been updated based on the new state of the "Show Played" toggle

#### 2. Mark as Played not reflected in episode list

1. Have the "Show Played" toggle turned off so played episodes should not be played.
2. Start playing an episode of a podcast
3. Go to the episode list for that podcast
4. Tap on the miniplayer to open the full screen player
5. Mark the episode as played
6. Go back to the episode list from the full screen player
7. Observe that the episode that was marked as played is shown even though the "Show Played" toggle is off
8. Back out of the episode list and reload that list
9. Observe that the episode that was marked as played is no longer shown

geekygecko:
> I am seeing an issue where I have to reload the episode list screen to get the completed episodes "removed".

Unfortunately, I don't believe it's possible to fix this issue. The Automotive OS pages seem to work like remote pages, they are requested from us rather than us having control over them.

PR #391, opened 2022-10-10
Tracks for filters part 6 (filter_list_reordered)

mchowning:
| 📘 Project: #261 |
| --- |

## Description

Adds the `filter_list_reordered` event.

## To Test

1. Open the filters tab
2. Long press on one of the filters and then move it to a new position
3. ✅  Observe that a `filter_list_reordered` event is fired
4. Long press on one of the filters but leave it in the same position
5. ✅ Observe that a `filter_list_reordered` event is _not_ fired

# Checklist

### DIFF
--- a/modules/services/analytics/src/main/java/au/com/shiftyjelly/pocketcasts/analytics/AnalyticsEvent.kt
+++ b/modules/services/analytics/src/main/java/au/com/shiftyjelly/pocketcasts/analytics/AnalyticsEvent.kt
@@ -267,4 +267,5 @@ enum class AnalyticsEvent(val key: String) {
     PLAYBACK_EFFECT_TRIM_SILENCE_TOGGLED("playback_effect_trim_silence_toggled"),
     PLAYBACK_EFFECT_TRIM_SILENCE_AMOUNT_CHANGED("playback_effect_trim_silence_amount_changed"),
     PLAYBACK_EFFECT_VOLUME_BOOST_TOGGLED("playback_effect_volume_boost_toggled"),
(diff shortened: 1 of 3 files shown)


ashiagr (approved):
This is working as described 👍

modules/services/analytics/src/main/java/au/com/shiftyjelly/pocketcasts/analytics/AnalyticsEvent.kt
```diff
@@ -267,4 +267,5 @@ enum class AnalyticsEvent(val key: String) {
     PLAYBACK_EFFECT_TRIM_SILENCE_TOGGLED("playback_effect_trim_silence_toggled"),
     PLAYBACK_EFFECT_TRIM_SILENCE_AMOUNT_CHANGED("playback_effect_trim_silence_amount_changed"),
     PLAYBACK_EFFECT_VOLUME_BOOST_TOGGLED("playback_effect_volume_boost_toggled"),
+    FILTER_LIST_REORDERED("filter_list_reordered"),
```
ashiagr, line 270: I realized after merging that `FILTER_LIST_REORDERED` event can be grouped with the `Filters screen` events. Can you take care of it in your upcoming filters PR?

mchowning: > Can you take care of it in your upcoming filters PR?

Yes, of course! Thanks for catching that. 🙇

PR #343, opened 2022-09-23
Track playback

ashiagr:
| 📘 Project: #261 |
|:---:|

Add tracks for:

* Play
* Pause
* Skip back
* Skip forward

### Screens

1. Run the app
2. Tap "Podcasts" tab
3. Tap a podcast
4. Play and pause any episode
5. ✅ Check that `play ["source": "podcast_screen"]` and `pause ["source": "podcast_screen"]` is tracked
6. Tap Filters
7. Access any list with podcasts
8. Tap play and pause on any episode
9. ✅ Check that `play ["source": "filters"]` and `pause ["source": "filters"]` is tracked
10. Tap Discover
11. Scroll down till you see a featured trailer
12. Tap "Play Trailer" twice (one to play and one to pause)
13. ✅ Check that `play ["source": "discover"]` and `pause ["source": "discover"]` is tracked
14. Tap Profile > Downloads
15. Play and pause any episode
16. ✅ Check that `play ["source": "downloads"]` and `pause ["source": "downloads"]` is tracked
14. Tap Profile > Files
15. Play and pause any episode
16. ✅ Check that `play ["source": "files"]` and `pause ["source": "files"]` is tracked
14. Tap Profile > Starred
15. Play and pause any episode
16. ✅ Check that `play ["source": "starred"]` and `pause ["source": "starred"]` is tracked
14. Tap Profile > Listening History
15. Play and pause any episode
16. ✅ Check that `play ["source": "listening_history"]` and `pause ["source": "listening_history"]` is tracked

### Miniplayer

1. Tap play and pause on the miniplayer
2. ✅ Check that `play ["source": "miniplayer"]` and `pause ["source": "miniplayer"]` is tracked
3. Tap skip back on the miniplayer
3. ✅ Check that `skip_back ["source": "miniplayer"]` is tracked
3. Tap skip forward on the miniplayer
3. ✅ Check that `skip_forward ["source": "miniplayer"]` is tracked

### Player

1. Open the player screen (tap the podcast cover on the miniplayer)
1. Tap play and pause
2. ✅ Check that `play ["source": "player"]` and `pause ["source": "player"]` is tracked
3. Tap skip back on the player
3. ✅ Check that `skip_back ["source": "player"]` is tracked
3. Tap skip forward on the player
3. ✅ Check that `skip_forward ["source": "player"]` is tracked

### Media Button Broadcast Action

1. Long press the app icon, choose widgets and add podcast widget to your device
2. Tap play and then pause on the widget player
[...]

PR #392, opened 2022-10-11
Identify and skip/track few unknown playback sources

ashiagr:
| 📘 Project: #261 | 📘 Continuation of: #343) |
|:---:|:---:|

> **Warning**
> The PR targets release/7.24 branch

This PR 
1. Explicitly passes `PlaybackSource` instead of setting it on the `PlaybackManager` so that we're sure of where those sources are coming from - da6367c3555639746a7cfad0bd4760ed256e684d
2. Explicitly passes `PlaybackSource` from `warningsHelper.streamingWarningDialg(...)` as suggested in https://github.com/Automattic/pocket-casts-android/pull/343#discussion_r983907170
3. Identifies a few `unknown` playback sources: `up_next`, `media_button_broadcast_search_action`, `chromecast`, `auto_play`, `auto_pause`. 
    - `auto_play`,  `auto_pause` sources are skipped from tracking.
    -  For `chromecast`, I'm only tracking the cast-connected scenario. Current screen source is tracked irrespective of whether the player is a simple player or a cast player. I can take it up separately if we decide to track it differently.

These changes might not eliminate all the unknowns. Hopefully, it is a step forward in identifying some of the missing ones. 

#### Test-1 Refactoring 
- Go through the testing steps from https://github.com/Automattic/pocket-casts-android/pull/343
- ✅ Verify that events are tracked as before

#### Test-2 Streaming Warning Dialog 
- Disable wifi and enable mobile network
- Make sure warn before using data is enabled in Profile -> Settings -> Storage & data use
- Play an episode so that `You're not on WiFi` alert is displayed
- Tap on `Stream Anyway` 
- ✅ Verify that the current screen source is tracked
- Re-enable wifi 

#### Test-3.1 Up Next
- Disable Profile -> Settings -> Play Up Next episode on tap 
- Play an episode
- Add a few episodes to the `Up Next` queue
- Long press on one of the episodes in `Up Next` queue
- ✅ Verify that the selected episode starts playing and is tracked with the source `up_next`
- Enable Profile -> Settings -> Play Up Next episode on tap 
- Tap on one of the episodes in `Up Next` queue 
- ✅ Verify that the selected episode starts playing and is tracked with the source `up_next`

#### Test-3.2 media_button_broadcast_search_action
- Close the application
- Test the search using the following terminal command
  `adb shell am start -a android.media.action.MEDIA_PLAY_FROM_SEARCH -p au.com.shiftyjelly.pocketcasts.debug --es query "The\ Daily"`
- ✅ Verify that an episode from a podcast title starting with "The\ Daily" starts playing and is tracked with the source `media_button_broadcast_search_action`

#### Test-3.3 Multiple episodes - play all
- Go to filters tab
- Select a filter having episodes
- Select options drop down in the tool bar and tap on Play all
- Tap Play x episodes from the "Play all" confirmation dialog
-  ✅ Verify that an episode starts playing and is tracked with the source `filters`

#### Test-4 Auto-play/ auto-pause skip tracking
- Add few episodes to up-next queue
- Play an episode
- Long press on the mini-player
- Tap on Mark played
- ✅ Verify that no playback event is tracked
- Expand the mini player 
- Drag the seek-bar to the end so that next episode is played
- ✅ Verify that no playback event is tracked
- While the episode is playing, background the app and play a song from another app
- Reopen the app
- ✅ Verify that the episode that was playing is paused and no playback event is tracked

# Checklist
N/A

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
@@ -1864,6 +1864,7 @@ open class PlaybackManager @Inject constructor(
         PLAYER("player"),
         NOTIFICATION("notification"),
         FULL_SCREEN_VIDEO("full_screen_video"),
+        UP_NEXT("up_next"),
         MEDIA_BUTTON_BROADCAST_ACTION("media_button_broadcast_action"),
         CHROMECAST("chromecast"),
         AUTO_PLAY("auto_play"),
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
@@ -1850,9 +1850,6 @@ open class PlaybackManager @Inject constructor(
         }
     }
 
-    private fun PlaybackSource.skipTracking() =
-        this in listOf(PlaybackSource.AUTO_PLAY, PlaybackSource.AUTO_PAUSE)
-
     enum class PlaybackSource(val analyticsValue: String) {
         PODCAST_SCREEN("podcast_screen"),
         FILTERS("filters"),
@@ -1874,6 +1871,8 @@ open class PlaybackManager @Inject constructor(
         CHROMECAST("chromecast"),
         AUTO_PLAY("auto_play"),
         AUTO_PAUSE("auto_pause"),
-        UNKNOWN("unknown"),
+        UNKNOWN("unknown");
+
+        fun skipTracking() = this in listOf(AUTO_PLAY, AUTO_PAUSE)
     }
 }
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
@@ -206,7 +206,7 @@ open class PlaybackManager @Inject constructor(
         withContext(Dispatchers.Default) {
             upNextQueue.playNext(autoPlayEpisode, downloadManager) {
                 launch {
-                    loadCurrentEpisode(autoPlay)
+                    loadCurrentEpisode(play = autoPlay, playbackSource = PlaybackSource.AUTO_PLAY)
                 }
             }
         }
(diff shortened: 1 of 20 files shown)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
```diff
@@ -1863,7 +1867,13 @@ open class PlaybackManager @Inject constructor(
         PLAYER("player"),
         NOTIFICATION("notification"),
         FULL_SCREEN_VIDEO("full_screen_video"),
+        UP_NEXT("up_next"),
         MEDIA_BUTTON_BROADCAST_ACTION("media_button_broadcast_action"),
+        MEDIA_BUTTON_BROADCAST_SEARCH_ACTION("media_button_broadcast_search_action"),
+        PLAYER_BROADCAST_ACTION("player_broadcast_action"),
```
ashiagr, line 1873: I added `player_broadcast_action` source to track any event fired from `PlayerBroadcastReceiver`. I wasn't able to receive any event in that receiver. Seems like those events are now received in `MediaSessionManager.onMediaButtonEvent(...)` and tracked with the source `media_button_broadcast_action`. I still thought to add this source to be sure we're not missing it.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
```diff
@@ -608,10 +607,10 @@ open class PlaybackManager @Inject constructor(
                 onCompletion(episode.uuid)
             }
         }
-        trackPlayback(AnalyticsEvent.PLAYBACK_SKIP_FORWARD)
+        trackPlayback(AnalyticsEvent.PLAYBACK_SKIP_FORWARD, playbackSource)
     }
 
-    fun skipBackward() {
+    fun skipBackward(playbackSource: PlaybackSource = PlaybackSource.UNKNOWN) {
```
mchowning, line 613: Just food for thought... I have mixed feelings about defaulting `PlaybackSource` arguments to be `UNKNOWN`. I kind of like the idea of not having a default so that it feels kind of icky everytime you see a call with an `UNKNOWN` (or <gasp> add a call with an `UNKNOWN`) and helps give us a nudge to get rid of those unknowns when we notice that one doesn't make sense. Having it just be a default paramger kind of feels like it becomes hidden.

On the other hand, losing the default parameter adds quite a bit of boilerplate. Like I said, I feel conflicted, and I just wanted to share my thinking. This is totally fine as-is.

mchowning (approved):
These are some great improvements. Thanks @ashiagr !

PR #395, opened 2022-10-11
Turn off Sentry's file IO tracing

### LINKED ISSUE #394: In-app logs not updating appropriately
### Description

The in-app logs that get sent to support seem to have started not updating correctly. In particular, it seems that the logs work fine up to a given point in time, but then there is a gap until the very last log line. It's almost as if instead of appending to the logs, new logs are instead just overwritten onto the last log line.

For example, I experienced an unrelated bug this morning, but when I sent the logs to myself, they looked like this:

```
I 04/10 13:28:20 Worker Downloading episode Jeremiah 90cf4c8b-aafe-4cb1-8ff9-5732053aad66
I 04/10 13:28:20 UpNextSyncJob - scheduled
I 11/10 08:54:45 stopForeground state: 2 (update notification)
```

Note that those logs show only a single entry since October 4th. It was inside of that gap that I encountered the bug I was trying to report, so I know I was using the phone during that time. If you're running a build of the app that never properly saved the logs, then it seems you will always only have a single line of logs.

This issue did not exist in the 7.22.0 release.

internal ref: p1665491925690749/1665479608.840769-slack-C02A333D8LQ

1. Open the app
2. Do some things that would generate logs
3. Take a look at the log file on the phone (`data/data/au.com.shiftyjelly.pocketcasts.debug/files/logs/`)
4. Observe that there is a gap in the logs (or there is only a single log line)

_No response_

Pixel 6 Pro, Android 13, PC 7.23
Pixel 3, Android 12, PC main branch at https://github.com/Automattic/pocket-casts-android/commit/0765a874332d75c183d2f3c8bb45cbb61ab0b90c

mchowning:
## Description

This fixes a problem where the in-app log files were not getting updated correctly. Instead of appending to the log file it seems that the last line of the log file was just getting replaced.

It seems this started occurring because Sentry does [byte-code manipulation to some of the File I/O classes](https://docs.sentry.io/platforms/android/configuration/integrations/file-io/). I'm not clear on exactly what they're doing that is causing this problem, but turning that off seems to get our logs working again.

> **Note**
> I believe we should include this as a beta-fix since it basically makes the app logs that get sent with support requests useless. For that reason, I'm targeting the `release/7.24 branch`. Let me know if you disagree.

Fixes #394 

## To Test
1. Open the app
2. Do some things that would generate logs
3. Take a look at the log file on the phone (data/data/au.com.shiftyjelly.pocketcasts.debug/files/logs/)
4. Observe that the logs are being updated appropriately since this fix has been applied (there may be missing logs from before you applied this fix).

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
         ([#342](https://github.com/Automattic/pocket-casts-android/pull/342)).
     *   Add a helpful alert if the user is signed out in the background.
         ([#340](https://github.com/Automattic/pocket-casts-android/pull/340)).
+    *   Fix in-app logs not updating appropriately.
+    *   ([#395](https://github.com/Automattic/pocket-casts-android/pull/395)).
 *   Health:
     *   Migrate app to Android 13 by targeting the SDK version 33.
         ([#312](https://github.com/Automattic/pocket-casts-android/pull/312)).
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,3 +1,5 @@
+import io.sentry.android.gradle.extensions.InstrumentationFeature
+
 plugins {
     id 'io.sentry.android.gradle'
 }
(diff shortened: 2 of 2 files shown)


ashiagr (approved):
Tested an earlier version where I could reproduce the issue. 

I can confirm that these changes fix the issue. Thanks for taking care of it @mchowning!

PR #396, opened 2022-10-12
Track player shelf events

ashiagr:
| 📘 Project: #261  | 
|:---:|

iOS PR: https://github.com/Automattic/pocket-casts-ios/pull/293

This adds events for the "player shelf" actions:

- `player_shelf_action_tapped`: When the user taps on one of the actions located in the player shelf or overflow menu
- `player_shelf_overflow_menu_shown`: When the user taps on the vertical `...` button and the overflow menu is displayed
- `player_shelf_overflow_menu_rearrange_started`: When the user taps the Edit button and the overflow menu enters rearrange mode
- `player_shelf_overflow_menu_rearrange_action_moved`: When the user moves an action in during a rearrange
- `player_shelf_overflow_menu_rearrange_finished`: When the user exits the rearrange mode

## To test

1. Play an episode if you're not already
3. Tap each of the 4 actions located in the shelf at the bottom of the player view
4. ✅ Verify you see the following in console: `🔵 Tracked: player_shelf_action_tapped ["from": "shelf", "action": "NAME"]` - Where name is the name of the action that was tapped
5. Tap the ... to open the overflow menu
6. ✅ `🔵 Tracked: player_shelf_overflow_menu_shown`
7. Tap the actions located in the overflow menu
8. ✅ `🔵 Tracked: player_shelf_action_tapped ["from": "overflow_menu", "action": "NAME"]`
9. Open the overflow menu again
10. Tap the edit button
11. ✅ `🔵 Tracked: player_shelf_overflow_menu_rearrange_started`
12. Move an item out of the "Shortcut on player" section
13. ✅ `🔵 Tracked: player_shelf_overflow_menu_rearrange_action_moved ["action": "NAME", "moved_to": "overflow_menu", "moved_from": "shelf", "position": POSITION]`
- NAME: Is the name of the action that was moved
- POSITION is the new position in section it was moved to
14. Move an action from the "More actions" section to the "Shortcut on player" section
15. ✅ `🔵 Tracked: player_shelf_overflow_menu_rearrange_action_moved ["moved_from": "overflow_menu", "action": "NAME", "moved_to": "shelf", "position": POSITION]`
16. Move an item to a higher position in the Shortcut On player section
17. ✅ `🔵 Tracked: player_shelf_overflow_menu_rearrange_action_moved ["moved_from": "shelf", "action": "NAME", "moved_to": "shelf", "position": POSITION]`
18. Tap the Done button
19. ✅ `🔵 Tracked: player_shelf_overflow_menu_rearrange_finished`

# Checklist
N/A

### DIFF
--- a/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/ShelfBottomSheet.kt
+++ b/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/ShelfBottomSheet.kt
@@ -6,6 +6,8 @@ import android.view.View
 import android.view.ViewGroup
 import androidx.fragment.app.activityViewModels
 import androidx.recyclerview.widget.LinearLayoutManager
+import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsEvent
+import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsTrackerWrapper
 import au.com.shiftyjelly.pocketcasts.player.databinding.FragmentShelfBottomSheetBinding
 import au.com.shiftyjelly.pocketcasts.player.viewmodel.PlayerViewModel
 import au.com.shiftyjelly.pocketcasts.repositories.chromecast.CastManager
@@ -20,6 +22,7 @@ import javax.inject.Inject
 @AndroidEntryPoint
 class ShelfBottomSheet : BaseDialogFragment() {
     @Inject lateinit var castManager: CastManager
+    @Inject lateinit var analyticsTracker: AnalyticsTrackerWrapper
 
     override val statusBarColor: StatusBarColor? = null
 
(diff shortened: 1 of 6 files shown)


modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/ShelfBottomSheet.kt
```diff
@@ -79,16 +96,28 @@ class ShelfBottomSheet : BaseDialogFragment() {
...
             }
-            is ShelfItem.Archive -> { playerViewModel.archiveCurrentlyPlaying(resources)?.show(parentFragmentManager, "archive") }
-            else -> {}
+            is ShelfItem.Archive -> {
+                analyticsAction = ShelfItem.ShelfItemId.ARCHIVE.analyticsValue
+                playerViewModel.archiveCurrentlyPlaying(resources)?.show(parentFragmentManager, "archive")
+            }
+            else -> {
+                analyticsAction = AnalyticsProp.Value.UNKNOWN
+            }
```
mchowning, line 115: I think the only missing branch here is for `Download`, so what if Instead of having an `else` branch here we added that branch specifically. I feel like that would be a bit more informative about which items we're not expecting here and if we add a new `ShelfItem` in the future, we would get a compile error here and force us to explicitly decide how we want to handle the new item instead of it just getting caught in the else. wdyt?

(note that the commit I linked to in my other comment includes this change)

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/ShelfItem.kt
```diff
@@ -81,4 +81,21 @@ sealed class ShelfItem(val id: String, var title: (Playable?) -> Int, var iconRe
...
+        CAST("cast", "chromecast"),
+        PLAYED("played", "mark_as_played"),
+        ARCHIVE("archive", "archive"),
+        DOWNLOAD("download", "download");
+
+        companion object {
+            fun fromId(id: String?) =
+                ShelfItemId.values().find { it.value == id }
+        }
+    }
```
mchowning, line 100: Instead of having this separate `ShelfItemId` class that we use to map from ids to analytics values, what would you think about adding the analytics value to the `ShelfItem` class itself?

I tried this out to see if it worked, and it seemed to. You can take a look at what I mean [here](https://github.com/Automattic/pocket-casts-android/commit/8b3145ffa159345cac5cebd5b639c7e248634c6f).

mchowning (approved):
Looks good and works well! I left a couple of small comments on the code, but no blockers.

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/ShelfItem.kt
ashiagr: That looks better! Thanks for the commit, I've cherry-picked it.

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/ShelfBottomSheet.kt
ashiagr: Makes sense to me 👍
Accepted in b5c5a8e

PR #400, opened 2022-10-12
Add token info to user.env-example

mchowning:
## Description

Adds some info about the kind of token needed to the example user env file so the user knows what kind of tokens they need to add. If you think there's a better place to provide this information, please feel free to just update this PR directly.

I do not know if these are the correct token permissions, but I was able to successfully run the `new_beta_release` script with them. I based these permissions on https://github.com/woocommerce/woocommerce-android/blob/trunk/scripts/release-management-prechecks.sh#L53-L55.

## Checklist

### DIFF
--- a/fastlane/env/user.env-example
+++ b/fastlane/env/user.env-example
@@ -3,5 +3,5 @@ GITHUB_TOKEN=<GitHub access token>
 
 # Requires:
 #   Organizations: 'Automattic',
-#   REST Scopes: 'read_repos', 'write_repos'
+#   REST Scopes: 'write_builds'
 BUILDKITE_TOKEN=<Buildkite Personal Access Token>
--- a/fastlane/env/user.env-example
+++ b/fastlane/env/user.env-example
@@ -1,3 +1,7 @@
+# Requires: 'repo'
 GITHUB_TOKEN=<GitHub access token>
 
+# Requires:
+#   Organizations: 'Automattic',
+#   REST Scopes: 'read_repos', 'write_repos'
 BUILDKITE_TOKEN=<Buildkite Personal Access Token>


geekygecko:
Thanks for adding this helpful message. I might let @oguzkocer or @mokagio approve this, as they will know if these are the correct permissions.

mchowning:
Looks like there is info about this [in the release checklist](https://github.com/Automattic/mobile-toolbox/blob/trunk/android/pocketcastsandroid-release-scenario.sh#L31-L34). I still think it would be nice to have this in the example file though.

> - [\`GITHUB_TOKEN\`](https://github.com/settings/tokens) added to your environment (with "repo" scope)
> - [\`BUILDKITE_TOKEN\`](https://buildkite.com/user/api-access-tokens/new) added to your environment
>     - It should have the scopes \`read_builds\`, \`write_builds\`, \`read_build_logs\`, \`read_organizations\`, \`read_pipelines\`.
> - Alternatively, you can copy the example environment file with \`cp fastlane/env/user.env-example \$HOME/.pocketcastsandroid-env.default\` then adding the necessary tokens using the above instructions.

> **Warning**
> Based on the instructions in the checklist, it looks like I may have not be granting enough permission to my buildkite token. 🤔

fastlane/env/user.env-example
```diff
@@ -1,3 +1,7 @@
+# Requires: 'repo'
 GITHUB_TOKEN=<GitHub access token>
 
+# Requires:
+#   Organizations: 'Automattic',
+#   REST Scopes: 'read_repos', 'write_repos'
```
mokagio, line 6: I don't see any `_repos` REST scopes on Buildkite. Could they have been `*_builds`?

image

Also, I'm 99% confident the only scope we need is the write one, `write_builds`.

```suggestion
#   REST Scopes: 'write_builds'
```

We use the token in the `buildkite_trigger_build_action` from [release toolkit](https://github.com/wordpress-mobile/release-toolkit/, which uses the [`buildkit` gem](https://github.com/Shopify/buildkit internally. The gem [calls the `POST /builds` endpoint for us](https://github.com/Shopify/buildkit/blob/742e2e2ad8fc255f1e26143bf5502b017d754da9/lib/buildkit/client/builds.rb#L67-L85). The endpoint [documentation](https://buildkite.com/docs/apis/rest-api/builds#create-a-build) lists `write_builds` as the only required scope:

image

mokagio (commented):
These comments are very helpful!

fastlane/env/user.env-example
mchowning: Yes, I got my wires crossed there. Thanks! 🙇

mokagio (approved):
I think we can merge this and eventually amend if we discover that the information is inadequate.

mchowning:
👋 @geekygecko , can I get an approval from you on this so I can merge?

geekygecko (approved):

Issue #348, opened 2022-09-28
Android Automotive: Starred episode list


geekygecko:
A user can star an episode in Android Automotive but there is no way to view the starred episodes. This change should add the starred episode list to the "Filters" tab.

We used to include a default starred filter in the app but it was moved into the "Profile" tab as we wanted to include podcasts that you aren't subscribed to. I wonder if we should have two sections in the "Filters" tab, one for your filters and one for system filters (but a better name). It would be good if the starred list does the same as the version on the Profile tab and included unsubscribed episodes.  

If we could also add the listening history episode list that would be ideal. 

![Screenshot_20220928_211222](https://user-images.githubusercontent.com/308331/192771008-5592a289-5fb5-4d3f-bfb1-6b6e8c7eb822.png)

PR #403, opened 2022-10-13
Added new episode lists to Automotive OS. Starred, Listening History, and Files.

geekygecko:
This change adds a new section Profile section to Android Automotive. The section includes the profile episode lists of Starred and Listening History. If the user is signed into Plus then they also see Files, the user uploaded files.

Having four tabs doesn't look great in some of the emulators but it seems fine in the Polestar emulator. It makes more sense to match the mobile apps and have the lists in a Profile section, rather than put the lists in the Filters section. The Filters section should be left for just subscribed podcasts or it may cause confusion for the user.

Fixes https://github.com/Automattic/pocket-casts-android/issues/348

### Profile episode lists
1. Tap the Discover tab
2. Tap a podcast
3. Tap an episode
4. In the full screen player tap the three dots next to the skip forward button.
5. Tap the star icon
6. Close the full screen page and podcast page
7. Tap the Profile tab
8. Tap the Starred item
9. ✅ Verify you see the episode you starred in the list
10. Tap the back arrow
11. Tap the Listening History item
12. ✅ Verify you see the episode you played in the list

## Files episode list
1. Tap the settings cog
2. Sign into a Pocket Casts account with Plus
3. Tap the Podcasts tab
4. Tap the Filters tab
5. ✅ Verify you see the Files item in the list


ashiagr (approved):
Nice work, @geekygecko! I tested Profile options and found everything working. I agree with your decision to keep these new options under Profile to be consistent with the mobile version and to avoid confusion.

Code changes look good too 👍

PR #404, opened 2022-10-13
Track episode details events

ashiagr:
| 📘 Project: #261 | 
|:---:|

iOS PR: https://github.com/Automattic/pocket-casts-ios/pull/297

This adds events to the Episode Detail view:

- `episode_detail_shown`:  When the episode details view is displayed to the user
- `episode_detail_show_notes_link_tapped`: When the user taps a link from the show notes section
- `episode_detail_podcast_name_tapped`: When the user taps the podcast name link/button
- `episode_detail_dismissed`: When the episode details view is dismissed
## To test

1. Launch the app
2. Tap on the Podcasts tab
3. Select a podcast
4. Tap on an episode
5. ✅ `🔵 Tracked: episode_detail_shown ["source": "podcast_screen"]`
6. Tap on a link that's located in the show notes
7. ✅ `🔵 Tracked: episode_detail_show_notes_link_tapped ["episode_uuid": "UUID", "source": "podcast_screen"]` - Where UUID is the uuid of the episode that is being presented
8. Go back
9. Tap on the name of the Podcast under the episode name
10. ✅ `🔵 Tracked: episode_detail_podcast_name_tapped ["source": "podcast_screen"]`
11. Tap on an episode again and manually dismiss it
12. ✅ `🔵 Tracked: episode_detail_dismissed ["source": "podcast_screen"]`

Repeat the steps on the following views:

1. Go to the Filters tab
2. Tap on a Filter
3. ✅  Follow the steps from above and verify the source is `filters`
4. Tap on the Profile tab
5. Tap on Downloads
6. ✅  Follow the steps from above and verify the source is `downloads`
7. Tap on Starred
8. ✅  Follow the steps from above and verify the source is `starred`
9. Tap on Listening History
10. ✅  Follow the steps from above and verify the source is `listening_history`
11. Go to discover
12. Locate the "SINGLE EPISODE TEST" item, and tap on it
13. ✅ Verify the source is `discover`
14. Sign in if you're not signed in, trigger new episode notification from `Profile -> Settings -> Developer`, and background the app
15. Tap on the new episode notification when it arrives
16. ✅ Verify the source is `notification`
17. Tap share from the episode details screen and copy the share url
18. Open the URL on your device browser, and tap Open in Pocket Casts
19. ✅ Verify the source is `share`

# Checklist
N/A

### DIFF
--- a/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/episode/EpisodeFragment.kt
+++ b/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/episode/EpisodeFragment.kt
@@ -344,15 +344,15 @@ class EpisodeFragment : BaseDialogFragment() {
                             binding.lblAuthor.compoundDrawableTintList = ColorStateList.valueOf(iconColor)
                         }
                         binding.lblAuthor.setOnClickListener {
+                            analyticsTracker.track(
+                                AnalyticsEvent.EPISODE_DETAIL_PODCAST_NAME_TAPPED,
+                                mapOf(
+                                    AnalyticsProp.Key.EPISODE_UUID to state.episode.uuid,
+                                    AnalyticsProp.Key.SOURCE to EpisodeViewSource.PODCAST_SCREEN.value
+                                )
+                            )
                             dismiss()
                             if (!overridePodcastLink) {
-                                analyticsTracker.track(
-                                    AnalyticsEvent.EPISODE_DETAIL_PODCAST_NAME_TAPPED,
-                                    mapOf(
-                                        AnalyticsProp.Key.EPISODE_UUID to state.episode.uuid,
-                                        AnalyticsProp.Key.SOURCE to EpisodeViewSource.PODCAST_SCREEN.value
-                                    )
-                                )
                                 (listener as FragmentHostListener).openPodcastPage(state.podcast.uuid)
                             }
                         }
--- a/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/episode/EpisodeFragment.kt
+++ b/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/episode/EpisodeFragment.kt
@@ -25,11 +25,14 @@ import androidx.core.view.isVisible
 import androidx.core.view.updateLayoutParams
 import androidx.fragment.app.viewModels
 import androidx.lifecycle.Observer
+import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsEvent
+import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsTrackerWrapper
 import au.com.shiftyjelly.pocketcasts.analytics.FirebaseAnalyticsTracker
 import au.com.shiftyjelly.pocketcasts.localization.helper.TimeHelper
 import au.com.shiftyjelly.pocketcasts.models.entity.Episode
 import au.com.shiftyjelly.pocketcasts.models.type.EpisodePlayingStatus
 import au.com.shiftyjelly.pocketcasts.models.type.EpisodeStatusEnum
+import au.com.shiftyjelly.pocketcasts.models.type.EpisodeViewSource
 import au.com.shiftyjelly.pocketcasts.podcasts.databinding.FragmentEpisodeBinding
 import au.com.shiftyjelly.pocketcasts.preferences.Settings
 import au.com.shiftyjelly.pocketcasts.repositories.images.PodcastImageLoader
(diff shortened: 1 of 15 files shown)


modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/type/EpisodeViewSource.kt
```diff
@@ -0,0 +1,19 @@
+package au.com.shiftyjelly.pocketcasts.models.type
+
+enum class EpisodeViewSource(val value: String) {
+    DISCOVER("discover"),
```
ashiagr, line 4: I added this class to `services.models` module. Adding it in the `feature.podcasts` module where the `EpisodeFragment` resides made it inaccessible from the `services.ui` module's `openEpisodeDialog(...)` due to cyclic dependency.

Other places I can think of adding it are the settings or the analytics modules. Or I can add the source as a string in `openEpisodeDialog(...)` but we'll lose type-safety in this case.

Let me know if you can think of a better home for the class.

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/episode/EpisodeFragment.kt
```diff
@@ -170,6 +207,9 @@ class EpisodeFragment : BaseDialogFragment() {
 
     override fun onDestroyView() {
         super.onDestroyView()
+        if (!viewModel.isFragmentChangingConfigurations) {
+            analyticsTracker.track(AnalyticsEvent.EPISODE_DETAIL_DISMISSED)
```
mchowning, line 211: I think this is supposed to have a "source" property.

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/type/EpisodeViewSource.kt
mchowning: Adding this to `models` like you have is probably what I would have done too.

> ...Or I can add the source as a string in openEpisodeDialog(...) but we'll lose type-safety in this case.

Yeah, I wouldn't want to lose the type-safety.

mchowning:
> 7. Tap on the Files
> 8. ✅ Follow the steps from above and verify the source is files

I don't think it's possible to get an episode details view for a locally downloaded file, so I didn't complete these test steps. Let me know if I'm missing something.

mchowning (approved):
Thanks for taking care of this Ashita! I noticed a few issues, but overall this works well.

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/episode/EpisodeFragment.kt
```diff
@@ -306,6 +346,13 @@ class EpisodeFragment : BaseDialogFragment() {
                         binding.lblAuthor.setOnClickListener {
                             dismiss()
                             if (!overridePodcastLink) {
+                                analyticsTracker.track(
+                                    AnalyticsEvent.EPISODE_DETAIL_PODCAST_NAME_TAPPED,
+                                    mapOf(
+                                        AnalyticsProp.Key.EPISODE_UUID to state.episode.uuid,
+                                        AnalyticsProp.Key.SOURCE to EpisodeViewSource.PODCAST_SCREEN.value
+                                    )
+                                )
```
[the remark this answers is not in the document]
ashiagr: Good catch, moved it outside in 70d1aad. It doesn't have to be inside the if-check as the podcast page is already behind the episode details screen and doesn't need to be opened explicitly in this case.

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/episode/EpisodeFragment.kt
ashiagr: Added in 237b285

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/episode/EpisodeFragment.kt
```diff
@@ -158,6 +194,7 @@ class EpisodeFragment : BaseDialogFragment() {
 
         episodeUUID?.let { episodeUuid ->
             podcastUuid?.let { podcastUuid ->
+                analyticsTracker.track(AnalyticsEvent.EPISODE_DETAIL_SHOWN, mapOf(AnalyticsProp.Key.SOURCE to episodeViewSource.value))
```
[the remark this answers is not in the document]
ashiagr: Fixed in 237b285.

ashiagr:
> I don't think it's possible to get an episode details view for a locally downloaded file (instead you get a file details modal), so I didn't complete these test steps. Let me know if I'm missing something.

There's a corresponding `user_file_detail_shown` event for this scenario so yes, you're right, it can't be fired for files. I've updated the instructions. 

> I noticed a few issues, but overall this works well.

Sorry about the issues, I submitted the PR towards the end of my day without giving it a second pass. 😞 I'll be careful next time.

PR #405, opened 2022-10-14
Update custom media action order for mediastyle notification

ashiagr:
# Description

> **Warning**
> Targets release/7.24

Starting Android 13, there can be [more playback controls](https://developer.android.com/about/versions/13/behavior-changes-13#playback-controls) in the `mediastyle` notification. With the current set of custom actions, `mark_as_played` action has started displaying at the end of the notification. An accidental tap on it can mark the episode as played causing a poor user experience. 

This PR reorders `mark_as_played` action below the `star` action so that the `star` action takes precedence. 

Slack convo: p1665681166247159-slack-C02A333D8LQ

Before | After
-----| ----
before | after

# Checklist
N/A
- [ ] Should this change be included in the release notes? If yes, please add a line in CHANGELOG.md - We changed app's target to Android 13 in 7.24. Since this is a beta fix, the change doesn't need to be included in the CHANGELOG.

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/MediaSessionManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/MediaSessionManager.kt
@@ -353,8 +353,6 @@ class MediaSessionManager(
             stateBuilder.addCustomAction(APP_ACTION_CHANGE_SPEED, "Change speed", drawableId)
         }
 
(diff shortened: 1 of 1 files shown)


mchowning (approved):
Thanks for this Ashita! I think this will avoid a lot of frustration from users accidentally archiving episodes.

Tested on both Android 12 and Android 13 and this looks good to me.

mchowning:
> **Warning**
> Targets release/7.24

Just noting that this PR actually was still targeting main. This branch was branched off of the `release/7.24` branch, so I was able to just update the PR to target `release/7.24` and everything looked good.

PR #407, opened 2022-10-17
Fix status bar colors

geekygecko:
This change fixes the issue on the light theme that the status bar icons are hidden because they are the same color as the bar. I noticed this when I was working on the Google Sign In but it also affects the share podcasts page. 

| Before | After |
| --- | --- |
| Screenshot_20221017_152738 | Screenshot_20221017_153316 |
| Screenshot_20221017_152745 | Screenshot_20221017_153323 |
| Screenshot_20221017_152757 | Screenshot_20221017_153245 |

### DIFF
--- a/modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/fragments/BaseFragment.kt
+++ b/modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/fragments/BaseFragment.kt
@@ -43,19 +43,24 @@ open class BaseFragment : Fragment(), CoroutineScope, HasBackstack {
         view.isFocusable = true
     }
 
-    @Suppress("DEPRECATION")
-    override fun setUserVisibleHint(visible: Boolean) {
-        super.setUserVisibleHint(visible)
+    override fun onResume() {
+        super.onResume()
 
         // Need to make sure we are the top fragment before updating the status bar
-        val backstack = activity?.supportFragmentManager?.backStackEntryCount
-        if (visible && (backstack == 0 || activity?.supportFragmentManager?.fragments?.last() == this)) {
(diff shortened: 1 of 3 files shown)


modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/fragments/BaseFragment.kt
```diff
@@ -43,19 +43,24 @@ open class BaseFragment : Fragment(), CoroutineScope, HasBackstack {
...
-        super.setUserVisibleHint(visible)
+    override fun onResume() {
+        super.onResume()
 
         // Need to make sure we are the top fragment before updating the status bar
-        val backstack = activity?.supportFragmentManager?.backStackEntryCount
-        if (visible && (backstack == 0 || activity?.supportFragmentManager?.fragments?.last() == this)) {
+        val fragmentManager = activity?.supportFragmentManager
+        if (fragmentManager != null && (fragmentManager.backStackEntryCount == 0 || fragmentManager.fragments.last() == this)) {
             updateStatusBar()
```
geekygecko, line 52: `setUserVisibleHint` wasn't firing when the account fragment opened. I have moved this to the onResume method, as the documentation advises. Unfortunately, this meant I had to add the suppress annotation to the child fragments overriding this. We should also fix up these warnings in a future PR.

> Deprecated
If you are manually calling this method, use FragmentTransaction.setMaxLifecycle(Fragment, Lifecycle.State) instead. If overriding this method, behavior implemented when passing in true should be moved to onResume(), and behavior implemented when passing in false should be moved to onPause().

ashiagr (approved):
LGTM! Thanks for fixing it, @geekygecko!

PR #412, opened 2022-10-18
Merge release/7.24.1 into main

ashiagr:
Binary has just been submitted to PlayStore.


geekygecko (approved):

version.properties
```diff
@@ -1,3 +1,3 @@
 # Version Information for Vanilla / Release builds
-versionName=7.25-rc-1
-versionCode=9040
+versionName=7.24.1
```
mchowning, line 2: Should we be updating the version name "back" to a 7.24 version?

ashiagr: I need to upload a new beta for `7.25__` with an upgraded version code which should supersede this change and the `7.24_` release on Play Console. I'm waiting to do that until `7.24_` gets approved.

Issue #410, opened 2022-10-17
Project: End of Year


ashiagr:
## Description

Show end of year stats.

### Initial work

- [x] Add modal structure #413
- [x] Add stories structure #428
- [x] Add profile card structure and badge #498
- [x] Load a dummy story + add gestures #486

### Presenting and sharing Stories

- [x] Generate a story with dummy data #495
- [x] Generate an image based on the story shown #505
- [x] Share the generated image through a share sheet #505

### Stories data
- [x] Show total listening time in 2022 https://github.com/Automattic/pocket-casts-android/pull/522
- [x] Show number of different categories they’ve listened to https://github.com/Automattic/pocket-casts-android/pull/531
- [x] Show top 5 categories and how many podcasts in each of these they’ve listened to https://github.com/Automattic/pocket-casts-android/pull/531
- [x] Show total number of podcasts and episodes they’ve listened to https://github.com/Automattic/pocket-casts-android/pull/536
- [x] Show number 1 podcast for 2022 https://github.com/Automattic/pocket-casts-android/pull/537
- [x] Show top 5 podcasts of the year https://github.com/Automattic/pocket-casts-android/pull/537
- [x] Show longest episode they listened to in 2022 https://github.com/Automattic/pocket-casts-android/pull/541
- [x] Show the first and last story (with the option to replay) https://github.com/Automattic/pocket-casts-android/pull/542
- [x] Check eligibility and if the user has the whole 2022 Listening History, if not, grab their 2022 listening history from the server https://github.com/Automattic/pocket-casts-android/pull/551

#### Known Issues/ Improvements
- Check swipe down motion to close sensitivity (disabled to allow scrolling for large font size)
- [x] ~~Address podcasts tap events capture~~  Addressed in #505
- [x] ~~Address weird state noticed in https://github.com/Automattic/pocket-casts-android/pull/495#issuecomment-1292676258~~ Addressed in #505
- [x] ~~Sudden speed up in Stories~~ (noticed in #505) Addressed in https://github.com/Automattic/pocket-casts-android/pull/531

### Design

- [x] Finish the modal and profile's card design https://github.com/Automattic/pocket-casts-android/pull/552
- [x] Finish the stories view design
    - [x] Intro story https://github.com/Automattic/pocket-casts-android/pull/553
    - [x] Listening time story https://github.com/Automattic/pocket-casts-android/pull/556
    - [x] Listened categories story https://github.com/Automattic/pocket-casts-android/pull/559
[...]

PR #413, opened 2022-10-18
End of Year: Add modal structure

ashiagr:
| 📘 Project: #410 |
|:---:|

This PR adds a `Compose` modal bottom sheet that prompts the user to check their End of Year. Design (and tablet) is not covered in this task and will be done in a future PR.

## To test

2. Run the app
3. ✅ Check that a bottom sheet appears with two buttons: "View My 2022" and "Not Now"
4. ✅ Tap "Not Now" and ensure the bottom sheet collapses
5. Re-run the app
6. When the bottom sheet appears, tap "View My 2022"
7. ✅ Ensure that nothing happens ("View My 2022" is not yet connected to any screen)
8. Rotate the device to landscape mode
9. ✅ Ensure that modal bottom sheet does not collapse and spreads to landscape mode
10. Rotate the device to portrait mode and tap outside the modal view
11. ✅ Ensure that modal bottom sheet collapses
12. Re-run the app
13. When the bottom sheet appears, perform a back gesture or press device back button
14. ✅ Ensure that modal bottom sheet collapses
15. Perform a back gesture or press device back button again
16.  ✅ Ensure that app goes to background

## Checklist

- Should this change be included in the release notes? If yes, please add a line in CHANGELOG.md N/A
- Are all the strings localized? - Added to strings xml
- Could you have written any new tests? N/A

### DIFF
--- a/modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/bottomsheet/BottomSheetContent.kt
+++ b/modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/bottomsheet/BottomSheetContent.kt
@@ -82,7 +82,6 @@ fun BottomSheetContent(
             TextH40(
                 text = content.titleText,
                 color = MaterialTheme.theme.colors.primaryText01,
-                maxLines = 1
             )
 
             Spacer(modifier = modifier.height(16.dp))
--- a/modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/bottomsheet/BottomSheetContent.kt
+++ b/modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/bottomsheet/BottomSheetContent.kt
@@ -1,0 +1,143 @@
+package au.com.shiftyjelly.pocketcasts.compose.bottomsheet
+
+import androidx.compose.foundation.BorderStroke
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.wrapContentHeight
+import androidx.compose.foundation.layout.wrapContentWidth
+import androidx.compose.material.Button
+import androidx.compose.material.MaterialTheme
+import androidx.compose.material.OutlinedButton
+import androidx.compose.material.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.tooling.preview.PreviewParameter
+import androidx.compose.ui.unit.dp
+import au.com.shiftyjelly.pocketcasts.compose.AppThemeWithBackground
+import au.com.shiftyjelly.pocketcasts.compose.components.TextH40
+import au.com.shiftyjelly.pocketcasts.compose.components.TextP50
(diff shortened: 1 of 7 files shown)


modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/bottomsheet/ModalBottomSheet.kt
```diff
@@ -0,0 +1,53 @@
...
+import androidx.compose.material.ModalBottomSheetValue
+import androidx.compose.material.rememberModalBottomSheetState
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.rememberCoroutineScope
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.unit.dp
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+
+@OptIn(ExperimentalMaterialApi::class)
```
ashiagr, line 17: I'm using [ModalBottomSheetState](https://developer.android.com/reference/kotlin/androidx/compose/material/ModalBottomSheetState) which is experimental right now and I had to opt-in for `ExperimentalMaterialApi`. It made things really simple, I also tested it works on older Android versions (API level 23). But let me know if it's a deal breaker.

modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/bottomsheet/BottomSheetContent.kt
```diff
@@ -0,0 +1,143 @@
...
+            horizontalAlignment = Alignment.CenterHorizontally
+        ) {
+            val content = state.content
+
+            Spacer(modifier = modifier.height(16.dp))
+
+            TextH40(
+                text = content.titleText,
+                color = MaterialTheme.theme.colors.primaryText01,
+                maxLines = 1
```
mchowning, line 85: Do we need a maxLines here? I'm just wondering if we want this text ellipsized if it extends wider than the screen on some phone/screen/display-settings configurations.

mchowning (approved):
Looks good! I'm excited that you're starting in on the EOY project. 😄

modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/bottomsheet/ModalBottomSheet.kt
mchowning: I think that's fine. With compose still relatively young I don't think we should be too hesitant about using experimental apis when they make sense because a lot of fairly fundamental stuff still seems to be experimental.

modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/bottomsheet/BottomSheetContent.kt
```diff
@@ -0,0 +1,143 @@
...
+            val onClick: (() -> Unit)?
+
+            data class Primary(
+                override val label: String,
+                override val onClick: () -> Unit,
+            ) : Button
+
+            data class Secondary(
+                override val label: String,
+                override val onClick: (() -> Unit)? = null,
```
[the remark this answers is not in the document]
ashiagr: In the current implementation, [clicking the secondary button dismisses the bottom sheet](https://github.com/Automattic/pocket-casts-android/pull/413/files#diff-7039a52beb217dffc340f844cd92bd2dd2d3a21a7e228981b82fbb0def9ed2dfR109-R110). If that's the only behavior we need, we might not want to set any click listener. 

I'm open to making any changes in case we need to.

modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/bottomsheet/BottomSheetContent.kt
ashiagr: I was thinking to limit title length to give enough space to the summary text in case of a large display setting. But the check might not be needed as we'll know the title and summary beforehand and we can test how they look on large displays. I removed it in cf006e6.

PR #414, opened 2022-10-18
filter_created Tracks event

mchowning:
## Description

Adds the filter_created event

## To Test

Just do this a few times with the different possible filter configurations:
1. Create a new filter
2. Observe that a `filter_created` event is fired with the appropriate properties for these keys:
    1. all_podcasts
    2. color
    3. downloaded
    4. duration
    5. episode_status_played
    6. episode_status_in_progress
    7. episode_status_unplayed
    8. icon_name
    9. media_type
    10. release_date
    11. starred

# Checklist

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/PlaylistUpdateAnalytics.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/PlaylistUpdateAnalytics.kt
@@ -20,99 +20,118 @@ class PlaylistUpdateAnalytics @Inject constructor(
     ) {
 
         when {
-            isCreatingFilter -> {
-                sendPlaylistCreatedEvent(playlist)
-            }
+            isCreatingFilter -> sendPlaylistCreatedEvent(playlist)
 
             // If the playlist is a draft, then we are in the middle of the filter creation flow and
             // we don't want to send update events
-            !playlist.draft -> {
-                sendPlaylistUpdateEvent(userPlaylistUpdate)
-            }
+            !playlist.draft -> sendPlaylistUpdateEvent(userPlaylistUpdate)
         }
     }
 
     private fun sendPlaylistCreatedEvent(playlist: Playlist) {
         val properties = buildMap<String, Any> {
 
-            val icon = when (playlist.drawableId) {
-
-                IR.drawable.ic_filters_list,
-                IR.drawable.auto_filter_list,
-                IR.drawable.automotive_filter_list -> AnalyticsProp.Value.Icon.LIST
-
-                IR.drawable.ic_filters_headphones,
-                IR.drawable.auto_filter_headphones,
-                IR.drawable.automotive_filter_headphones -> AnalyticsProp.Value.Icon.HEADPHONES
-
-                IR.drawable.ic_filters_clock,
-                IR.drawable.auto_filter_clock,
-                IR.drawable.automotive_filter_clock -> AnalyticsProp.Value.Icon.CLOCK
-
-                IR.drawable.ic_filters_download,
-                IR.drawable.auto_filter_downloaded,
-                IR.drawable.automotive_filter_downloaded -> AnalyticsProp.Value.Icon.DOWNLOADED
-
-                IR.drawable.ic_filters_play,
-                IR.drawable.auto_filter_play,
-                IR.drawable.automotive_filter_play -> AnalyticsProp.Value.Icon.PLAY
-
-                IR.drawable.ic_filters_volume,
-                IR.drawable.auto_filter_volume,
-                IR.drawable.automotive_filter_volume -> AnalyticsProp.Value.Icon.VOLUME
-
-                IR.drawable.ic_filters_video,
-                IR.drawable.auto_filter_video,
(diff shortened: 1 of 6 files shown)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/PlaylistUpdateAnalytics.kt
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    private fun sendPlaylistCreatedEvent(playlist: Playlist) {
+        val properties = buildMap<String, Any> {
+
+            put(Key.ALL_PODCASTS, playlist.allPodcasts)
+            colorAnalyticsValue(playlist)?.let {
+                put(Key.COLOR, it)
+            }
+            put(Key.DOWNLOADED, playlist.downloaded)
```
geekygecko, line 38: This is missing downloading and not downloaded but it isn't in the spreadsheet so not an issue with this pull request. 

The download option of "all" has downloaded, downloading, and not downloaded set to true. So it might not be that much use without the other flags.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/PlaylistUpdateAnalytics.kt
```diff
@@ -0,0 +1,236 @@
...
+
+            else -> {
+                Timber.e("No matching analytics icon found")
+                null
+            }
+        }
+
+    private fun mediaTypeAnalyticsValue(playlist: Playlist) =
+        when (playlist.audioVideo) {
+            0 -> Value.MediaType.ALL
```
geekygecko, line 100:
```suggestion
            Playlist.AUDIO_VIDEO_FILTER_ALL -> Value.MediaType.ALL
```
Should this use this constant instead?

geekygecko (approved):

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/PlaylistUpdateAnalytics.kt
```diff
@@ -0,0 +1,236 @@
...
+            releaseDateAnalyticsValue(playlist)?.let {
+                put(Key.RELEASE_DATE, it)
+            }
+        }
+
+        analyticsTracker.track(AnalyticsEvent.FILTER_CREATED, properties)
+    }
+
+    private fun iconAnalayticsValue(playlist: Playlist) =
+        when (playlist.drawableId) {
```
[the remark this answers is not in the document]
mchowning: > I might be misunderstanding

You didn't misunderstand, I did. 🙂  I was thinking that I would cover the auto/automotive case here too, but you're right that only the app drawables come through the `drawableId` field. Updated in [a0cae1f](https://github.com/Automattic/pocket-casts-android/pull/414/commits/a0cae1fd8036b6be980f8fbf52570b5240a73bf8).

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/PlaylistUpdateAnalytics.kt
mchowning: Yes, I updated to use all AUDIO_VIDEO_FILTER constants in [ede79da](https://github.com/Automattic/pocket-casts-android/pull/414/commits/ede79da6905e3ae595b187d6b98e8a53d6e37b05). That's much better, thanks!

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/PlaylistUpdateAnalytics.kt
mchowning: Thanks for catching this!

I added a `not_downloaded` boolean prop in [1e311e4](https://github.com/Automattic/pocket-casts-android/pull/414/commits/1e311e46dd22424a656ea306c5dc372f85b1a933).

I'm less clear on what the `downloading` value does though. It looks like the download option of "all" [only sets `downloaded` and `notDownloaded` to `true`](https://github.com/Automattic/pocket-casts-android/blob/1e02c85fe568047f2dc95baf41268e70fed100ed/modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/filters/TimeOptionsFragment.kt#L209-L213).  It does not update `downloading`. In fact, I don't see any way that a user can alter the `downloading` property on a filter. Is there a bug here because the `downloading` property is never getting updated by the client?

geekygecko: Oh I got that wrong, we must have taken `downloading` out the UI at some point. I found this comment in the code:

```
// regular filters no longer have a downloading but the not downloaded one should show in progress downloads
```

https://github.com/Automattic/pocket-casts-android/blob/1e02c85fe568047f2dc95baf41268e70fed100ed/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/PlaylistManagerImpl.kt#L367

PR #416, opened 2022-10-19
Organize dependency declarations

mchowning:
# Description

Just a tiny bit of cleanup to make it easier to see whether a specific module dependency is already declared.

The only change here that is more than moving lines around is the removal of a duplicate declaration of the analytics dependency inside the profile module.

# Checklist

- [N/A] Should this change be included in the release notes? If yes, please add a line in CHANGELOG.md
- [N/A] Have you tested in landscape?
- [N/A] Have you tested accessibility with TalkBack?
- [N/A] Have you tested in different themes?
- [N/A] Does the change work with a large display font?
- [N/A] Are all the strings localized?
- [N/A] Could you have written any new tests?
- [N/A] Did you include Compose previews with any components?

### DIFF
--- a/modules/features/account/build.gradle
+++ b/modules/features/account/build.gradle
@@ -13,17 +13,20 @@ android {
 }
 
 dependencies {
-    implementation project(':modules:services:localization')
-    implementation project(':modules:services:preferences')
-    implementation project(':modules:services:utils')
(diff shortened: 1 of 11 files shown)


geekygecko (approved):
Very organised!

PR #418, opened 2022-10-19
Enable gradle build cache when building pocket casts

liutikas:
# Description

Build caching is off by default in Gradle

Details https://docs.gradle.org/current/userguide/build_cache.html#sec:build_cache_enable

# Checklist

- [x] Should this change be included in the release notes? If yes, please add a line in CHANGELOG.md
- [x] Are all the strings localized?
- [x] Could you have written any new tests?


geekygecko (approved):
Have I been waiting longer than I need to for builds all this time, ekk! Thanks for making this change.

PR #420, opened 2022-10-20
Disable unused Android Gradle Plugin features

liutikas:
# Description

Disable unused Android Gradle Plugin features to speed up builds

Defaults are documented here:
https://developer.android.com/reference/tools/gradle-api/7.3/com/android/build/api/dsl/BuildFeatures?hl=en


geekygecko (approved):
Thanks again

PR #421, opened 2022-10-20
Fix for fullscreen player button not animating

### LINKED ISSUE #207: The full screen player's play button no longer animates
This change should fix the full screen player's play button to animate between a play and pause icon. It should animate the same way as the mini player's play button. It used to animate so this isn't a new feature.

VGJohn:
# Description

Setting animate to true so the animation is played when isPlaying changes state

Fixes #207


geekygecko (approved):
Well, this is embarrassing! I thought this was going to be a complicated issue but you showed me.  🤦‍♂️

Thanks for helping.

VGJohn:
Haha no worries, nothing to be embarrassed about. I was a bit surprised that was it too but it seems to work from my testing. Nice first issue though 😋

Cheers for the approval, love the app, been using it for years! Awesome to see it go open source!

PR #422, opened 2022-10-20
Hide next episode date if unknown

TylerCarberry:
Some podcasts don't have an expected next episode date. This happens for shows with an inconsistent release schedule. Previously it would show "Next episode" with no date. Now that section is completely hidden.

Before:
Screen Shot 2022-10-19 at 10 18 59 PM

After:
Screen Shot 2022-10-19 at 10 16 59 PM

For context, this is what it looks like for episodes where the next date is predicted (both before and after this change)

Screen Shot 2022-10-19 at 10 22 46 PM


geekygecko (approved):
Great change! Thank you so much for your help. 🙏

PR #425, opened 2022-10-20
Initial value for skip backwards setting uses the skip forwards value

VGJohn:
Just spotted this minor issue, looks like a bad copy paste maybe 😄

The initial value for the skip backwards setting was using the skip forwards value instead of the skip backwards value.


ashiagr:
👋@VGJohn, can you add a [changelog](https://github.com/Automattic/pocket-casts-android/blob/main/CHANGELOG.md)  entry something like: Fixed skip backwards settings?

I'll merge the PR once this is done.

VGJohn:
Done 😃

PR #426, opened 2022-10-20
migrate `FolderLocation` to Kotlin and make its properties immutable

wiryadev:
# Description
I migrated the FolderLocation to Kotlin's data class and make its properties Immutable since there is no usage of the setter function. The build and test is pass.


mchowning (approved):
This is such a nice improvement and drops 20 lines of code! 😄  Thanks for the contribution @wiryadev !

PR #427, opened 2022-10-20
migrate `CustomTabsBroadcastReceiver` to Kotlin and create string res…

wiryadev:
# Description

migrate `CustomTabsBroadcastReceiver` to Kotlin and create string resource for the Toast message to open translation possibility in the future


mchowning:
Thanks for this @wiryadev ! 🙇  It's great to get the last remaining bits of java converted over to Kotlin! 😄 

It looks to me like we're actually not using the `CustomTabsBroadcastReceiver` at all anymore, so it could just be removed. What do you think? 

cc: @geekygecko since you're more familiar with this very old code than I am and might know of something I'm overlooking.

geekygecko:
@mchowning I found the old commit that is using it. 

```
    private PendingIntent createPendingCopyLinkIntent() {
        Intent intent = new Intent(context.getApplicationContext(), CustomTabsBroadcastReceiver.class);
        return PendingIntent.getBroadcast(context.getApplicationContext(), 0, intent, PendingIntent.FLAG_UPDATE_CURRENT);
    }
```

I think we are safe to remove it. I will raise a pull request. 

@wiryadev thanks for converting it anyway. It's awesome to get rid of the old Java code. We should get rid of XML views next!

PR #432, opened 2022-10-20
Fix for different background in Compose theme

### LINKED ISSUE #251: Background colors
It seems like some of the background colors aren't quite right. This might have be introduced either #235 or #215. Most of the more recent designs have tried to blend the toolbar with the list background color.

| Production | Now |
| --- | --- |
| Screenshot_20220819_120809 | Screenshot_20220819_120959 |
| Screenshot_20220819_121122 | Screenshot_20220819_121113 |

VGJohn:
Background color in the compose theme was slightly different to the background color used in non-compose screens

Fixes #251 

Attached matching screenshots from the ones seen in the linked issue with the correct background color:

Screenshot_20221020_171321
Screenshot_20221020_171301


geekygecko (approved):

PR #440, opened 2022-10-21
migrate `TimberDebugTree` to Kotlin

wiryadev:
# Description

Migration of `TImberDebugTree` to Kotlin.


geekygecko (approved):
Thanks for tackling more Java code!

PR #441, opened 2022-10-21
migrate `StorageOptions` to Kotlin

wiryadev:
# Description

migrate `StorageOptions` class to Kotlin


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/file/StorageOptions.kt
```diff
@@ -0,0 +1,94 @@
...
+        var externalSDCardCount = 1
+        val firstMountPoint = confirmedMountPoints[0]
+        confirmedMountPoints.removeAt(0)
+
+        val resources = context.resources
+        // the first mount point is different to the rest
+        if (!Environment.isExternalStorageRemovable()
+            || Environment.isExternalStorageEmulated()
+        ) {
+            folderLocations!!.add(
```
ashiagr, line 64: Nitpick: We might want to avoid using the double-bang(!!) operator inside `determineFolderLocations(...)`. You can add a `requireNotNull(...)` with an explicit message that folder location cannot be null. 

This is just a personal preference and I don't feel strongly about it.

ashiagr (approved):
Thank you so much for your contribution! 
I just left a single nitpick. I'm approving the PR but not merging yet it to let you address it if you want to. 

It's really great to be getting closer to 100% Kotlin 🥳

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/file/StorageOptions.kt
wiryadev: Hi, i see that explicit message is helpful. Pushed the new one

PR #445, opened 2022-10-21
Update feature request issue template label

mchowning:
I updated the label's name from "feature" to "feature request" to make it more descriptive in GitHub, and that seems broke this template so that it no longer adds the "feature"/"feature request" label. I assume this change will fix that.


geekygecko (approved):
🤞

Issue #114, opened 2022-07-14
Help & Feedback: Get in touch: Buttons overlap out of the container


geekygecko:
When looking at the Get in touch area of the app, the second button is overlapping out of the container:

<img  width="300" src="https://user-images.githubusercontent.com/308331/179106253-d700cd61-0db0-415f-ae43-1f677ade699f.png" />

I am testing on a Samsung Galaxy S21 FE

<img width="33%" alt="image" src="https://user-images.githubusercontent.com/16253818/178987669-9f53083d-59a6-4420-9a4b-35c63988a30b.png"><img width="33%" alt="image" src="https://user-images.githubusercontent.com/16253818/178987709-d1014790-050f-434e-9d9f-c90d8b26cf2c.png"><img width="33%" alt="image" src="https://user-images.githubusercontent.com/16253818/178987763-48d80c76-5861-40c6-b8a0-85dcd61d6839.png">

neeldoshii:
I don't find this issue. My device is at 200% zoom.

I guess this issue is old and have might been fixed. Please close this issue if its fixed then.

<img src="https://github.com/Automattic/pocket-casts-android/assets/60827173/e36eea76-547b-40df-af24-10ceccbd9032" width="30%" height ="30%">


Thanks

mchowning:
I'm not seeing this issue anymore either. Looks like it was fixed by #446 most likely. Thanks for checking this @neeldoshii ! 🙇

PR #446, opened 2022-10-21
Sets textZoom to 100 for webview

josepatinob:
# Description

The webview in the Help & Feedback section was affected by the user's display size settings, which was causing the buttons in the Get in touch modal to overlap. Setting the textZoom to 100 will alleviate the problem on that particular webview. 

Fixes # (Help & Feedback: Get in touch: Buttons overlap out of the container #114)

# Checklist

Screenshot (Oct 21, 2022 3 41 18 PM)


geekygecko (approved):
Thanks for fixing this! I didn't know that setting existed.

PR #472, opened 2022-10-24
Trust user certificates on debug builds

mchowning:
## Description

Makes it easier to use something like Charles proxy to inspect our https calls on debug builds.

I followed the instructions from these two sources:
- [Android documationation](https://developer.android.com/training/articles/security-config#TrustingDebugCa)
- [Charles Proxy documentation](https://www.charlesproxy.com/documentation/using-charles/ssl-certificates/)

When testing this on release builds I did get [this crash](https://sentry.io/organizations/a8c/issues/3693580607/?project=6711064&referrer=release-issue-stream) a couple of times, but then it went away and I couldn't recreate it. Not sure if I was just doing something weird with my build or what was going on there.

##### Debug Build
1. Follow instructions for installing Charles certificate on your device
2. Set your device up to proxy through Charles (as part of this, make sure you install the Charles SSL cert on your device)
4. Turn on SSL proxying for one of the app's https calls
5. Have the app make that call
6. Confirm that you can inspect the call through Charles

##### Release Build
1. Follow steps 1 through 4 above
2. Observe that the app crashes because we don't trust the Charles certificate on release builds (I was a bit surprised to see the app crash here, I thought we just wouldn't be able to inspect the call. I think I'm doing everything right though).
3. Set your device up to _not_ proxy through Charles
4. Have the app make an https call
5. Confirm that you cannot inspect the call in Charles. _I had one release build that crashed in this scenario, but I can't recreate it, so I think I just did something weird with that build. Definitely worth keeping an eye out for though._

# Checklist

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -38,6 +38,7 @@
             android:supportsRtl="false"
             tools:replace="android:supportsRtl"
             android:theme="@style/ThemeDark"
+            android:networkSecurityConfig="@xml/network_security_config"
             tools:ignore="UnusedAttribute">
 
         <meta-data android:name="com.google.android.gms.car.application"
--- a/app/src/main/res/xml/network_security_config.xml
+++ b/app/src/main/res/xml/network_security_config.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<network-security-config>
+<debug-overrides>
+    <trust-anchors>
+        <!-- Trust user added CAs while debuggable only -->
+        <certificates src="user" />
+    </trust-anchors>
+</debug-overrides>
+</network-security-config>


mchowning:
> I'm sure that I am worrying about nothing

I've seen this change on quite a few other apps and I was a bit worried about this change too, so I definitely understand. There's nothing worse than making a change solely to ease development and then to find out it's causing problems for users. I'm pretty confident that this is safe, though.

> ...this doesn't increase the certificate checking in production does it? I know producer episode hosting isn't always the best so it would be good to keep that as flexible as possible.

I don't think it should at all. As [the documentation for `debug-overrides`](https://developer.android.com/training/articles/security-config#debug-overrides) notes, this will be completely ignored in release builds:

> Overrides to be applied when [android:debuggable](https://developer.android.com/guide/topics/manifest/application-element#debug) is "true", which is normally the case for non-release builds generated by IDEs and build tools. Trust anchors specified in debug-overrides are added to all other configurations, and certificate pinning is not performed when the server's certificate chain uses one of these debug-only trust anchors. If [android:debuggable](https://developer.android.com/guide/topics/manifest/application-element#debug) is "false", then this section is completely ignored.

PR #476, opened 2022-10-24
improve nullability of OpmlExporter when sharing files

tobiasschuerg:
# Description

Improve nullability of OpmlExporter by using the already null-checked file instead of global var when sharing files - this is in preparation of converting `FileUtil` to Kotlin.

Fixes # (issue)

# Checklist

- [x] Should this change be included in the release notes? If yes, please add a line in CHANGELOG.md
- [x] Are all the strings localized?
- [x] Could you have written any new tests?


geekygecko (approved):
Very nice! I didn't think to add the @NonNull annotation before converting code to Kotlin. Thanks for the change.

PR #477, opened 2022-10-24
Use same date format for next episode and previous

TylerCarberry:
# Description
Use the same date format for the predicted episode date and the dates in the list. I'm in the US, I think that's why the dates are in the different order

Before & After
Before  After


modules/services/localization/src/main/java/au/com/shiftyjelly/pocketcasts/localization/helper/RelativeDateFormatter.kt
```diff
@@ -43,7 +43,7 @@ class RelativeDateFormatter(val context: Context) {
             }
         } else if (Locale.getDefault().language == "en") {
             // if using an old Android version but using English then add Today and Yesterday
-            val alternativeString = formatCloseDaysOld(calendar, resources)
+            val alternativeString = formatCloseDaysOld(calendar, context.resources)
```
TylerCarberry, line 46: I also removed this `resources` param since there's already a field for `context`

geekygecko (commented):
Thanks for looking making the next episode much nicer to read and cleaning up the context code. I think there is an issue with our RelativeDateFormatter not supporting future dates as November 6 isn't this Sunday but the next. Did you want to take a look at this? If not I'm happy to.

TylerCarberry:
Fixed! Just double check this is the format you want to show. I think it should be clear from the context whether "sunday" means last sunday (episode release date) or next sunday (prediction date)

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/Podcast.kt
```diff
@@ -204,15 +205,17 @@ data class Podcast(
...
         val now = System.currentTimeMillis()
         val sevenDaysAgo = now - sevenDaysInMs
-        val sevenDaysFuture = now + sevenDaysInMs
+
+        val resources = context.resources
         return when {
             expectedTime < sevenDaysAgo -> null
             DateUtils.isToday(expectedTime) -> resources.getString(LR.string.podcast_next_episode_today)
             DateUtils.isToday(expectedTime - DateUtils.DAY_IN_MILLIS) -> resources.getString(LR.string.podcast_next_episode_tomorrow)
             expectedTime in sevenDaysAgo..now -> resources.getString(LR.string.podcast_next_episode_any_day_now)
```
ashiagr, line 214: Since we show "any day now" for a date between now and 7 days ago, 
`expectedTime in sevenDaysAgo..now -> resources.getString(LR.string.podcast_next_episode_any_day_now)`

we might want to reduce time period from
`calendar.after(sevenDaysAgo) && calendar.before(sevenDaysFromNow) -> DateUtils.FORMAT_SHOW_WEEKDAY`
to 
`calendar.after(now) && calendar.before(sevenDaysFromNow) -> DateUtils.FORMAT_SHOW_WEEKDAY`
inside [RelativeDateFormatter](https://github.com/Automattic/pocket-casts-android/pull/477/commits/4328b37a725852c18d7dca5af12986248b150a95#diff-1715dbdb50e2293050b7f90f7e612032209eab4e476fd622aa9677bb9f37df6cR50). 

Also, we might want to match [iOS date formatting](https://github.com/Automattic/pocket-casts-ios/blob/trunk/podcasts/Podcast+Formatting.swift#L55-L56) and instead of adding seven days to now, we can just add six days to be consistent in both platforms.

modules/services/localization/src/main/java/au/com/shiftyjelly/pocketcasts/localization/helper/RelativeDateFormatter.kt
```diff
@@ -22,18 +22,14 @@ import java.util.Locale
 class RelativeDateFormatter(val context: Context) {
 
     private val sevenDaysAgo: Calendar = Calendar.getInstance().apply { add(Calendar.DATE, -7) }
+    private val sevenDaysFromNow: Calendar = Calendar.getInstance().apply { add(Calendar.DATE, 7) }
```
ashiagr, line 25: Can we also update the comment above to say?

`* "Monday" if less than a week from now`

ashiagr (approved):
Thank you so much for the changes, @TylerCarberry 🙏 and really sorry for taking so long to review. 

I just left a couple of super minor suggestions. Other than that, date formatting looks much better now! 🎉  

Also, it'll be awesome to add a [CHANGELOG](https://github.com/Automattic/pocket-casts-android/blob/main/CHANGELOG.md) entry to 7.28 for the fix.

ashiagr (approved):
Thank you! 🙏

PR #481, opened 2022-10-25
Update PR template

mchowning:
# Description

The main motivation for this PR was to add a section for testing instructions to the PR template. I think that this makes reviewing a PR much easier, particularly when it comes from an open source contributor where there may be less context around the change.

This is just a proposal. I'm definitely open to modifying this or not merging these changes.

--------------

# What this template looks like ⬇️ 

## Description

Fixes # 

## Checklist
 
- [ ] with different themes

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,16 +1,25 @@
-# Description
+## Description
+<!-- Please include a summary of what this PR is changing and why these changes are needed. -->
 
-Please include a summary of the changes and the related issue.
+Fixes # <!-- issue number, if applicable -->
 
-Fixes # (issue)
+## Testing Instructions
+<!-- Please include step by step instructions on how to test this PR. -->
+<!-- 1. Tap on the Filters tab -->
+<!-- 2. Tap on a filter -->
+<!-- 3. etc. -->
 
-# Checklist
+## Screenshots or Screencast 
+<!-- if applicable -->
 
-- [ ] Should this change be included in the release notes? If yes, please add a line in CHANGELOG.md
-- [ ] Have you tested in landscape?
-- [ ] Have you tested accessibility with TalkBack?
-- [ ] Have you tested in different themes?
-- [ ] Does the change work with a large display font?
-- [ ] Are all the strings localized?
-- [ ] Could you have written any new tests?
-- [ ] Did you include Compose previews with any components?
+## Checklist
+- [ ] If this is a user-facing change, I have added an entry in CHANGELOG.md
+- [ ] I have considered whether it makes sense to add tests for my changes
+- [ ] Any jetpack compose components I added or changed are covered by compose previews
+ 
+#### I have tested any UI changes...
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -1,16 +1,25 @@
...
-- [ ] Have you tested in landscape?
-- [ ] Have you tested accessibility with TalkBack?
-- [ ] Have you tested in different themes?
-- [ ] Does the change work with a large display font?
-- [ ] Are all the strings localized?
-- [ ] Could you have written any new tests?
-- [ ] Did you include Compose previews with any components?
\ No newline at end of file
+## Checklist
+- [ ] If this is a user-facing change, I have added an entry in CHANGELOG.md
```
mchowning, line 16: I wanted to include information about when a Changelog entry is needed.

.github/pull_request_template.md
```diff
@@ -1,16 +1,25 @@
...
+- [ ] I have considered whether it makes sense to add tests for my changes
+- [ ] Any jetpack compose components I added or changed are covered by compose previews
+ 
+#### I have tested any UI changes...
+<!-- If this PR does not contain UI changes, ignore these items -->
```
mchowning, line 21: I've noticed some contributors seem unsure about how to handle checklist items that aren't applicable. This is an attempt to reduce that confusion, but there may be a better way to handle this.

.github/pull_request_template.md
```diff
@@ -1,16 +1,25 @@
...
-# Checklist
+## Screenshots or Screencast 
+<!-- if applicable -->
 
-- [ ] Should this change be included in the release notes? If yes, please add a line in CHANGELOG.md
-- [ ] Have you tested in landscape?
-- [ ] Have you tested accessibility with TalkBack?
-- [ ] Have you tested in different themes?
-- [ ] Does the change work with a large display font?
-- [ ] Are all the strings localized?
```
mchowning, line 14: I dropped the localized strings checklist items. I'm not sure the benefit from this is worth the noise of another checklist item since this will almost certainly be caught during PR review and it's easy to fix. Let me know if you think we should keep it.

geekygecko: Instead of removing "Are all the strings localized?" maybe it could say something like:
```
- [ ] Are all the strings which need to be localized in modules/services/localization/src/main/res/values/strings.xml?
```
Or something more friendly. I [noticed a PR](https://github.com/Automattic/pocket-casts-android/pull/427/files) with a string to be localized in another module.

geekygecko (approved):

.github/pull_request_template.md
mchowning: Adding the location where the localized strings should go is a good idea. Added in [bdf4159](https://github.com/Automattic/pocket-casts-android/pull/481/commits/bdf415920b13db1efbddd94f64a957714fdf2f0a).

PR #484, opened 2022-10-25
Fix Plus features not working after cancelled but not expired.

geekygecko:
When a user has paid for Plus the sponsors are removed from the discover section. There was an issue that these sponsors were shown after a user had cancelled their paid Plus subscription but it hadn't expired. The change fixes this issue.

We could start checking the expiry date but then we would need to rely on the device time. A safe way seems to be to let the Plus expire the next time the server is called. 

I had trouble testing this as I could buy Plus from my development builds, so instead I wrote some tests.

## Test steps
1. Subscribed to monthly Plus 
1. Confirmed paid placements were removed in Discover
1. Cancelled Plus subscription
1. The Paid Placements had reappeared in Discover


mchowning (approved):
Tested and this works as expected—canceled subscriptions show up as Plus paid if they haven't expired now. I really like the tests you added for this!

PR #494, opened 2022-10-26
Fix podcast filter title and hide filter option if no filters available

### LINKED ISSUE #49: Podcast Settings -> Included in filters page changes required
There's a bug on this page where the heading says "Choose podcasts", but it should say "Select filters"

Android:

vs how it should look from iOS:

Additionally we should hide the "Included in filters" button if there are no filters it can be included in (eg: there aren't any that are currently not All Podcasts") otherwise you get a screen like this. Which doesn't really make a lot of sense. What can a user do here? What does it mean? We already hide this option on iOS.

VGJohn:
## Description
Fix for wrong title on the filter select screen "Choose podcasts" -> "Select filters" (match iOS design)
 - Added separate layout since the filter select screen was using the podcast select layout resulting in the incorrect title

Hide filters option in podcast settings if no filters available to assign the podcast to (match iOS behaviour)

Fixes #49 

## Screenshots or Screencast 
Screenshot_20221026_023355
Screenshot_20221026_023329
Screenshot_20221026_023258

## Checklist
 
- [x] with different themes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
         ([#482](https://github.com/Automattic/pocket-casts-android/pull/482)).
     *   Redesign of the fullscreen player share option
         ([#451](https://github.com/Automattic/pocket-casts-android/pull/451)).
+    *   Fix select filters title and hide filter option when applicable in the podcast settings
+        ([#494](https://github.com/Automattic/pocket-casts-android/pull/494)).
 *   Bug Fixes:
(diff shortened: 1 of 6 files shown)


ashiagr (approved):
Great fix!
I now know what to select on the screen 😄 

Thank you once again 🙇‍♀️

ashiagr:
👋 @VGJohn 

I'll merge the PR once the (changelog) conflict is resolved.

VGJohn:
@ashiagr I've fix that conflict in the changelog 😄

PR #495, opened 2022-10-26
End of Year: Generate stories with dummy data + support dynamic lengths

ashiagr:
| 📘 Project: #410 |
|:---:|

iOS PR: https://github.com/Automattic/pocket-casts-ios/pull/397

Creates "dummy" stories based on randomly generated podcasts and episode. This PR also adds support for dynamic story lengths.

Same as the previous tasks, the designs here are not final and are mainly placeholders.

## Testing Instructions

2. Run the app
3. When the prompt appears tap "View my 2022"
4. ✅ A list of podcasts should appear ¶
5. Wait for 2 seconds
6. ✅ Notice that the next story starts displaying a random episode as the longest played episode
7. Wait for 3 seconds (dynamic time set for the second story)
8. ✅ Notice that stories progress ends
9. Play around with previous/ next taps
10. ✅ Notice that stories are skipped to correct positions

¶ If you don't see any podcasts, tap some podcasts on Discover or login.

## Screenshots or Screencast 

https://user-images.githubusercontent.com/1405144/197961191-f5a20174-3301-4a6a-b1d5-f92c6ecfae21.mp4

## Checklist
N/A as only fake story views are added

- with different themes

### DIFF
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/PodcastDao.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/PodcastDao.kt
@@ -344,4 +344,7 @@ abstract class PodcastDao {
 
     @Query("UPDATE podcasts SET start_from = :startFromSecs, skip_last = :skipLastSecs, folder_uuid = :folderUuid, sort_order = :sortPosition, added_date = :addedDate WHERE uuid = :uuid")
     abstract suspend fun updateSyncData(uuid: String, startFromSecs: Int, skipLastSecs: Int, folderUuid: String?, sortPosition: Int, addedDate: Date)
+
+    @Query("SELECT * FROM podcasts LIMIT 5")
+    abstract fun findRandomPodcasts(): Flow<List<Podcast>>
 }
--- a/modules/features/endofyear/build.gradle
+++ b/modules/features/endofyear/build.gradle
@@ -13,6 +13,8 @@ dependencies {
     // services
     implementation project(':modules:services:compose')
     implementation project(':modules:services:localization')
+    implementation project(':modules:services:model')
+    implementation project(':modules:services:repositories')
     implementation project(':modules:services:ui')
     implementation project(':modules:services:utils')
     implementation project(':modules:services:views')
--- a/modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/SegmentedProgressIndicator.kt
+++ b/modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/SegmentedProgressIndicator.kt
@@ -12,11 +12,11 @@ import androidx.compose.ui.geometry.Offset
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.drawscope.DrawScope
 import androidx.compose.ui.unit.dp
+import au.com.shiftyjelly.pocketcasts.endofyear.StoriesViewModel.State.Loaded.SegmentsData
 
 private val StrokeWidth = 2.dp
-private val GapWidth = 8.dp
 private val SegmentHeight = StrokeWidth
-private const val IndicatorBackgroundOpacity = 0.24f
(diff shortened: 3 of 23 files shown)


modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/PodcastDao.kt
```diff
@@ -344,4 +344,7 @@ abstract class PodcastDao {
 
     @Query("UPDATE podcasts SET start_from = :startFromSecs, skip_last = :skipLastSecs, folder_uuid = :folderUuid, sort_order = :sortPosition, added_date = :addedDate WHERE uuid = :uuid")
     abstract suspend fun updateSyncData(uuid: String, startFromSecs: Int, skipLastSecs: Int, folderUuid: String?, sortPosition: Int, addedDate: Date)
+
+    @Query("SELECT * FROM podcasts LIMIT 5")
+    abstract fun findRandomPodcasts(): Flow<List<Podcast>>
```
mchowning, line 349: Are these database "random" methods just for testing purposes, or is it really supposed to return podcasts randomly?

mchowning (approved):
This is coming along nicely. 

The only issue I ran into is that the list items in the list of podcasts is capturing the tap events, which can make it tricky to go to the previous/next story. This can be addressed in a follow-up PR if you'd like, though. I'll hold off on merging in case you want to do that here and for the flaky tests you wanted to look into.


https://user-images.githubusercontent.com/4656348/198133749-8c23fdde-e692-4515-b78d-e299203b352a.mp4

mchowning:
Two additional things I noticed when testing your follow-up PR:

On one occasion when I was playing around with landscape, the story view got messed up where it had "extra" stories, and the gaps in the top-of-the-screen indicator would change position (and not line up with the actual story changes consistently). I know that description is hard to follow, you've got to watch the video because it's pretty weird.

I reproduced this 4 times, but it was _very_ hard to reproduce and I don't know how I reproduced it other than that it seemed to occur when I was testing landscape and switching orientations.

https://user-images.githubusercontent.com/4656348/198142203-95674b51-5899-45d5-a699-2ab074d176c8.mp4

I also noticed that the swipe down motion to close the end of year screen is too sensitive to me. Touching the screen with the slightest downward movement seems to swipe the screen away for me. Certianly not something that has to be addressed in this PR.

https://user-images.githubusercontent.com/4656348/198141663-d6adb206-f866-45cd-aca3-24e16c3e316c.mp4

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/PodcastDao.kt
ashiagr: These are just for testing purposes and will be converted to return actual top podcasts and the longest episode.

ashiagr:
Thanks for testing it, @mchowning! 

> The only issue I ran into is that the list items in the list of podcasts is capturing the tap events, which can make it tricky to go to the previous/next story. This can be addressed in a follow-up PR if you'd like, though.

> I also noticed that the swipe down motion to close the end of year screen is too sensitive to me. Touching the screen with the slightest downward movement seems to swipe the screen away for me. Certianly not something that has to be addressed in this PR.

Yep, these were not in the scope of this PR, I'll address them in future PRs. 

> On a few occasions when I was playing around with landscape, the story view got messed up where it had "extra" stories, and the gaps in the top-of-the-screen indicator would change position (and not line up with the actual story changes consistently). I know that description is hard to follow, you've got to watch the video because it's pretty weird.

> I reproduced this 4 times, but it was very hard to reproduce and I don't know how I reproduced it other than that it seemed to occur when I was testing landscape and switching orientations. FWIW, once I got the phone into this weird state, it would display that state in both landscape and vertical orientation.

I wasn't able to reproduce it at all. Tried switching between portrait and landscape as well. Most probably we'll lock the screen to portrait just like in iOS. So I'll hold off going into this rabbit hole and I'll also discuss with you how exactly you got into this state.

Issue #471, opened 2022-10-24
Improve Tasker Integration UI


mchowning:
We've recently added integration with Tasker thanks to [an open source contribution](https://github.com/Automattic/pocket-casts-android/pull/431/). 🎉  The UI doe not fit well with the rest of Pocket Casts though. Let's keep iterating on the UI of this screen to make it more consistent with the rest of our app.

<img width="350px" src="https://user-images.githubusercontent.com/4656348/197546065-6d964346-7a4c-4c6d-91fe-f378bf2fc6c6.png" />

### Some possible improvements:
- [x] The screen could use some margins around the edge ([PR](https://github.com/Automattic/pocket-casts-android/pull/475))
- [x] It would be nice to use icons to help differentiate the different options. Ideally, the filter options would look very similar to how they look on the filters tab in the app ([PR](https://github.com/Automattic/pocket-casts-android/pull/509))
- [ ] The Dropdown component doesn't indicate whether it can be scrolled to display more items ([relevant discussion on StackOverflow](https://stackoverflow.com/questions/69782653/jetpack-compose-possible-to-set-dropdownmenu-height-to-display-next-item)). Would be good to find a better way to present the list of options.
- [ ] Upon first opening the screen, the user just sees a text box with an icon or two and an ok button. Maybe there is a way we could make the relevant options more obvious to the user? 🤔 
- [ ] Some of the fields accept Strings when only a number is a valid input. It would be great to require the correct input at the UI level. That would also allow us to pass around fewer raw Strings.

Please go ahead and suggest other improvements. These are just a few that immediately came to mind for me.

jacobherrington:
Hey @mchowning I'm new to mobile dev, pretty sure I could figure this out though. Do you mind if I run with it for a few days?

mchowning:
That would be great @jacobherrington ! 😄  Let me just ping @joaomgcd  in case he had any thoughts here since this is building on top of his work.

mchowning:
> Just a reminder though, the fields can't be number-only fields because...

👍  I crossed that one off

joaomgcd:
👍

jacobherrington:
Haha yes, I got distracted with job hunting - my few days have passed if anyone wants to make more progress. I am not knowledgeable around these things. :)

PR #500, opened 2022-10-26
New commands in Tasker Action "Control Playback": "Skip Forward" and "Skip Back"

joaomgcd:
## Description
Adds ability for the **Control Playback** action in Tasker to skip the number of provided seconds either forward or backward

## Testing Instructions
- Open Tasker
- Add **Pocket Casts Control Playback** action
- Set **Playback Command** to **Skip Forward**
- Set **Time To Skip** to **600**
- Press the **OK** button
- Start playing a podcast in Pocket Casts
- Run the Tasker action and check that it skips 10 minutes forward

## Screenshots or Screencast 
screencapture (16)

## Checklist
 
- [x] with different themes


mchowning (commented):
This works well. Thanks for the contribution @joaomgcd ! 🙇 

There are some formatting issues that are going go fail on CI. You can run `./gradlew spotlessCheck` to see the violations or `./gradlew spotlessApply` to automatically fix them. Also, if you want, you can install our pre-commit hook by running `./gradlew installGitHooks` which runs spotless on every commit. I find the pre-commit hook helpful, but manually running spotless works just fine too.

Long-term I think it would be good if we could tighten up the types in this module a bit so that we're passing around fewer raw Strings (for example, requiring that a skip-to-time be a number at the point of the user's input instead of passing it around as a String). You don't need to worry about that in this PR, but just something to think about as we move forward. I went ahead and added a note about this possible improvement to [the UI improvement issue description](https://github.com/Automattic/pocket-casts-android/issues/471).

modules/features/taskerplugin/src/main/java/au/com/shiftyjelly/pocketcasts/taskerplugin/controlplayback/InputControlPlayback.kt
```diff
@@ -11,13 +11,16 @@ import com.joaomgcd.taskerpluginlibrary.input.TaskerInputRoot
 class InputControlPlayback @JvmOverloads constructor(
     @field:TaskerInputField("command") var command: String? = null,
     @field:TaskerInputField("skipToChapter") var chapterToSkipTo: String? = null,
-    @field:TaskerInputField("timeToSkipToSeconds") var timeToSkipToSeconds: String? = null
+    @field:TaskerInputField("timeToSkipToSeconds") var timeToSkipToSeconds: String? = null,
+    @field:TaskerInputField("timeToSkipSeconds") var timeToSkipSeconds: String? = null
```
mchowning, line 15: I wonder if changing these to something shorter like `skipSeconds` and `skipToSeconds` might make it harder to mix these up. The fact that the only difference between them currently is that one has a second "To" in the name makes them look very similar (to my eyes at least). I don't feel strongly though, feel free to leave it as-is if you prefer.

mchowning:
Let's try merging the latest from `main` into this branch after [this PR](https://github.com/Automattic/pocket-casts-android/pull/502) is merged. I think that may resolve the build issues on CI.

joaomgcd:
Awesome! :) Thank you!

PR #501, opened 2022-10-26
Anonymous Bump Stats (part 1)

mchowning:
## Description

This adds the ability to anonymously bump a stat. We need this for purposes of allowing users to opt-out of analytics while still allowing us to get the data we need to be able to provide to paid sponsors. 

The only data that is provided with these events are:
1. An event name
2. The time of the event
3. Any specific properties about the event (i.e., the id of the list that the user interacted with in the discover feed)

The second part of this work will involve providing anonymous daily unique stats, which I will address in a follow-up PR. This sets us up to drop Firebase as an analytics provider, and allows us to do a better job of restricting the data we collect.

## Testing Instructions
1. Load the app
2. Go to the discover screen and perform the the following actions:
    1. Subscribe to a podcast
    2. Tap on "Show All" to see all of a list
    6. Play a featured episode from the Discover Screen
    7. Tap on a podcast on the Discover screen
    8. Tap. on an episode on the Discover screen
3. Put the app in the background and reload it
4. Observe that the relevant events are sent to Tracks

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/services/analytics/src/main/java/au/com/shiftyjelly/pocketcasts/analytics/AnalyticsTracker.kt
+++ b/modules/services/analytics/src/main/java/au/com/shiftyjelly/pocketcasts/analytics/AnalyticsTracker.kt
@@ -11,8 +11,8 @@ object AnalyticsTracker {
         trackers.forEach { it.clearAllData() }
     }
 
-    fun registerTracker(tracker: Tracker?) {
-        tracker?.let { trackers.add(tracker) }
+    fun register(vararg trackers: Tracker) {
+        this.trackers.addAll(trackers)
     }
 
     fun track(event: AnalyticsEvent, properties: Map<String, Any> = emptyMap()) {
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/AnonymousBumpStat.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/AnonymousBumpStat.kt
@@ -1,0 +1,114 @@
+package au.com.shiftyjelly.pocketcasts.models.entity
+
+import androidx.room.ColumnInfo
+import androidx.room.Entity
+import androidx.room.TypeConverter
+import com.squareup.moshi.FromJson
+import com.squareup.moshi.Moshi
+import com.squareup.moshi.ToJson
+import com.squareup.moshi.Types
+import timber.log.Timber
+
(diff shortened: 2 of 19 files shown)


modules/services/analytics/src/main/java/au/com/shiftyjelly/pocketcasts/analytics/AnalyticsTracker.kt
```diff
@@ -11,8 +11,8 @@ object AnalyticsTracker {
         trackers.forEach { it.clearAllData() }
```
mchowning, line 11: Should we be calling `clearAllData` every time we init the `AnalyticsTracker` (which occurs each time the app starts)? This doesn't effect this PR (because the BumpStatsTracker doesn't do anything when `clearAllData` gets called), but this didn't feel right to me and I was wondering if maybe I was missing something.

ashiagr: Good catch! I think this was added when we regenerated anonID every 30 minutes and on application init. We should remove this now.

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/AnonymousBumpStat.kt
```diff
@@ -0,0 +1,114 @@
...
+import com.squareup.moshi.Moshi
+import com.squareup.moshi.ToJson
+import com.squareup.moshi.Types
+import timber.log.Timber
+
+@Entity(tableName = "bump_stats", primaryKeys = ["name", "event_time", "custom_event_props"])
+data class AnonymousBumpStat(
+    @ColumnInfo(name = "name") var name: String,
+    @ColumnInfo(name = "event_time") var eventTime: Long = System.currentTimeMillis(),
+    @ColumnInfo(name = "custom_event_props") var customEventProps: Map<String, Any> = emptyMap()
```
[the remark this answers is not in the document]
ashiagr: I can't think of any suggestions right now. Also, this doesn't seem to be causing any issues. So feel free to get rid of this warning later when you can, it doesn't need to be addressed right now.

ashiagr (approved):
Tremendous job, @mchowning! I tested the changes and confirm that events appended with `_bump ` are tracked without any user-specific data.

Code changes look good too. 

Looking forward to the next PR. 🙌

PR #512, opened 2022-10-29
PC-I182: Change Text When Upgrading Account From 'Account Created' to 'Account Upgraded'

### LINKED ISSUE #182: After upgrading an account to Plus, the confirmation screen says "Account Created"
When upgrading an already existing free account to Plus, the confirmation screen says "Account Created". It feels like "Account Upgraded" would be more appropriate since this could be an extremely old account.

Would also be nice to fix the extra space above the close button at the top of the screen.

owenlejeune:
## Description
Change text when upgrading from free to premium from 'Account Created' to 'Account Upgraded'

Fixes #182 

## Testing Instructions
1. Upgrade a free account to premium

## Screenshots or Screencast 
Screenshot_20221029-084941

### DIFF
--- a/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/CreateDoneFragment.kt
+++ b/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/CreateDoneFragment.kt
@@ -100,15 +100,17 @@ class CreateDoneFragment : BaseFragment() {
 
     private fun processCreated(subscriptionType: SubscriptionType?) {
         var resourceId = 0
-        val title = getString(LR.string.profile_account_created)
+        var title = ""
         var detail = ""
         if (subscriptionType != null) {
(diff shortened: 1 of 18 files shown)


mchowning (commented):
This looks good and works well. Thanks for the contribution @owenlejeune ! 🙇 

The only question I have is about where the translations came from. We [use glotpress to get human-translated versions of our strings](https://translate.wordpress.com/projects/pocket-casts/android) imported into Pocket Casts. Did you get these translations from something like Google Translate and add them manually? If so, I really appreciate you going to that effort, but let's remove all but the English version of the String from this PR and let GlotPress handle the translations for us.

mchowning:
> My bad! I wasn't sure what the process for adding translations was so I added them in myself

No worries, we need to document this better. Thanks for the update.

mchowning (approved):

Issue #52, opened 2022-07-06
Tapping the play button flickers back to a pause icon


geekygecko:
The episode play button changes to a play icon, back to a pause icon, and then quickly back to a play icon again.

**Steps to repeat**
1. Open a podcast page
2. Tap an episode play button

https://user-images.githubusercontent.com/308331/177546412-5a1c30b2-4895-4bd4-b7e8-3e0e263980fc.mov

ezebecke:
https://user-images.githubusercontent.com/24919469/194363026-a373baae-b4c4-47fb-9b9c-8edd31a492f4.mp4

Same issue but tapping play on the mini player. 

Reported on 5558791-zd

AdamGEmerson:
I don't seem to be able to reproduce this issue on my device.  Is it possible that it has been resolved by the changes in #523?

ashiagr:
This does seem to be resolved by the changes in https://github.com/Automattic/pocket-casts-android/issues/523.

I'll mark it as closed.

PR #513, opened 2022-10-29
PC-I25: Prevent Mini Player From Animating When Changing Theme

### LINKED ISSUE #25: When toggling themes the miniplayer animates back into place
If the miniplayer is present when switching themes, it briefly disappears and then animates back into place on the screen. We should update this so it just remains in place.

https://user-images.githubusercontent.com/4656348/176779590-7f85a656-5bba-401f-8c97-8c6f28ff0cca.mp4

owenlejeune:
## Description
Store the state of the mini player when recreating the MainActivity so that the mini player animation can be disabled when changing the theme.

Fixes #25 

## Testing Instructions
1. Navigate to Appearance settings
2. Select a theme different from the currently applied theme
3. Mini player should not animate

## Screenshots or Screencast 
https://user-images.githubusercontent.com/20568035/198835613-b27c9045-0dd0-426f-bc44-8c3008ef6778.mp4

### DIFF
--- a/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
+++ b/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
@@ -120,6 +120,7 @@ import au.com.shiftyjelly.pocketcasts.views.R as VR
 import com.google.android.material.R as MR
 
 private const val SAVEDSTATE_PLAYER_OPEN = "player_open"
+private const val SAVEDSTATE_MINIPLAYER_SHOWN = "miniplayer_shown"
 
 @AndroidEntryPoint
 class MainActivity :
@@ -219,7 +220,10 @@ class MainActivity :
             activity = this
         )
 
-        setupPlayerViews()
+        val showMiniPlayerImmediately = savedInstanceState?.getBoolean(SAVEDSTATE_MINIPLAYER_SHOWN, false) ?: false
+        binding.playerBottomSheet.isVisible = showMiniPlayerImmediately
+
+        setupPlayerViews(showMiniPlayerImmediately)
 
         if (savedInstanceState != null) {
             val videoComingToPortrait =
@@ -313,6 +317,7 @@ class MainActivity :
     override fun onSaveInstanceState(outState: Bundle) {
         super.onSaveInstanceState(outState)
         outState.putBoolean(SAVEDSTATE_PLAYER_OPEN, binding.playerBottomSheet.isPlayerOpen)
+        outState.putBoolean(SAVEDSTATE_MINIPLAYER_SHOWN, binding.playerBottomSheet.isShown)
     }
 
     override fun overrideNextRefreshTimer() {
@@ -477,7 +482,7 @@ class MainActivity :
 
     @OptIn(DelicateCoroutinesApi::class)
     @Suppress("DEPRECATION")
-    private fun setupPlayerViews() {
+    private fun setupPlayerViews(showMiniPlayerImmediately: Boolean) {
         binding.playerBottomSheet.listener = this
 
         viewModel.playbackState.observe(this) { state ->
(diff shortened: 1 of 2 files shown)


modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/PlayerBottomSheet.kt
```diff
@@ -134,14 +134,14 @@ class PlayerBottomSheet @JvmOverloads constructor(context: Context, attrs: Attri
...
-    fun setUpNext(upNext: UpNextQueue.State, theme: Theme) {
+    fun setUpNext(upNext: UpNextQueue.State, theme: Theme, shouldAnimateOnAttach: Boolean) {
         binding.miniPlayer.setUpNext(upNext, theme)
 
         // only show the mini player when an episode is loaded
         if (upNext is UpNextQueue.State.Loaded) {
             if ((isHidden() || !hasLoadedFirstTime)) {
                 show()
-                if (!shouldPlayerOpenOnAttach) {
+                if (!shouldPlayerOpenOnAttach && shouldAnimateOnAttach) {
```
mchowning, line 144: This is preventing `listener?.onMiniPlayerVisible()` from getting called on line 148 after changing themes. As a result, the padding isn't being updated to handle the mini player being visible, and we're losing content under the miniplayer. I was able to see this by following these steps:

1. Open the app and go to Profile tab -> Gear Icon
2. Scroll to the bottom of the screen and observe that the last two settings are "Privacy" followed by "About"
3. Tap on Appearance and change the theme
4. Tap back and scroll to the bottom
5. Observe that Privacy is now the last visible setting (because "About" is now hidden behind the miniplayer).

https://user-images.githubusercontent.com/4656348/199075257-fcca4e34-0277-4626-92bc-4056d0a82cfb.mp4

Similarly, it seems like we should also be updating the `hasLoadedFirstTime` field here, although I'm not sure that not updating it is actually causing any problems currently.

app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
```diff
@@ -219,7 +220,10 @@ class MainActivity :
             activity = this
         )
 
-        setupPlayerViews()
+        val showMiniPlayerImmediately = savedInstanceState?.getBoolean(SAVEDSTATE_MINIPLAYER_SHOWN, false) ?: false
+        binding.playerBottomSheet.isVisible = showMiniPlayerImmediately
```
mchowning, line 224: Am I understanding correctly that we're updating the visibility here even though it will get updated later in order to avoid (or at least reduce) the miniplayer "flickering" when changing themes? If so, it might be worth adding a comment here briefly explaining that.

mchowning (commented):
Thanks for the contribution @owenlejeune ! 🙇  This is looking nice and it fixes the animation issue. 🎉 

I left a comment below about an issue with the padding not being updated that I noticed. It would be good to address that before we merge this.

One other thing I noticed is that now, if a podcast is playing, the miniplayer's play button quickly switch back and forth between play and pause a few times when we switch themes. I suspect this was probably happening before, but the animation just hid it. 

| Before | After |
| --- | --- |
| <video src="https://user-images.githubusercontent.com/4656348/199078613-f11cab28-e6d1-44c8-a1fa-c7abe9fb69b8.mp4"/> | <video src="https://user-images.githubusercontent.com/4656348/199077534-1502e5b8-c2b9-4960-9f35-cfd434b68b01.mp4"/> |


We can log an issue for this (I wonder if it might have a similar cause as https://github.com/Automattic/pocket-casts-android/issues/52), and not address it in this PR if you'd like though because it's not directly related to the animation issue you're addressing here. Just let me know what you'd like to do.

owenlejeune:
@mchowning good catch! The padding issue should be addressed now. I don't seem to be able to reproduce the play button animation on my end though, may be best to open a new issue to address that separately.

mchowning (approved):
Looks good. Thanks so much for taking care of this @owenlejeune!

> I don't seem to be able to reproduce the play button animation on my end though, may be best to open a new issue to address that separately.

👍 Opened here: https://github.com/Automattic/pocket-casts-android/issues/523

PR #514, opened 2022-10-30
Support RTL notes in the webview

### LINKED ISSUE #109: Incorrect text alignment in episode description and notes for RTL languages
The text alignment for podcasts in RTL languages (e.g. Hebrew) is aligned to the left in episode details and notes. The user expects the text is aligned to the right like the podcast titles and description.

Sample podcast: https://pca.st/a8hgnlud

App version : 7.19.2 (7442)
Phone: Google - sunfish - Pixel 4a - sunfish
Android version: 12 SDK 31

Screenshot taken from the episode details screen

VGJohn:
## Description
Added RTL support to the notes that display in the webview of the EpisodeFragment and NotesFragment. `dir="rtl"` needs to set on the notes html for it to display correctly so check if the device locale is RTL and adding the appropriate tag to the html body.

Fixes #109 

## Testing Instructions
1. Set the device language to RTL language
2. Open podcast episode
3. See that the notes in the webview are aligned RTL

## Screenshots or Screencast 
Screenshot_20221030_050715
Screenshot_20221030_050638

## Checklist
 
- [x] with different themes


modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/helper/ShowNotesFormatter.kt
```diff
@@ -61,7 +64,10 @@ class ShowNotesFormatter(settings: Settings?, private val context: Context) {
             }
 
             addShowNotesHead(html)
-            html.append("<body>")
+
+            val isRtl = Locale.getDefault().layoutDirection == LayoutDirection.RTL
```
TylerCarberry, line 68: I think this should be checking for the language of the show notes, not of the device. Otherwise if you are subscribed to podcasts in different languages, half of them will always be wrong

VGJohn: Good point! Using `Bidi` instead to determine the direction of the show notes and switch the `WebView` to rtl, should match the behaviour of the `TextView` which shows the title rtl even when the device locale is set to a language that is not rtl 😃 

Cheers!

ashiagr (approved):
This is looking good! 
Thank you @VGJohn, @TylerCarberry

PR #519, opened 2022-10-31
Added basic support for Android 13 per-app language preferences

VGJohn:
## Description
Added basic support for Android 13 per-app language preferences
https://developer.android.com/guide/topics/resources/app-languages#app-language-settings

## Testing Instructions
Users can select their preferred language for each app through the system settings. They can access these settings in two different ways:

Access through the System settings:
Settings > System > Languages & Input > App Languages > (select an app)

Access through Apps settings:
Settings > Apps > (select an app) > Language

## Checklist
 
- [ ] with different themes


geekygecko (approved):
Thanks, it's great to see this in action! For anyone else looking at this pull request.

Before


After

PR #522, opened 2022-11-01
End of Year: Show total listening time for the year

ashiagr:
| 📘 Project: #410 |
|:---:|

iOS PR: https://github.com/Automattic/pocket-casts-ios/pull/411

## Description

Show approximate total listening time in 2022. (Design not covered)

## Testing Instructions

1. Login to an existing account
2. Run the app
4. ✅ Check that your total listening time in 2022 is shown

## Checklist

N/A
-  If this is a user-facing change, I have added an entry in CHANGELOG.md
 
- with different themes


modules/services/localization/src/main/res/values/strings.xml
```diff
@@ -1389,6 +1389,7 @@
     <string name="end_of_year_prompt_card_title" translatable="false">@string/end_of_year_launch_modal_title</string>
     <string name="end_of_year_prompt_card_summary">See your top podcasts, categories, listening stats and more.</string>
     <string name="end_of_year_share_via" translatable="false">@string/podcasts_share_via</string>
+    <string name="end_of_year_listening_time">In 2022, you spent %1$s listening to podcasts.</string>
```
mchowning, line 1392: Nitpick: I don't think we're actually telling them how much time they spent listening to podcasts because we're not taking account of skipping forward/backward or playback speed. I think we're more telling them that they listened to XXX hours/minutes of podcasts.

For example, I can make the total time go down if I rewind a podcast or make a previously played podcast as unplayed. 

Just raising this as something worth keeping in mind. Definitely not a blocker.

mchowning (approved):
Looks good. Nice work @ashiagr !

FWIW, I did see the Stories cycling faster than expected a few more times while testing this (not sure why, but I found this was _much_ easier to reproduce today compared to yesterday). The speedup generally seemed to be 2x-4x as fast.

https://user-images.githubusercontent.com/4656348/199256656-240165ca-8e94-491c-8952-8be62708ad75.mp4

One interesting thing I noticed is that when the speedup occurs, clicking the Share button does not stop the Stories from continuing to progress. The right Story is still shared, but in the background I can see the UI continuing to move through the stories. This does not seem to occur for me when the Stories are playing at the correct speeds though (i.e., the Stories immediately stop progressing when I tap Share). I wonder if this might give a hint as to the underlying cause. 🤔 

| Without speedup | With speedup |
| --- | --- |
| <video src="https://user-images.githubusercontent.com/4656348/199257411-7291e9cb-490d-4658-b31e-6e54efc4765b.mp4"/> | <video src="https://user-images.githubusercontent.com/4656348/199257429-519aacf4-3d38-4169-9752-db30483066cc.mp4"/> |

This isn't anything that needs to be addressed in this PR.

modules/services/localization/src/main/res/values/strings.xml
ashiagr: I understand your point, @mchowning. AFAIU, it is just an approximate listening time, does not take into account changes to played time due to skipping, and matches iOS behavior. 

I'll keep it in mind for sure. 👍

ashiagr:
> FWIW, I did see the Stories cycling faster

Thanks for sharing the video and observations! I wasn't able to check that issue yet but they'll definitely be handy when I'll come to it. 🙇‍♀️

Issue #524, opened 2022-11-01
Project: Onboarding Flow


mchowning:
## Description

Add a new onboarding flow to help users get up to speed more quickly.

### Initial Setup
- [X] Put feature behind feature flag during development ([#525](https://github.com/Automattic/pocket-casts-android/pull/525))
- [X] Create basic navigation framework with empty screens ([#525](https://github.com/Automattic/pocket-casts-android/pull/525))
- [X] Only show onboarding to users a single time ([#603](https://github.com/Automattic/pocket-casts-android/pull/603))

### Initial choose whether to Log In or Sign Up Screen
- [X] Implement designs ([#555](https://github.com/Automattic/pocket-casts-android/pull/555), [#617](https://github.com/Automattic/pocket-casts-android/pull/617))
    - [X] Make podcasts in image dynamic based on the user's subscribed podcasts ([#665](https://github.com/Automattic/pocket-casts-android/pull/665))
- [X] Add functionality (buttons should link to appropriate screens) ([#525](https://github.com/Automattic/pocket-casts-android/pull/525))
- [X] Analytics ([#555](https://github.com/Automattic/pocket-casts-android/pull/555), [#564](https://github.com/Automattic/pocket-casts-android/pull/564), [#642](https://github.com/Automattic/pocket-casts-android/pull/642))

### Log in screen
- [X] Implement design  ([#555](https://github.com/Automattic/pocket-casts-android/pull/555))
- [X] Enable log in  ([#555](https://github.com/Automattic/pocket-casts-android/pull/555))
- [X] Handle errors  ([#555](https://github.com/Automattic/pocket-casts-android/pull/555))
- [X] Forgot password screen ([#560](https://github.com/Automattic/pocket-casts-android/pull/560))
- [X] Analytics  ([#555](https://github.com/Automattic/pocket-casts-android/pull/555), [#560](https://github.com/Automattic/pocket-casts-android/pull/560), [#646](https://github.com/Automattic/pocket-casts-android/pull/646))
- [X] [v2] Migrate old log in screen to use new screen written in compose ([#694](https://github.com/Automattic/pocket-casts-android/pull/694))

### Create Account screen
- [X] Implement design  ([#555](https://github.com/Automattic/pocket-casts-android/pull/555))
- [X] Disable submission button until input is valid  ([#555](https://github.com/Automattic/pocket-casts-android/pull/555))
[...]

Issue #338, opened 2022-09-21
application_opened analytics event is delayed


mchowning:
The application_opened analytics event appears to be just a bit delayed resulting in it firing _after_ the various screen shown events are fired. Seems like this could cause some difficulties with determining what the user's flow was through the app when they open the application.

```
2022-09-21 13:12:58.765 I  🔵 Tracked: podcast_tab_opened, Properties: {...}
2022-09-21 13:12:58.864 I  🔵 Tracked: podcasts_list_shown, Properties: {...}
2022-09-21 13:12:58.981  I  🔵 Tracked: application_opened, Properties: {...}
```

#### Actual Behavior
application_opened fires _after_ the relevant screen_shown events are fired when opening the app

#### Expected Behavior
application_opened fires _before_ the relevant screen_shown events are fired when opening the app

PR #525, opened 2022-11-01
Add basic navigation for new onboarding flow

mchowning:
| 📘 Project: #524  |
| --- |

## Description
This PR adds the basic high-level framework for the new onboarding flow.

## Testing Instructions

1. Open the app
2. ✅ Verify that the onboarding flow does NOT open
3. Update the `ONBOARDING_ENABLED` build flag in `base.gradle` to be "true"
4. ✅ Verify that the onboarding flow opens
5. Admire the amazing UI that has been implemented
6. ✅ Verify that tapping "Not Now" exits the onboarding flow and takes you to the app
7. For each of the other buttons ("Continue with Google", "Sign Up Free", and "Log in"):
    1. ✅ Verify that tapping the button opens a new screen
    2. Again, admire the amazing UI of this new screen
    3. ✅ Verify that swiping back (or tapping the back arrow if there is one) returns to the top-level screen and that swiping back a second time exits the app.

## Checklist
- [ ] Any jetpack compose components I added or changed are covered by compose previews (will handle this when implementing the screens)
 
- [X] with different themes


app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
```diff
@@ -265,6 +270,18 @@ class MainActivity :
         if (BuildConfig.END_OF_YEAR_ENABLED) {
             setupEndOfYearLaunchBottomSheet()
         }
+
+        binding.onboardingFrame.setContent {
```
mchowning, line 274: Instead of doing this, would I be better off starting a new Activity? wdyt?

ashiagr (approved):
Great start, @mchowning! 🎉 

I liked how you split your commit, realizing that I should be renaming (Stories) compose screen to be appended with `Page` and not `Screen`.

I also liked how you directly used a composable without adding a fragment to the `onboardingFrame` Compose view.

It's really nice to see more Compose in the codebase. 🙂 Looking forward to the next PR. 🙌

mchowning:
> One thing that might concern us is that <initial>_tab_opened is fired as soon as we open the app, even before onboarding flow is complete and we land on the actual main screen. I wonder if it could be due to https://github.com/Automattic/pocket-casts-android/issues/338 that we noticed earlier.

That's a good point @ashiagr . I suspect that even if we resolved that issue, the tab opened event would still get fired with onboarding because I'm still loading the tab behind the onboarding screen. I'll keep this in mind going forward and see if I can think of a good way to address it.

PR #527, opened 2022-11-01
Provide MiniPlayerState to save/restore isPlaying state for MiniPlayer View

### LINKED ISSUE #523: When changing themes while playing the miniplayer play icon switches back and forth
### Description

If a podcast is playing, the miniplayer's play button quickly switch back and forth between play and pause a few times when we switch themes. This became apparent after [we fixed an animation issue when changing themes](https://github.com/Automattic/pocket-casts-android/pull/513), but I suspect the issue existed before and was just hidden by the animation.

I wonder if this might have a similar underlying cause as https://github.com/Automattic/pocket-casts-android/issues/52.

1. Start playing a podcast
2. Switch themes in the app
3. 💥 Observe that the play button quickly switches back and forth a few times between play and pause

<video src="https://user-images.githubusercontent.com/4656348/199077534-1502e5b8-c2b9-4960-9f35-cfd434b68b01.mp4"/> 

Samsung A13, Android 12, local Pocket Casts build of https://github.com/Automattic/pocket-casts-android/pull/513

mrmike:
## Description
Provide `MiniPlayerState` to restore `playing` value after `MiniPlayer` view is recreated. Currently `MiniPlayer` view after recreation always starts with the state `playing=state` even if podcast is playing. This may lead to different glitches with animating play/pause state (like that one described in https://github.com/Automattic/pocket-casts-android/issues/523). I was considering using [BaseSavedState](https://developer.android.com/reference/android/view/View.BaseSavedState) instead of plain `Parcelable` but I think in this case it's fine since `MiniPlayer` won't be extended by other classes.

Fixes #523 

## Testing Instructions
1. Start playing a podcast
2. Switch themes in the app
3. Play button should not switch back and forth a few times between play and pause 

## Screenshots or Screencast 
https://user-images.githubusercontent.com/529635/199326831-6208dbd1-e1e7-4e26-8c81-4207869f53d8.mp4

## Checklist
 
- [x] with different themes


modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/MiniPlayer.kt
```diff
@@ -60,8 +62,17 @@ class MiniPlayer @JvmOverloads constructor(context: Context, attrs: AttributeSet
 
     override fun onLayout(changed: Boolean, left: Int, top: Int, right: Int, bottom: Int) {
         super.onLayout(changed, left, top, right, bottom)
-        // set the initial play button icon to the play triangle. This is run after the button has inflated or the Lottie drawable is null and it won't work.
```
mrmike, line 63: This comment is no longer valid since we have restored state and don't have to start with play button. But I'm open to suggestion if you want to keep this comment in any form.

ashiagr (approved):
Thank you for fixing that annoying animation! 🙏

PR #536, opened 2022-11-03
End of Year - Show number of listened podcasts and episodes 

ashiagr:
| 📘 Project: #410 |
|:---:|

iOS PR: https://github.com/Automattic/pocket-casts-ios/pull/414

## Description
Adds a story that shows the number of listened podcasts and episodes. (Designs not covered)

## Testing Instructions
1.  Set `END_OF_YEAR_ENABLED` feature flag to `true` in `base.gradle`
4. Skip to the fourth story
5. ✅ Check that the number of played podcasts and episodes is shown

## Checklist
N/A (Added placeholder layouts and strings)

- [ ] with different themes

### DIFF
--- a/modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/stories/EndOfYearStoriesDataSource.kt
+++ b/modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/stories/EndOfYearStoriesDataSource.kt
@@ -13,8 +13,9 @@ class EndOfYearStoriesDataSource @Inject constructor(
         return combine(
             endOfYearManager.getTotalListeningTimeInSecsForYear(YEAR),
             endOfYearManager.findListenedCategoriesForYear(YEAR),
+            endOfYearManager.findListenedNumbersForYear(YEAR),
             endOfYearManager.findRandomPodcasts(),
-        ) { listeningTime, listenedCategories, podcasts ->
+        ) { listeningTime, listenedCategories, listenedNumbers, podcasts ->
             val stories = mutableListOf<Story>()
 
             listeningTime?.let { stories.add(StoryListeningTime(it)) }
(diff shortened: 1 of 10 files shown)


modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/stories/EndOfYearStoriesDataSource.kt
```diff
@@ -13,15 +13,19 @@ class EndOfYearStoriesDataSource @Inject constructor(
...
-        ) { listeningTime, listenedCategories, podcasts ->
+        ) { listeningTime, listenedCategories, listenedNumbers, podcasts ->
             val stories = mutableListOf<Story>()
 
             listeningTime?.let { stories.add(StoryListeningTime(it)) }
             if (listenedCategories.isNotEmpty()) {
                 stories.add(StoryListenedCategories(listenedCategories))
                 stories.add(StoryTopListenedCategories(listenedCategories))
             }
+            if (listenedNumbers.numberOfEpisodes > 0 && listenedNumbers.numberOfPodcasts > 0) {
```
mchowning, line 26: I konw the strings are just placeholders, but noting that if we're going to display this screen with 1 episode or 1 podcast, we'll need to have strings to accommodate those numbers. It might be better to just not show this screen if the number of episodes is 1 since that doesn't really provide any value.

mchowning (approved):
Looks good!

While testing this, I ran across [this issue](https://github.com/Automattic/pocket-casts-android/issues/538), which can cause the listening history to get out of sync with an account. It's not related to your changes, but I wanted to mention it since it can affect the data displayed in the EOY stats.

modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/stories/EndOfYearStoriesDataSource.kt
ashiagr: Good point, I updated the condition in f6675c6.

PR #537, opened 2022-11-03
End of Year: Show top 1 podcast and top 5 podcasts

ashiagr:
| 📘 Project: #410  |
|:---:|

iOS PR: https://github.com/Automattic/pocket-casts-ios/pull/415

## Description
Adds the story for the top 1 podcast and the top 5 podcasts. 

PS: Ignore hard-coded strings and placeholder layouts/ design in StoryViews.

## Testing Instructions
4. Skip to the 5th story
5. ✅ Check that it shows your most listened podcast, together with the number of episodes and listening time
6. Skip to the 6th story
7. ✅ Check that it shows your top 5 podcasts

## Screenshots or Screencast 
| Top 1 podcast | Top 5 podcasts |
| -------------- | --------------- |
 | 

## Checklist
N/A (added placeholder layouts + strings)

- [ ] with different themes


modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/views/stories/StoryTopFivePodcastsView.kt
```diff
@@ -55,9 +55,9 @@ fun StoryFake1View(
             modifier = modifier.fillMaxWidth(),
             verticalArrangement = Arrangement.Center
         ) {
-            items(story.podcasts.size) { index ->
+            items(story.topPodcasts.size) { index ->
```
mchowning, line 58: Not a big deal, but I think we could simplify this just a bit with:

``` kotlin
import androidx.compose.foundation.lazy.items

// ...

items(story.topPodcasts) { topPodcast ->
    PodcastItem(
        podcast = topPodcast.toPodcast(),
        // ...
    )
}

```

Certainly not a blocker.

mchowning (approved):
Looks good. 👍 

While testing this I did notice that there is a bit of a flicker when switching to the stories with 5 list items (happen on both the top podcast and top categories ones). It's quick (you really have to slow down the video to see it), but it is definitely noticeable on my device.


https://user-images.githubusercontent.com/4656348/199739762-63b4ab98-d176-4b2f-bfef-cf2f496749a6.mp4


My guess is that coil is loading the podcast images asynchronously. If it's not easy to delay displaying the screen until the images are loading, perhaps we could just include some kind of placeholder so the text doesn't jump (ideally, the `PodcastImage` component itself would take care of that, or at least have an option to to take care of that). That would probably be enough to completely hide this. I haven't tested any of that, so all of that might be completely wrong.

I'll go ahead and approve because this doesn't need to be addressed in this PR.

modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/views/stories/StoryTopFivePodcastsView.kt
ashiagr: Nice tip, applied in 793134d. Thank you!

ashiagr:
Thanks for reviewing it, @mchowning! I'll address the flicker issue in a future PR.

PR #543, opened 2022-11-04
Fixed crashes with Tasker plugin actions when using minified code

joaomgcd:
## Description
The Tasker plugin code has an issue that only occurs when code is minified (ie in release builds) that is fixed here. This bug is preventing the Tasker actions from being used at all.

## Testing Instructions
1. Build app with ```minifyEnabled true```
2. Try using any Pocket Casts Tasker action from Tasker
3. Check that it doesn't crash anymore

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,16 @@
 7.27
 -----
     *   Added ability to set playback effects in Tasker "Control Playback" action.
-        ([#415](https://github.com/Automattic/pocket-casts-android/pull/509)).
+        ([#509](https://github.com/Automattic/pocket-casts-android/pull/509)).
+    *   Fixed crashes with Tasker plugin actions when using minified code.
+        ([#543](https://github.com/Automattic/pocket-casts-android/pull/543)).
 
 7.26
 -----
 
 *   New Features:
     *   Added Tasker integration with "Play Filter" and "Control Playback" actions.
-        ([#415](https://github.com/Automattic/pocket-casts-android/pull/431)).
+        ([#431](https://github.com/Automattic/pocket-casts-android/pull/431)).
     *   Import OPML from a URL
         ([#482](https://github.com/Automattic/pocket-casts-android/pull/482)).
     *   Redesign of the fullscreen player share option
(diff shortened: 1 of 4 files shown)


mchowning (approved):
Interesting. I didn't even think about the class files getting stripped out when I was reviewing this initially. Good reminder to be careful about that. Thanks for tracking down the issue and fixing this!

I'm not sure we'll be able to get this into 7.26, so could you add a 7.27 release note, and then we can merge this?

joaomgcd:
Ok, updated. Thanks.

A shame the first release with Tasker integration doesn't work. 😣Is the version with Tasker already available for everyone? Or just in beta?

Thanks again!

mchowning:
> Is the version with Tasker already available for everyone? Or just in beta?

It's in beta. 

If we have to update the beta for any other reason, we'll try to get this in too, but since this isn't a regression (on account of it never working in release builds 😞 ) I don't think this alone quite justifies cutting a new beta release (there's a fair bit of work around cutting a new release, and it would mean that all of our beta users would be downloading another apk, which we try to minimize).

joaomgcd:
Sure, I understand. Since it's in beta I don't think it's that big of an issue. I was getting worried that everyone in PR would have a broken Tasker integration. 😅 Hope it works now!

mchowning:
> I was getting worried that everyone in PR would have a broken Tasker integration. 

They will when the beta gets released as 7.26 in a week or so. 😞 Fortunately, they never had Tasker integration before, so it's at least not breaking existing functionality. We'll keep this in mind though, and see if we can get it in before 7.26 is released.

joaomgcd:
Oh I see. That is unfortunate... Maybe it's best that this is tested on a real release build then, just to make sure that we didn't miss anything else related to it being built that way.

Unfortunately I'm not able to do a proper release build, I always get the following error:

```
Execution failed for task ':app:processReleaseGoogleServices'.
> No matching client found for package name 'au.com.shiftyjelly.pocketcasts'
```
Any way around that? Would be great to make sure that everything is working in a proper installation from a release APK.

mchowning:
> Any way around that? Would be great to make sure that everything is working in a proper installation from a release APK.

I'm not sure there's a way around it short of you creating your own google project with our release package name and getting your own google-services. json file from that (to be clear, I'm not suggesting you actually do that 😄 ). But I did test this with a release build when I reviewed the PR this time, and it worked fine, so I think we're good.

joaomgcd:
Awesome! :) Thank you for testing!

ashiagr:
This is included in a new beta `7.26-rc-4`.

PR #551, opened 2022-11-07
End of Year: Check user eligibility and listening history

ashiagr:
| 📘 Project: #410 |
|:---:|

iOS PR: https://github.com/Automattic/pocket-casts-ios/pull/433

## Description

This PR 
* Checks if a user is eligible to see their EoY stats (i.e. they listened to at least one episode for more than 30 minutes this year)
* Checks user Listening History, if their history is not complete, call a placeholder sync method.

Note that the "listening history" sync endpoint is not yet integrated.

#### Checking Listening History

Right now, the app assumes that the user has the full listening history based on:

1. If their oldest item on Listening History is from a previous year
2. If the above is false check if their total number of episodes in Listening History is smaller or equal to 100 (the server limit)

## Testing Instructions

2. Do a clean install of the app and run it
3. ✅ Check that nothing happens, no modal is shown and there's no card under Profile
4. Login to account in which you have a Listening History with a few episodes
5. ✅ Check that as soon as sync completes, the modal is shown and the card is now visible on Profile

## Checklist
 
- with different themes


leandroalonso:
@ashiagr JFYI, I recently changed the logic on iOS to 5 minutes, instead of 30.

There's a huge amount of podcasts with many episodes with less than 30 minutes, so a user that listened to, let's say, 10 episodes with 29min would be excluded from EoY and we don't want that. :)

ashiagr:
Got it. I'll update it when I integrate the listening history sync endpoint. Thanks for keeping a check on Android side. 🥇

modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/StoriesDataSource.kt
```diff
@@ -4,5 +4,8 @@ import au.com.shiftyjelly.pocketcasts.endofyear.stories.Story
 import kotlinx.coroutines.flow.Flow
 
 interface StoriesDataSource {
+    suspend fun isEligibleForStories(): Flow<Boolean>
+    suspend fun hasFullListeningHistory(): Flow<Boolean>
+    suspend fun syncListeningHistory(): Flow<Boolean>
```
MiSikora, line 9: Functions that return `Flow` shouldn't be suspending. Suspending mechanism is already built into `Flow`. Any `suspend fun foo(): Flow<T>` can be expressed as `fun foo(): Flow<T>`. This change should be propagated all the way downstream to view models and wherever else these functions are used directly or indirectly.

app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
```diff
@@ -190,8 +195,17 @@ class MainActivity :
...
+            repeatOnLifecycle(Lifecycle.State.STARTED) {
+                viewModel.isEndOfYearStoriesEligible().collect { isEligible ->
+                    if (isEligible) {
+                        setupEndOfYearLaunchBottomSheet()
+                        if (settings.getEndOfYearShowBadge2022()) {
+                            binding.bottomNavigation.getOrCreateBadge(VR.id.navigation_profile)
+                        }
+                    }
+                }
+            }
```
MiSikora, line 208: This could be refactored to use `.flowWithLifeycle()` and `.launchIn()` extensions to avoid nesting once functions exposing `Flow` in `StoriesDataSource` are no longer suspending.

modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/StoriesViewModel.kt
```diff
@@ -49,25 +49,42 @@ class StoriesViewModel @Inject constructor(
...
-                        currentStory = result[currentIndex],
-                        segmentsData = SegmentsData(
-                            xStartOffsets = List(numOfStories) { getXStartOffsetAtIndex(it) },
-                            widths = storyLengthsInMs.map { it / totalLengthInMs.toFloat() },
-                        )
-                    )
+            storiesDataSource.hasFullListeningHistory().stateIn(viewModelScope)
+                .collect { isFullListeningHistory ->
+                    if (!isFullListeningHistory) {
+                        storiesDataSource.syncListeningHistory().collect { success ->
```
MiSikora, line 55: I'm not sure I understand this logic. The `hasFullListeningHistory()` stream will be suspended until `syncListeningHistory()` is finished collecting all items, and won't process anything from the upstream. Is this what should happen?

ashiagr: I haven't hooked up `syncListeningHistory(...)` yet, and this might need some refactoring. The idea is to check if the user doesn't have a full listening history, get it from the server and then load all stories. 

You can leave it as is, it'll be taken care of while integrating the listening history endpoint.

MiSikora: Would `syncListeningHistory()` be a continuous operation ore just a one shot one? If it is hitting an HTTP endpoint it most likely shouldn't be exposed as a `Flow` but rather as a suspending function. In any case it can be taken care of once the integration is in place.

ashiagr: Correct, that should just be a suspending function.

modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/StoriesDataSource.kt
ashiagr: Removed in 35f1d40

app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
ashiagr: Nice tip, applied in ebe4f45

modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/StoriesViewModel.kt
ashiagr: I added a TODO in 1ceb61f

MiSikora (approved):

geekygecko (approved):
This PR has already been reviewed by @MiSikora so just approving so it can be merged.

PR #555, opened 2022-11-10
Onboarding: Login and Signup

mchowning:
| 📘 Project: #524  |
| --- |

### Description
Adds the login and create account pages for the new onboarding flow.

The image on the opening onboard screen has a placeholder because I do not have that asset yet. Otherwise, the design should be accurate. 

Note that the [Figma designs](PviUP0aiZctOprDuuYRwpb-fi-649%3A27097) do not include any theming, but this implementation does reflect the current theme, which is what Adam requested.

This PR includes a change to use a new Activity for the onboarding flow. I think this is cleaner and it avoids an issue where TalkBack was picking up on UI elements from the Main Activity that were not visible during onboarding.

### Testing Instructions

Before testing this, make sure that update the `ONBOARDING_ENABLED` build flag in `base.gradle` to be "true".

Also, before each of these tests, make sure the app is not logged into an account—the onboarding flow will only load if the app is not logged into an account.

##### 1.1 Happy path

1. Open the app
2. ✅  Observe that the main onboarding screen design is correct, and matches the relevant theme (apart from the placeholder for the main image)
3. ✅ Observe the event `🔵 Tracked: setup_account_shown`
4. Tap "Log in"
5. ✅ Observe the event `🔵 Tracked: setup_account_button_tapped` with the `button` property of "sign_in", followed by the event `🔵 Tracked: signin_shown`
6. ✅ Verify that either swiping back or tapping the back button on the toolbar takes you back to the initial screen and logs the event `🔵 Tracked: signin_dismissed` and `🔵 Tracked: setup_account_shown`
7. ✅ Observe that the login screen design is correct ([Figma designs](PviUP0aiZctOprDuuYRwpb-fi-649%3A27097))
8. ✅ Observe that the keyboard opens and the email field is focused
9. Type in an account email address
10. ✅ Observe that tapping Next on the keyboard moves focus to the password field
11. Type in the password for the account
12. ✅ Observe that either (1) tapping Done on the keyboard or (2) tapping the Log in button will advance the user to a recommendations page (just a placeholder at this time)
13. ✅ Observe the event ` 🔵 Tracked: user_signed_in, Properties: {"sign_in_source":"onboarding", ...}`
14. Swipe back
15. ✅ Observe that the user is taken to the main screen of the app (and is not taken back to the log in screen)
16. ✅ Observe that the app has been logged into the relevant account

##### 1.2 Error cases
1. Get to the onboarding "log in" screen
2. Without typing anything, tap the login button
3. ✅ Observe the the outline of both text fields turns red and there is red text below each text field advising the user of an error.
4. Add an email address
5. ✅ Observe that as soon as the entry constitutes a valid email address the error color and message for that field goes away.
7. Add a password that is at least 6 characters long
8. ✅ Observe that as soon as 6 characters have been typed the error color and message for that field goes away
9. ✅ Observe that if you edit the email or password fields, as soon as those fields are no longer valid, the error state for that field is shown (red outline + error message)
10. Edit the email address so that it is an email address without an associated Pocket Casts account
11. Tap Log in
12. ✅ Observe that the error from the server advising that there is not an account for this email is displayed
13. Change the email to be one with a Pocket Casts account, but enter the wrong password and tap Log in
17. ✅ Observe that the error from the server advising that the password is wrong is displayed

##### 1.3 Autofill
Observe that when you enter the login screen, any autofill credentials that you have saved with the Android OS are presented as options, and that you can successfully submit those credentials.

##### 1.1 Happy path

1. Open the app
2. ✅  Observe that the main onboarding screen design is correct, and matches the relevant theme (apart from the placeholder for the main image)
3. Tap `Create free account`
4. ✅ Observe the event `🔵 Tracked: setup_account_button_tapped, Properties: {"button":"create_account", ...}`, followed by the event `🔵 Tracked: create_account_shown`
5. ✅ Verify that either swiping back or tapping the back button on the toolbar takes you back to the initial screen and logs the event `🔵 Tracked: create_account_dismissed` and `🔵 Tracked: setup_account_shown`
6. ✅ Observe that the create account screen design is correct ([Figma designs](PviUP0aiZctOprDuuYRwpb-fi-649%3A27097))
7. ✅ Observe that the keyboard opens and the email field is focused
8. Type in an email address
9. ✅ Observe that tapping Next on the keyboard moves focus to the password field
10. Type in the password for the account
11. ✅ Observe that the newsletter is toggled on by default
12. [optional] switch the newsletter toggle to be off
13. ✅ Observe that either (1) tapping Done on the keyboard or (2) tapping the Create account button will advance the user to a recommendations page (which is just a placeholder at this time)
14. ✅ Observe the event `🔵 Tracked: user_account_created`, followed by the event `🔵 Tracked: newsletter_opt_in_changed, Properties: {"source":"account_created","enabled":[true/false], ...}`
15. Swipe back
18. ✅ Observe that the user is taken to the main screen of the app (and is not taken back to the log in screen)
19. ✅ Observe that the app has been logged into the relevant account
20. ✅ Observe that on the user's profile screen (Profile Tab -> tap on Account Block toward top of screen), the newsletter toggle matches the selection that was made on the Create Account screen

##### 1.2 Error cases
1. Get to the onboarding Create Account screen
2. Without typing anything, tap the Create Account button
3. ✅ Observe that the outline of both text fields turns red and there is red text below each text field advising the user of an error.
4. Add an email address
5. ✅ Observe that as soon as the entry constitutes a valid email address the error color and message for that field goes away.
7. Add a password that is at least 6 characters long, but is NOT the correct password
8. ✅ Observe that as soon as 6 characters have been typed the error color and message for that field goes away
9. ✅ Observe that if you edit the email or password fields, as soon as those fields are no longer valid, the error state for that field is shown (red outline + error message)
10. Attempt to log in with the incorrect password entered on step 6.
11. ✅ Observe that the password error message from the server is returned, but because this is a server error (as opposed to a local error due to the password not meeting our password criteria), the password field does not turn red (same would happen if you had entered an invalid email address).

##### 1.3 Autofill
Autofill doesn't seem to be saving credentials automatically from this screen the way I would expect it to. It appears that this is a bug with Jetpack compse: https://issuetracker.google.com/issues/176949051

#### Onboarding opening screen
image

#### Log in

| Default state | with errors |
| --- | --- |
| image | image |

#### Create account
| Default state | with errors |
| --- | --- |
| image | image |

## Checklist
 
- [X] with different themes


modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingFlowComposable.kt
```diff
@@ -2,25 +2,41 @@ package au.com.shiftyjelly.pocketcasts.account.onboarding
...
 
-        // If the BackHandler is disabled, then the normal back pressed behavior
-        // takes over and closes the app
-        val enableBackHandler = navController.backQueue.isNotEmpty()
-        BackHandler(enabled = enableBackHandler) {
-            navController.popBackStack()
+        val backStackEntry by navController.currentBackStackEntryAsState()
+        val currentDestination = backStackEntry?.destination?.route
+
+        BackHandler {
```
ashiagr, line 28: Looks like it was tricky to fix 😄

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/viewmodel/OnboardingCreateAccountViewModel.kt
```diff
@@ -0,0 +1,124 @@
...
+class OnboardingCreateAccountViewModel @Inject constructor(
+    private val auth: AccountAuth,
+    private val analyticsTracker: AnalyticsTrackerWrapper,
+    private val subscriptionManager: SubscriptionManager,
+    private val settings: Settings,
+) : ViewModel(), CoroutineScope {
+    override val coroutineContext: CoroutineContext
+        get() = Dispatchers.Default
+
+    private val _stateFlow = MutableStateFlow(
```
ashiagr, line 30: Should we use an underscore for mutable fields or prefix them with `mutable`? I know I used both conventions, so wanted to make sure that I use the correct one.

ashiagr (approved):
Everything works as described, @mchowning. Great job!

Just in the create account flow, I found that the password error message text was not in red:

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/viewmodel/OnboardingCreateAccountViewModel.kt
mchowning: I don't have a strong preference either way. If you have a preference, let me know because I'm fine adopting that.

Sounds like this may not be a problem [much longer](https://github.com/Kotlin/KEEP/blob/explicit-backing-fields-re/proposals/explicit-backing-fields.md). 🤞

mchowning:
>  in the create account flow, I found that the password error message text was not in red

It was like that because that text always displayed, so it wasn't error text, but I like the idea of giving it a red state when the user tries to submit a password that is too short. I've made that change in [8a16556](https://github.com/Automattic/pocket-casts-android/pull/555/commits/8a165565ebad2a7f0fb6ec73e85fbe63e933f263). Thanks for suggesting that!

PR #559, opened 2022-11-10
End of Year: Listened categories design 

ashiagr:
| 📘 Project: #410 |
|:---:|

iOS PR: https://github.com/Automattic/pocket-casts-ios/pull/453

## Description
Adds the design for the "listened categories" story.

* The podcasts cover displayed are determined by getting the list of top categories and reversing it, picking the top 1 podcast for each of them
* Background is picked from the middle podcast in the stack (without any gradient)

## Testing Instructions
4. Navigate to the third story
4. ✅ Check the design

## Screenshots or Screencast 

| Phone (Pixel 5) |
| -------- | 
|  |

| Tablet (Nexus 9) |
| -------- | 
| |

P.S. -> Story speed issue has re-surfaced after `StoriesFragment` (bottom sheet fragment) was removed in favor of Compose dialog in [this commit](https://github.com/Automattic/pocket-casts-android/commit/07478904b503f0b7718ff120a72d920f28a6e4d4). I tried to fix in 67bf509e11297d02b28c549b851dc78f67a692d2 but I can still the issue. I'll try to fix it in the next PR.

## Checklist
- \If this is a user-facing change, I have added an entry in CHANGELOG.md
 
- [x] with different themes

### DIFF
--- a/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
+++ b/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
@@ -497,7 +497,6 @@ class MainActivity :
             var showDialog by rememberSaveable { mutableStateOf(false) }
             if (showDialog) {
                 StoriesPage(
-                    showDialog = true,
                     theme = theme,
                     onCloseClicked = { showDialog = false },
                 )
--- a/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
+++ b/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
@@ -38,7 +38,6 @@ import au.com.shiftyjelly.pocketcasts.compose.AppTheme
 import au.com.shiftyjelly.pocketcasts.databinding.ActivityMainBinding
 import au.com.shiftyjelly.pocketcasts.discover.view.DiscoverFragment
 import au.com.shiftyjelly.pocketcasts.endofyear.StoriesPage
-import au.com.shiftyjelly.pocketcasts.endofyear.StoriesViewModel
 import au.com.shiftyjelly.pocketcasts.endofyear.views.EndOfYearLaunchBottomSheet
 import au.com.shiftyjelly.pocketcasts.filters.FiltersFragment
 import au.com.shiftyjelly.pocketcasts.localization.helper.LocaliseHelper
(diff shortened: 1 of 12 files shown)


app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
```diff
@@ -499,8 +497,7 @@ class MainActivity :
             var showDialog by rememberSaveable { mutableStateOf(false) }
             if (showDialog) {
                 StoriesPage(
-                    viewModel = storiesViewModel,
-                    showDialog = showDialog,
+                    showDialog = true,
```
mchowning, line 500: Do we still need this`showDialog` parameter? It looks like we're always passing `true` now, but maybe that's just temporary?

modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/views/stories/StoryListenedCategoriesView.kt
```diff
@@ -1,36 +1,141 @@
...
+    modifier: Modifier = Modifier,
+) {
+    TextP40(
+        text = stringResource(id = LR.string.end_of_year_story_listened_to_categories_subtitle),
+        textAlign = TextAlign.Center,
+        color = story.tintColor,
+        fontWeight = FontWeight.Bold,
+        modifier = modifier
+            .fillMaxWidth()
+            .alpha(0.8f)
```
mchowning, line 138: With some tint colors, this `alpha` can make the text a bit hard to read (to my eyes at least).

mchowning (approved):
LGTM! 👍

app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
ashiagr: Not needed anymore, removed in 0ac80aa.

modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/views/stories/StoryListenedCategoriesView.kt
ashiagr: I'll try adding a background gradient to improve this in a future PR. 

I also realized we need to display only three podcast covers. This is fixed in 82c7c52.

PR #560, opened 2022-11-10
Onboarding Flow: Forgot Password

mchowning:
| 📘 Project: #524  |
| --- |

> **Note**
> This PR builds on https://github.com/Automattic/pocket-casts-android/pull/555

## Description
Adds the forgot password screen to the onboarding flow.

## Testing Instructions

Before testing this, make sure that update the ONBOARDING_ENABLED build flag in base.gradle to be "true".

1. Open the app when not logged into an account
2. Tap `Log in`
3. Tap `Forgot password?`
4. Type text that is not a valid email address (i.e., "123")
5. Tap either the keyboard "Done" button or the "Confirm" button
6. ✅ Verify that the text field's border turns red and an appropriate error message (which comes from the server) is displayed
7. Update the email field to have a valid email address that is associated with a Pocket Casts account
8. ✅ Observe that while the call is in progress the text field and confirm button are disabled
9. ✅ When the call completes, observe that the keyboard is dismissed and a basic dialog is shown confirming that the link has been sent and asking the user to check their email (I'm using this simple dialog without styling because it is what we were already using in the `[ResetPasswordFragment](https://github.com/Automattic/pocket-casts-android/blob/5dc3ae9fa1f354a5de6ec2fd4be1256db4efad17/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/ResetPasswordFragment.kt#L84-L93)`)
10. Tap `OK`
11. ✅ Observer that the onboarding flow is closed and the user is placed on the main screen of the app.

## Screenshots or Screencast 

| Empty | Error from server |
| --- | --- |
| Screen Shot 2022-11-10 at 2 44 17 PM | Screen Shot 2022-11-10 at 2 47 52 PM |

## Checklist
 
- [X] with different themes


geekygecko (approved):
Tested this and it works well.
Minor suggestion, it might be nicer to clear the error state as soon as they start typing in the text box.

PR #561, opened 2022-11-11
End of Year: Top podcast and longest episode stories design 

ashiagr:
| 📘 Project: #410 |
|:---:|

iOS PRs: 
https://github.com/Automattic/pocket-casts-ios/pull/455
https://github.com/Automattic/pocket-casts-ios/pull/463

## Description
Add the design for the top podcast and longest episode stories which were similar in design.

The layers of the displayed podcast and episode are achieved by using the podcast cover colors informed by the server.

PS: I also extracted a few components and modifiers (e.g. `PodcastCover`, `Modifier.podcastDynamicBackground`). This has impacted previous stories and increased no. of lines in this PR.

## Testing Instructions

4. Navigate to the top podcast story
5. ✅ Check the design
6. Navigate to the longest episode story
7. ✅ Check the design

## Screenshots or Screencast 

|Top Podcast (Pixel 2) | Longest Episode (Pixel 5) |
|------|----|
|   | |

## Checklist
 
- with different themes


modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/utils/StoryModifierExtensions.kt
```diff
@@ -0,0 +1,33 @@
...
+
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.drawWithContent
+import androidx.compose.ui.graphics.BlendMode
+import androidx.compose.ui.graphics.Brush
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.graphicsLayer
+import au.com.shiftyjelly.pocketcasts.models.entity.Podcast
+
+fun Modifier.podcastDynamicBackground(podcast: Podcast) =
```
ashiagr, line 11: This is not perfect. I'll try to improve in a future PR.

modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/components/PodcastCover.kt
```diff
@@ -0,0 +1,75 @@
...
+import androidx.compose.ui.graphics.graphicsLayer
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+import au.com.shiftyjelly.pocketcasts.compose.components.PodcastImage
+
+private const val PodcastCoverRotationAngle = -30f
+private const val PodcastCoverSkew = 0.45f
+
+@Composable
+fun PodcastCover(
```
ashiagr, line 23: It uses a card elevation instead of the gradients provided. Again, this will be improved in a future PR.

modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/views/stories/StoryListeningTimeView.kt
```diff
@@ -44,76 +36,58 @@ fun StoryListeningTimeView(
...
-
-    val timeText = StatsHelper.secondsToFriendlyString(story.listeningTimeInSecs, context.resources)
-    val funnyText = FunnyTimeConverter().timeSecsToFunnyText(
-        story.listeningTimeInSecs,
-        context.resources
-    )
-
-    val backgroundColor = story.podcasts[0].toPodcast().getTintColor(false)
-
     Column(
```
mchowning, line 39: This isn't related to the changes in this PR, but what about making this Column scrollable? That might help prevent overlap for users who use large font sizes.

mchowning (approved):
Looks good. 👍 

One thing that is becoming more obvious to me as you add more content is how the current speed of the stories feels too fast for me. I figure this is probably something that is not in its final state yet, but figured I'd go ahead and mention it.

Another thing I noticed is that the language "podcasts creators" language on the last story screen seems off to me. Maybe podcasts should be singular, so "podcast creators"?

modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/views/stories/StoryListeningTimeView.kt
ashiagr: Column was scrollable actually. I increased cover row height in [66568da](https://github.com/Automattic/pocket-casts-android/pull/561/commits/66568da32662b429637948cf69349540912da17e) to accommodate transformed image height. 

https://user-images.githubusercontent.com/1405144/201461621-63d7a02b-3f60-4de8-9cf1-61b11b150988.mp4

ashiagr:
> One thing that is becoming more obvious to me as you add more content is how the current speed of the stories feels too fast for me. I figure this is probably something that is not in its final state yet, but figured I'd go ahead and mention it.

I kept story time smaller to switch stories faster during development. But I increased it to 5 secs to match with iOS in b92f796 so that I don't forget. Thanks for pointing it out. 🙏 

> Another thing I noticed is that the language "podcasts creators" language on the last story screen seems off to me. Maybe podcasts should be singular, so "podcast creators"?

String.xml has the [correct string](https://github.com/Automattic/pocket-casts-android/blob/e46b5347fdbef44aa4e3afad7c8be0f46471f925/modules/services/localization/src/main/res/values/strings.xml#L1441) for this. I'll use that one when I finalize the epilogue story design.

PR #562, opened 2022-11-11
End of Year: Sync the missing listening history

geekygecko:
## Description

The change checks if there is any missing listening history at the start of the End of Year flow. If you have lots of missing listening history this might take awhile. I will take a look at the speed of this in a future change.

## Testing Instructions

2. Run the `debugProd` build variant 
3. Login to your account which will sync 100 listening history episodes
4. When the prompt appears, tap "View My 2022"
5. The app will download all your episode listening history for 2022
6. ✅ Check the results seem correct


modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/EpisodeDao.kt
```diff
@@ -363,11 +363,11 @@ abstract class EpisodeDao {
         """
         SELECT * 
         FROM episodes
-        WHERE episodes.last_playback_interaction_date IS NOT NULL AND episodes.last_playback_interaction_date < :fromEpochMs
+        WHERE episodes.last_playback_interaction_date IS NOT NULL AND episodes.last_playback_interaction_date < :fromEpochMs AND episodes.last_playback_interaction_date != 0
```
geekygecko, line 366: If an episode is deleted from listening history this is set to 0 so this ignores that episode.

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/EpisodeDao.kt
```diff
@@ -376,5 +376,5 @@ abstract class EpisodeDao {
         WHERE episodes.last_playback_interaction_date IS NOT NULL AND episodes.last_playback_interaction_date > :fromEpochMs AND episodes.last_playback_interaction_date < :toEpochMs
         """
     )
-    abstract fun findEpisodesCountInListeningHistory(fromEpochMs: Long, toEpochMs: Long): Flow<Int>
+    abstract suspend fun findEpisodesCountInListeningHistory(fromEpochMs: Long, toEpochMs: Long): Int
```
geekygecko, line 379: I switched this to get the result once instead of a Flow as we only need to know the value at the start of the flow.

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/Episode.kt
```diff
@@ -53,7 +53,7 @@ data class Episode(
     // Removed but leaving the column until we definitely aren't using it again
     @ColumnInfo(name = "cleanTitle") var cleanTitle: String? = title,
     @ColumnInfo(name = "last_playback_interaction_date") var lastPlaybackInteraction: Long? = null,
-    @ColumnInfo(name = "last_playback_interaction_sync_status") var lastPlaybackInteractionSyncStatus: Long = 1,
+    @ColumnInfo(name = "last_playback_interaction_sync_status") var lastPlaybackInteractionSyncStatus: Long = LAST_PLAYBACK_INTERACTION_SYNCED,
```
geekygecko, line 56: I moved the history sync logic and switched this value to a constant at the same time to clean up.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/endofyear/EndOfYearManagerImpl.kt
```diff
@@ -13,47 +17,72 @@ import au.com.shiftyjelly.pocketcasts.repositories.endofyear.stories.StoryTopFiv
...
+    private val settings: Settings,
 ) : EndOfYearManager, CoroutineScope {
+
+    companion object {
+        private const val YEAR = 2022
+        private const val EPISODE_MINIMUM_PLAYED_TIME_IN_MIN = 30L
+    }
+
+    private val yearStart = epochAtStartOfYear(YEAR)
+    private val yearEnd = epochAtStartOfYear(YEAR + 1)
```
geekygecko, line 49: I moved the repeating logic below to these variables.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/HistoryManager.kt
```diff
@@ -0,0 +1,89 @@
...
+    override val coroutineContext: CoroutineContext
+        get() = Dispatchers.Default
+
+    /**
+     * Read the server listening history response.
+     * @param response The server response.
+     * @param updateServerModified Set to true when this is latest listening history, rather than part of the user's history.
+     */
+    suspend fun processServerResponse(response: HistorySyncResponse, updateServerModified: Boolean) = withContext(Dispatchers.IO) {
+        if (!response.hasChanged(0) || response.changes.isNullOrEmpty()) {
```
geekygecko, line 35: I moved this logic from the SyncHistoryTask so the End of Year logic could share it.

ashiagr (approved):
Tested with an account having more than 100 listening history records. I found everything working. 

Thank you so much for this effort, @geekygecko! 🙏

PR #565, opened 2022-11-12
End of Year: Number of listened podcasts/episodes story design

ashiagr:
| 📘 Project: #410 |
|:---:|

iOS PR: https://github.com/Automattic/pocket-casts-ios/pull/454

## Description
Add the design for the "number of podcasts/episodes listened" story.

The podcasts cover displayed is determined by getting the top 10 podcasts and reversing it. We then display the last 6 available.

## Testing Instructions
4. Navigate to the 5th story
4. ✅ Check the design

## Screenshots or Screencast 

Phone (Pixel 6)  | Tablet (Nexus 7 - 2012)
-----| ------
 | 

## Checklist
 
- [x] with different themes


geekygecko (approved):
This worked really well! Amazing work. 

My only comment would be that the text might look better with the same line breaks as the design, rather than just letting the width do it.

Screenshot 2022-11-14 at 11 43 25 am

ashiagr:
I added the line break feedback to the backlog. As discussed, I'll take it up later.

mchowning:
> My only comment would be that the text might look better with the same line breaks as the design, rather than just letting the width do it.

I wonder how line breaks will look in other languages though. 🤔 I always try to avoid line breaks for that reason, but maybe they don't actually cause problems.

PR #567, opened 2022-11-14
End of Year: Epilogue story design

ashiagr:
| 📘 Project: #410 |
|:---:|

iOS PR: https://github.com/Automattic/pocket-casts-ios/pull/467

## Description

- Adds the design for the epilogue story.
- Improves pause/ resume behavior (now the story pause on press instead of long press earlier)

## Testing Instructions
4. Navigate to the last story
5. ✅ Check the design
6. Tap "Replay"
7. ✅ Stories should replay
8. Tap on a story
9.  ✅ Story should pause immediately
10. Release tap event
11.   ✅ Story should resume

## Screenshots or Screencast 

P.S. I picked the background color from the iOS app and applied the existing gradient on other screens.

## Checklist
 
- [x] with different themes


modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/views/stories/StoryEpilogueView.kt
```diff
@@ -1,89 +1,136 @@
...
-        )
-        TextH30(
-            text = "Don't forget to share with your friends and give a shout out to your favorite podcasts creators",
-            color = story.tintColor,
-            textAlign = TextAlign.Center,
-            modifier = modifier.padding(16.dp)
-        )
+        Spacer(modifier = modifier.height(40.dp))
+
+        Spacer(modifier = modifier.weight(1f))
```
mchowning, line 52: This is fine, but fwiw I would have probably combined these Spacers with something like:

``` kotlinSpacer(
    Modifier
        .heightIn(min = 40.dp)
        .weight(1f)
)
```

mchowning (approved):
Looking good.

The only issue I noticed was that, with the pausing behavior, if your finger moves at all, the story continues playing. Seems like it would be nice for the pause to persist as long as the finger is held down particularly since I found it easy to accidentally move my finger just a bit when I was trying to pause.

ashiagr:
Merging this PR, tried to fix it in https://github.com/Automattic/pocket-casts-android/pull/581.

PR #577, opened 2022-11-15
Onboarding Flow: Feature flag the Continue with Google button

geekygecko:
## Description

This change hides the "Continue with Google" single sign on button behind a feature flag. The SSO may be finished after we release the new onboarding, that is why it's a separate flag.
The button is also hidden if the user hasn't installed Google Play Services. 

## Testing Instructions

1. Update the ONBOARDING_ENABLED and SINGLE_SIGN_ON_ENABLED build flags in base.gradle to be "true".
2. Open the app on an emulator with Google Play Services installed.
3. ✅ Verify the "Continue with Google" button is visible.
4. Open the app on an emulator that doesn't have Google Play Services installed.
5. ✅ Verify the "Continue with Google" button is not visible.
6. Update the SINGLE_SIGN_ON_ENABLED build flags in base.gradle to be "false". 
7. Open the app on an emulator with Google Play Services installed.
8. ✅ Verify the "Continue with Google" button is not visible.

### DIFF
--- a/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/viewmodel/OnboardingLoginOrSignUpViewModel.kt
+++ b/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/viewmodel/OnboardingLoginOrSignUpViewModel.kt
@@ -11,7 +11,7 @@ import javax.inject.Inject
 
 @HiltViewModel
 class OnboardingLoginOrSignUpViewModel @Inject constructor(
-    @ApplicationContext private val context: Context
+    @ApplicationContext context: Context
 ) : ViewModel() {
 
     val showContinueWithGoogleButton = BuildConfig.SINGLE_SIGN_ON_ENABLED && GoogleApiAvailability.getInstance().isGooglePlayServicesAvailableSuccess(context)
--- a/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/viewmodel/OnboardingLoginOrSignUpViewModel.kt
+++ b/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/viewmodel/OnboardingLoginOrSignUpViewModel.kt
@@ -1,0 +1,18 @@
+package au.com.shiftyjelly.pocketcasts.account.viewmodel
+
+import android.content.Context
+import androidx.lifecycle.ViewModel
+import au.com.shiftyjelly.pocketcasts.account.BuildConfig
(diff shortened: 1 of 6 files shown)


modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/viewmodel/OnboardingLoginOrSignUpViewModel.kt
```diff
@@ -0,0 +1,18 @@
...
+import dagger.hilt.android.qualifiers.ApplicationContext
+import javax.inject.Inject
+
+@HiltViewModel
+class OnboardingLoginOrSignUpViewModel @Inject constructor(
+    @ApplicationContext private val context: Context
+) : ViewModel() {
+
+    val showContinueWithGoogleButton = BuildConfig.SINGLE_SIGN_ON_ENABLED && GoogleApiAvailability.getInstance().isGooglePlayServicesAvailableSuccess(context)
+}
```
geekygecko, line 18: It might not of been worth it creating a view model for just this but I will need one for the Google Sign sign in logic.

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/viewmodel/OnboardingLoginOrSignUpViewModel.kt
```diff
@@ -0,0 +1,18 @@
...
+import au.com.shiftyjelly.pocketcasts.account.BuildConfig
+import au.com.shiftyjelly.pocketcasts.utils.extensions.isGooglePlayServicesAvailableSuccess
+import com.google.android.gms.common.GoogleApiAvailability
+import dagger.hilt.android.lifecycle.HiltViewModel
+import dagger.hilt.android.qualifiers.ApplicationContext
+import javax.inject.Inject
+
+@HiltViewModel
+class OnboardingLoginOrSignUpViewModel @Inject constructor(
+    @ApplicationContext private val context: Context
```
mchowning, line 14: Nitpick, but we don't actually need to hold onto the context (at least we don't need to yet), so we could drop the `private val`. If we want to keep ahold of it, perhaps it would be worth adding `@SuppressLint("StaticFieldLeak")` to get rid of the lint warning since it's not possible for a ViewModel to "leak" the application's context.

geekygecko: Oops. Thanks, I have removed `private val`.

PR #579, opened 2022-11-15
End of Year: Top categories story design 

ashiagr:
| 📘 Project: #410 |
|:---:|

iOS PR: https://github.com/Automattic/pocket-casts-ios/pull/468

## Description
Adds the design for the top categories story.

## Testing Instructions
4. Navigate to the 4th story
4. ✅ Check the design

## Screenshots or Screencast 

| Pixel 5 | Nexus 7 (2012)|
|----|---|
 | 

## Checklist
 
- [x] with different themes

### DIFF
--- a/modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/components/TextStyles.kt
+++ b/modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/components/TextStyles.kt
@@ -237,7 +237,7 @@ fun TextH70(
     Text(
         text = text,
         color = color,
-        fontSize = 12.sp,
+        fontSize = if (disableScale) 12.nonScaledSp else 12.sp,
         fontWeight = fontWeight,
         lineHeight = if (disableScale) 14.nonScaledSp else 14.sp,
         letterSpacing = if (disableScale) .25f.nonScaledSp else .25.sp,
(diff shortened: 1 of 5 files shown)


modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/components/TextStyles.kt
```diff
@@ -90,14 +102,16 @@ fun TextH40(
     modifier: Modifier = Modifier,
     textAlign: TextAlign? = null,
     color: Color = MaterialTheme.theme.colors.primaryText01,
-    maxLines: Int = Int.MAX_VALUE
+    maxLines: Int = Int.MAX_VALUE,
+    fontWeight: FontWeight = FontWeight.Medium,
+    disableScale: Boolean = false
```
ashiagr, line 107: As the category pillar title and duration are limited to small widths, I'm using non-scaled font size for both otherwise they spread into multiple lines on a large display screen.

mchowning: That makes sense. 

I am still seeing some issues related to that though. In particular, "Technology" still gets wrapped (going to be hard to avoid that since there maybe even longer categories). Also, the time values still appear to be scaled when I increase my font size.

mchowning (approved):
So much fancy matrix work here! 😄 Nice job! I think there are still some small issues with the font size and the durations. Slightly more concerning is that I saw the "Technology" word wrapping even  without any font scaling. Might be good to see if there's a way we can make allow that text to be wider, but I realize that's probably a really tricky problem to solve since we don't know how wide the text will be.

modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/components/TextStyles.kt
ashiagr: I fixed font size for the duration and increased text widths a bit as we have a little space available between pillars. This is a tricky problem to solve I bet.

PR #581, opened 2022-11-15
End of Year: Fix pause and vm issues

ashiagr:
## Description

This PR attempts to fix following two issues

#### Pause behavior: https://github.com/Automattic/pocket-casts-android/pull/567#pullrequestreview-1179288351

For this, I added an `AwaitPointerEventScope` extension inspired from [this post](https://stackoverflow.com/a/68878910/193545). - 68e8851f333da7fc5564923c2b13f739430c2a75

#### `StoriesViewModel` Scope: https://github.com/Automattic/pocket-casts-android/pull/570#discussion_r1021785677

Due to [viewmodels not being scoped to a composable](https://issuetracker.google.com/issues/165642391), I changed the way the dialog is shown. - 9c140aa3e47936301290a31d4d3a889381c02425

## Checklist
 
- [ ] with different themes

### DIFF
--- a/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
+++ b/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
@@ -12,10 +12,6 @@ import androidx.activity.viewModels
 import androidx.appcompat.app.AppCompatActivity
 import androidx.appcompat.app.AppCompatDelegate
 import androidx.appcompat.widget.Toolbar
-import androidx.compose.runtime.getValue
-import androidx.compose.runtime.mutableStateOf
-import androidx.compose.runtime.saveable.rememberSaveable
-import androidx.compose.runtime.setValue
 import androidx.core.view.isVisible
 import androidx.core.view.updatePadding
 import androidx.fragment.app.Fragment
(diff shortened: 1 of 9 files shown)


mchowning (commented):
I really like the pause behavior when you hold now. It feels really naturally. I also noticed that it pauses when you scroll as well, which I also like.

### 1. Flinging scroll cycles the story

If you scroll slowly, the story doesn't change, but if you do a quick fling to scroll, then the story changes. This is only likely to come up with large font/display settings where scrolling is more necessary, but it did throw me because I initially didn't realize my fling is what was triggering the story to change and it seemed like the stories were just rotating faster, so I could imagine this might confuse a user.

Since this does not really come up on normal display sizes,imo this probably isn't the highest priority issue to address.

https://user-images.githubusercontent.com/4656348/201948844-d7933f40-afea-4bcc-8538-9dc940a4d7bb.mp4

### 2. Story continuing to play during share

If I did the tap-share-then-tap-the-story-really-fast move, then the Story would not stop playing like it normally does when I tap Share. The video below shows this issue.

### 3. Pause stopping working

I had a few times where the pause behavior stopped working. The only way I was able to recreate this consistently was with the tap-share-then-tap-the-story-really-fast move. I did cause the issue without tapping the share button, but I wasn't able to find any other way to consistently recreate it.


https://user-images.githubusercontent.com/4656348/201950184-bb2cfffd-14e1-48fc-9f6d-bfd1e89a4c25.mp4

ashiagr:
👋 @mchowning 

For 1.  tried to improve the behavior for this in 53a18b5 but it's just a workaround. It is tricky to detect tap/ long tap/ fling / scroll, I'll need more time to figure it out.

For 2 & 3, I converted db functions to suspend from flow as now the sync listening history endpoint downloads podcasts and episodes on login before we show stories. This fixes stories state changes when episode 's `played_up_to` is updated in db triggering another start.

Can you give it another try? I think we're very close to fixing it.🤞

mchowning:
I'm still able to recreate all 3 issues. FWIW, 2 & 3 feel like edge cases to me if the only way we can reproduce them is with the tap-share-and-then-quickly-tap-the-story action. I still haven't found another way to recreate it although I'm sure I caused it without sharing before (I only saw the issue with sharing this time, but I'm not sure if that's because it's better or because I just didn't do whatever is causing it).

> For 1. tried to improve the behavior in https://github.com/Automattic/pocket-casts-android/commit/53a18b5e09b2a5bf475debc97c3cdb9b727190dc but it's just a workaround. It is tricky to detect tap/ long press/ fling / scroll,

Yeah, it does seems like it would be really hard to handle all those actions well.

mchowning (approved):

mchowning:
Going ahead and approving in case you'd like to merge this and try to do some improvements in separate PRs.

ashiagr:
Quite a few PRs are dependent on it, so I'll merge it and come up with a fix if I can reproduce the scenario.
Thanks so much for testing it, @mchowning! I really appreciate it.

PR #583, opened 2022-11-15
Added Tasker Actions: Query Podcasts, Query Podcast Episodes, Query Filters, Query Filter Episodes, Add To Up Next

joaomgcd:
## Description
Added 6 new Tasker Actions that allows users to query data from Pocket Casts and manage the Up Next Queue: 

- Query Podcasts
- Query Podcast Episodes
- Query Filters
- Query Filter Episodes
- Add To Up Next
- Query Up Next Episodes

This allows users to setup their own up next queue manager in Tasker like for example  as shown in these demos: 

- creating a simple Pocket Casts Android Wear app: https://youtu.be/opzeAOuLOjI
- creating a notification that automatically plays your favorite podcasts on the tap of a button: https://youtu.be/ifpf-D8w0b4
- creating a playlist manager: https://youtu.be/aaKggzKK2QM

More output info can be added to the Query actions very easily if needed.

**Note**: If you could test this in a release build and check that there no crashes that would be great :)

## Testing Instructions
1. Open Tasker
2. Add the **Pocket Casts Query Podcasts** action
3. Add a flash action with the variables you want from the above action
4. Add the **Pocket Casts Query Podcast Episodes**
5. Select any podcast you want
6. Again, check that the variables are correctly populated
7. Do the same for **Pocket Casts Query Filters** and **Pocket Casts Query Filter Episodes** actions
8. Add a **Pocket Casts Add To Up Next** action
9. Use the result UUID of a podcast episode as this action's input (for example, to use the first result from a query above use the **%episode_uuid(1)** value
10.  Check if the queue was set correctly in Pocket Casts

## Checklist
 
- [x] with different themes
- x] with the device set to have a large display and font size


mchowning:
👋  @joaomgcd ! Wanted to let you know that this is on my radar. I've had a lot on my plate lately, so I haven't had a chance to get to this yet, but I'm hoping to get to it next week. Thanks so much for continuing to improve the Tasker integration. It's really great to see.

joaomgcd:
@mchowning no problem :) Whenever you have time! Glad you like the Tasker integration!

Since you haven't looked at it yet I quickly added in another action that I forgot to add before: **Query Up Next**: allows you to get episode info from the current queue in Pocket Casts. With this action users can now automate based on what is already on there :)

Thanks again!

mchowning (commented):
👋  @joaomgcd ! FInally got to this, sorry for the delay. This is looking good (and I really like the videos 😄 )! I ran into a couple of issues when testing it, but maybe I'm just misunderstanding something because I don't have much experience creating actions like this in Tasker.

#### 1. Query Filter Episodes

When using this action, I noticed that tapping on the search icon in the configuration section did not bring up my filters--instead, it brought up my subscribed podcasts. Selecting one of those would then, not surprisingly, cause the action to fail. If I manually typed in one of my filters it would work though.

#### 2. Query Podcast Episodes -> Add To Up Next

If I set up the Add To Up Next Action to run after the Query Podcast Episodes action, tapping on the tag icon in the configuration section for Add To Up Next did not present the `%episode_uuid()` variable as an option. Manually entering `%episode_uuid()` did work though.

joaomgcd:
Hi again Matt :) No problem, glad you're liking the actions!

I just fixed the first issue you mentioned: was merely an issue with the labels and the function used to get the suggestions.

About the second issue, I can't actually replicate it.

If I 

- Add a **Query Podcast Episdoes** action
- Configure and accept it
- Add a **Add To Up Next** action
- Configure it
- Click the **tag** icon in any of the inputs

I can see all the variables that the **Query Podcast Episodes** action outputs.

Do you not see any of the variables at all?

Thanks again!

mchowning:
> I just fixed the first issue you mentioned: was merely an issue with the labels and the function used to get the suggestions.

Yep, that fixed it! 👍 

>  Is it maybe just the issue you mentioned before that you have to scroll down to see more variables? 😅

You'd think I'd learn, but apparently not—that was the issue. I am seeing both `%episode_uuid` and `%episode_uuid()` in the list, any idea why that might be?

Screenshot 2022-12-19 at 2 11 42 PM

If that isn't something you can reproduce, I think we can merge this once the branch is up-to-date with `main` and move the release note to the `7.30` section.

joaomgcd:
> You'd think I'd learn, but apparently not—that was the issue.

Well, that goes to show that it really should be changed somehow if even we sometimes don't remember there are more values... 😅


>  I am seeing both %episode_uuid and %episode_uuid() in the list, any idea why that might be?

In some cases that does make sense in Tasker-land, but out of simplicity I changed it so that only the array variant will show if both exist :)

Hope everything's ok with it now! Thanks again!

mchowning (approved):

madiele:
@joaomgcd I've build the debug apk and I've been testing the plugin the last few days, it seems to work with no issue, I've just one small gripe, the query episodes actions do not return the id of the parent podcast which means you can't use it to get the thumb url and other podcast info easily, at the moment I'm using the publish date as a makeshift foreign key which works only if you list the latest podcast episodes releases.

Other than that great work! This pull request made me switch from podcast addict back to pocket casts again after 5 years!

joaomgcd:
Hi @madiele ! Thanks for the suggestion! :) Ok, added the podcast uuid to the query episode actions' outputs. Let me know if it works as you need it to! :)

PR #591, opened 2022-11-17
End of Year: Require login

ashiagr:
| 📘 Project: #410 |
|:---:|

iOS PR: https://github.com/Automattic/pocket-casts-ios/pull/474
## Description

In order to provide better End of Year stats next year, it's required to create an account to see 2022 End of Year stats.

## To test

3. Logout
4. Re-run the app
5. ✅ The 2022 prompt should appear, even though you're logged out
6. Tap "View my 2022"
7. You should be prompted to login or create an account
8. Login (or create an account)
11. ✅ The stories should appear

You should see the same behavior when tapping the card under Profile. If you're logged in, the stories should appear. If not, the login/sign up screen should appear.

I also updated min played time eligibility from 30min to 5min based on this comment https://github.com/Automattic/pocket-casts-android/pull/551#issuecomment-1305493401.

P.S. 
1. Remote feature flag is not added in this PR due to issues in retrieving it from Firebase.
2. Apologies for using quite a few booleans in this PR, I'll create a tech-debt task to clean those up.

## Checklist
 
- [ ] with different themes

### DIFF
--- a/base.gradle
+++ b/base.gradle
@@ -29,6 +29,7 @@ android {
 
         // Feature Flags
         buildConfigField "boolean", "END_OF_YEAR_ENABLED", "false"
+        buildConfigField "boolean", "END_OF_YEAR_REQUIRE_LOGIN", "true"
         buildConfigField "boolean", "ONBOARDING_ENABLED", "false"
 
         testInstrumentationRunner project.testInstrumentationRunner
--- a/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
+++ b/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
@@ -503,15 +503,30 @@ class MainActivity :
                 EndOfYearLaunchBottomSheet(
                     shouldShow = shouldShow,
                     onClick = {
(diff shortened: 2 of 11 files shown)


base.gradle
```diff
@@ -29,6 +29,7 @@ android {
 
         // Feature Flags
         buildConfigField "boolean", "END_OF_YEAR_ENABLED", "false"
+        buildConfigField "boolean", "END_OF_YEAR_REQUIRE_LOGIN", "true"
```
mchowning, line 32: Why do we want this flag for requiring login?

mchowning (approved):
This works well in my testing. 👍 

> ...using quite a few booleans in this PR

I think that if you moved some of the logic from `MainActivity` to the view model, that would make it easier to simplify the logic some and reduce the back and forth between the activity and the view model.

What you have is working though, so I'll go ahead and approve.

base.gradle
ashiagr: iOS uses a remote FF to control whether you need an account login to see stories.
> There's a remote flag that dictates if an account is required or not to see the stories.

I have put a placeholder FF here, and try to move it to Firebase remote config if I can.

PR #593, opened 2022-11-18
End of Year - Update queries

ashiagr:
| 📘 Project: #410 |
|:---:|

## Description
Updates db queries to match with iOS:

1. Use distinct on queries: https://github.com/Automattic/pocket-casts-ios/pull/528

2. Exclude empty categories from top categories: https://github.com/Automattic/pocket-casts-ios/pull/515. 

    I tested it by running old and new queries that included podcast uuids with empty and non-empty categories in the where clause. Old query returned the category with empty title and new one did not.

    exclude category

## Checklist
 
- [ ] with different themes


mchowning (approved):
Nice! 👍

PR #594, opened 2022-11-18
Onboarding upgrade selection screen

mchowning:
| 📘 Project: #524  |
| --- |

## Description
This adds the upgrade screens for the onboarding flow.

Sorry about the big PR. I'll do a better job of keeping these smaller in the future.

## Issues

There were two issues I ran into that I had trouble resolving.

### 1. ~Flicker on non-selected subscription~

~When switching between the different subscriptions, there is frequently a ripple effect on one of the non-selected subscriptions. Let me know if you have any idea what might be causing this or how we might be able to fix it.~

<details>
<summary><s>Video where the middle button "ripples" even though it's not being touched</s></summary>

https://user-images.githubusercontent.com/4656348/202799223-be275cc1-c091-4a1e-80c4-023b31233a28.mp4

</details>

UPDATE: This flicker issue [has been resolved](https://github.com/Automattic/pocket-casts-android/pull/594#issuecomment-1322690834).

### 2. Keyboard opening from Google Payment flow

When Google's payment flow completes and returns the user to our app, the keyboard is opening for some reason. I don't know why, so I just have the composable checking if the keyboard is open and closing it if it is. This works ok, but there is a pretty noticeable flicker/jank when the keyboard opens and is then immediately closed. I'd love to find a better way to handle this.

I noticed that this also happens when you return from opening the Privacy Policy or Terms of Service (the links to open those doesn't seem to work on debug builds that are set up to receive subscriptions for some reason, so they have to be tested with a release build).

<details>
<summary>Video</summary>

https://user-images.githubusercontent.com/4656348/202799360-5dba664c-2f51-4063-8439-6570be5d9906.mp4

</details>

### 1. Log into free account
1. Use a build where you are able to retrieve subscriptions from Google Play and where you have altered the build so that it receives our test subscription with a 4 day free trial. Ping me if you would like some help doing this or check the message I left in Slack.
2. In `base.gradle` turn on the onboarding feature.
3. Make sure you are not logged into an account (otherwise the onboarding flow won't start)
4. Start the app
5. Tap "Log in"
7. Log into an account that does NOT have Plus (or create a new account)
8. Tap "Not now" on the placeholder recommendations screen
9. Tap "Unlock all features" on the initial upgrade screen
10. ✅ Verify that the UI matches the Figma designs
11. ✅  Observe that all available plans are displayed.
12. ✅ If you have a plan with a free trial available, that plan should be selected by default
13. ✅ Verify that swiping back or tapping above the bottom sheet closes the bottom sheet
14. ✅ Verify that tapping on "Privacy Policy" and "Terms and Conditions" links open the appropriate pages in a web browser. NOTE: This did not work for me on debug builds with the changes I needed to pull the subscriptions from Google Play for some reason (I had the same problem trying to open those web URLs from the account screen in the app too), so you may need to test this on a release build.
15. Tap a plan that does not have a free trial
16. ✅ Observe that the "4 DAY FREE TRIAL" banner and the "Recurring payments will begin..." text animate out of view if they were visible.
17. Tap a plan that does have a free trial
18. ✅  Observe that the "4 DAY FREE TRIAL" banner and the "Recurring payments will begin..." text animate into view.
19. Tap "Start Free Trial & Subscribe"
20. ✅ Observe that the google subscription flow is shown
21. Swipe back or otherwise cancel the google subscription flow
22. ✅ Observe that there is NOT an error message displayed
23. Tap "Start Free Trial & Subscribe"
24. Attempt to subscribe with a card that always declines
25. Tap Ok in the error sheet from google
26. ✅ Observe the view now has an error message "Could not create subscription. Please try again"
27. Tap on a different subscription
28. ✅ Observe that the error message goes away
29. Tap "Start Free Trial & Subscribe"
30. Subscribe with a card that will approve
31. ✅ Observe that the upgrade screen is dismissed and the placholder welcome screen is displayed (It just has the text "Welcome"). There is a brief moment where the upgrade screen is still displayed before the Welcome screen shows up here that I don't like. Let me know if you have any suggestions for how to improve that.
32. Swipe back
33. ✅ Observe that you are taken to the app's main screen.

### 2. Plus Account
1. Perform the same steps 1-6 as the previous test
2. Log into an account that is already Plus
3. Tap "Not now" on the placeholder recommendations screen
4. ✅ Observe that the upgrade screen is not displayed, instead the placholder welcome screen is displayed
5. Swipe back
6. ✅ Observe that you are taken to the app's main screen.

## Screenshots or Screencast 

https://user-images.githubusercontent.com/4656348/202809306-1279545e-a0b3-4de1-96f3-7bbdc6f5c9da.mp4

## Checklist
 
- [X] with different themes

### DIFF
--- a/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingFlowComposable.kt
+++ b/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingFlowComposable.kt
@@ -141,7 +141,16 @@ fun OnboardingFlowComposable(
                         navController.popBackStack()
                     },
                     onUpgradePressed = { analyticsTracker.track(AnalyticsEvent.ONBOARDING_UPGRADE_UNLOCK_ALL_FEATUERS_TAPPED) },
-                    onNotNowPressed = { analyticsTracker.track(AnalyticsEvent.ONBOARDING_UPGRADE_NOT_NOW_TAPPED) },
+                    onNotNowPressed = {
+                        analyticsTracker.track(AnalyticsEvent.ONBOARDING_UPGRADE_NOT_NOW_TAPPED)
+                        navController.navigate(OnboardingNavRoute.welcome)
+                    },
+                    onComplete = { navController.navigate(OnboardingNavRoute.welcome) },
+                )
+            }
+            composable(OnboardingNavRoute.welcome) {
+                OnboardingWelcomePage(
+                    onContinue = completeOnboarding,
                 )
             }
         }
@@ -166,4 +175,5 @@ private object OnboardingNavRoute {
     const val forgotPassword = "forgot_password"
     const val recommendations = "recommendations"
     const val plusFeatures = "upgrade_features"
+    const val welcome = "welcome"
 }
--- a/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingFlowComposable.kt
+++ b/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingFlowComposable.kt
@@ -142,7 +142,6 @@ fun OnboardingFlowComposable(
                     },
                     onUpgradePressed = { analyticsTracker.track(AnalyticsEvent.ONBOARDING_UPGRADE_UNLOCK_ALL_FEATUERS_TAPPED) },
                     onNotNowPressed = { analyticsTracker.track(AnalyticsEvent.ONBOARDING_UPGRADE_NOT_NOW_TAPPED) },
-                    onCompletePlus = { /* TODO */ }
                 )
             }
         }
--- a/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingFlowComposable.kt
+++ b/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingFlowComposable.kt
@@ -135,16 +135,8 @@ fun OnboardingFlowComposable(
 
             composable(OnboardingNavRoute.plusFeatures) {
                 OnboardingPlusFeaturesFlow(
-                    onShown = { analyticsTracker.track(AnalyticsEvent.ONBOARDING_UPGRADE_SHOWN) },
-                    onBackPressed = {
-                        analyticsTracker.track(AnalyticsEvent.ONBOARDING_UPGRADE_DISMISSED)
-                        navController.popBackStack()
-                    },
-                    onUpgradePressed = { analyticsTracker.track(AnalyticsEvent.ONBOARDING_UPGRADE_UNLOCK_ALL_FEATUERS_TAPPED) },
-                    onNotNowPressed = {
-                        analyticsTracker.track(AnalyticsEvent.ONBOARDING_UPGRADE_NOT_NOW_TAPPED)
-                        navController.navigate(OnboardingNavRoute.welcome)
-                    },
+                    onBackPressed = { navController.popBackStack() },
+                    onNotNowPressed = { navController.navigate(OnboardingNavRoute.welcome) },
                     onComplete = { navController.navigate(OnboardingNavRoute.welcome) },
                 )
             }
--- a/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingFlowComposable.kt
+++ b/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingFlowComposable.kt
@@ -141,7 +141,8 @@ fun OnboardingFlowComposable(
                         navController.popBackStack()
                     },
                     onUpgradePressed = { analyticsTracker.track(AnalyticsEvent.ONBOARDING_UPGRADE_UNLOCK_ALL_FEATUERS_TAPPED) },
-                    onNotNowPressed = { analyticsTracker.track(AnalyticsEvent.ONBOARDING_UPGRADE_NOT_NOW_TAPPED) }
+                    onNotNowPressed = { analyticsTracker.track(AnalyticsEvent.ONBOARDING_UPGRADE_NOT_NOW_TAPPED) },
+                    onCompletePlus = { /* TODO */ }
                 )
             }
         }
--- a/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingFlowComposable.kt
+++ b/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingFlowComposable.kt
@@ -124,7 +124,7 @@ fun OnboardingFlowComposable(
                     },
                     onBackPressed = {
                         analyticsTracker.track(AnalyticsEvent.RECOMMENDATIONS_DISMISSED)
-                        navController.popBackStack()
+                        completeOnboarding()
                     },
                     onComplete = {
                         // TODO analytics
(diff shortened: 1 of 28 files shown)


modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingPlusFeaturesPage.kt
```diff
@@ -74,18 +63,34 @@ import au.com.shiftyjelly.pocketcasts.localization.R as LR
...
     val viewModel = hiltViewModel<OnboardingPlusFeaturesViewModel>()
     val state by viewModel.state.collectAsState()
 
-    LaunchedEffect(Unit) { onShown() }
-    BackHandler { onBackPressed() }
+    @Suppress("NAME_SHADOWING")
+    val onUpgradePressed = {
+        viewModel.onUpgradePressed()
+        onUpgradePressed()
+    }
```
mchowning, line 79: I'm curious what you think about this pattern. I've always been skeptical of shadowing variable names, but this seemed like a nice way to guarantee that `onUpgradePressed()` cannot be called without also calling `viewModel.onUpgradePressed`.

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingFlowComposable.kt
```diff
@@ -124,24 +126,30 @@ fun OnboardingFlowComposable(
...
-                        navController.popBackStack()
-                    },
-                    onUpgradePressed = { analyticsTracker.track(AnalyticsEvent.ONBOARDING_UPGRADE_UNLOCK_ALL_FEATUERS_TAPPED) },
-                    onNotNowPressed = { analyticsTracker.track(AnalyticsEvent.ONBOARDING_UPGRADE_NOT_NOW_TAPPED) }
+            composable(OnboardingNavRoute.plusUpgrade) {
+                OnboardingPlusUpgradeFlow(
+                    onBackPressed = { navController.popBackStack() },
+                    onNotNowPressed = { navController.navigate(OnboardingNavRoute.welcome) },
+                    onComplete = { navController.navigate(OnboardingNavRoute.welcome) },
+                )
```
mchowning, line 148: It felt to me like this navigation component shouldn't be so concerned with analytics, so I moved the analytics logic into the composable's view model. I'm planning to make this change to the other components this navigation component navigates to in a follow-up PR, so let me know if you think this isn't a good change.

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingPlusFeaturesPage.kt
```diff
@@ -134,11 +142,13 @@ fun OnboardingPlusFeaturesPage(
                 Spacer(Modifier.weight(1f))
                 Spacer(Modifier.height(36.dp))
 
-                PlusRowButton(
-                    text = stringResource(LR.string.onboarding_upgrade_unlock_all_features),
-                    onClick = onUpgradePressed,
-                    modifier = Modifier.padding(horizontal = 24.dp),
-                )
+                if (canUpgrade) {
```
mchowning, line 145: This `canUpgrade` variable should never be false because it will only be false if there aren't any subscriptions available. I have it just as a bit of a safety check so that if for some reason a user doesn't see any subscriptions instead of asking them to select from a list of 0 subscriptions, the only option we give them is to tap on "Not Now". That's still not a great experience, but it seems a lot better than the alternative and, like I mentioned, it shouldn't really ever come up.

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/viewmodel/OnboardingPlusBottomSheetViewModel.kt
```diff
@@ -0,0 +1,161 @@
...
+    private val subscriptionManager: SubscriptionManager,
+    private val analyticsTracker: AnalyticsTrackerWrapper,
+) : ViewModel() {
+
+    private val _state = MutableStateFlow<OnboardingPlusBottomSheetState>(Loading)
+    val state: StateFlow<OnboardingPlusBottomSheetState> = _state
+        .map {
+            // If selected subscription has trialPricingPhase, update mostRecentlySelectedTrialPhase
+            (it as? Loaded)?.selectedSubscription?.trialPricingPhase?.let { trialPhase ->
+                it.copy(mostRecentlySelectedTrialPhase = trialPhase)
```
mchowning, line 40: Updating the `mostRecentlySelectedTrialPhase` field here in the flow means that we don't have to remember to check and see if we need to update it anytime we change the `selectedSubscription`. Instead, the `mostRecentlySelectedTrialPhase` gets updated automatically by the Flow.

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/viewmodel/OnboardingPlusBottomSheetViewModel.kt
```diff
@@ -0,0 +1,161 @@
...
+                selectedSubscription = defaultSelected,
+                purchaseFailed = false,
+            )
+        }
+    }
+}
+
+sealed class OnboardingPlusBottomSheetState {
+    object Loading : OnboardingPlusBottomSheetState()
+    object NoSubscriptions : OnboardingPlusBottomSheetState()
```
mchowning, line 141: `No Subscriptions` should never really happen.

ashiagr:
I didn't do a full review but did go through the "Flicker on non-selected subscription" issue that you mentioned. 

I noticed that by replacing `Column` with `LazyColumn`, the ripple effect is gone altogether.
```
LazyColumn(
    verticalArrangement = Arrangement.spacedBy(16.dp)
) {
    items(subscriptions) {subscription ->
    if (subscription == state.selectedSubscription) {
        PlusOutlinedRowButton(...)
    } else {
        UnselectedPlusOutlinedRowButton(...)
    }
}
```

https://user-images.githubusercontent.com/1405144/203066322-5834ff24-8fc3-41c5-98a5-8a0aa6cedca7.mp4

But maybe we don't need a ripple as the selected `PlusOutlinedRowButton` already has its own style.
I also came across [this issue](https://issuetracker.google.com/issues/189971666), not sure if you're already aware of it. It mentions about issues in touch feedback when list items don't have a key set on them.

I'll continue with the full review tomorrow unless @geekygecko beat me to it.

mchowning:
> I noticed that by replacing Column with LazyColumn, the ripple effect is gone altogether.

That is interesting. This got me digging into this a bit more, and I think I figured out the problem and fixed it in [0885682](https://github.com/Automattic/pocket-casts-android/pull/594/commits/08856828c8bbd48f2685fb848aa334b1d86a16e0). I believe the issue is that the `MutableInteractionSource` instance was getting applied to different items in the column when the views were being selected/unselected. In essence, I think that compose was behaving as though we were moving views around on the column as opposed to us just changing the state of the item at a given position. 

The way to avoid this and ensure that the same `MutableInteractionSource` is getting applied to the same position in the column was to remember the `MutableInteractionSource` at the level of the column instead of the default behavior of them being remembered [at the level of the Button component](https://github.com/androidx/androidx/blob/androidx-main/compose/material/material/src/commonMain/kotlin/androidx/compose/material/Button.kt#L95).

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingFlowComposable.kt
ashiagr: LGTM 👍

ashiagr:
> The fact that I accidentally had a duplicate call to viewModel.onBackPressed() is a bit concerning. I still think I may like it better than the alternatives though. 🤔

I have no strong opinion so let's go with what you like 🙂

PR #596, opened 2022-11-21
Speed up the listening history sync

geekygecko:
## Description

My listening history sync for the end of year stats was taking 94 seconds, this change speeds it up to 14 seconds. I did this by adding podcasts in parallel and instead using the full podcast JSON to load the episodes rather than calling the server for each individual episode.
Also I have changed the loading page from a circle progress spinner to a linear progress bar. This makes it feel like there is an end rather than with the circular spinner it feels like it's never going to end. 

## Testing Instructions
2. Fresh install the app
3. Login to your sync account
4. Re-run the app
5. Tap "View my 2022"
6. ✅ The linear progress bar should appear, followed by the end of year story

## Screenshots or Screencast 

https://user-images.githubusercontent.com/308331/202955649-914ac5fa-5976-4032-af4e-9656650d8bdc.mp4


ashiagr (approved):
Sync speed is remarkably increased, thank you for this effort @geekygecko!

PR #598, opened 2022-11-21
End of Year: Enable feature

ashiagr:
| 📘 Project: #410 |
|:---:|

This PR enables EoY feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,9 @@
 -----
 
 *   New Features:
-    *   Support Android 13 per-app language preferences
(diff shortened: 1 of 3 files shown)


mchowning (approved):
Retested the flow, and it looks good. Great work on this! 

Only question I have is whether we would be better off just removing the feature flag entirely. It's unlikely we're ever going to need it again, and It would be easy enough to add back in if we did need to turn this off for some reason.

ashiagr:
Feature flag removed, also added changelog entry.

Thank you for the final review! 🎉

PR #604, opened 2022-11-23
Onboarding: Recommendations Search

mchowning:
| 📘 Project: #524  |
| --- |

## Description

Adding search button functionality to the onboarding recommendations screen

### 1. Triggering search immediately

I've added the ability for the user to avoid the debounce on the search by tapping the search key on the soft keyboard in 02dd06635dfe33cb3b53ec2992b8a7a9a25900e8. This is a new behavior for the app, but I thought that if the user went to the trouble of tapping the search key, it would be a better user experience and not too much of a strain on our servers to kick off the search immediately.

### 2. Subscription state slow to update

When tapping on a podcast to subscribe, It can take a bit of time for the podcast manager to reflect that the podcast has been subscribed to, particularly on slow internet connections. Since the UI is listening for these updates from the podcast manager, the UI can be slow to update.

We could avoid this delay by just immediately updating the UI without waiting for the podcast manager to update, but I am hesitant to risk the UI getting out of sync with the podcast manager (i.e., if the subscription update failed for some reason such as a lack of an internet connection, the UI would indicate it had succeeded), so I didn't do this. Let me know if you have any thoughts here though.

## Testing Instructions

1. Turn on the onboarding flag in `base.gradle`
2. When not logged into the app, start the app and create a new account so the recommendations screen is loaded. (Please feel free to cheat here by making [these changes](https://gist.github.com/mchowning/4d9a754b93a7b57e1d7f98c0910bd191) so you can be logged in and go straight to the recommendations screen).
3. Tap on the search field
4. ✅  Observe that the full screen search view is opened, with the search bar focused and the keyboard open
5. Type one letter in the search field
6. ✅ Observe that the search does not start
7. Type a second letter into the search field
8. ✅ Observe that after a delay, the spinning loading indicator is shown until the search results are displayed (it's hard to see the loading indicator if your internet connection is fast)
9. Type a third letter in the search field
10. ✅ Observe that after a delay, a subtle linear loading indicator is shown at the top of the search results until the search results are updated. This is just something I added because I've been confused a few times thinking the app wasn't searching when it was--I thought this seemed like a good way to give some small visual indicator that something was happening without being distracting, but let me know what you think.
11. Type a fourth letter in the search field and immediately tap the search time action button on the soft keyboard
12. ✅ Observe that the search begins immediately (i.e., there is no debounce delay). **This immediate search is a new behavior for the app**, but I thought that if the user went to the trouble of tapping the search key, it would be a much better user experience and not too much of a strain on our servers to kick off the search immediately.
13. Tap on a few podcasts to subscribe and unsubscribe them.
14. ✅ Observe that the podcast row updates with a checkmark when it is subscribed and that the checkmark is removed when it is unsubscribed (note my concern about about this being slow when the internet connection is slow).
15. Swipe back twice to return to the main app.
16. ✅ Observe that the subscription changes you made in the search view are reflected in the podcasts tab of the main app.

## Screenshots or Screencast 

https://user-images.githubusercontent.com/4656348/203468654-fe2fea3f-2c1e-4c9b-9e18-8ad4fc665bf2.mp4

## Checklist
 
- [X] with different themes

### DIFF
--- a/modules/features/account/build.gradle
+++ b/modules/features/account/build.gradle
@@ -16,6 +16,7 @@ dependencies {
     // features
     implementation project(':modules:features:cartheme')
     implementation project(':modules:features:settings')
+    implementation project(':modules:features:search')
 
     // services
     implementation project(':modules:services:analytics')
--- a/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingFlowComposable.kt
+++ b/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingFlowComposable.kt
@@ -7,6 +7,7 @@ import androidx.navigation.compose.NavHost
 import androidx.navigation.compose.composable
 import androidx.navigation.compose.currentBackStackEntryAsState
 import androidx.navigation.compose.rememberNavController
+import au.com.shiftyjelly.pocketcasts.account.onboarding.recommendations.OnboardingRecommendationsFlow
 import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsEvent
 import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsTrackerWrapper
 import au.com.shiftyjelly.pocketcasts.compose.AppThemeWithBackground
@@ -76,7 +77,7 @@ fun OnboardingFlowComposable(
                         navController.popBackStack()
                     },
                     onAccountCreated = {
-                        navController.navigate(OnboardingNavRoute.recommendations) {
+                        navController.navigate(OnboardingNavRoute.recommendationsFlow) {
                             // clear backstack after account is created
                             popUpTo(OnboardingNavRoute.logInOrSignUp) {
                                 inclusive = true
@@ -113,8 +114,8 @@ fun OnboardingFlowComposable(
                 )
             }
 
-            composable(OnboardingNavRoute.recommendations) {
-                OnboardingRecommendations(
+            composable(OnboardingNavRoute.recommendationsFlow) {
+                OnboardingRecommendationsFlow(
                     onShown = {
                         analyticsTracker.track(AnalyticsEvent.RECOMMENDATIONS_SHOWN)
                     },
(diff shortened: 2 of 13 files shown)


ashiagr (approved):
Nice work here! 

> Observe that after a delay, a subtle linear loading indicator is shown at the top of the search results until the search results are updated. This is just something I added because I've been confused a few times thinking the app wasn't searching when it was--I thought this seemed like a good way to give some small visual indicator that something was happening without being distracting, but let me know what you think.

I really like the linear loading indicator on top of the existing search results. It was very thoughtful of you to consider adding separate progress indicators based on whether search results are shown or not.

> I've added the ability for the user to avoid the debounce on the search by tapping the search key on the soft keyboard in https://github.com/Automattic/pocket-casts-android/commit/02dd06635dfe33cb3b53ec2992b8a7a9a25900e8. This is a new behavior for the app, but I thought that if the user went to the trouble of tapping the search key, it would be a better user experience and not too much of a strain on our servers to kick off the search immediately.

Makes sense to me 👍 

> When tapping on a podcast to subscribe, It can take a bit of time for the podcast manager to reflect that the podcast has been subscribed to, particularly on slow internet connections. Since the UI is listening for these updates from the podcast manager, the UI can be slow to update.

> We could avoid this delay by just immediately updating the UI without waiting for the podcast manager to update, but I am hesitant to risk the UI getting out of sync with the podcast manager (i.e., if the subscription update failed for some reason such as a lack of an internet connection, the UI would indicate it had succeeded), so I didn't do this. Let me know if you have any thoughts here though.

While I'm fine with a small delay here, it wasn't very clear to me that tapping the search result row means subscribing to the podcast. In the designs, I see there's a plus button to indicate it. Can we add it to the search result rows?

Screenshot 2022-11-23 at 5 18 31 PM

mchowning:
>  In the designs, I see there's a plus button to indicate it. Can we add it to the search result rows?

Yes, good catch. Added in [1492adb](https://github.com/Automattic/pocket-casts-android/pull/604/commits/1492adb6e31b88d360a88d564426d2fa2c277966).

PR #610, opened 2022-11-24
Google Sign-In call intent and handle result

geekygecko:
## Description

This change calls an Intent for the user to sign into Pocket Casts with their Google account. It doesn't include the next part of using their Google ID token to sign into Pocket Casts, this will be part of a separate PR.

The One Tap for Android library is used. Sign in doesn't work when no Google accounts are set up on the device. In this case, fallback to the legacy Google Sign-In for Android.

## Testing Instructions

1. Update the ONBOARDING_ENABLED and SINGLE_SIGN_ON_ENABLED build flags in base.gradle to be "true".
2. Open the app on an emulator with Google Play Services installed, without a Google account logged in.
3. Tap the "Continue with Google" button.
4. ✅ Verify you are taking to sign in with a Google account. (This will use the legacy Google Sign-In)
5. Deploy the app to a real device
6. Tap the "Continue with Google" button.
7. ✅ Verify you are taking to sign in with a Google account. (This will use the new One Tap Google sign-in)


ashiagr:
Nice work, @geekygecko!  I tested the changes and found everything as described.

I had to add the missing `googleSignInServerClientId` inside `dependencies.gradle` (similar to other properties like `pocketcastsSentryDsn`) in my local setup, otherwise, I got this error:

`MissingPropertyException: Could not get unknown property 'googleSignInServerClientId' for project ':app' of type org.gradle.api.Project.`

It might also be causing CI checks to fail. Can you too check if adding the missing property is all we need to do?

geekygecko:
@ashiagr hopefully I have fixed this issue now.

ashiagr (approved):
LGTM! 👍

PR #619, opened 2022-11-30
End of Year: Design tweaks (part ii)

ashiagr:
| 📘 Project: #410 |
|---|

## Description

Applies small design tweaks:

Intro Screen 
- Moved of to the second line to avoid widow in listening... ("en" lang only)
- Used an exclamation point instead of ellipses
- The 2022 logo moved downwards so it’s optically better centered

Generic
- Replaced apostrophe in Let's with a typographic one
- Moved the Pocket Casts logo downwards, so there isn’t such a big empty space between it and the Share button
- Covers made bigger `ListenedCategories`, `LongestEpisode`, `TopPodcast` and `ListenedNumbers` stories

Top Categories
- Applied 3D effect to pillar

|Top Categories|
|----|
||

Epilogue
- Added pulsating animation to the heart vector
- Replaced Pocket Casts logo (with a red circle)

I also captured an `UnknownFormatConversionException` in [7.27](https://sentry.io/organizations/a8c/issues/3771206667/?project=6711064) in e4fb26e. It is for the Arabic Letter Meem (م). The letter is used inside values-ar/strings.xml in several places. I'll get that fixed in a separate PR after consulting someone from the i18n team.


modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/views/stories/StoryEpilogueView.kt
```diff
@@ -52,7 +59,7 @@ fun StoryEpilogueView(
 
         Spacer(modifier = modifier.weight(1f))
 
-        HeartImage()
+        PulsatingHeart()
```
mchowning, line 62: Nice!

mchowning (approved):
I'll let @adamzelinski take a look at this from a UI perspective, but the code looks good.

The only thing I noticed that seemed off was that on the two screens where we removed the Share button, the content does not expand to take up the Share Button space. This is visible with a normal font size (because the content does not have a black background like the space where the share button does), but it is much more obvious when a larger font size is used where the user can scroll.



https://user-images.githubusercontent.com/4656348/204876063-11012035-83c6-46e4-b7d6-3e3d688da421.mp4

Also seems like the spacing gets rather crowded with the larger font size.

modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/views/stories/StoryEpilogueView.kt
CookieyedCodes: The only thing I think it needs matt is the last story if you try to skip the story it just loops the last story it should close the story 🤔 😉

(other then a story top 15-25 podcasts the current story should say add the time for the each of the top 5😉, theirs allways next year tho),
 (last story should also collect all the story's and allows sharing of the 9 as a group as thats better for social engagement )

adamzelinski:
Thanks for these updates @ashiagr it's looking great and I love the new heart animation! As for the text size on the screenshot, yes it looks funny if it scrolls. On iOS, we're using dynamic type but there is also a max size set so maybe we could look at doing something similar?

Also, in iOS on the first and last screen now that the share button is hidden we have expanded those stories 
2022 Year In Podcasts 1
2022 Year In Podcasts 10

ashiagr:
Thank you, for the feedback!
> Also, in iOS on the first and last screen now that the share button is hidden we have expanded those stories

To match iOS, removed share button on Android for those screens. - eff6611

> The only thing I noticed that seemed off was that on the two screens where we removed the Share button, the content does not expand to take up the Share Button space. This is visible with a normal font size (because the content does not have a black background like the space where the share button does), but it is much more obvious when a larger font size is used where the user can scroll.

> On iOS, we're using dynamic type but there is also a max size set so maybe we could look at doing something similar?

To improve experience, disabled text scaling if the device font scale is beyond 1.15 (popular values: 1.0, 1.15, 1.3) - 1e2ed1f

Available font scales from Tracks:
Screenshot 2022-12-01 at 2 55 47 PM

Intro | Epilogue
-----|-----
 | 



https://user-images.githubusercontent.com/1405144/205017355-0d84835b-781d-47a0-a7b0-99e863c78275.mp4



Feedback from Slack DM 

- Increase padding between artwork and text on `ListenedNumbers` story - a5a065c

Normal | Scaled
-----|----
| 


- Revert to the original logo on Epilogue screen - 71326bd

modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/views/stories/StoryEpilogueView.kt
ashiagr: Your feedback is valuable, @CookieyedCodes! Thanks for sharing it. 🙏  I'll add it to the backlog.

modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/components/TextStyles.kt
```diff
@@ -57,13 +57,16 @@ fun TextH20(
...
+    disableScale: Boolean = false,
+    fontSize: TextUnit = 22.sp,
+    lineHeight: TextUnit = 30.sp,
 ) {
     Text(
         text = text,
         color = color,
-        fontSize = 22.sp,
-        lineHeight = 30.sp,
+        fontSize = if (disableScale) fontSize.value.nonScaledSp else fontSize,
```
mchowning, line 68: One minor issue that I think this logic creates is that someone with a font scaled to 1.1 will have a bigger font than someone whose font is set to be 1.2 because the 1.2 gets reset to 1.0. _Very_ minor issue, but I wanted to mention it.

PR #623, opened 2022-12-01
End of Year: Fix modal keeps popping up when opening the app (logged out state)

### LINKED ISSUE #622: End of Year Review: Keeps popping up when opening the app
### Description

If you aren't logged into a Pocket Casts account, each 2nd time you open the app, you are prompted to view the 'Your Year in Podcasts' page. Pressing 'Not Now' only temporarily hides it until you open the app again. This does not happen when you are signed into an account

1. Sign out of your Pocket Casts account
2. Close the app
3. Re-open it and notice the pop-up
4. Press 'Not Now'
5. Close the app
6. Re-open the app, close it again
7. Notice on the second time you re-open the app, the Pop-up is back

https://user-images.githubusercontent.com/16253818/205007370-16ae526a-0eb3-426a-9e47-bf8a16066d2e.mp4

Samsung S21 FE, Android 12, Pocket Casts v7.27-rc-3

ashiagr:
## Description
This fixes EoY modal that keeps popping up when opening the app when the user is not logged into the app.

Fixes #622

## Testing Instructions

1. Sign out of your Pocket Casts account
2. Close the app
3. Re-open it and notice the pop-up
4. Dismiss the popup
5. Close the app
6. Re-open the app, close it again
7. Notice on the second time you re-open the app, the Pop-up is not reshown

## Screenshots or Screencast 

https://user-images.githubusercontent.com/1405144/205042056-b57bc883-a258-4bb3-b2f3-e9e5848854d5.mp4

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/user/UserManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/user/UserManager.kt
@@ -108,6 +108,7 @@ class UserManagerImpl @Inject constructor(
 
         settings.setMarketingOptIn(false)
         settings.setMarketingOptInNeedsSync(false)
+        settings.setEndOfYearModalHasBeenShown(false)
         analyticsTracker.track(AnalyticsEvent.USER_SIGNED_OUT, mapOf(KEY_USER_INITIATED to wasInitiatedByUser))
         analyticsTracker.flush()
         analyticsTracker.clearAllData()
--- a/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
+++ b/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
@@ -594,8 +594,6 @@ class MainActivity :
                 } else if (!settings.getEndOfYearModalHasBeenShown()) {
                     viewModel.updateStoriesModalShowState(true)
                 }
-            } else {
-                settings.setEndOfYearModalHasBeenShown(false)
             }
 
             if (signinState.isSignedInAsPlus) {


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/user/UserManager.kt
```diff
@@ -108,6 +108,7 @@ class UserManagerImpl @Inject constructor(
 
         settings.setMarketingOptIn(false)
         settings.setMarketingOptInNeedsSync(false)
+        settings.setEndOfYearModalHasBeenShown(false)
```
ashiagr, line 111: This reset is still needed to matches iOS behavior to show the modal again when a user login (or creates an account): https://github.com/Automattic/pocket-casts-ios/pull/474/commits/2903fada733fb0b86a9543726aee27f269b28cb2

The difference now is that it is done on explicit signout rather than when `SignInState` is `SignedOut` when the app is restarted.

mchowning: I ran across 2 edge cases here that I wasn't sure about.

#### 1. Signing in frequently doesn't trigger the dialog

Just noting that this only works if the user does not close the app between signing out and log-in/creating-an-account. If they do restart the app, then the pop-up displays when the app restarts with them signed out, but does not display when they sign in.

Similarly, if a user:
1. Freshly installs the app
2. Listens to 5+ minutes of podcast without signing in
3. Restarts the app
4. They will see the stats popup
5. Signs in with an account that adds a bunch of new listening history
6. Restarts the app
7. They will not see the pop-up

These seem a bit like edge cases, so I don't think they're blockers at all. And maybe the current behavior is actually what we want, I just wasn't sure since a user who signs in could have a lot of new data.

#### 2. Signing in and then Signing out immediately shows the dialog

1. Open the app with more than 5 minutes of listening time
2. See the End of Year dialog
3. Sign into an account
4. Sign out of the account
5. Observe that the EOY dialog is immediately shown after sign out
6. Sign into an account
7. Observe that the EOY dialog is not shown

Would we rather the dialog show at step 7 instead of step 5?

mchowning (approved):
This does address the dialog keeps presenting issue for logged out users, so I'm approving it. I did leave a comment regarding some edge cases that I'm not sure we're handling the way we want though.

ashiagr:
Thanks, @mchowning! I'll let this fix go into 7.27.1.

There's also another observation I noted in Slack (p1669896264509019-slack-C041BGTFQ5R) regarding modal display.  I'll go through these edge cases and if not the intended behavior, will fix them in a separate PR.

PR #624, opened 2022-12-01
Add disable scale to replay button

mchowning:
## Description
Disables scaling to large font sizes on the replay button so that it is consistent with the rest of the story view. Follow-up on changes in https://github.com/Automattic/pocket-casts-android/pull/619. If this was intended behavior, just close this PR. I just figured it would be quicker to put up a PR in case we do want that change. cc: @adamzelinski 

> **Note**
> This PR is targeting `release/7.21.1`, but I also think it would be fine to delay this to the next release since this is a minor issue.

## Testing Instructions
1. Set device font size to a large setting
2. View EOY epilogue Story
3. ✅  Observer that the replay button has a consistent size with the rest of the text

## Screenshots or Screencast 

| Before | After |
| --- | --- |
|  image | image |

## Checklist
 
- [X] with different themes


ashiagr (approved):
Thank you! LGTM.

PR #625, opened 2022-12-01
Merge/release 7.27.1 to main

ashiagr:
Includes #623  and #619


version.properties
```diff
@@ -1,3 +1,3 @@
 # Version Information for Vanilla / Release builds
-versionName=7.28-rc-1
-versionCode=9054
+versionName=7.27.1
```
mchowning, line 2: This won't mess up the remaining 7.28 release, right? I can't remember.

mchowning (approved):

version.properties
ashiagr: I'll shortly submit `7.28-rc-2` merging these changes with a higher version code. So it won't mess up. 🙂

PR #632, opened 2022-12-06
Allow syncing of skip time settings

### LINKED ISSUE #303: Unable to permanently change "Skip back time" settings
I'm grouping a few recent reports of users on Android not being able to permanently save the skip settings. 

I haven't been able to replicate this on my end. 

[5516359-zd](https://woothemes.zendesk.com/agent/tickets/5516359)

App version : 7.21.0
Phone: samsung - Galaxy S8 - SM-G950U 
Android version: 9 SDK 28

> In the screenshots below, you can see where I changed the "skip back time" from 12 seconds to 25 seconds. I've tried making the change at least 10-15 times now, and restarted my phone, and it always reverts back.

[5534334-zd](https://woothemes.zendesk.com/agent/tickets/5534334)

App version : 7.20.3 
Phone: samsung - Galaxy Z Fold3 5G - SM-F926U1 
Android version: 12 SDK 31

> I keep changing the Skip backward from 30 sec to 15 and it isn't saving. Keeps going back to 30 as soon as I switch apps and it runs in the background.

[5539341-zd](https://woothemes.zendesk.com/agent/tickets/5539341)

App version : 7.21.0
Phone: Google - Pixel 4a - Pixel 4a 
Android version: 13 SDK 33

> The skip control settings, keep resetting to 30 forwards and 30 backwards after the phone screen is off for a few minutes.
I tried to set it to 30 forwards, and 15 backwards.

ashiagr:
Fixes #303

## Description

This allows syncing of skip time settings _to the server_ for logged-in users. The app [supported syncing of skip time settings](https://github.com/Automattic/pocket-casts-android/blob/main/modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/sync/SyncSettingsTask.kt#L15-L25) to the servers, but they were not synced as the `..NeedsSync` flags were not set for skip time settings. This PR sets those flags whenever skip time settings are changed. Those settings are synced to the server whenever syncing is triggered either by the user or after a pre-defined interval by the app.

Note that syncing _from the server_ worked fine which explains why the issue was resolved for those who updated the web player setting https://github.com/Automattic/pocket-casts-android/issues/303#issuecomment-1322919716.

P.S. For logged-out users, the app's preferences are auto-backed up by Google roughly every night when the device is idle according to their [backup schedule](https://developer.android.com/guide/topics/data/autobackup#BackupSchedule). If they change those settings after backup but before [restore](https://developer.android.com/guide/topics/data/autobackup#RestoreSchedule), which happens when the app is installed, either from the Play Store, or during device setup, they’re overridden with the last backup values. Backup and restore schedule in this scenario is controlled by Google, and looks like we cannot do much about it.

## Testing Instructions

1. Login to your `Pocket Casts` account on the web player and note down skip time settings. 
2. Install and launch the app.
3. Go to `Profile` -> `Settings` -> `General`.
4. Update `Skip forward time` and `Skip backward time` under player settings. 
5. Return to `Profile` home tab and tap on "Refresh now" at the bottom of the screen.
6. Check skip time settings on the web.
7. Notice that they're updated to values set in the Android app.


modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/PlaybackSettingsFragment.kt
```diff
@@ -155,7 +155,10 @@ class PlaybackSettingsFragment : BaseFragment() {
                         saved = settings.skipForwardInSecsObservable
                             .subscribeAsState(settings.getSkipForwardInSecs())
                             .value,
-                        onSave = settings::setSkipForwardInSec
+                        onSave = {
+                            settings.setSkipForwardNeedsSync(true)
+                            settings::setSkipForwardInSec.invoke(it)
```
mchowning, line 160: Why are we manually calling invoke instead of calling `settings.setSkipForwardInSec(it)`? Same question for the skip forward changes? What you have works fine, but it's harder to read IMO, so I was just curious if maybe there's a reason to prefer this that I wasn't thinking of.

mchowning (approved):
This works great and fixes a super annoying bug. Great work @ashiagr !

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/PlaybackSettingsFragment.kt
ashiagr: I don't prefer it either. Went ahead and switched to `settings.setSkipForwardInSec(it)`.

PR #635, opened 2022-12-06
Onboarding: Convert recommendations flow to a nested navigation graph

mchowning:
| 📘 Project: #524  |
| --- |

This PR includes the changes in https://github.com/Automattic/pocket-casts-android/pull/634.

## Description

Updating the recommendation flow to use a proper nested graph. This also hooks up the import button, and removes some logic for possibly skipping the upgrade screen after recommendations (no longer needed because of a previous change whereby the recommendations screen will always be followed by the upgrade screen because you now only get to the recommendations screen after creating a new account).

Note that the UI for th initial recommendations screen is not yet completed.

Also, feel free to ignore any code for Tracks. I will be making changes to that in a follow-up PR that focuses on Tracks.

## Testing Instructions

1. Turn on the onboarding feature in `base.gradle`
2. Ensure that navigation with the recommendation screen  works as expected
    1. Going back should return you to the main app's dashboard (not back to the sign-up screen)
    2. Tapping Import should take you to the import screens
    3. Tapping the Search field should take you to a full screen search view
    4. Tapping "Not Now" Should take you to the upgrade screens

## Checklist
- [X If this is a user-facing change, I have added an entry in CHANGELOG.md
 
- [X] with different themes

### DIFF
--- a/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingFlowComposable.kt
+++ b/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingFlowComposable.kt
@@ -60,13 +60,7 @@ fun OnboardingFlowComposable(
                     completeOnboarding()
                 },
                 onComplete = {
-                    navController.navigate(
-                        if (signInState?.isSignedInAsPlus == false) {
-                            OnboardingNavRoute.plusUpgrade
-                        } else {
-                            OnboardingNavRoute.welcome
-                        }
-                    )
+                    navController.navigate(OnboardingNavRoute.plusUpgrade)
                 },
                 navController = navController,
             )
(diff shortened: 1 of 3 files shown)


ashiagr (approved):
Tested recommendations flow navigation and found it working as described in the testing steps.

Can you remind me what should be the flow when I login with an account without Plus from the onboarding launch screen? I'm assuming that the recommendations screen should be shown according to [these testing steps](https://github.com/Automattic/pocket-casts-android/pull/594) but looks like the app takes me straight to the discover tab. 🤔 With a new account, I could see the recommendations screen.

mchowning:
> Can you remind me what should be the flow when I login with an account without Plus from the onboarding launch screen? I'm assuming that the recommendations screen should be shown according to https://github.com/Automattic/pocket-casts-android/pull/594 but looks like the app takes me straight to the discover tab. 🤔 With a new account, I could see the recommendations screen.

Sorry I didn't call this out more clearly. Those testing steps were a mistake on my part because users who log in should go straight to the app dashboard; only users who are creating a new account should go through the rest of the flow (at least that's what it looks like to me from the Figma designs for iOS, @adamzelinski please let me know if I'm mistaken). I've fixed this now (as you observed).

I did add a warning to the PR description on that PR to hopefully help reduce this confusion going forward:


image

adamzelinski:
Yes thanks @mchowning only users who are creating a new account should go through the rest of the flow

PR #636, opened 2022-12-06
Onboarding: FIx plus upgrade page to alwasy be full screen

mchowning:
| 📘 Project: #524  |
| --- |

## Description
This fixes a bug where the Plus upgrade screen's content would not always expand to the full-height of the screen. Apparently this is due to a bit of weirdness in how heights are calculated inside a scrollable composable (see this [Stackoverflow discussion](https://stackoverflow.com/questions/69394543/fillmaxsize-modifier-not-working-when-combined-with-verticalscroll-in-jetpack-co)).

## Testing Instructions
1. Turn on the onboarding flow in `base.gradle` and if you want to save yourself some time, update the startDestination [here](https://github.com/Automattic/pocket-casts-android/blob/2c6b39dccba6eb67aa643e17e59fdc3df59db437/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingFlowComposable.kt#L49) to be `OnboardingNavRoute.plusUpgrade`. 
2. Update your phone to have the smallest possible font size
3. ✅  Verify that the plus screen button(s) are at the bottom of the screen
4. Update your phone to have the largest possible font size
5. ✅ Verify that the plus screen can be scrolled

## Screenshots or Screencast 

| Before | After |
| --- | --- |
| image | image |
* Note that only the "Not Now" button is showing in these screen shots because this was a debug build that could not retrieve any subscriptions from the Google Play store.

## Checklist
 
- [X] with different themes

### DIFF
--- a/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingFlowComposable.kt
+++ b/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingFlowComposable.kt
@@ -8,6 +8,7 @@ import androidx.navigation.compose.composable
 import androidx.navigation.compose.currentBackStackEntryAsState
 import androidx.navigation.compose.rememberNavController
 import au.com.shiftyjelly.pocketcasts.account.onboarding.recommendations.OnboardingRecommendationsFlow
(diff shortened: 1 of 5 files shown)


ashiagr (approved):
Nice fix!

PR #645, opened 2022-12-09
Onboarding: remove feature flag

mchowning:
| 📘 Project: #524  |
| --- |

## Description
This turns on the onboarding feature so it can go out to our open beta testers in 7.29.

I considered leaving this turned off for debug builds by default and having a feature flag that could be used to turn it on for debug builds in order to not make development more difficult, but since this only gets presented a single time after a fresh install (and can be dismissed with a single button tap), I decided that probably wasn't necessary. Happy to add a flag if you disagree though.

## Testing Instructions
Just make sure the feature loads up on a fresh install. Everything else has already been tested, so no other specific testing is needed for this change. Don't hesitate to give the feature a bit of smoke-testing if you're up for it though.

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/Settings.kt
+++ b/modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/Settings.kt
@@ -604,6 +604,6 @@ interface Settings {
     fun getEndOfYearModalHasBeenShown(): Boolean
     fun endOfYearRequireLogin(): Boolean
 
-    fun getHasCompletedOnboarding(): Boolean
+    fun hasCompletedOnboarding(): Boolean
     fun setHasCompletedOnboarding()
 }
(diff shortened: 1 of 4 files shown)


modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/Settings.kt
```diff
@@ -604,6 +604,6 @@ interface Settings {
     fun getEndOfYearModalHasBeenShown(): Boolean
     fun endOfYearRequireLogin(): Boolean
 
-    fun getHasCompletedOnboarding(): Boolean
+    fun hasCompletedOnboarding(): Boolean
```
mchowning, line 607: Just made this change because I thought the slightly shorter name read better.

ashiagr (approved):

mchowning:
Updated this with [a09bb85](https://github.com/Automattic/pocket-casts-android/pull/645/commits/a09bb85cdee65f46d8e85c23d87ef76e2af7c1f9) to only create the onboarding activity if `savedInstanceState` is null. This is to avoid creating multiple onboarding activities if `MainActivity::onCreate` is called twice. This has only been recreated on emulators (in my testing, an "identical" physical device did not show the issue whereas the emulator did), so this may just be an emulator issue, but this seemed like a safe check to put in to help protect us.

I believe this was the cause of the [double-back issue](https://github.com/Automattic/pocket-casts-android/pull/603#pullrequestreview-1191350711) you found earlier @ashiagr .

I'm re-requesting your review because I want to make sure this change doesn't introduce any issues.

ashiagr:
After this change, I've verified that multiple onboarding activities are not created on the emulator. Thanks for taking care of it. 👍

PR #646, opened 2022-12-10
Onboarding analytics: sign in, create account, forgot password, and recommendations

mchowning:
| 📘 Project: #524  |
| --- |

## Description
This is moving the analytics logic out of the navigation composable and into the view models for the specific pages. I'm making that change because I don't think it makes sense for the navigation composable to be responsible for analytics like I had it before. Not only does it make the navigation composable more complex, but it also makes it more difficult to reuse the composable views, which we'll be doing once we replace the main app account-creation and plus-upgrade screens with the new screens from the onboarding flow.

### Sign up (Create Account) Screen
1. On load, the `create_account_shown` event is sent
2. Either (a) swiping back or (b) tapping the back arrow sends `create_account_dismissed`

### Log in Screen
1. On load, the `signin_shown` event is sent
2. Either (a) swiping back or (b) tapping the back arrow sends `sigin_dismissed`

### Forgot Password Screen (accessible from Log in screen)
1. On loead, the `forgot_password_shown` event is sent
2. Either (a) swiping back or (b) tapping the back arrow sends `forgot_password_dismissed`

### Recommendation Screen (accessible after creating a new account)
1. On load, the `recommendations_shown` event is sent
4. Swiping back sends `recommendations_dismissed, Properties: {"subscriptions":###`
5. Tapping the search field sends `recommendations_search_tapped`
6. Tapping the import button sends `recommendations_import_tapped`
7. Tapping the not-now/continue button sends `recommendations_continue_tapped, Properties: {"subscriptions":###`

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingActivity.kt
+++ b/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingActivity.kt
@@ -8,7 +8,6 @@ import androidx.appcompat.app.AppCompatActivity
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
 import au.com.shiftyjelly.pocketcasts.account.onboarding.OnboardingActivityContract.OnboardingFinish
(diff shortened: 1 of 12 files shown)


ashiagr (approved):
LGTM 👍

PR #651, opened 2022-12-13
EoY: Update initial value for the launch modal to check listening history

ashiagr:
## Description

Adds a safety check so that EoY modal is not shown to newly signed-up users (who do not have a listening history) after the onboarding flow is complete.

While this check is added in MainActivity [here](https://github.com/Automattic/pocket-casts-android/blob/main/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt#L202), I noticed that EoY modal was shown after creating a new account through the onboarding flow. I updated the initial value for showing the launch modal to consider the listening history eligibility.

## Testing Instructions
1. Create a new account using the onboarding flow.
2. Proceed to the welcome screen and click `Done` button.
3. Notice that EoY modal is not shown. 

## Checklist
 
- with different themes


mchowning (approved):
Good catch. This does work better.

PR #657, opened 2022-12-15
Enable autolink for podcast descriptions

TylerCarberry:
## Description
Enable autolink on podcast descriptions to allow you to click on urls. This should also support email addresses, but I haven't found any podcasts to test that on.

## Testing Instructions
1. Open a podcast with a link in their description. Up First and Waveform have them.
2. Ensure you can tap on them to open the browser
3. Ensure the color of these links matches the tintColor of the podcast

## Checklist

## Before & After
Before  After

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,9 +37,9 @@
 
 *   New Features:
     * Add End of Year stats
-      ([#410](https://github.com/Automattic/pocket-casts-android/issues/410)).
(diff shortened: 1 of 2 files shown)


CookieyedCodes:
Nice instead of the typical blue could we get links to match the theam not, you might have allready done this tho 🤔 😉

ashiagr (approved):
Awesome additon! Thank you so much 🙌 

Updated screenshots:

Before| Now
---|---
 |

PR #659, opened 2022-12-15
Reduce size of "Not Now" button

mchowning:
| 📘 Project: #524  |
| --- |

## Description
I noticed that the designs call for the "Not Now" button to be H40 instead of H30. This makes that update, but take a look and let me know what you think because although the designs specify H30, the button in the designs uses a slightly larger font (16) than we have for H30 [in the app](https://github.com/Automattic/pocket-casts-android/blob/0db6f36f113d771ece6b5f5a62139cf4d27f0afe/modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/components/TextStyles.kt#L138) (15).

Related discussion: p1671091615388669-slack-C02A333D8LQ

## Testing Instructions
Just see how the button looks. Let's hold off on merging this until @adamzelinski chimes in with his thoughts.

## Screenshots or Screencast 
| Before (H30) | After (H30) |
| --- | --- |
| Screenshot 2022-12-15 at 8 00 39 AM | Screenshot 2022-12-15 at 7 59 33 AM |

## Checklist
 
- [X] with different themes


ashiagr (approved):

mchowning:
Closing this in favor of https://github.com/Automattic/pocket-casts-android/pull/666.

PR #665, opened 2022-12-16
Onboarding: Support login dynamic artwork

ashiagr:
| 📘 Project: #524  |
| --- |

## Description

This populates login artwork icons from local DB podcasts if they're available. 

## Testing Instructions

1. Comment out `lines 191, 193, and 195` in `MainActivity` so that onboarding flow is launched every time when app restarts
2. Login with an account with some podcasts
3. Restart the app
4. Notice that 
    - Artwork is populated with icons from random podcasts available in local DB
    - If local DB podcasts count is less than the max count (6), then default icons are shown

## Checklist
 
- with different themes


mchowning (approved):
This works great! Thanks for taking care of this!

PR #668, opened 2022-12-19
Lower version of accompanist

mchowning:
## Description
Using the `0.28.0` version of Accompanist's SystemUiController was increasing our compose version transitively from `1.2.1` to `1.3` (see note [here](https://github.com/google/accompanist#compose-versions)). I didn't intend to make that change in a beta release, so bumping that version down to `0.25.1` to get us back to `1.2.1`.

## Testing Instructions
Just do a quick smoke test of the onboarding flow to verify that it still works as expected.

## Checklist

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -79,7 +79,8 @@ project.ext {
 
     // Library versions
     versionBillingClient = '5.0.0'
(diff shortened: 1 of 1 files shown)


ashiagr (approved):
LGTM!

PR #675, opened 2023-01-04
Add categories to recommendations screen

mchowning:
| 📘 Project: #524  |
| --- |

## Description
Adds rows for our categories to the onboarding recommendations screen based on @adamzelinski 's [feedback](https://github.com/Automattic/pocket-casts-android/issues/524#issuecomment-1347960181).

I implemented this so that the sections are loaded one at a time. I thought this would be good for users with slow connections so that they didn't have to either wait for all the sections to load or have the sections moving around on them as the network responses came back. Let me know if you think there's a better way to handle this.

## Testing Instructions
1. Open a fresh install of the app so that the onboarding flow is triggered. You may want to also [make this change](https://gist.github.com/mchowning/f60082a45df993dd7c0fffbecb60585b) so that the onboarding flow just goes straight to the recommendations screen. Otherwise, navigate through the flow to the recommendations screen.
2. From the recommendations screen, verify that 6 podcasts are shown for each section, first "Trending" and then one section for each "category" from our Discover feed.
3. Verify that tapping the "More ____" button adds an additional 6 podcasts to that section and that the following event is sent: `recommendations_more_tapped, Properties: {"section":"trending","number_visible":6, ...`.
4. Verfiy that landscape mode also looks good and the "More ___" button works as expected.
5. Change your device's language
6. Observe that the titles of the sections change to the new language
7. Observe that the `section` part of the analytics event that gets sent when you tap one of the "More ____" buttons does _**not**_ get translated (so they remain consistent regardless of the device's language).

## Screenshots or Screencast 

https://user-images.githubusercontent.com/4656348/210627440-ff23f1e3-f1ea-43d6-bd88-db4385f81685.mov

## Checklist
 
- [X] with different themes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
         ([#657](https://github.com/Automattic/pocket-casts-android/pull/657)).
     *   Added new Tasker "Query Podcasts", "Query Podcast Episodes", "Query Filter", "Query Filter Episodes" and "Add To Up Next" actions.
         ([#583](https://github.com/Automattic/pocket-casts-android/pull/583)).
+    *   Add categories to recommendations screen
+        ([#675](https://github.com/Automattic/pocket-casts-android/pull/675)).
 
 7.29
 -----
--- a/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/recommendations/OnboardingRecommendationsStartPage.kt
+++ b/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/recommendations/OnboardingRecommendationsStartPage.kt
@@ -14,12 +14,15 @@ import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.heightIn
 import androidx.compose.foundation.layout.navigationBars
 import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.statusBars
 import androidx.compose.foundation.layout.windowInsetsPadding
 import androidx.compose.foundation.lazy.grid.GridCells
 import androidx.compose.foundation.lazy.grid.LazyGridScope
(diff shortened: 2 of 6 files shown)


ashiagr:
Nice work, @mchowning!

> I implemented this so that the sections are loaded one at a time. I thought this would be good for users with slow connections so that they didn't have to either wait for all the sections to load or have the sections moving around on them as the network responses came back. Let me know if you think there's a better way to handle this.

I don't have any issues with the approach.

> > Including "Featured" would be good, though!

mchowning:
Thanks for the review @ashiagr 🙇 

> I wonder if we should include Featured podcasts also as per @EmmelineKB's https://github.com/Automattic/pocket-casts-android/issues/524#issuecomment-1357205731.

I'll look into adding that in a separate PR. I left that out because I'm trying to keep my PRs a bit smaller. 🙂 

> Noticed one issue: After loading more and clicking a podcast to subscribe, number of podcasts displayed for that section reduces to the original count. Sometimes even clicking load more does the same thing but I wasn't able to reproduce it consistently.

I'm so glad you caught this! 

It was being caused because any time the subscriptions or sections flows updated, I was recreating the entire state flow object. Since the number to show was only present in the state flow object, we would lose that value (i.e., it would go back to the default) anytime one of the source flows changed. That's why changing subscriptions would cause the number to show to revert back to the default. Similarly, I suspect that the "random" times the number to show would revert would be because one of the API calls would return for one of the categories, which would cause the state flow to be recreated.

I think the issue is resolved now, but please give it a test and let me know if you still see it.

ashiagr:
>  I left that out because I'm trying to keep my PRs a bit smaller. 🙂

👍

> It was being caused because any time the subscriptions or sections flows updated [...]

Thank you for the clear explanation. All of it makes sense to me. 🙌

> I think the issue is resolved now, but please give it a test and let me know if you still see it.

I tested and found no issues this time. Good job!

ashiagr (approved):

PR #679, opened 2023-01-07
Adding support for HLS streams

### LINKED ISSUE #670: Add support for HLS (m3u8 streams)

Some podcasts are published using the HTTP Live Streaming protocol.

Those streams can currently be used in Pocket Cast for iOS, but it's not working on Android.

Implementing support of the protocol would remove the feature disparity.

smsisko:
## Description
Add support for HLS streams when destination URI is a m3u8 file.

Fixes #670 

## Testing Instructions

1. Add an RSS link to produces HLS sources (e.g. https://ohdieux.ligature.ca/rss?programme_id=672)
2. Try to listen to one of the episodes
3. Before the changes, you would get a "File cannot be played. Try converting to another format" error, after it will play the episode normally


ashiagr (approved):
Thanks for supporting HLS streams, @smsisko! It's a great addition and brings feature parity with iOS.

I'll approve this PR but not merge it right away. As you mentioned, let's also show some kind of informative message that playing from downloaded HLS files is not supported and disable download, at least on the episode details screen. I'll create another PR shortly that will address it and both PRs can be merged together.

Meanwhile, you can run `./gradlew :spotlessApply` to auto-fix formatting issues and add a [change log](https://github.com/Automattic/pocket-casts-android/blob/main/CHANGELOG.md) entry in v7.31 describing the change.

CookieyedCodes:
Just want to test this, dose anyone have an rss that uses these type files I'm just interested, the one listed in the route appears normal for me  🤔

PR #683, opened 2023-01-10
Disable download for newly supported HLS streams

ashiagr:
## Description

This PR disables download for [newly supported HLS stream](https://github.com/Automattic/pocket-casts-android/pull/679) on the episode details screen.

Parent Issue #670

## Testing Instructions

1. Add an RSS link that produces HLS sources (e.g. https://ohdieux.ligature.ca/rss?programme_id=672).
2. Tap on one of the HLS source episodes to go to the episode details screen.
3. Notice that the download button is disabled and displays a message that playback from downloaded HLS files is not supported.

## Checklist
 
- with different themes


ashiagr:
Thanks for the review, @mchowning! 

> [...] with Talkback enabled it wasn't possible for me to get TalkBack to read the error message. [...] I don't think this has anything to do with your changes, so its fine with me if you'd like to just file an issue for this.

I've created a new issue for this so it can be tackled separately.

PR #685, opened 2023-01-11
Switched to using the new user login and register

geekygecko:
## Description
As part of the Single Sign-On change, this upgrades from the legacy user login and register endpoints. The iOS app has been using these new endpoints for a long time, so I don't see any danger in switching to them.

This change also switches over to using Retrofit and Coroutines. Hopefully, we can get rid of the [ServerManager](https://github.com/Automattic/pocket-casts-android/blob/main/modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/ServerManager.kt) class as it feels very dated when compared with Retrofit.

## Testing Instructions
1. Clean install the app
2. Tap 'Sign up' on the welcome screen
3. Enter an existing Pocket Casts account email and any password
4. Tap 'Create Account'
5. ✅  Verify an error message is displayed
6. Enter a new email which doesn't exist
7. Tap 'Create Account'
8. ✅  Verify you are taken to the "Find your favorite podcasts" screen.
9. Go to the Profile tab and sign out of the account
10. Tap the 'Set Up Account' and 'Sign in'
11. Enter an existing email and password
12. ✅  Verify you are signed in

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 7.31
 -----
-
+*   Health:
+    *   Switched to using the new user login and register endpoints.
+        ([#685](https://github.com/Automattic/pocket-casts-android/pull/685)).
 
 7.30
 -----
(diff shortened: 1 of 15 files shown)


ashiagr (approved):
Found no issues in my testing. Code changes look good too. 👍

PR #696, opened 2023-01-12
Setup Google SSO server call

geekygecko:
## Description
This change does the following:
- takes the Google id token and sends it to our sync server
- adds logic to handle refresh and access tokens
- updates the forgotten password call to use Retrofit

## Testing Instructions

I recommend using an actual device as I found it more reliable than an emulator with the Google Sign-In framework.
It would be good to test both using the staging and production servers as they are running different code and we will be releasing the Android app before the updated server code is released to production.

### Staging testing setup
1. Change SINGLE_SIGN_ON_ENABLED in base.gradle to `true`
2. Make sure the Build Variant is 'debug'

### Test Google sign in, with an existing account
1. Clean install the app
2. Tap 'Continue with Google'
3. Pick your Google account with an email that exists in the app
4. ✅ Verify you are taken into the app and on the profile tab your email should be displayed

### Test Google sign in, without an existing account
1. Clean install the app
2. Tap 'Continue with Google'
3. Pick your Google account with an email that doesn't exist (if you need delete the account in the staging user admin area)
4. ✅ Verify you are taken to the next welcome page to pick some podcasts

### Test email sign in
1. Clean install the app
2. Tap 'Sign up'
3. Enter a new email and password
4. Tap 'Create Account'
5. ✅ Verify you are taken to the next welcome page to pick some podcasts

### Test creating an account with email
1. Clean install the app
2. Tap 'Sign up'
3. Enter a new email and password
4. Tap 'Create Account'
5. ✅ Verify you are taken to the next welcome page to pick some podcasts

### Test forgotten password
1. Clean install the app
2. Tap 'Log in'
3. Enter an existing username and password
4. Tap 'Log in'
5. ✅ Verify you are taken to the next welcome page to pick some podcasts

### Production testing setup
1. Change SINGLE_SIGN_ON_ENABLED in base.gradle to `true`
2. Make sure the Build Variant is 'debugProd'
3. Repeat the tests above, excluding the Google sign in tests, as the server endpoints aren't available yet.

### DIFF
--- a/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/AccountAuth.kt
+++ b/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/AccountAuth.kt
@@ -6,21 +6,18 @@ import android.content.Context
 import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsEvent
 import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsTrackerWrapper
 import au.com.shiftyjelly.pocketcasts.analytics.TracksAnalyticsTracker
-import au.com.shiftyjelly.pocketcasts.localization.helper.LocaliseHelper
 import au.com.shiftyjelly.pocketcasts.preferences.AccountConstants
 import au.com.shiftyjelly.pocketcasts.preferences.Settings
 import au.com.shiftyjelly.pocketcasts.repositories.podcast.PodcastManager
 import au.com.shiftyjelly.pocketcasts.repositories.refresh.RefreshPodcastsThread
-import au.com.shiftyjelly.pocketcasts.servers.ServerCallback
 import au.com.shiftyjelly.pocketcasts.servers.ServerManager
 import au.com.shiftyjelly.pocketcasts.servers.model.AuthResultModel
 import au.com.shiftyjelly.pocketcasts.servers.sync.SyncServerManager
 import au.com.shiftyjelly.pocketcasts.servers.sync.parseErrorResponse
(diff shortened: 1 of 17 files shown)


ashiagr (approved):
So happy to see this PR! 🎉

### debug build

1. Test Google sign in, with an existing account ✅ 
2. Test Google sign in, without an existing account  ✅ 
3. Test email sign in  ✅ 
4. Test creating an account with email  ✅ 
5. Test forgotten password  ✅ 

PS: I think "Test email sign in" and "Test forgotten password" instructions need to be updated. 

### debugProd build

1. Test email sign in  ✅ 
2. Test creating an account with email  ✅ 
3. Test forgotten password  ✅ 

Just to be sure, we cannot yet set a password for the new account created using Google sign-in, right?  I get an "Email not found" (instead of an incorrect password) error if I try to log in with this account using email and an "incorrect" password.

geekygecko:
> Just to be sure, we cannot yet set a password for the new account created using Google sign-in, right? I get an "Email not found" (instead of an incorrect password) error if I try to log in with this account using email and an "incorrect" password.

That's a great point. It is on my project thread todo list, and I won't enable the button in the app until this is fixed.
- Remove “Change password” and “Change email” when using Google Sign-In

I have been wondering if removing these options when using Google Sign-In is the correct change. Maybe the change password feature should let you set a password if you used Single Sign-On to create the account. We still have platforms like Sonos, which might not have SSO when this is released.
Similarly, with "Change email" maybe it should let you change your Google account.

geekygecko:
> PS: I think "Test email sign in" and "Test forgotten password" instructions need to be updated.

Oops, thanks for picking that up.

Issue #161, opened 2022-07-21
Notifications unresponsive both Locked and Unlocked


mchowning:
We've gotten reports of difficulties with Pocket Casts' responsiveness on the lock screen since Android 12's release:

110663-zen:
> Hello, just a minor thing I've noticed with pocketcasts. I'm running it on a Pixel 5 using Android 12 and since updating to android 12 I've not been able to use the player controls in the notification shade (regardless of whether device unlocked or not). Not sure if this is a widespread thing, I've rebooted my device to try and fix this and it's not done anything.

110671-zen:
>Hello! Ever since I upgraded on Android 12 on my Pixel 3, I intermittently can't pause pocketcasts from the widget on my lock screen. Sometimes it doesn't work at all, other times it takes around 30 seconds to register that I tapped pause or play, and sometimes it works fine. I don't have this issue with Spotify or any other audio app. Is this a known issue?

4547590-zen:

> I have a Google Pixel 5a 5g, and for the last couple of weeks the controls that appear in the notification bar don't work. Can't pause or anything from there like I used to be able to, have to touch it to open the app and then pause.

5301624-zen:

> latest version 7.19.2. has big notification problems.
> The Playback always stuck and the podcast shows the wrong episode I played before.
> The Version 7.19.1 doesn't have this issue.
> I use the Pixel Pro 6 with the latest Android version.

In addition, @azollafox observed this issue.

erricgunawan:
(Probably) similar report from 5733519-zen:

> For the past week or two pocketcasts has been frequently failing to play when pressing the play button in the android notification area (both in the notification shade and the lock screen).
>
> Specifically this seems to occur when pocketcasts has been dormant for a bit (over five or ten minutes). If pocketcasts has been active recently it works fine.

App version : 7.27-rc-1 (9050)
Phone: Google - Pixel 4a (5G) - Pixel 4a (5G) - bramble
Kernel: 4.19.239-g248fa42a73b2-ab9062766 - abfarm-release-rbe-64-00027
Android version: 13 SDK 33

ashiagr:
5767416-zd-woothemes

> Sometimes (after an unintentional timeout [?]), the Playback notification media player stops working. You can push the play icon and it will say it is playing, but it isn't. You click on the notification, which opens the app and it's still paused. Happens on both the Pixel 6 and 7. Never happened on the Motorola Edge I used previously. Maybe a Android 13 issue.

mdrockwell:
[...]

PR #709, opened 2023-01-17
Temporary fix for unresponsive media notifications in Android 13

ashiagr:
## Description

Temporarily fixes unresponsive media notifications in Android 13 (#161).

According to [this](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/app/Service.java#830), 

>  [...] If {@link #STOP_FOREGROUND_DETACH} is supplied, the service's association
with the notification will be severed. [...] 

For the steps outlined in https://github.com/Automattic/pocket-casts-android/issues/161#issuecomment-1191971420, where the service is stopped, I am unable to restart it from the notification media actions if  `STOP_FOREGROUND_DETACH` flag is passed to `stopForeground(...)`.

Until we figure out a proper fix, this PR reverts to using legacy flags with `stopForeground(...)`.

## Testing Instructions
1. Force close the app 
2. Start playback from within the app
3. Stop playback from within the app
4. Swipe app away from recent (don't force close the app because that removes the persistent media notification)
5. Click play on the persistent media notification 
6. ✅ Notice that playback is resumed

## Checklist
 
- with different themes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@
         ([#710](https://github.com/Automattic/pocket-casts-android/pull/710)).
     *   Make it easier to tap newsletter toggle
         ([#714](https://github.com/Automattic/pocket-casts-android/pull/714)).
+    *   Improved unresponsive media notifications
+        ([#709](https://github.com/Automattic/pocket-casts-android/pull/709)).
 
 7.30
 -----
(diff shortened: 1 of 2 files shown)


mchowning (approved):
> Until we figure out a proper fix, this PR reverts to using legacy flags with `stopForeground(...)`

I agree that this doesn't feel like the "proper" fix, but it does fix it for me as well, so I agree that this change is worth it until we find something better.

PR #716, opened 2023-01-18
Add drop shadow for recommendations button

mchowning:
| 📘 Project: #524  |
| --- |

## Description
Adds a drop shadow to the part of the recommendations screen with the button so it is visually cleaner as content scrolls under the button.

I picked the `8.dp` drop shadow because it looked about right to me, but let me know if you think another value would be better.

## Testing Instructions
1. Create a new account so you are taken to the recommendations screen
2. Confirm that the bottom part of the screen with the `Not Now`/`Continue` button has a slight drop shadow now.

## Screenshots or Screencast 

| Before | After |
| --- | --- |
| Screenshot 2023-01-18 at 1 26 04 PM | Screenshot 2023-01-18 at 1 25 00 PM |

## Checklist
- [ ] If this is a user-facing change, I have added an entry in CHANGELOG.md (I don't think this needs a changelog entry. Let me know if you disagree).
 
- [X] with different themes

### DIFF
--- a/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/recommendations/OnboardingRecommendationsStartPage.kt
+++ b/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/recommendations/OnboardingRecommendationsStartPage.kt
@@ -24,6 +24,7 @@ import androidx.compose.foundation.lazy.grid.items
 import androidx.compose.foundation.progressSemantics
 import androidx.compose.material.CircularProgressIndicator
 import androidx.compose.material.MaterialTheme
+import androidx.compose.material.Surface
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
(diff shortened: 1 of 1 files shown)


ashiagr (approved):
8.dp looks fine. 

I was wondering if we could just have the shadow on the top. Can we try [clipRect](https://stackoverflow.com/a/71341254/193545)?

mchowning:
> I was wondering if we could just have the shadow on the top. Can we try [clipRect](https://stackoverflow.com/a/71341254/193545)?

😬 I somehow managed to completely not see that. Thanks for pointing it out. 🙇 

I was able to avoid the issue by just moving the navigation bar inset padding into the button component itself (https://github.com/Automattic/pocket-casts-android/pull/716/commits/e8b364dcf3b82346f4be831fa10a734933072ee6). Now the button padding goes all the way to the bottom of the screen, so any shadow on the bottom is off screen. Let me know what you think.

Screenshot 2023-01-19 at 12 05 41 PM

ashiagr:
Looks great!

PR #717, opened 2023-01-18
Immediately update subscribed UI when user taps

mchowning:
| 📘 Project: #524  |
| --- |

## Description
This PR updates the recommendation screen so that when the user taps to subscribe to a podcast, the UI doesn't wait for the db to reflect the subscription and instead immediately updates to reflect the new subscription state.

## Testing Instructions
1. Create a new account so you are taken to the recommendation screen
2. Tap on a podcast to change it's subscribe status
3. Observe that the UI updates to reflect the new subscription status without any delay

## Screenshots or Screencast 

https://user-images.githubusercontent.com/4656348/213270868-f17e2436-a4f6-483b-8e83-7b29af20a11b.mov

## Checklist
- [ ] If this is a user-facing change, I have added an entry in CHANGELOG.md (I don't think this change needs a changelog entry, but let me know if you disagree)
 
- [ ] with different themes


ashiagr:
Can we apply the same effect when subscribing to a podcast in search results?

https://user-images.githubusercontent.com/1405144/213364208-5996ba36-281b-42dd-91a8-76f6ff0e87e4.mp4

I was also thinking of reverting to an unsubscribed state if there was a network failure. But it might add to the complexity, also I don't see it elsewhere in the app so we can be consistent with that here as well. Feel free to ignore this feedback.

ashiagr (approved):

Issue #590, opened 2022-11-17
Add support for new media notification controls configuration to support 3 icons (if room).


geekygecko:
### Description

Recently the ability to [configure your media notification controls](https://github.com/Automattic/pocket-casts-android/pull/540) was added. Unfortunately, it changed the notification from three to two controls, so user's have mentioned they have lost the mark as played control.

<img  width="300" src="https://user-images.githubusercontent.com/308331/202375262-bdbddcb2-2005-4d3f-9b2a-7956363d043d.png" />

We used to offer three controls, playback speed, star, and mark as played. The Android playback notification only shows two controls but Android Auto uses three controls. 

Before
<img width="912" alt="Screenshot 2022-11-17 at 4 00 23 pm" src="https://user-images.githubusercontent.com/308331/202375541-53f0692c-c88c-40b8-a104-afb382d1e3e5.png">

After
<img width="912" alt="Screenshot 2022-11-17 at 4 58 52 pm" src="https://user-images.githubusercontent.com/308331/202375506-3be5881a-9f62-4054-8254-b45874cdd6f6.png">

It would be good to support three controls but it might be more difficult to implement the configuration page as we might want to let the user choose the positions so the user's least favourite control is removed.

<img  width="300" src="https://user-images.githubusercontent.com/308331/202375950-fe40c289-a7b9-4b6a-a413-5f70da5e53ac.png" />

### Step-by-step reproduction instructions

1. Follow the steps to set up an Android Auto emulator https://developer.android.com/training/cars/testing#test-auto
2. Deploy the app to a phone
3. Connect the phone using USB
4. Play an episode
5. Open the full screen playing screen

The mark as played control will be missing.

ksog66:
Hey @geekygecko,
We can provide the user to choose 3 options and give a sorting option using up and down arrow to choose the position.
like in this image
![image](https://user-images.githubusercontent.com/70816971/204342375-658f9aeb-4cc6-4237-913a-3560b8f81740.png)

and we can also provide something like preview for mobile notif and android auto notif.( while choosing the position and media control)

geekygecko:
@ksog66 I wasn't sure the best way for the user to give us their preference, but this looks like a great solution! Thanks for sharing your thoughts on this.

npaladin2000:
The icons are being hidden on Android 12 as well, and we are not being shown the menu to manually change which icons we get.  On a Galaxy Z Fold 4 myself here. Really need my Mark as Played button...

[...]

PR #719, opened 2023-01-19
Add ability to hide custom media actions on media notification

ashiagr:
## Description
This is a small PR that adds the ability to hide custom media actions (available on the Android 13 media style notification and Android Auto).
See comment https://github.com/Automattic/pocket-casts-android/issues/590#issuecomment-1354402436

## Testing Instructions

1. Play an episode and check that it displays two custom actions at the bottom right of the media notification.
2. Go To Profile > Settings > General > Defaults.
3. Click on "Media notification controls" under `Defaults` settings.
4. ✅ Notice a new settings row titled "Hide custom media actions".
5. Turn "Hide custom media actions" settings on.
6. ✅ Notice that "Prioritize ..." section is disabled and the two custom actions are hidden in the media notification.
7. Turn "Hide custom media actions" settings off.
8. ✅ Notice that the two custom actions are shown in the media notification.
9. Rearrange custom media actions from "Prioritize ..." section.
10. ✅ Notice that custom actions are rearranged in the media notification.

** Since these actions are also shown on Android Auto, they're also controlled by this new setting.
You can setup an `Android Auto` emulator https://developer.android.com/training/cars/testing#test-auto to test it. I've also included screenshots below from my testing.

## Screenshots or Screencast 

https://user-images.githubusercontent.com/1405144/213417484-b00afe83-80c3-449b-89fa-da3643125a95.mp4

Android Auto 

Shown | Hidden
---|---
Screenshot 2023-01-19 at 3 47 27 PM |  Screenshot 2023-01-19 at 3 47 39 PM

## Checklist
 
- [x] with different themes


CookieyedCodes:
Is their a way to make sure it's allways on if android auto is connected? That would probably avoid confusion 🤔

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/MediaSessionManager.kt
```diff
@@ -112,6 +113,17 @@ class MediaSessionManager(
...
     }
 
+    private fun observeHideCustomMediaActions() {
+        launch {
+            settings.maxCustomMediaActionsCountFlow.collect {
+                withContext(Dispatchers.Main) {
+                    val playbackStateCompat = getPlaybackStateCompat(playbackManager.playbackStateRelay.blockingFirst(), currentEpisode = playbackManager.getCurrentEpisode())
+                    updatePlaybackState(playbackStateCompat)
+                }
+            }
```
mchowning, line 123: This is just listening to the flow in order to recreate the playback state anytime the flow updates, right? If so, it might be worth adding a comment here clarifying that it is just listening for changes (but not actually using updated values). Just a thought I had since I had to read this chunk of code a few times before I realized I wasn't just overlooking where the updated value was being used. I don't feel strongly.

mchowning:
> Is their a way to make sure it's allways on if android auto is connected? That would probably avoid confusion

That's an interesting idea. I'm concerned that might just be trading one kind of confusion ("Where did my Android Auto controls go?") for another kind of confusion ("Why do my phone's media notification show custom media controls _some_ of the time?"). 

I'd be inclined to avoid the additional complexity of changing the behavior when AA is connected until we're sure that users are getting confused. I don't feel strongly though.

Hopefully, Android will eventually give us a way to have separate settings for the media notification and Android Auto. 🤞

mchowning (approved):
This is looking pretty good. Left a couple of comments but nothing major.

While looking into this a bit, I noticed that the Android documentation [here](https://developer.android.com/about/versions/13/behavior-changes-13#playback-controls) says that these controls can also be used on Android TV. I tested using a Chromecast device and a Google Nest Hub though, and I did not find these media controls showing up regardless of whether the custom controls setting was turned on or off. 

Obviously that could change at any time though, and if it does we might want to consider updating the wording for this setting to indicate that it also affects those devices. Just something we may want to keep in mind in the future.

modules/services/localization/src/main/res/values/strings.xml
```diff
@@ -1023,6 +1023,10 @@
     <string name="settings_lock_image">lock image</string>
     <string name="settings_manage_downloads_include_starred">Include starred</string>
     <string name="settings_manage_downloads_total">Total</string>
+    <string name="settings_media_actions_prioritise">Prioritize your custom media actions</string>
+    <string name="settings_media_actions_customise">Customise media actions</string>
+    <string name="settings_media_actions_hide_title">Hide custom media actions</string>
```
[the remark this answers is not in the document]
ashiagr: Good point. Updated toggle to reflect show instead of hide in 8ec4a77.

modules/services/localization/src/main/res/values/strings.xml
```diff
@@ -1023,6 +1023,10 @@
...
+    <string name="settings_media_actions_hide_subtitle">This will override below settings and hide custom media actions.</string>
```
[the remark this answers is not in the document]
ashiagr: Updated texts:

modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/Settings.kt
```diff
@@ -605,4 +606,7 @@ interface Settings {
 
     fun hasCompletedOnboarding(): Boolean
     fun setHasDoneInitialOnboarding()
+
+    fun getVisibleCustomMediaActionsCount(): Int
+    fun setVisibleCustomMediaActionsCount(value: Int)
```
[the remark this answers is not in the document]
ashiagr: I was speculating we might have to configure showing 1 or 2 custom actions. But with existing requirements, it makes sense to just save a boolean. Updated in 8ec4a77.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/MediaSessionManager.kt
ashiagr: Added a comment in e8e7d1e

ashiagr:
Thanks for providing thoughtful suggestions, @mchowning 🙏 I've accepted them. 

> While looking into this a bit, I noticed that the Android documentation [here](https://developer.android.com/about/versions/13/behavior-changes-13#playback-controls) says that these controls can also be used on Android TV. I tested using a Chromecast device and a Google Nest Hub though, and I did not find these media controls showing up regardless of whether the custom controls setting was turned on or off.

> Obviously that could change at any time though, and if it does we might want to consider updating the wording for this setting to indicate that it also affects those devices. Just something we may want to keep in mind in the future.

I'll keep this in mind and update the wording if needed. Thanks for the heads-up. 🙇‍♀️

This is ready for another look.

modules/services/localization/src/main/res/values/strings.xml
```diff
@@ -1023,6 +1023,11 @@
     <string name="settings_lock_image">lock image</string>
     <string name="settings_manage_downloads_include_starred">Include starred</string>
     <string name="settings_manage_downloads_total">Total</string>
+    <string name="settings_media_actions_prioritize_title">Prioritize your custom media actions</string>
+    <string name="settings_media_actions_prioritize_subtitle">Your top actions will be available.</string>
+    <string name="settings_media_actions_customise">Customise media actions</string>
+    <string name="settings_media_actions_hide_title">Show custom media actions</string>
+    <string name="settings_media_actions_hide_subtitle">Show custom media actions in Android 13 media notifications and Android Auto.</string>
```
mchowning, line 1030: Want to also update these string names from "hide" to "show"?

mchowning (approved):
Thanks for the updates. This is working great. I just had one minor naming comment, but it's not a blocker.

modules/services/localization/src/main/res/values/strings.xml
mchowning: I went ahead and added this change in [753d545](https://github.com/Automattic/pocket-casts-android/pull/719/commits/753d5451ce0e5f55e751f1bd5ae6d82e51ae81fb).

PR #722, opened 2023-01-19
Update UI immediately when tapping on search result

mchowning:
| 📘 Project: #524  |
| --- |

## Description
This is a follow-up to https://github.com/Automattic/pocket-casts-android/pull/717. It adds the same immediately-update-the-UI logic to the recommendations search screen, as suggested [here](https://github.com/Automattic/pocket-casts-android/pull/717#issuecomment-1396468512).

## Testing Instructions
1. Create a new account so you are taken to the recommendation screen
2. Perform a search from the recommendations screen
3. Tap to subscribe-to or unsubscribe-from one of the search results
4. Observe that the UI updates to reflect the new subscription status without any delay

## Screenshots or Screencast 

https://user-images.githubusercontent.com/4656348/213539741-91dfc7e1-0085-4160-9bd8-2303c02c446b.mov

## Checklist
- [ ] If this is a user-facing change, I have added an entry in CHANGELOG.md (I don't think this needs a changelog entry)


ashiagr (approved):
Works well 👍

PR #726, opened 2023-01-23
Wear App: Play selected episode

ashiagr:
## Description

This PR 
- Displays episodes for a podcast
- Connects "Up Next" and "Podcast" episodes with "Now Playing" screen 
- Plays selected episode
- Updates now playing episode title on the main screen list

Ignore all styling, consider it only as a skeleton that connects mentioned screens and performs basic streaming for an episode.

## Testing Instructions

1. Install and launch the wear app.
2. Log in, navigate to the Podcasts screen, and select a podcast.
3. ✅ Notice that episodes are shown for the selected podcast.
4. Tap on an episode.
5. ✅ Notice that the selected episode starts playing.
6. Return to the main screen. 
7. ✅ Notice that "Now Playing" displays the currently playing media title.

## Screenshots or Screencast 

https://user-images.githubusercontent.com/1405144/214042390-22352c65-769f-4e63-ac27-858583d4e0ae.mp4


wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/PocketCastsWearApplication.kt
```diff
@@ -33,10 +40,23 @@ class PocketCastsWearApplication : Application() {
...
 
     private fun setupAnalytics() {
         AnalyticsTracker.init(settings)
     }
+
+    override fun getWorkManagerConfiguration(): Configuration {
+        return Configuration.Builder()
+            .setWorkerFactory(workerFactory)
+            .setExecutor(Executors.newFixedThreadPool(3))
+            .setJobSchedulerJobIdRange(1000, 20000)
```
mchowning, line 59: Just curious, did we have to set these values because otherwise the defaults caused problems?

mchowning (approved):
Looks good!

wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/PocketCastsWearApplication.kt
ashiagr: I'm not sure about the history around it. To be safe, I kept the values same as were set in mobile and automotive apps. I think it is a decent range that works for us.

PR #729, opened 2023-01-26
Media custom action: Allow odd speeds

ashiagr:
## Description
This allows odd playback speeds using media speed custom action available on android app media notification/ android auto/ android automotive.

(Internal discussion: https://a8c.slack.com/archives/C02A333D8LQ/p1674725427583819)

## Testing Instructions
1. Install automotive app.
2. Play an episode.
3. Tap on playback speed action on the player screen.
4. Notice that odd speeds 1.1x, 1.3x, 1.5x ... are now available.

## Screenshots or Screencast 
https://user-images.githubusercontent.com/1405144/214810449-0722383a-f3ed-4b74-91d3-446283dd64ce.mp4

## Checklist
 
- with different themes

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/MediaSessionManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/MediaSessionManager.kt
@@ -631,10 +631,15 @@ class MediaSessionManager(
         launch {
             val speed = playbackManager.getPlaybackSpeed()
             val newSpeed = when {
+                speed < 1.1 -> 1.1
                 speed < 1.2 -> 1.2
+                speed < 1.3 -> 1.3
                 speed < 1.4 -> 1.4
+                speed < 1.5 -> 1.5
                 speed < 1.6 -> 1.6
+                speed < 1.7 -> 1.7
                 speed < 1.8 -> 1.8
+                speed < 1.9 -> 1.9
                 speed < 2 -> 2.0
                 else -> 1.0
             }


CookieyedCodes:
Maybe add a long press to reset back to default 1.0, jump .25 (note 1) or the local settings -is that read & sent or separate?🤔-, feels like it needs a fast action for car safety 😉

-note 1: as were talking about speed I personally in the main app I would like the option in settings that  + - on the speed control would do the .1 increments but tapping the number increase the speed by +.25 at that if it messes with the numbers too much ignore me it would be rare but it would be a nice to have for just getting to places faster with sometimes the finer control that I would occasionally like-

-note 2: also I've heard some blind people can listen at up to 5X speed maby have that as an unlock in setting for those who would like it-

mchowning:
I'm concerned this could create a slight safety issue by requiring/encouraging users to do too much tapping on our app to cycle through the playback speeds. I know this is a big concern for Google when reviewing Android Auto apps, and I see that in the original PR setting the available playback speeds for AA (PR 185 from before we open-sourced the app) @geekygecko stated in the description "Trying to stop our app being pulled by the Android Auto team.". Maybe you have some more context here Phil?

> as were talking about speed I personally in the main app I would like the option in settings that + - on the speed control would do the .1 increments but tapping the number increase the speed by +.25 a

Something different could work, but we want to be careful that we don't create a distraction or too much cognitive overhead by making the Android Auto UI more complex. When it comes to Android Auto simplicity really is king imo, even though it often comes at the cost of convenience, and having multiple complex gesture/long-press interactions feels like too much complexity for Android Auto to me.

> also I've heard some blind people can listen at up to 5X speed maybe have that as an unlock in setting for those who would like it

I hadn't thought about this being important for the visually impaired before, thanks for mentioning that. We have also heard some requests for this and it's on our radar, but we don't have any immediate plans.

geekygecko:
@mchowning like you mentioned, in the early days of Android Automotive you only had a certain number of taps before the screen gave you a warning. It was the same with scrolling down a long list. I wonder if this is no longer the case. Having the extra increments would be good.

mchowning:
> in the early days of Android Automotive you only had a certain number of taps before the screen gave you a warning. It was the same with scrolling down a long list. I wonder if this is no longer the case

FWIW, I wasn't able to get any warning with a lot of tapping and scrolling on the headunit emulator just now, so maybe that isn't a limitation anymore. I do still think it could come up during a google review of our automotive app though (admittedly I'm wildly speculating there, who really knows 🤷 ).

> Having the extra increments would be good.

I agree, but I'm not convinced we should do it like this. I think the thing that I'm most hung up on is that it would take 9 taps for someone to reduce their playback speed by `.1`, and if they accidentally tap the button 10 times instead of 9, that means they have to do 9 _more_ taps. That just feels like too much of a driving distraction to me.

One other option that came to mind for me based on @CookieyedCodes comment is that long pressing could go down instead of up, but (1) I'm not sure whether we can capture a long-press and (2) that would not be very discoverable at all IMO.

Another option would be to see if we could have the media button open a little pop-up that had both a `-` and `+` button so it wouldn't take so many taps.

Yet another option would be to add a setting where the user can select which 4 playback speeds they want to be available in Android Auto (maybe they want 3x, 2x, 1x, and 0.5x). I probably like this idea best because it gives the user control if they want it, but it's more work and I'm worried that we might be adding too many settings to the app and making things confusing for most of our (non-power) users.

Pinging @adamzelinski in case he might have any thoughts here.

CookieyedCodes:
Re long press it really depends on the type of touch screen & car company's love skimping on touch screens -many aren't multitouch still & even then it can be rough on responsiveness, getting better because of tech company pressure tho-

Is their any chance it could take the local settings if that's set up for the pod & just not give the option to change the setting if it's set to local (give a notice that it syncs from the app) & if the podcast has global playback effects have it adjustable by +.25 on tap of the number with +- on eitherside that should reduce taps -this global shouldn't sync back to the app-  

 (also might be good to have a setting to flip sides for this setting so people don't have to reach if they drive on the other side of the road)

adamzelinski:
I'm ultimately happy to defer to what everyone thinks is best here as I haven't had the opportunity to use the app through Android Auto. My first thought though is to just have the playback speed set to whatever the listener has it set to on their other devices. I also like @mchowning idea of adding a setting where the user can select which 4 playback speeds they want to toggle between in Android Auto. Or, I think we could even do something similar to the skip forward / skip back settings where they chose the increment they would want it to go up and down by.

TylerCarberry:
Anecdotally I'd prefer to not have to tap the screen 10 times to change from 1x to 2x while driving. Tap and hold also seems dangerous - I don't know of any other android auto apps that use touch and hold

CookieyedCodes:
Might it be better then to input have a long press but then the user can manually put in a number 🤔 (& have a block/notification that it can not be accessed while driving-ie if automotive knows the car is moving it won't allow it 🤔)

ashiagr:
I personally don't prefer long press either as I think it will not be discoverable just like @mchowning mentioned. Also, I'd favor predefined speed options rather than manually entering any number. 

I'm closing this PR for now and will consider re-opening it once we can visualize a better UI/ UX targeted for Automotive, and if we receive more requests to change the existing behavior.

Thanks for all the nice suggestions, everyone!

PR #730, opened 2023-01-26
Track multi select view events

ashiagr:
## Description
This adds events to the "multi select view" that appears when the user

`multi_select_view_overflow_menu_shown`: When the user taps the more button (... or ︙) button
`multi_select_view_overflow_menu_rearrange_started`: When the user taps the Edit button and the overflow menu enters rearrange mode
`multi_select_view_overflow_menu_rearrange_action_moved`: When the user moves an action in during a rearrange
`multi_select_view_overflow_menu_rearrange_finished`: When the user exits the rearrange mode

iOS PR: https://github.com/Automattic/pocket-casts-ios/pull/296

## Testing Instructions
1. Launch the app
2. Tap on the Podcasts tab
3. Select a Podcast
4. Tap and hold to enter the multi select mode
5. Tap the ... button
6. ✅ Verify you see `🔵 Tracked: multi_select_view_overflow_menu_shown ["source": "podcast_screen"]`
7. Tap the Edit button
8. ✅  `🔵 Tracked: multi_select_view_overflow_menu_rearrange_started ["source": "podcast_screen"]`
9. Drag an item out of the Shortcut in Action bar section into the In Menu section
10. ✅ `🔵 Tracked: multi_select_view_overflow_menu_rearrange_action_moved ["moved_from": "shelf", "moved_to": "overflow_menu", "position": POSITION, "action": "ACTION", "source": "podcast_screen"]`
   - POSITION is the position that it was moved to relative to the section
   - ACTION is the name of the action you moved
11. Tap the Done button
12. ✅ `🔵 Tracked: multi_select_view_overflow_menu_rearrange_finished ["source": "podcast_screen"]`

Repeat the steps on the following views:
1. Go to the Filters tab
2. Tap on a Filter
3. ✅  Follow the steps from above and verify the source is `filters`
4. Tap on the Profile tab
5. Tap on Downloads
6. ✅  Follow the steps from above and verify the source is `downloads`
7. Tap on the Files
8. ✅  Follow the steps from above and verify the source is `files`
9. Tap on Starred
10. ✅  Follow the steps from above and verify the source is `starred`
11. Tap on Listening History
12. ✅  Follow the steps from above and verify the source is `listening_history`

### DIFF
--- a/modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/multiselect/MultiSelectFragment.kt
+++ b/modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/multiselect/MultiSelectFragment.kt
@@ -4,9 +4,13 @@ import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import androidx.core.os.bundleOf
 import androidx.lifecycle.LiveDataReactiveStreams
 import androidx.recyclerview.widget.ItemTouchHelper
 import androidx.recyclerview.widget.SimpleItemAnimator
+import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsEvent
+import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsSource
+import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsTrackerWrapper
 import au.com.shiftyjelly.pocketcasts.preferences.Settings
 import au.com.shiftyjelly.pocketcasts.ui.helper.FragmentHostListener
 import au.com.shiftyjelly.pocketcasts.views.databinding.FragmentMultiselectBinding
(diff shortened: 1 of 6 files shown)


modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/multiselect/MultiSelectFragment.kt
```diff
@@ -105,10 +121,72 @@ class MultiSelectFragment : BaseFragment(), MultiSelectTouchCallback.ItemTouchHe
...
     }
 
-    override fun onItemTouchFinished() {
+    override fun onItemTouchFinished(position: Int) {
         settings.setMultiSelectItems(items.filterIsInstance<MultiSelectAction>().map { it.groupId })
+        trackItemMovedEvent(position)
+    }
+
+    private fun sectionTitleAt(position: Int) =
+        if (position < MultiSelectToolbar.MAX_ICONS) AnalyticsProp.Value.SHELF else AnalyticsProp.Value.OVERFLOW_MENU
```
mchowning, line 134: I think we need this to be `<=` instead of `<`. Without that change the events are acting like my shelf is one item smaller than it actually is (i.e., an item at the "end" of my shelf is treated as if it is in the overflow menu).

```suggestion
        if (position <= MultiSelectToolbar.MAX_ICONS) AnalyticsProp.Value.SHELF else AnalyticsProp.Value.OVERFLOW_MENU
```

mchowning (approved):
I ran into one small tweak I think we need to make, but with that change I think this is good to go. Thanks! 🙇

modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/multiselect/MultiSelectFragment.kt
ashiagr: Oops, thanks for catching it 🙏

PR #733, opened 2023-01-27
Playback - Add seek to accuracy setting

ashiagr:
## Description

This adds a setting enabling more accurate playback seeking to overcome an [ExoPlayer seekTo limitation](https://rb.gy/rojbz2) for large MP3 files. 

ℹ️ Note that enabling more accuracy can make seeking slower so it is not enabled by default and it is mentioned in the setting summary.

Internal discussion: p1674202802570419-slack-C02A333D8LQ

## Testing Instructions

Steps to reproduce
1. Play a long episode in MP3 format, e.g. https://pca.st/xrfsu987 (I tested in `debugProd` env.)
2. Do forward/ backward seeking a few times. (I was able to reproduce the issue more reliably by doing this.)
3. Seek to almost the end of the episode until a few secs left for it to complete.
4. Listen till the end of the episode.
5. ✅ Notice that the episode keeps playing for a few more seconds than the original duration (see in screenshots).

Test new behavior

7. Go to Profile -> Settings -> General
8. Enable "Seek to accuracy" (feel free to improve wording).
9. Reopen the same episode. 
10. Repeat steps 1-4.
11. ✅ Notice that episode ends almost at the original duration.

## Screenshots or Screencast 

Original duration | Episode plays beyond duration | New Setting
--|--|--
duration | issue | new setting

P.S. This new setting is not available in the automotive app. I can add it separately once this PR is approved.

## Checklist
 
- with different themes

### DIFF
--- a/modules/services/localization/src/main/res/values/strings.xml
+++ b/modules/services/localization/src/main/res/values/strings.xml
@@ -1078,6 +1078,8 @@
     <string name="settings_row_action_download" translatable="false">@string/download</string>
     <string name="settings_row_action_play" translatable="false">@string/play</string>
     <string name="settings_row_action_summary">For episodes that aren\'t downloaded yet, lists will show %s buttons.</string>
+    <string name="settings_seek_to_accuracy">Seek to accuracy</string>
+    <string name="settings_seek_to_accuracy_summary">If on, seeking will be more accurate but it can make seeking slower.</string>
     <string name="settings_show_archived" translatable="false">@string/show_archived</string>
     <string name="settings_show_archived_action_hide">Hide</string>
     <string name="settings_show_archived_action_show">Show</string>
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SimplePlayer.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SimplePlayer.kt
@@ -230,7 +230,13 @@ class SimplePlayer(val settings: Settings, val statsManager: StatsManager, val c
             .setUserAgent("Pocket Casts")
             .setAllowCrossProtocolRedirects(true)
         val dataSourceFactory = DefaultDataSource.Factory(context, httpDataSourceFactory)
-        val extractorsFactory = DefaultExtractorsFactory().setMp3ExtractorFlags(Mp3Extractor.FLAG_ENABLE_CONSTANT_BITRATE_SEEKING)
+        val extractorsFactory = DefaultExtractorsFactory().setMp3ExtractorFlags(
+            if (settings.seekToAccuracy()) {
+                Mp3Extractor.FLAG_ENABLE_INDEX_SEEKING
+            } else {
+                Mp3Extractor.FLAG_ENABLE_CONSTANT_BITRATE_SEEKING
+            }
+        )
         val location = episodeLocation
         if (location == null) {
             onError(PlayerEvent.PlayerError("Episode has no source"))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 7.32
 -----
-
+*   Bug Fixes:
+    *   Added a new setting to enable more accurate seeking
+        ([#733](https://github.com/Automattic/pocket-casts-android/pull/733)). 
 
 7.31
 -----
--- a/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/cloud/AddFileActivity.kt
+++ b/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/cloud/AddFileActivity.kt
@@ -699,7 +699,13 @@ class AddFileActivity :
             .setUserAgent("Pocket Casts")
             .setAllowCrossProtocolRedirects(true)
         val dataSourceFactory = DefaultDataSource.Factory(this, httpDataSourceFactory)
-        val extractorsFactory = DefaultExtractorsFactory().setMp3ExtractorFlags(Mp3Extractor.FLAG_ENABLE_CONSTANT_BITRATE_SEEKING)
+        val extractorsFactory = DefaultExtractorsFactory().setMp3ExtractorFlags(
+            if (settings.seekToAccuracy()) {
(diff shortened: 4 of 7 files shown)


mchowning:
Thanks for working on this @ashiagr ! I'm not sure I think this is a change we should make though. In my testing without this fix, 

Without this change, I tested jumping to 10 different places in the podcast you link to above, then skipping forward to near the end of the podcast. In my testing, the podcast ran 10-30 seconds "past" the predicted end of the episode without this fix. That is obviously a bug, and seems like a nuisance, but I'm not certain of the scenario where it's a big problem to be 30 seconds wrong on the end of a ~2 hour podcast. Maybe it's just not a big deal to me though and it's more important to other people.

The setting does fix the problem of the podcast running over but, like you noted, it slows down seek behavior. I tested by starting the podcast you linked to above (streaming, not downloaded) and then immediately tapping the playback scrubber to jump to the middle of the podcast. With this setting turned on, it took 3-8s for the podcast to play after I did the seek. I suspect this is very dependent on the speed of the internet connection since it is downloading much more of the episode to get the more accurate seeking (subsequent seeks on the same episode were much quicker). Without this setting turned on, it was always (<1 second).

My concern is that the much slower seeking and higher data usage for users who stream are (imo) far more significant downsides compared to having long, variable-bit-rate podcasts run up to 30 seconds past the "end" of the episode. For that reason, my inclination would be to not offer this option to our users. I'm not strongly opposed though, so let's get some other opinions (👋  @geekygecko and @adamzelinski ).

If we keep this setting, I think we should give a stronger warnings against turning it on since I feel like it makes the experience worse for the vast majority of users. Something like "Prioritizing seek accuracy can significantly slow down seek responsiveness and cause Pocket Casts to consume more data". Even better might be having an `Advanced` or `Experimental` settings section where we basically say that we don't recommend any of these settings but here they are if you know what you're doing (we could even require the user to turn on a toggle to open up that section). That could come in handy in the future when there inevitably are more settings that may be useful to a small subset of users but which we don't want to distract most of our users with.

modules/services/localization/src/main/res/values/strings.xml
```diff
@@ -1078,6 +1078,8 @@
     <string name="settings_row_action_download" translatable="false">@string/download</string>
     <string name="settings_row_action_play" translatable="false">@string/play</string>
     <string name="settings_row_action_summary">For episodes that aren\'t downloaded yet, lists will show %s buttons.</string>
+    <string name="settings_seek_to_accuracy">Seek to accuracy</string>
```
mchowning, line 1081: What would you think about changing this to "Prioritize seek accuracy"?

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SimplePlayer.kt
```diff
@@ -230,7 +230,13 @@ class SimplePlayer(val settings: Settings, val statsManager: StatsManager, val c
             .setUserAgent("Pocket Casts")
             .setAllowCrossProtocolRedirects(true)
         val dataSourceFactory = DefaultDataSource.Factory(context, httpDataSourceFactory)
-        val extractorsFactory = DefaultExtractorsFactory().setMp3ExtractorFlags(Mp3Extractor.FLAG_ENABLE_CONSTANT_BITRATE_SEEKING)
+        val extractorsFactory = DefaultExtractorsFactory().setMp3ExtractorFlags(
+            if (settings.seekToAccuracy()) {
+                Mp3Extractor.FLAG_ENABLE_INDEX_SEEKING
+            } else {
+                Mp3Extractor.FLAG_ENABLE_CONSTANT_BITRATE_SEEKING
+            }
```
mchowning, line 238: Would it be possible to only do this for variable bit rate files? I'm guessing that by the time we know the file is variable bit rate it's probably too late, but figured I'd ask in case I'm mistaken.

ashiagr:
👋

I really like your suggestions:
1. Add an experimental setting section 
2. Give a stronger warning against turning it on

Just wanted to add more context: The issue is not limited to playback overshooting the actual duration. It is also about inaccurate in-between seeking that can cause an unpleasant experience (noted from an existing ticket 5728271-zd-woothemes):

> I have noticed lately when I use the skip controls that it doesn't slip properly. The app thinks I'm farther into the podcast and when I skip back it actuslly takes me forward into the podcast.

>  I have noticed that the time in the podcast doesnt seem with the audio I am hearing does not always align. So the app thinks I am at 10:00 but in reality I am at 9:00 into the podcast. So when I skip backwards it seems like I skip forward. And then I have to backstep multiple times to find my spot.

I don't strongly feel about adding this setting either, just wanted us to test it out and see if it can make any difference to listening experience for users who let's say want to take notes by seeking several times and not being able to get to the point they really want to be at.

We can obviously add it, track it and dump or keep it based on data and feedback.

mchowning:
> So the app thinks I am at 10:00 but in reality I am at 9:00 into the podcast.

That's interesting. I didn't get anywhere near that large of a discrepancy in my testing around seeking. The 30-seconds off was the worst I saw and that wasn't at all consistent even with a lot of seeking. I wasn't listening to long stretches in my testing though, so maybe that leads to worse behavior that I didn't see.

adamzelinski:
Hey there @ashiagr @mchowning, while I don’t fully understand everything that’s happening here technically to fix the seek issues. I do agree that from a user experience perspective, it’s a bit weird to add a setting to “Fix” issues we’re having with the scrubbing. it also doesn’t seem like a great experience if when seeking to a new position you then have to wait a lot longer for the podcast to resume. Maybe there’s an extra check or something we can do at the beginning before the first play or on download to check that the times are going to be correct.

ashiagr:
Thanks for the feedback, @adamzelinski!
 
>  Maybe there’s an extra check or something we can do at the beginning before the first play or on download to check that the times are going to be correct.

The problem here is that we don't know beforehand whether a user really needs accuracy over speed. While speed works for the majority of users, a small subset of users might prefer more accuracy and I was thinking of giving them a way to do that (with an explicit warning about the limitations linked with it, as @mchowning pointed out). 

I think we can wait to add it until we hear about the issue from more users and explore better options for fixing it. I'll create a separate issue to track it.

ashiagr:
Issue created: https://github.com/Automattic/pocket-casts-android/issues/744. Closing this PR for now.

PR #742, opened 2023-01-31
Storage settings tracks

mchowning:
## Description
Adding tracks for the storage settings.

I also did a minor refactor here so that instead of passing around a `Pair<String, String>` (which would have become a `Triple<String, String, String>` in order to include the analytics info), we're now passing around the `FolderLocation` object, which contains those three strings. I think this is much clearer and makes it more difficult to mix the strings up.

## Testing Instructions
1. Open the app, go to Profile tab → ⚙️ → `Storage & data use`
2. ✅ Observe the event `settings_storage_shown`
3. Tap on "Clear download cache"
4. ✅ Observe the event: `settings_storage_clear_download_cache`
5. Tap on "Store podcasts on"
6. Select a storage option and tap "Ok". 
7. ✅ Observe the event `settings_storage_location, Properties: {"location":"sd_card|phone|custom", ... }`
8. On a device with API <29, you will have the `Custom Folder...` option when you tap on "Store podcasts on"—select that option and tap ok
9. ✅ Observe that this time the `settings_storage_location` event has a "location" property of "custom"
10. Tap on the new option for "Custom folder location" and then tap ok
11. ✅ Observe the event `settings_storage_set_folder_location`
12. Toggle background refresh
13. ✅ Observe the event `settings_storage_background_refresh_toggled, Properties: {"enabled":true|false, ... }`
14. Toggle "Warn before using data
15. ✅ Observe the event `settings_storage_warn_before_using_data_toggled, Properties: {"enabled":true|false, ... }`

## Checklist

### DIFF
--- a/modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/StorageSettingsPage.kt
+++ b/modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/StorageSettingsPage.kt
@@ -34,6 +34,7 @@ import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.tooling.preview.PreviewParameter
 import androidx.compose.ui.unit.dp
 import au.com.shiftyjelly.pocketcasts.compose.AppThemeWithBackground
+import au.com.shiftyjelly.pocketcasts.compose.CallOnce
 import au.com.shiftyjelly.pocketcasts.compose.bars.ThemedTopAppBar
 import au.com.shiftyjelly.pocketcasts.compose.components.DialogButtonState
 import au.com.shiftyjelly.pocketcasts.compose.components.DialogFrame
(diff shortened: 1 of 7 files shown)


ashiagr (approved):
Looks good! 

Just noticed one issue: `settings_storage_shown` is shown on device orientation changes as well.

mchowning:
> Just noticed one issue: settings_storage_shown is shown on device orientation changes as well.

Added a Composable for only calling a function one time despite configuration changes in [41f839d](https://github.com/Automattic/pocket-casts-android/pull/742/commits/41f839dcdcc59df8ca5491a12ce37b2c0fe3f492), and used it to address this issue in [4ed6e5e](https://github.com/Automattic/pocket-casts-android/pull/742/commits/4ed6e5e234a19e5bbe0916733841ba4d543e5e5e). 

I verified that with this change the shown event is not sent on configuration changes, but if you back out of the screen and return to it the event is sent again.

PR #756, opened 2023-02-03
Tracks for auto add to Up Next settings

mchowning:
## Description
Adds tracks for the auto add to Up Next settings.

### Adding `AutoAddUpNext` enum

While doing this, I also switched us to using an `AutoAddUpNext` enum in place of passing around `Int`s. To avoid that change requiring a database migration or cause any issues I am using a `TypeConverter` that converts the new enum into the same `Int` values that we were persisting in the database before. That means that there should be no difference as far as sqlite is concerned. 

With that said, I did separate out the parts of this change that directly impacted the database into a separate commit (81ce815ca809d6f8df6d1396bf00fdbd54eb4a64)—I realize that is the slightly riskier change and if you'd like me to revert that part of the PR, just let me know.

The most likely way I could have messed something up with this is by making a simple typo (i.e., passing the `PlayNext` enum when replacing `PlayLast`), so please keep that in mind for that when you look at the code changes.

### Extra "shown" event when returning to screen

I'm getting an extra `settings_auto_add_up_next_shown` event when I go into the "Choose podcasts" screen and then return to the "Auto add to Up Next". This doesn't happen on the Auto Download settings screen, and I'm not exactly sure what the cause is although I suspect it is due to how we're navigating to the PodcastSelectFragment differently in the [AutoAddSettingsFragment](https://github.com/Automattic/pocket-casts-android/blob/81ce815ca809d6f8df6d1396bf00fdbd54eb4a64/modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/AutoAddSettingsFragment.kt#L155) compared to the [AutoDownloadSettingsFragment](https://github.com/Automattic/pocket-casts-android/blob/aa9ae49b05e94fae69cdf7c1aa71241a8e0120f9/modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/AutoDownloadSettingsFragment.kt#L223-L226).

I didn't spend a ton of time looking into this because (1) it doesn't seem like a huge issue and (2) the issue will probably go away when these screens get converted to compose, but if you have any ideas for how to fix it, let me know (or if you think it's worth spending more time on this let me know, maybe it's a bigger deal than I'm realizing).

#### 1. Persistence of Auto Add to Up Next setting

Since this PR is touching the persistence of this setting, let's make sure I didn't break anything.

1. Go to Profile tab → ⚙️ → `Auto add to Up Next`
2. Tap on one of your subscribed podcasts and change "Auto Add To"  to a different value.
3. Tap on the Podcasts tab
4. Tap on the podcast you updated in step 2
5. Tap on the ⚙️ 
6. Observe that add to Up Next is toggled on and the Position matches the selection you made in step two ("Top" → "Play Next" and "Bottom" → "Play last").
7. Either (a) toggle "Add to Up Next" off or (b) change the "Position" to a different value.
8. Return to Profile tab → ⚙️ → `Auto add to Up Next`
9. Observe that the position for that podcast has been updated appropriately in the UI

#### 2. Tracks

1. Go to Profile tab → ⚙️ → `Auto add to Up Next`
2. ✅  Observe the event `settings_auto_add_up_next_shown`
3. Tap "Auto add limit" and select a new value
4. ✅ Observe the event `settings_auto_add_up_next_auto_add_limit_changed, Properties: {"value":[number], ... }`
5. Tap "If limit reached" and select a new value
6. ✅ Observe the event `settings_auto_add_up_next_limit_reached_changed, Properties: {"value":"stop_adding"|"only_add_top", ... }`
7. Tap on "Choose podcasts", change the selected podcasts, and tap back
8. ✅ Observe the event `settings_auto_add_up_next_podcasts_changed, Properties: {"number_selected":[number], ... }`
9. Tap on one of the podcasts in the "Auto-add podcasts" section, and change the setting
10. ✅ Oberve the event `settings_auto_add_up_next_podcast_position_option_changed, Properties: {"value":"add_first"|"add_last", ... }`

## Checklist

### DIFF
--- a/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/AppDatabaseTest.kt
+++ b/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/AppDatabaseTest.kt
@@ -6,6 +6,7 @@ import androidx.room.testing.MigrationTestHelper
 import androidx.sqlite.db.SupportSQLiteDatabase
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import androidx.test.platform.app.InstrumentationRegistry
+import au.com.shiftyjelly.pocketcasts.models.entity.Podcast
 import au.com.shiftyjelly.pocketcasts.models.type.EpisodesSortType
 import org.junit.After
 import org.junit.Assert.assertEquals
@@ -91,7 +92,7 @@ class AppDatabaseTest {
         assertEquals(true, podcast?.isSubscribed)
         assertEquals(true, podcast?.isShowNotifications)
         assertEquals(2, podcast?.autoDownloadStatus)
-        assertEquals(1, podcast?.autoAddToUpNext)
+        assertEquals(Podcast.AutoAddUpNext.PLAY_LAST, podcast?.autoAddToUpNext)
         assertEquals(0xF00000, podcast?.backgroundColor)
         assertEquals(0xFF0000, podcast?.tintColorForLightBg)
         assertEquals(0xFFF000, podcast?.tintColorForDarkBg)
--- a/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/podcast/PodcastSettingsFragment.kt
+++ b/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/podcast/PodcastSettingsFragment.kt
@@ -164,7 +164,7 @@ class PodcastSettingsFragment : BasePreferenceFragment(), CoroutineScope, Filter
 
             preferenceAddToUpNext?.isChecked = !podcast.isAutoAddToUpNextOff
             preferenceAddToUpNextOrder?.isVisible = !podcast.isAutoAddToUpNextOff
-            preferenceAddToUpNextOrder?.summary = if (podcast.autoAddToUpNext == Podcast.AUTO_ADD_TO_UP_NEXT_PLAY_NEXT) getString(LR.string.play_next) else getString(LR.string.play_last)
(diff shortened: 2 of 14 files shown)


ashiagr (approved):
> I am using a TypeConverter that converts the new enum into the same Int values that we were persisting in the database before. That means that there should be no difference as far as sqlite is concerned.

It looks fine to me. 

> I'm getting an extra settings_auto_add_up_next_shown event when I go into the "Choose podcasts" screen and then return to the "Auto add to Up Next". This doesn't happen on the Auto Download settings screen, and I'm not exactly sure what the cause is although I suspect it is due to how we're navigating to the PodcastSelectFragment differently in the [AutoAddSettingsFragment](https://github.com/Automattic/pocket-casts-android/blob/81ce815ca809d6f8df6d1396bf00fdbd54eb4a64/modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/AutoAddSettingsFragment.kt#L155) compared to the [AutoDownloadSettingsFragment](https://github.com/Automattic/pocket-casts-android/blob/aa9ae49b05e94fae69cdf7c1aa71241a8e0120f9/modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/AutoDownloadSettingsFragment.kt#L223-L226).

I agree, it looks like the way we're navigating from the screens is causing this difference. I wonder if moving `viewModel.onShown()`  from `onCreateView(...)` to `onCreate(...)` helps. I gave it a try and it didn't fire duplicate shown events. It also matched iOS behavior. 

I'm approving the PR, feel free to merge after considering the suggestion for `settings_auto_add_up_next_shown` event.

mchowning:
>  I wonder if moving viewModel.onShown() from onCreateView(...) to onCreate(...) helps. I gave it a try and it didn't fire duplicate shown events. It also matched iOS behavior.

 Great idea! Works like a charm! 😄

Issue #752, opened 2023-02-03
Search history


ashiagr:
## Description 

- A list of recent searches will be displayed now when the user focus the search input box. 
- Whenever a user selects an item in a search results list, the selected item will be saved in the search history, and displayed in the list described in this task (p1665038946092499/1664975388.327359-slack-C03JCLLR9C3).

#### Data layer (https://github.com/Automattic/pocket-casts-android/pull/758)

- [x] Add a migration to add new search history table to the database saving:
       - search term
       and 
       - podcast uuid 
       - episode uuid
       - folder uuid 
       along with essentials details to render search history for items selected in a search results list
- [x] Add CRUD operations for search history item
   - [x] Load podcast/ episode/ folder search history item
- [x] Add delete operation for all search history items

#### Display results (https://github.com/Automattic/pocket-casts-android/pull/765)

- [x] Create a new screen to display the search history
- [x] Add selected podcast from the search results list to search history 
- [x] Add selected folder from the search results list to search history 
- [x] Add search term to search history
- [x] Show search results
- [x] Show X button + click action

#### Other
- [x] Add search history row click action ([#767](https://github.com/Automattic/pocket-casts-android/pull/767))
- [x] Show clear all button in search result header and add click action ([#767](https://github.com/Automattic/pocket-casts-android/pull/767))
- [x] Fine-tune UI ([#769](https://github.com/Automattic/pocket-casts-android/pull/769))
- [x] Add analytics ([#778](https://github.com/Automattic/pocket-casts-android/pull/778))

ashiagr:
@BKrnk1337 yes, this will be integrated in iOS as well.

ashiagr:
#### Todo (this will be added once episodes are supported in search results)
- [ ] Add selected episode from the search results list to search history

PR #758, opened 2023-02-06
Search history: Data layer

ashiagr:
Part of: #752

## Description
This PR
- Adds a new table `search_history` (+ migration) to the database
   - Adds unique indexes to restrict duplicate `search term`/ `podcast`/ `folder`/ `episode` search history item insertion
   - Includes `podcast`/ `folder`/ `episode` search history item details as embedded objects (with minimum details to render corresponding search history item) so that search history can be displayed for them even when there's no corresponding local `podcast`/ `folder`/ `episode` in the database.
- Adds `SearchHistoryDao` for search history item operations
   - findAll
   - insert (or replace)
   - delete
   - deleteAll

Note: Targets a feature branch to allow modifying the `search_history` table until the feature is ready.

#### Tests
- Ensure that tests in `SearchHistoryDaoTest` cover important scenarios:
   - Single search history item insert and delete 
   - Inserting same `search term`/ `podcast`/ `folder`/ `episode` search history item replaces the previous item
   - Multiple unique `search term`/ `podcast`/ `folder`/ `episode` search history items are inserted
   - Last inserted/ replaced search history item is at the top of the search history
   - Delete all search history

#### Migration
- Install the app from the `main` branch
- Install the app from this PR branch
- Go to database inspector and notice that `search_history` table is created

#### Fresh Install 
- Clean install the app from this PR branch
- Go to database inspector and notice that `search_history` table is created

## Checklist
 
- with different themes

### DIFF
--- a/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/SearchHistoryDaoTest.kt
+++ b/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/SearchHistoryDaoTest.kt
@@ -87,9 +87,8 @@ class SearchHistoryDaoTest {
 
             val result = searchHistoryDao.findAll()
             assertEquals("Insert should replace, count should be 1", 1, result.size)
-            assertEquals(
+            assertTrue(
                 "Replaced search term should be on top",
-                true,
                 result.first().modified > modifiedPrevious
             )
         }
@@ -121,9 +120,8 @@ class SearchHistoryDaoTest {
 
             val result = searchHistoryDao.findAll()
             assertEquals("Same podcast search insert should replace, count should be 1", 1, result.size)
-            assertEquals(
+            assertTrue(
                 "Replaced podcast search history item should be on top",
-                true,
                 result.first().modified > modifiedPrevious
             )
         }
@@ -157,9 +155,8 @@ class SearchHistoryDaoTest {
 
             val result = searchHistoryDao.findAll()
             assertEquals("Same folder search insert should replace, count should be 1", 1, result.size)
-            assertEquals(
+            assertTrue(
                 "Replaced folder search should be on top",
-                true,
                 result.first().modified > modifiedPrevious
             )
         }
@@ -193,9 +190,8 @@ class SearchHistoryDaoTest {
 
             val result = searchHistoryDao.findAll()
             assertEquals("Same episode insert should replace, count should be 1", 1, result.size)
-            assertEquals(
+            assertTrue(
                 "Replaced episode search should be on top",
-                true,
                 result.first().modified > modifiedPrevious
             )
         }
--- a/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/SearchHistoryDaoTest.kt
+++ b/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/SearchHistoryDaoTest.kt
@@ -1,0 +1,274 @@
+package au.com.shiftyjelly.pocketcasts.models.db
+
+import androidx.room.Room
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.test.platform.app.InstrumentationRegistry
+import au.com.shiftyjelly.pocketcasts.models.db.dao.SearchHistoryDao
+import au.com.shiftyjelly.pocketcasts.models.entity.SearchHistoryItem
+import au.com.shiftyjelly.pocketcasts.models.entity.SearchHistoryItem.Folder
+import au.com.shiftyjelly.pocketcasts.models.entity.SearchHistoryItem.Podcast
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.runTest
+import org.junit.After
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import java.util.Date
+import java.util.UUID
+
+private const val SEARCH_TERM_TEST1 = "test1"
+private const val SEARCH_TERM_TEST2 = "test2"
+
+@OptIn(ExperimentalCoroutinesApi::class)
+@RunWith(AndroidJUnit4::class)
+class SearchHistoryDaoTest {
+    lateinit var searchHistoryDao: SearchHistoryDao
+    lateinit var testDb: AppDatabase
+
+    @Before
+    fun setupDb() {
+        val context = InstrumentationRegistry.getInstrumentation().targetContext
+        testDb = Room.inMemoryDatabaseBuilder(context, AppDatabase::class.java).build()
+        searchHistoryDao = testDb.searchHistoryDao()
+    }
+
+    @After
+    fun closeDb() {
+        testDb.close()
+    }
+
+    /* INSERT */
+    @Test
+    fun testInsertSearchTerm() = runTest {
+        searchHistoryDao.insert(createTermSearchHistoryItem(SEARCH_TERM_TEST1))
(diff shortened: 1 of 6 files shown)


modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/SearchHistoryItem.kt
```diff
@@ -0,0 +1,44 @@
...
+        val uuid: String,
+        val title: String,
+        val color: Int,
+        val podcastIds: String,
+    )
+    data class Episode(
+        val uuid: String,
+        val title: String,
+        val publishedDate: Date,
+        val duration: Double
```
ashiagr, line 42: These fields might change based on how an episode's recent search (logged when user taps an item from the episode search results) is displayed.

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/SearchHistoryDao.kt
```diff
@@ -0,0 +1,23 @@
...
+import androidx.room.Delete
+import androidx.room.Insert
+import androidx.room.OnConflictStrategy
+import androidx.room.Query
+import au.com.shiftyjelly.pocketcasts.models.entity.SearchHistoryItem
+
+@Dao
+abstract class SearchHistoryDao {
+    @Query("SELECT * FROM search_history ORDER BY modified DESC LIMIT :limit")
+    abstract fun findAll(limit: Int = 10): List<SearchHistoryItem>
```
mchowning, line 13: Did you intend to not make this a suspended function? If you want to make that change, feel free to address that in one of your follow-up PRs when you start using this DAO.

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/SearchHistoryItem.kt
mchowning: Am I understanding correctly that the idea here is that each Search History Item would be for either a `Podcast`,`Folder`, or `Episode`, so the embedded fields for only one of those three would not be `null` in any given `SearchHistoryItem`?

If that's right, it might be worth considering at some point using a sealed class for this so we don't have so many nullable fields to deal with in our code. I'm not sure I'd do that at the db level necessarily (what you have here is nice and straightforward), but perhaps when we start using this part of the db, it would be worth creating a class for accessing the database that would convert all of the entries to/from an appropriate sealed class. Very rough pseudo code of what I'm thinking of:

<details>
<summary>pseudo code</summary>

``` kotlin
sealed class SearchHistoryEntry(
    val modified: Long,
    /* ... */
) {

    class Podcast(
        modified: Long,
        val author: String,
        // ...
        ) : SearchHistoryEntry(modified /* ... */)

    class Folder(
        modified: Long,
        val color: Int
        // ...
    ): SearchHistoryEntry(modified /* ... */)

    class Episode(
        modified: Long,
        val publishedDate: Date,
        // ...
    ): SearchHistoryEntry(modified, /* ... */)
}

class SearchHistoryDbUtil(private val searchHistoryDao: SearchHistoryDao) {
    suspend fun insert(entry: SearchHistoryEntry) {
        val item: SearchHistoryItem = when (entry) {
            is SearchHistoryEntry.Episode -> // create SearchHistoryItem from SearchHistoryEntry.Episode
            is SearchHistoryEntry.Folder -> // create SearchHistoryItem from SearchHistoryEntry.Folder
            is SearchHistoryEntry.Podcast -> // create SearchHistoryItem from SearchHistoryEntry.Podcast
        }
        searchHistoryDao.insert(item)
    }

    suspend fun findAll(limit: Int = 10): List<SearchHistoryEntry> {
        val entries = searchHistoryDao.findAll(limit)
        return entries.map { it.toEntry() }
    }

    private fun SearchHistoryItem.toEntry(): SearchHistoryEntry {
        // create appropriate SearchHistoryEntry from SearchHistoryItem
    }
}
```

</details>

We may figure out that this is actually a bad idea once we start implementing this, but I just wanted to mention it as a possibility in case you might find that it would actually be useful as you start using the db.

app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/SearchHistoryDaoTest.kt
```diff
@@ -0,0 +1,274 @@
...
+            val modifiedPrevious = searchHistoryDao.findAll().first().modified
+            searchHistoryDao.insert(createPodcastSearchHistoryItem(uuid = uuid))
+
+            val result = searchHistoryDao.findAll()
+            assertEquals("Same podcast search insert should replace, count should be 1", 1, result.size)
+            assertEquals(
+                "Replaced podcast search history item should be on top",
+                true,
+                result.first().modified > modifiedPrevious
+            )
```
mchowning, line 128: Could just use `assertTrue`. Certainly not a big deal though.

mchowning (approved):
Looks good! 👍

app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/SearchHistoryDaoTest.kt
```diff
@@ -0,0 +1,274 @@
...
+import org.junit.runner.RunWith
+import java.util.Date
+import java.util.UUID
+
+private const val SEARCH_TERM_TEST1 = "test1"
+private const val SEARCH_TERM_TEST2 = "test2"
+
+@OptIn(ExperimentalCoroutinesApi::class)
+@RunWith(AndroidJUnit4::class)
+class SearchHistoryDaoTest {
```
geekygecko, line 26: It's very cool to see all these tests! ❤️

app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/SearchHistoryDaoTest.kt
ashiagr: I had a different actual vs expected earlier that needed an equals assertion, `assertTrue` is better now, replaced in - 8f4ddd7

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/SearchHistoryDao.kt
ashiagr: Added in 24ea15c, thanks for pointing it out!

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/SearchHistoryItem.kt
ashiagr: > Am I understanding correctly that the idea here is that each Search History Item would be for either a Podcast,Folder, or Episode, so the embedded fields for only one of those three would not be null in any given SearchHistoryItem?

That's correct.

> If that's right, it might be worth considering at some point using a sealed class for this so we don't have so many nullable fields to deal with in our code. I'm not sure I'd do that at the db level necessarily (what you have here is nice and straightforward), but perhaps when we start using this part of the db, it would be worth creating a class for accessing the database that would convert all of the entries to/from an appropriate sealed class. Very rough pseudo code of what I'm thinking of:

I did try to add relations and different tables for each sub-entity and foreign key and all, but it looked complicated. So I didn't use it at the db level to not over-optimize it considering that embedded objects are small. I like your suggestion to use a sealed class at another point and will use it in another PR.

I might also add a query to keep a limited number of records in the table at a later point.

PR #764, opened 2023-02-08
About screen tracks

mchowning:
## Description
This adds the shown event for the about sceen.

## Testing Instructions
1. Open the app and go to the Profile tab → ⚙️ → `About`
2. ✅ Observe the event `settings_about_shown`
3. Verify that rotating the screen does not resend the event.

## Checklist


ashiagr (approved):
LGTM!

PR #765, opened 2023-02-09
Search history: Display results

ashiagr:
Part of #752 

## Description

This PR 
- Creates a new page to display the search history
- Adds selected podcast from the search results list to search history
- Adds selected folder from the search results list to search history 
   (** folder search history is displayed conditionally based on subscription state and selected tab - Podcasts or Discover)
- Adds search term to search history
- Displays search results
- Adds X button click action

Notes:
- Ignore UI, it does not meet the spec yet
- Episode search, search history row click action, clear all are not added, they'll be taken up separately 

#### Folder Search - Plus Account (Podcasts and Discover tabs)
- Login with a Plus account having folders
- Go to Podcasts tab and search for a folder 
- Tap folder in search results
- Go back to Podcasts screen and tap search button again
- ✅ Notice that folder is shown in search history
- Go to Discover tab
- Tap search button
- ✅ Notice that folder although present in db is not shown in search history as Discover tab doesn't display folders

#### Folder Search - Free Account
- Remove Plus subscription for the above account
- Continue from the last test and go to Podcasts tab
- Tap search button
-  ✅ Notice that folder although present in db is not shown in search history as free account doesn't have access to folders

#### Podcast Search
- Go to Podcasts tab and search for a podcast 
- Tap a podcast in search results
- Go back to Podcasts screen and tap search button again
- ✅ Notice that podcast is shown in search history

#### Search Term
- Go to Podcasts tab and search a term
- ✅ Notice that search term is saved on debounce
- Go back to Podcasts screen and tap search button again
- ✅ Notice that searched term is shown in search history

#### Orientation change
- Go to Podcasts tab and search a term
- Go to search history
- Rotate device
- ✅ Notice that searched history is retained
- Search a term
- When result results appear, clear search query
- ✅ Notice that searched history is reshown

#### Remove search history
- Go to search history
- Tap X on a search history row
- Notice that the item is removed from the search history

## Screenshots or Screencast 

https://user-images.githubusercontent.com/1405144/217818207-f8ae86a3-f7b1-46aa-ad21-07e06eea7653.mp4

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/SearchHistoryEntry.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/SearchHistoryEntry.kt
@@ -54,7 +54,7 @@ sealed class SearchHistoryEntry(
                 uuid = uuid,
                 title = title,
                 color = color,
-                podcastIds = podcastIds.joinToString()
+                podcastIds = podcastIds.joinToString(separator = ",")
             )
         )
 
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/SearchHistoryEntry.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/SearchHistoryEntry.kt
@@ -1,0 +1,133 @@
+package au.com.shiftyjelly.pocketcasts.models.to
+
+import au.com.shiftyjelly.pocketcasts.models.entity.SearchHistoryItem
+import java.util.Date
+import au.com.shiftyjelly.pocketcasts.models.entity.Episode as EpisodeModel
+import au.com.shiftyjelly.pocketcasts.models.entity.Folder as FolderModel
+import au.com.shiftyjelly.pocketcasts.models.entity.Podcast as PodcastModel
+
+sealed class SearchHistoryEntry(
+    val id: Long? = null,
+) {
+    class Episode(
+        id: Long? = null,
+        val uuid: String,
+        val title: String,
+        val publishedDate: Date,
+        val duration: Double,
+    ) : SearchHistoryEntry(id = id)
+
+    class Folder(
+        id: Long? = null,
+        val uuid: String,
+        val title: String,
+        val color: Int,
+        val podcastIds: List<String>,
+    ) : SearchHistoryEntry(id = id)
+
+    class Podcast(
+        id: Long? = null,
+        val uuid: String,
+        val title: String,
+        val author: String,
+    ) : SearchHistoryEntry(id = id)
+
+    class SearchTerm(
+        id: Long? = null,
+        val term: String,
+    ) : SearchHistoryEntry(id = id)
+
+    fun toSearchHistoryItem() = when (this) {
+        is Episode -> SearchHistoryItem(
+            id = id,
+            episode = SearchHistoryItem.Episode(
+                uuid = uuid,
+                title = title,
+                publishedDate = publishedDate,
+                duration = duration
+            )
+        )
+
+        is Folder -> SearchHistoryItem(
+            id = id,
+            folder = SearchHistoryItem.Folder(
+                uuid = uuid,
+                title = title,
+                color = color,
+                podcastIds = podcastIds.joinToString()
+            )
+        )
+
+        is Podcast -> SearchHistoryItem(
+            id = id,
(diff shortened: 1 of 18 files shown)


CookieyedCodes:
@ashiagr Would it be an idea to have a bar of pills to enable & disable (show & hide basically) types of search terms 🤔 
(Global -all-, Podcast, Folder, Episode)

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/SearchHistoryEntry.kt
```diff
@@ -0,0 +1,133 @@
...
+                val podcast = item.podcast as SearchHistoryItem.Podcast
+                Podcast(
+                    id = item.id,
+                    uuid = podcast.uuid,
+                    title = podcast.title,
+                    author = podcast.author
+                )
+            }
+
+            else -> SearchTerm(id = item.id, term = item.term ?: "")
```
mchowning, line 130: For search terms, should the `item.term` never be `null`? If so, do you think it's worth checking that specifically here (instead of using `else ->`) and instead use the `else ->` branch to log an error message or throw an exception so it'll be easier to catch it if a bug gets introduced here in the future?

mchowning (approved):
Great job getting this working so quickly. It works great and there were only a few small issues I noticed.

### 1. Padding a bit tight?

The padding between the search field and the first entry feels to tight to me, and it is smaller than it is once a search is completed. I think it would be worth adding some space here.

| History | New Search |
| --- | ---|
| Screenshot 2023-02-09 at 3 43 32 PM | Screenshot 2023-02-09 at 3 43 01 PM |

### 2. Placeholder Images

I ran into a scenario where the folder graphic in a search history row showed the placeholder image even though the images were available. Unfortunately I'm not exactly sure how I made that happen.

https://user-images.githubusercontent.com/4656348/217933660-c93b16b6-a500-4812-bcdd-ef5e1d869aa9.mov

3. Different sizes for folder images

I noticed that the folder images are the same size for the history, but they're a bit bigger on a normal search (relative to the size of a podcast image). I'm guessing that difference between a search and history is intentional, but just wanted to note it in case it isn't.

| History | New Search |
| --- | --- |
| Screenshot 2023-02-09 at 3 46 41 PM | Screenshot 2023-02-09 at 3 46 51 PM |

ashiagr:
👋 @CookieyedCodes 

> @ashiagr Would it be an idea to have a bar of pills to enable & disable (show & hide basically) types of search terms 🤔
(Global -all-, Podcast, Folder, Episode, Author?)

Right now, we're not showing filters for search history. In phase 1, actual search results will be grouped based on their type. We'll keep this in mind and try to improve it in a later phase.

CookieyedCodes:
@ashiagr I didn't per say mean search for filters sorry for the confusion, I meant something like the bar of pills from filters that would behave like the starred filter pill ie a way to show & hide spefic results quickly 🤔

But ya I look forward to understanding the current project further ☺️

ashiagr:
> I didn't per say mean search for filters sorry for the confusion, I meant something like the bar of pills like filters that would behave like the starred filter pill ie a way to show & hide spefic results quickly 🤔

> Might also be nice to have a pin search function 🤔 

They're really nice suggestions, @CookieyedCodes! I'll share them with @adamzelinski and try to incorporate them at some point.

> But ya I look forward to understanding the current project further ☺️

Thank you for your interest in the project! :) We'll be adding a global search for podcasts and episodes as compared to current behavior where episodes are only available on the podcast page. We're also working on increasing the relevancy, giving faster search results, and providing recommended podcasts.

We'll soon add tasks for the client-side work which will include more details.

CookieyedCodes:
No bothers allways thinking of higher functionality ❤️😉

ashiagr:
Thanks for the feedback, @mattchowing! I fixed the placeholder image issue in 56eb0d3.



UI is not as per specs, there's a separate task to fine-tune it. I keep your feedback in mind when I get to it:
> 1. Padding a bit tight?
> 3. Different sizes for folder images

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/SearchHistoryEntry.kt
ashiagr: Good idea, added in f6f2367

PR #766, opened 2023-02-09
Chromecast tracks

mchowning:
## Description
Adding tracks for chromecasting from the app.

#### 1. Verify full flow
1. With a chromecast device available on your local wireless network...
2. Go to the Podcasts tab
3. Tap the chromecast icon in the toolbar (this will only appear if there are chromecast devices available)
4. ✅ Observe the event `chromecast_view_shown, Properties: {"is_connected":false, ... }`
5. Select a speaker to cast to
6. ✅ Observe the event `chromecast_started_casting`
7. Tap the chromecast icon again
8. ✅ Observe the event `chromecast_view_shown, Properties: {"is_connected":true, ... }`
9. Tap "Stop casting"
10. ✅ Observe the event `chromecast_stopped_casting`

#### 2. Verify other places we can chromecast from

In each of the following, verify that tapping on the chromecast icon sends the event `chromecast_view_shown, Properties: {"is_connected":true|false, ... }`.

1. Toolbar on the page for a specific podcast (similar to the toolbar on the Podcasts tab)
2. Full screen player with the chromecast button in the bottom shelf (you may have to move it here from the overflow menu)
3. Full screen player with the chromecast button in the overflow menu (you may have to move it here from the bottom shelf)

2 & 3 are listed separately because they involve different code paths. It looks like these are all the places that you can start chromecasting from, but let me know if you know of any others.

<details>
<summary>Here's a video of me triggering the relevant views.</summary>

https://user-images.githubusercontent.com/4656348/217907487-11bbfde3-1c55-4ceb-bbc7-350c7ca328bb.mov

</details>

## Checklist

### DIFF
--- a/modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/filters/FilterEpisodeListFragment.kt
+++ b/modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/filters/FilterEpisodeListFragment.kt
@@ -45,6 +45,7 @@ import au.com.shiftyjelly.pocketcasts.utils.extensions.dpToPx
 import au.com.shiftyjelly.pocketcasts.views.dialog.ConfirmationDialog
 import au.com.shiftyjelly.pocketcasts.views.dialog.OptionsDialog
 import au.com.shiftyjelly.pocketcasts.views.fragments.BaseFragment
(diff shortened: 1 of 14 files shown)


ashiagr (approved):
LGTM 👍 Thanks for knocking out all these events!

PR #767, opened 2023-02-10
Search history: Navigation and Clear All

ashiagr:
Part of #752
## Description

This adds 
- Navigation for search history item
- Clear all header and a confirmation dialog with a search history threshold (size > 3).

Note: Ignore UI, it doesn't exactly match the specs.

## Testing Instructions
- Build a search history of different types (podcast/ folder/ search term) of at least 4 items 
- Tap on each search history item separately
- ✅ Notice that the appropriate page is opened for podcast, folder and the search is re-triggered for the search term
- Return to search history and tap on Clear All
- ✅ Notice that a confirmation dialog is shown (count > search history threshold)
- Tap on Delete button in the dialog
- ✅ Notice that search history is cleared
- Search another query
- Tap on Clear All
- ✅ Notice that search history is cleared (no dialog shown since count < threshold)

## Screenshots or Screencast 
https://user-images.githubusercontent.com/1405144/218105561-de46cbd4-9fc6-49f4-a85b-7c2b5449fbdb.mp4

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchFragment.kt
+++ b/modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchFragment.kt
@@ -116,6 +116,15 @@ class SearchFragment : BaseFragment() {
         viewModel.onFragmentPause(activity?.isChangingConfigurations)
     }
 
+    private fun navigateFromSearchHistoryEntry(entry: SearchHistoryEntry) {
+        when (entry) {
+            is SearchHistoryEntry.Episode -> Unit // TODO
+            is SearchHistoryEntry.Folder -> listener?.onSearchFolderClick(entry.uuid)
+            is SearchHistoryEntry.Podcast -> listener?.onSearchPodcastClick(entry.uuid)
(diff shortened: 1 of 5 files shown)


mchowning (approved):
Looks good! 👍 

Only question I had is why we [limit the search history to 5 item](https://github.com/Automattic/pocket-casts-android/blob/06b9f2333641b14d7066b27c14bb56366b4e2442/modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/searchhistory/SearchHistoryViewModel.kt#L15-L14)? The UI strongly implies that there isn't any more search history (to my eyes at least) since the list only takes up part of the screen, and it feels a bit odd when I have more than 5 items in my history that deleting one seems to make a new one appear. Certainly not a blocker on merging this PR.

<details>
<summary>Video of what I'm talking about</summary>

https://user-images.githubusercontent.com/4656348/218146447-6b8b3c20-9c41-4111-bec9-b1476ad51238.mov

</details>

ashiagr:
> when I have more than 5 items in my history that deleting one seems to make a new one appear.

I have seen this behavior in other apps, but I see your point. I'll raise it during the design review and update it accordingly.
Thank you for all the reviews!

Issue #762, opened 2023-02-08
App Crashes When Opening Links without Browser Access


igotdes:
### Description

A user has reported a crash in Android when attempting to open links from the app when no browser can be used, such as when there is a device filter that blocks browser access.

The user suggested we display a pop-up notification similar to what WhatsApp does here:

<details>
<summary>Details</summary>

![Screenshot_20230203_164007_WhatsApp](https://user-images.githubusercontent.com/74066241/217462105-380bae04-297c-43f8-92b9-ba94ce9f7c79.jpg)

</details>

### Step-by-step reproduction instructions

1. Disable browser access for the device using a filter or similar
2. Click on a link in the app, such as one in an episode's notes
3. The app will crash

### Screenshots or screen recording

<details>
<summary>Details</summary>

https://user-images.githubusercontent.com/74066241/217461665-84dff4a4-b1e0-4ffa-ac34-91f11dcbcbde.mp4

</details>

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

Reported on:

* App version : 7.30 (9070)
* Phone: samsung - b0q - SM-S908U1 - b0q
* Android version: 13 SDK 33

mchowning:
Fixed by https://github.com/Automattic/pocket-casts-android/pull/772

PR #772, opened 2023-02-14
Adds toast message in case of ActivityNotFoundException on link click

josepatinob:
## Description
When clicking on a URL, the app attempts to launch the default browser Activity. In cases where the browser is disabled the app crashes because the ActivityNotFoundException is not handled. 

Fixes App Crashes When Opening Links without Browser Access #762 

## Testing Instructions
1. Disable browser access for the device using a filter or similar
2. Click on a link in the app, such as one in an episode's notes
3. The app should no longer crash and show a toast message instead

## Screenshots or Screencast 
image

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/services/localization/src/main/res/values/strings.xml
+++ b/modules/services/localization/src/main/res/values/strings.xml
@@ -1605,4 +1605,5 @@
     <string name="filter_id_description">An unique ID that represents the filter. Can be used on the \'Query Filter Episodes\' action.</string>
     <string name="filter_title">Filter Title</string>
     <string name="filter_id_or_title">Filter ID or Title</string>
+    <string name="error_opening_links_activity_not_found">You may not have the proper app for viewing this content.</string>
 </resources>
--- a/modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/helper/IntentUtil.kt
+++ b/modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/helper/IntentUtil.kt
@@ -4,9 +4,11 @@ import android.content.ActivityNotFoundException
 import android.content.Context
 import android.content.Intent
 import android.net.Uri
+import android.widget.Toast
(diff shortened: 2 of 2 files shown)


modules/services/localization/src/main/res/values/strings.xml
```diff
@@ -1605,4 +1605,5 @@
     <string name="filter_id_description">An unique ID that represents the filter. Can be used on the \'Query Filter Episodes\' action.</string>
     <string name="filter_title">Filter Title</string>
     <string name="filter_id_or_title">Filter ID or Title</string>
+    <string name="error_opening_links_activity_not_found">You may not have the proper app for viewing this content.</string>
```
mchowning, line 1608: I know this is the language that whatsapp uses, but "may" feels a little weak and confusing to users. What would you think about making this more direct with something along the lines of "Cannot find an app capable of viewing this content"? (feel free to improve the wording there)

mchowning (commented):
Thanks for fixing this crash @josepatinob ! 🎉 

I'm seeing some minor spotless violations, you can fix those by running `./gradlew spotlessApply` and committing the result (or you can just see the issues by running the `spotlessCheck` task).

Also, would you mind adding an entry for this fix to the [changelog](https://github.com/Automattic/pocket-casts-android/blob/main/CHANGELOG.md) section for 7.33? After that, I think this will be ready to merge.

Thanks for your contribution! ❤️

modules/services/localization/src/main/res/values/strings.xml
jpatino37: I have made the requested changes. Thanks for the feedback. @mchowning

josepatinob: I have made the requested changes. Thanks for the feedback. @mchowning

mchowning (approved):
Thanks for the updates. Everything looks great. And thanks again for contributing to Pocket Casts!

PR #774, opened 2023-02-15
Track signout on token refresh failure notification

mchowning:
## Description
Tracks when the you-ve-been-signed-out screen. is presented because the token refresh fails.

This turned out to be trickier than I expected because this notification originates from the settings module, but the analytics module depends on the settings module, so I can't directly make the analytics call from the settings module. Instead, I'm passing a function from the server module to be invoked when the UI is presented.

## Testing Instructions
1. The easiest way to trigger this condition is to build with [this change](https://gist.github.com/mchowning/7e56ca55560fa6007c6c147284dcb455) so that we request an invalid token.
2. Attempt to log in to an account from in the app
3. Observe the sign in error notification is presented
4. Observe the event: `signed_out_alert_shown`. Note that the event actually gets fired three times, but I'm not sure there's a good way to avoid that, and it looks like it is very possibly a consequence of how we're forcing the error—you'll see that there are also three `user_signed_in` events, but if you sign in without the code change from step 1, then only one user_signed_in event gets fired. If you know of a better way to test this error scenario that might not cause this side effect, let me know. Just noting that if we really want to avoid the multiple events here, [this](https://gist.github.com/mchowning/333b28beaa59efabefe3e77a4468bc44) is one possible (but hacky) approach that seemed to work in some limited testing.

## Checklist

### DIFF
--- a/modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/sync/SyncServerManager.kt
+++ b/modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/sync/SyncServerManager.kt
@@ -327,12 +327,6 @@ open class SyncServerManager @Inject constructor(
         cache.evictAll()
     }
 
-    fun cancelSupporterSubscription(subscriptionUuid: String): Single<Response<Void>> {
-        return getCacheTokenOrLogin {
-            server.cancelSubscription(addBearer(it), SupporterCancelRequest(subscriptionUuid))
-        }
-    }
-
     private suspend fun <T : Any> getCacheTokenOrLoginSuspend(serverCall: suspend (token: String) -> T): T {
         if (settings.isLoggedIn()) {
             return try {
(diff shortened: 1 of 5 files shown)


modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/sync/SyncServerManager.kt
```diff
@@ -325,16 +327,10 @@ open class SyncServerManager @Inject constructor(
         cache.evictAll()
     }
 
-    fun cancelSupporterSubscription(subscriptionUuid: String): Single<Response<Void>> {
-        return getCacheTokenOrLogin {
-            server.cancelSubscription(addBearer(it), SupporterCancelRequest(subscriptionUuid))
-        }
-    }
-
```
mchowning, line 333: This change is not tied to the goal of this PR, I just noticed the method was not used and removed it.

ashiagr:
That's really a tricky one 😀

> [...] Note that the event actually gets fired three times [...] Just noting that if we really want to avoid the multiple events here, [this](https://gist.github.com/mchowning/333b28beaa59efabefe3e77a4468bc44) is one possible (but hacky) approach that seemed to work in some limited testing.

I'd prefer avoiding multiple events. I actually like the idea of using a debounce to limit such events so you have my approval on that.

mchowning:
> I'd prefer avoiding multiple events. I actually like the idea of using a debounce to limit such events so you have my approval on that.

Added in [bec98f0](https://github.com/Automattic/pocket-casts-android/pull/774/commits/bec98f04a0e4178c3f1da04ba9d085826988efb0)

ashiagr (approved):

PR #776, opened 2023-02-15
Add track for when discover is shown

mchowning:
## Description
Adds the `discover_shown` event.

## Testing Instructions
1. Open the PC app
2. Navigate to the discover page if you're not already there
3. ✅ Observe the event `discover_shown`
4. Change the screen orientation
5. Observe that the `discover_shown` event is _not_ sent
6. Background the app
7. Reopen the app
8. ✅ Observe the event `discover_shown` is sent again

## Checklist

### DIFF
--- a/modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/view/DiscoverFragment.kt
+++ b/modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/view/DiscoverFragment.kt
@@ -48,7 +48,6 @@ class DiscoverFragment : BaseFragment(), DiscoverAdapter.Listener, RegionSelectF
 
     override fun onResume() {
         super.onResume()
-        viewModel.onShown()
     }
 
     override fun onPause() {
@@ -130,6 +129,8 @@ class DiscoverFragment : BaseFragment(), DiscoverAdapter.Listener, RegionSelectF
         if (viewModel.state.value !is DiscoverState.DataLoaded) {
             viewModel.loadData(resources)
         }
+
+        viewModel.onShown()
 
         return binding?.root
     }
--- a/modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/view/DiscoverFragment.kt
+++ b/modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/view/DiscoverFragment.kt
@@ -46,6 +46,16 @@ class DiscoverFragment : BaseFragment(), DiscoverAdapter.Listener, RegionSelectF
     private var adapter: DiscoverAdapter? = null
     private var binding: FragmentDiscoverBinding? = null
 
+    override fun onResume() {
+        super.onResume()
+        viewModel.onShown()
+    }
+
+    override fun onPause() {
+        super.onPause()
+        viewModel.onFragmentPause(activity?.isChangingConfigurations)
+    }
+
     override fun onPodcastClicked(podcast: DiscoverPodcast, listUuid: String?) {
         val fragment = PodcastFragment.newInstance(podcastUuid = podcast.uuid, fromListUuid = listUuid)
         (activity as FragmentHostListener).addFragment(fragment)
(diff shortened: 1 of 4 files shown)


modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/view/DiscoverFragment.kt
```diff
@@ -46,6 +46,16 @@ class DiscoverFragment : BaseFragment(), DiscoverAdapter.Listener, RegionSelectF
     private var adapter: DiscoverAdapter? = null
     private var binding: FragmentDiscoverBinding? = null
 
+    override fun onResume() {
+        super.onResume()
+        viewModel.onShown()
+    }
```
mchowning, line 52: I used `onResume` instead of `onCreate` because with `onCreate` I couldn't ever get the event to be sent a second time without killing the app. This way, backgrounding the app would resend the event. Let me know if you don't think that's desired behavior and I can switch it back to `onCreate`.

ashiagr: We're not tracking similar events like `podcasts_list_shown` when app is resumed from the background. Also, I checked that in iOS we track `discover_shown` even when returning from other screens like podcast details screen which is not tracked in Android if we use `onResume`.

Can we use `onCreateView` to track it?

mchowning: > I checked that in iOS we track discover_shown even when returning from other screens like podcast details screen which is not tracked in Android if we use onResume.

My thinking with `onResume` was that it gets us closer to that, but admittedly we'd still be a long ways from having the same behavior.

> We're not tracking similar events like podcasts_list_shown when app is resumed from the background. 

Good point. Changed to use `onCreateView` in [09f0948](https://github.com/Automattic/pocket-casts-android/pull/776/commits/09f0948de334f14256a948812945e8fc51f8910b) to be consistent with the other tabs.

ashiagr: Unlike other tabs, using `onCreateView` for `onShown` doesn't fire events when returning from other fragments. Not sure how `DiscoverFragment` is different from other tabs like `FiltersFragment` or `PodcastsFragment`. 🤔

mchowning: > Unlike other tabs, using onCreateView for onShown doesn't fire events when returning from other fragments.

Yeah, this is what sent me down the path of using `onResume` initially. I think you're right though that using `onResume` doesn't really give us much better data.

> how DiscoverFragment is different from other tabs like FiltersFragment or PodcastsFragment

Discover is the only tab where the fragment is set up to [not be detachable](https://github.com/Automattic/pocket-casts-android/blob/f8c55d3048191464bc156bc129222005c9e3a068/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt#L288)...

``` kotlin
val tabs = mapOf(
    VR.id.navigation_podcasts to { FragmentInfo(PodcastsFragment(), true) },
    VR.id.navigation_filters to { FragmentInfo(FiltersFragment(), true) },
    VR.id.navigation_discover to { FragmentInfo(DiscoverFragment(), false) }, // <---
    VR.id.navigation_profile to { FragmentInfo(ProfileFragment(), true) }
)
```

The commit making that change was titled "Discover flashing" and I do notice that if I make the fragment detachable there is a much more noticeable flash every time I switch to that tab as it loads the screen. On the flip side, the app might generally perform just a bit better if it's not always keeping the `DiscoverFragment` in memory.

I'm not inclined to make any changes here at this time since we don't have a clear need to make this analytic event work better now—it seems like we can probably extrapolate most of the data we need from `discover_tab_opened`. If it turns out we really need better data here, then we can spend some more time trying to fix this, but I think it will require either making the app perform worse or making some fairly large scale changes to how the app handles navigation.

Let me know what you think though.

ashiagr: I see. Let's not make large-scale changes and keep it the way it is. Thanks for looking into it.

ashiagr (approved):

PR #781, opened 2023-02-16
Track pull to refresh

mchowning:
## Description
Add track for pull to refresh.

## Testing Instructions
1. Start the app
2. Log in with a Plus account
3. Make sure you are subscribed to at least 1 podcast
4. Go to the podcasts tab
5. Pull to refresh
6. ✅ Observe the event. `pulled_to_refresh, Properties: {"source":"podcasts_list", ... }`
7. Go to the Profile tab → Files
8. Pull to refresh
9. ✅ Observe the event `pulled_to_refresh`. The `source` property will vary depending on whether you have any files. If you have files, it should be "files" and if you do not it should be "no_files".

## Checklist


ashiagr (approved):
Looks good 👍

Issue #787, opened 2023-02-20
Inline search results, segmented by podcasts/episodes


ashiagr:
The search results will be divided in two areas: podcasts and episodes. Each area will display a "Show all" link that will carry the user to the podcasts/episodes search result screen.

#### View Related
- [x] Initial prep ([#789](https://github.com/Automattic/pocket-casts-android/pull/789))
    - [x] Extract existing result views to Compose
    - [x] Add feature flag to show new search results view
    - [x] Add section headers with show all buttons for new search results view
- [x] Add new podcasts results section ([#792](https://github.com/Automattic/pocket-casts-android/pull/792))
- [x] Add episodes results section ([#792](https://github.com/Automattic/pocket-casts-android/pull/792))

#### Data
- [x] Show episodes from new episode search result API ([#792](https://github.com/Automattic/pocket-casts-android/pull/792))

#### Other
- [x] Add episode to search history on episode search result row click action ([#794](https://github.com/Automattic/pocket-casts-android/pull/794))
- [x] Add subscribe click action ([#794](https://github.com/Automattic/pocket-casts-android/pull/794))
- [x] Episode history row click action ([#794](https://github.com/Automattic/pocket-casts-android/pull/794))

PR #792, opened 2023-02-21
Inline search results - podcasts and episodes

ashiagr:
Part of: #787

## Description

This PR 
- Adds podcasts and episodes layout to the new search view
- Adds episodes from the search episodes endpoint

#### Feature flag - On

1. Enable search improvements feature flag in `base.gradle`
2. Select `debug` build variant
3. Go to `Podcasts` tab
4. Search a term
5. ✅ Notice that the first 20 podcasts search results are shown under `Podcasts` heading
    * Subscribed podcasts and folders are shown under the Podcasts section (pdeCcb-Xe-p2#comment-1431)
6. ✅ Notice that the first 20 (server) episodes search results are shown under `Episodes` heading
7. Go to `Discover` tab
8. ✅ Notice that only server results are shown for both podcasts and episodes 

#### Feature flag - Off

1. Disable search improvements feature flag in `base.gradle`
2. ✅ Notice that old search works as expected

Note that
* Local episodes not added
* Following actions are not working
   * subscribe 
   * show all 
   * episode click
* Search failed/ No results found views are not updated 
* I'll create a separate PR for a final design review.

## Checklist
-  If this is a user-facing change, I have added an entry in CHANGELOG.md
 
- [x] with different themes

### DIFF
--- a/modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchResultsPage.kt
+++ b/modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchResultsPage.kt
@@ -66,8 +66,8 @@ fun SearchResultsPage(
     val state = viewModel.searchResults.collectAsState(
         SearchState.Results(
             searchTerm = "",
-            list = emptyList(),
-            episodeItems = emptyList(),
+            podcasts = emptyList(),
+            episodes = emptyList(),
             error = null,
             loading = false
         )
@@ -80,7 +80,7 @@ fun SearchResultsPage(
                 val result = state.value as SearchState.Results
                 if (result.error == null || !onlySearchRemote || result.loading) {
                     if (BuildConfig.SEARCH_IMPROVEMENTS_ENABLED) {
-                        if (result.list.isNotEmpty()) {
+                        if (result.podcasts.isNotEmpty()) {
                             SearchResultsView(
                                 state = state.value as SearchState.Results,
                                 onPodcastClick = onPodcastClick,
@@ -141,7 +141,7 @@ private fun SearchResultsView(
         item {
             LazyRow(contentPadding = PaddingValues(horizontal = 8.dp),) {
                 items(
-                    items = state.list,
+                    items = state.podcasts,
                     key = { it.adapterId }
                 ) { folderItem ->
                     when (folderItem) {
@@ -164,7 +164,7 @@ private fun SearchResultsView(
             }
         }
         item { SearchResultsHeaderView(title = stringResource(LR.string.episodes)) }
-        state.episodeItems.forEach {
+        state.episodes.forEach {
             item {
                 SearchEpisodeItem(
                     episode = it,
@@ -224,7 +224,7 @@ private fun OldSearchResultsView(
             .nestedScroll(nestedScrollConnection)
     ) {
         items(
-            items = state.list,
+            items = state.podcasts,
             key = { it.adapterId }
         ) { folderItem ->
             when (folderItem) {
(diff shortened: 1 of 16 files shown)


modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchResultsPage.kt
```diff
@@ -107,10 +121,65 @@ fun SearchResultsPage(
...
+            )
+        }
+        item { SearchResultsHeaderView(title = stringResource(LR.string.episodes)) }
+        state.episodes.forEach {
+            item {
+                SearchEpisodeItem(
+                    episode = it,
+                    onClick = {},
+                )
+            }
```
mchowning, line 181: Do we want to use `items` here instead of `forEach { item {} }`? If not, I think we could replace the `LazyColumn` with a `Column`.

modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/components/TextStyles.kt
```diff
@@ -305,6 +305,26 @@ fun TextH70(
...
+    text: String,
+    modifier: Modifier = Modifier,
+    maxLines: Int = Int.MAX_VALUE
+) {
+    Text(
+        text = text.uppercase(Locale.getDefault()),
+        color = MaterialTheme.theme.colors.primaryText02,
+        fontFamily = FontFamily.SansSerif,
+        fontSize = 13.sp,
+        fontWeight = FontWeight.Bold,
```
mchowning, line 319: I prefer using `FontWeight.W700` style of constants because I find that to be clearer and easier to match up with the Figma designs. But what do you think? Definitely not a blocker.

mchowning (approved):
These changes look good. Left a couple of small comments, but they're nothing big.

I did notice what I think is a regression, but not a regression in this PR. As I am typing out my search, I have some local results appear and then disappear. I see this happening on `main` without the feature flag enabled as well. FWIW, I don't see this behavior in `7.32`. Feel free to address this in a separate PR.

| Feature flag on | Feature flag off |
| --- | --- |
| <video src="https://user-images.githubusercontent.com/4656348/220387342-87574231-ba60-45e4-a347-4da89e5ce111.mov" /> | <video src="https://user-images.githubusercontent.com/4656348/220386969-43ded3ca-5c28-4579-9398-882a68cc3137.mov" /> |

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchResultsPage.kt
ashiagr: Oh yes, used items in c8f7cab. Thanks for pointing.

modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/components/TextStyles.kt
ashiagr: I picked bold style for `C50` from Figma but don't have any issues in using the other style. 
Replaced in 22ab72a.

ashiagr:
> I did notice what I think is a regression, but not a regression in this PR. As I am typing out my search, I have some local results appear and then disappear. I see this happening on main without the feature flag enabled as well. FWIW, I don't see this behavior in 7.32. Feel free to address this in a separate PR.

This seems to happen when the server response has not been triggered and no results are present in the local result. I can see this behavior also in `7.32`. I'll keep an eye on it and fix it separately.

PR #793, opened 2023-02-21
Update add/wear branch

mchowning:
## Description
Bringing the `add/wear` branch up-to-date by:

1. Merging in the latest from `main`
2. Updating horologist fro `0.2.7` to `0.3.2`
3. Updating wear-compose from `1.1.2` to `1.2.0-alpha4`

From `0.3.0+` horologist is using the `1.2.0-alpha` of wear-compose. Although those are alphas, they are alpha because of potential API changes, not because of stability concerns. For that reason, I think our best bet is to go ahead with the latest version of horologist.

I did break the dependency upgrades into separate commits for each version bump (9f1b3e07d9beb1f483a240a0371e87b5063634a1, 856f9b7e25f34f83b62b98c6bb805eb8b2d77560, 6329f9c0d18817ca1aa856dee328e1996fb9c1fe, 8da93e8f19ffd91e0e980274fb5d753acb1c7a39, b0b5e500fe61fff98602fa77b5f9243b0e4b7b47), so it shouldn't be too hard if we end up deciding that the latest-and-greatest is not so great for us and we want to go back to a specific version.

## Testing Instructions
Just do some smoke testing to make sure that the basic functionality mostly works. This all wasn't working super smoothly before, so as long as there's nothing too egregious happening, I think we should be good to merge this.

## Checklist


ashiagr (approved):
I was able to load the screens and didn't encounter any crashes in my basic testing which is what I think is the scope of this PR, so I'll go ahead and approve the PR.

Nice thinking on breaking the commits for each version bump. 👍

PR #797, opened 2023-02-24
Correct request parameter name for user/token endpoint

mchowning:
## Description
Just a simple update to a parameter name to match what the sync server is expecting.

## Testing Instructions
I don't believe we're actively using the user/token in the app yet, so I don't think there's anything to test. Just check this parameter name against the swagger docs and server code.

## Checklist


mchowning:
Closing because this change is in https://github.com/Automattic/pocket-casts-android/pull/712

Issue #786, opened 2023-02-19
Automotive: Improve sign in or create account pages


geekygecko:
### Description

The Automotive app sign in or create account pages could be improved.
- The rest of the app is in a dark theme it would be good for the account pages to also be dark.
- The email text field prompt is small when entering the email.

### Step-by-step reproduction instructions

1. Deploy the automotive app to the Automotive emulator
2. Open Pocket Casts
3. Tap the settings cog
4. Tap "Set Up Account" option

### Screenshots or screen recording

<img src="https://user-images.githubusercontent.com/308331/219937598-1e8715f5-e78a-4f4d-a929-abeca5cde587.png" width="600" />

<img src="https://user-images.githubusercontent.com/308331/219937608-06774ee1-48e5-4690-8829-a67e47d8d649.png" width="600" />

<img src="https://user-images.githubusercontent.com/308331/219937624-56855726-68de-47b6-ac4d-61a8fe35fec8.png" width="600" />

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

PR #798, opened 2023-02-24
Improving the Automotive account pages

geekygecko:
## Description
This change does the following:
- Forces the Automotive app into car mode as some car companies don't set this and our car layout resources aren't picked up.
- The font sizes for the hint text have increased on the account pages.
- Improved the UI on the forgotten password page.
- Made sure only the dark theme is used in the Automotive version.

Fixes https://github.com/Automattic/pocket-casts-android/issues/786

## Testing Instructions
- Deploy the automotive app to the Automotive emulator
- Open Pocket Casts
- Tap the settings cog
- Tap "Set Up Account" option
- Tap "Create account"
- ✅ Verify the page looks better
- Tap back and the "Sign in"
- ✅ Verify the page looks better
- Tap "I forgot my password"
- ✅ Verify the page looks better

## Screenshots 
| Before | After |
| --- | --- |
| Screenshot_20230224_155050 | Screenshot_20230224_155241 |
| Screenshot_20230224_155058 | Screenshot_20230224_155250 |

### DIFF
--- a/automotive/src/main/java/au/com/shiftyjelly/pocketcasts/AutomotiveApplication.kt
+++ b/automotive/src/main/java/au/com/shiftyjelly/pocketcasts/AutomotiveApplication.kt
@@ -2,7 +2,9 @@ package au.com.shiftyjelly.pocketcasts
 
 import android.annotation.SuppressLint
 import android.app.Application
+import android.app.UiModeManager
 import android.util.Log
+import androidx.core.content.getSystemService
 import androidx.hilt.work.HiltWorkerFactory
 import androidx.work.Configuration
 import au.com.shiftyjelly.pocketcasts.account.AccountAuth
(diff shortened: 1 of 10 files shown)


automotive/src/main/java/au/com/shiftyjelly/pocketcasts/AutomotiveApplication.kt
```diff
@@ -82,6 +84,9 @@ class AutomotiveApplication : Application(), Configuration.Provider {
         userEpisodeManager.monitorUploads(applicationContext)
         downloadManager.beginMonitoringWorkManager(applicationContext)
         userManager.beginMonitoringAccountManager(playbackManager)
+
+        // force the Automotive app into car mode as some car companies send the UI mode as normal, this makes sure the car resources such as layout-car are used.
+        this.getSystemService<UiModeManager>()?.enableCarMode(0)
```
mchowning, line 89: I don't think I've ever seen a function whose parameter only had a single valid value before. 😆 

image

mchowning (approved):
Works well and the code looks good! 👍

PR #800, opened 2023-02-24
Automotive: When no browser is available display the URL with a QR code

geekygecko:
## Description
It seems common that Automotive platforms still need to provide a web browser. This solution offers a fallback which shows the user a page with the URL address and a QR code for them to open the page on their phones. 

The other option was for us to embed a web browser but pages like our Terms of Service link off to other sites and it could start getting complicated figuring out which sites to allow.

This takes inspiration from the Google Automotive link viewer.

## Testing Instructions
1. Comment out the code so only the fallback code in the catch block runs.
2. Deploy the Automotive build to an Automotive emulator
3. Tap the settings cog
4. Scroll down and tap About
5. Tap Terms of Service
6. ✅  Verify the URL is shown with the QR code

## Screenshots 
Screenshot_20230225_080901


ashiagr (approved):
Works as described, thanks for adding this nice improvement 👍

Issue #468, opened 2022-10-23
Wear OS version


yechielw:
### What problem does this address?

Currently, there is no way to listen to podcasts when the wear OS watch is not connected to the phone via bluetooth

### What is your proposed solution?

Publishing a trimmed down version of the app targeted for wear OS which will enable the options to stream and Download episodes directly from the watch.

at the moment, none of the big players (besides Spotify) offer a watch version of their podcast app.

### Did you search for existing feature suggestions?

- [X] I have searched for existing feature suggestions.

mchowning:
We have [gotten a start on this](https://github.com/Automattic/pocket-casts-android/tree/add/wear), but it's really just a start. Other priorities are keeping us from spending much time on it at this time. Contributions would be welcome though! 😄

Salzian:
@mchowning Is there some coordination on this effort already? I would love to contribute.

mchowning:
>  I would love to contribute.

That would be great @Salzian ! 

> Is there some coordination on this effort already?

We were mostly coordinating on private channels before now, but going forward this thread is probably the best place to coordinate now that we want to do that publicly (I'm open to other suggestions though).

We're largely looking to replicate the functionality from Pocket Casts' iOS watch app to start—not because we want the WearOS to be exactly the same as the iOS watch app, but more because that seems like the easiest way to create our first iteration for WearOS. You can do an iOS build yourself to explore the watch app on a simulator, but in case you can't (or don't want) to do that, this is what it looks like.

![Apple Watch](https://user-images.githubusercontent.com/4656348/201352755-470b9340-44b3-429d-b6f6-c04210c310be.png)

Our plan was to support offline playback as suggested in https://github.com/Automattic/pocket-casts-android/issues/490.

My inclination would be to focus on the Now-Playing/Up-Next functionality first since that seems like the core functionality. We might even be able to do a release with WearOS support with only that functionality. 🤔  If there's something else you're interested in working on though, feel free to tackle that.

Salzian:
I'm going to review progress on branch `add/wear` first. I agree that for the beginning, we should focus on getting playback working on Wear OS, especially for standalone users.

mchowning:
> Mostly initial setup of the app and the UI layout of it, am I right?

That's right

>  think a first step would be to establish the connection between watch and phone to remotely control playback on the phone.

That makes sense to me. 👍

Salzian:
[...]

PR #801, opened 2023-02-25
Sending refresh token from phone to watch

mchowning:
## Description
This PR automatically authenticates a user on a paired watch when they are logged into Pocket Casts on their phone by:

1. Retrieving a refresh token on the phone.
2. Sending that refresh token to the watch using the [Wearable Data Layer API](https://developer.android.com/training/wearables/data/data-layer).
3. When the watch receives a refresh token, the watch will use it to log in with that token.

One issue with the current implementation is that it's pretty pushy about logging you in on the watch if you're logged in on a connected phone. For example, if you're logged in on both devices, and you log out on the watch, then the next time you reopen the app you'll be logged back in on the watch. I'm not sure how I feel about this, it doesn't seem ideal, but it feels like an edge case for a user to want to be logged in on their phone but not their watch. If a user did want to keep them separate, they would just need to log into a different account on the watch. To me the current implementation feels good enough for now, but let me know if you have any thoughts.

https://user-images.githubusercontent.com/4656348/233832495-a3345807-9182-4fba-9239-ae877a930f13.mov

## Testing Instructions

When doing these tests, make sure you're using the same build variant on both the watch and the phone. If you don't, the syncing won't work. You may also want to comment out [this line](https://github.com/Automattic/pocket-casts-android/blob/d486b0ac7924402685a5f38422ce4d1796a29b8e/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/MainActivity.kt#L242) while you are testing so the notification doesn't get automatically dismissed too quickly. Note that I do have it set up so that the notification should show for a minimum of 5 seconds to ensure the user has time to see it and realize what is happening since this isn't a user initiated action but happens automatically when a refresh token is found.

1. Use a phone and a watch that are paired
2. Remove Pocket Casts from both the phone and the watch
3. Install the Pocket Casts app module from this branch onto the phone
5. Log into Pocket Casts on the phone
6. Install the Pocket Casts wear module from this branch onto the watch
7. ✅ Observe that the watch app is logged in pretty much immediately with a "logging in" screen shown while the refresh is in progress. This screen will show a spinning placeholder image if a gravatar image is not available for this account or if the gravatar image is still loading. I have this screen set to show for at least 5 seconds even if the refresh is finished earlier in order to make sure that there is enough time for the user to see this screen and realize that they are being logged in automatically.
8. ✅  After the refresh is complete, the "logging in" screen should be dismissed and the Now Playing and Up Next chips should be updated appropriately (assuming the account has a currently playing episode and it's Up Next queue is not empty)
9. Log out of the app on the phone
10. ✅  Observe that the watch app is still logged in
11. Log out of the watch app
12. Log into Pocket Casts on the phone
13. ✅ Observe that the watch app is logged in pretty much immediately

## Checklist


mchowning:
> 2. Sending that refresh token to the watch by using the [Wearable Data Layer API](https://developer.android.com/training/wearables/data/data-layer).

This [is Google's recommended approach](https://developer.android.com/training/wearables/apps/auth-wear#tokens), and has the benefit that the token gets stored so that once the phone sends the token to the data layer, the phone can disconnect from both the network and the watch (i.e., go into airplane mode), and the watch will still be able to retrieve the token if it has network access. 

The tradeoff here is that the token is synced to Google's servers (h/t  @NoahAndrews comment on the Pocket Casts Wear OS issue: https://github.com/Automattic/pocket-casts-android/issues/468#issuecomment-1442772267). 

I used the Data Layer API approach for this initial POC because it's the Google-recommended approach and seems to provide a better user experience. The other option would be to [send messages](https://developer.android.com/training/wearables/data/messages) between the phone and watch.

mchowning:
I have this PR labeled as a work in progress because there are still issues that need to be addressed in the code, but I believe the functionality described in the test steps is all working, so feel free to go ahead and test this out and let me know any issues you see.

wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/MainActivity.kt
```diff
@@ -32,13 +39,40 @@ class MainActivity : ComponentActivity() {
...
+                Lifecycle.Event.ON_START -> dataClient.addListener(viewModel.phoneSyncDataListener)
+                Lifecycle.Event.ON_STOP -> dataClient.removeListener(viewModel.phoneSyncDataListener)
+                else -> { /* do nothing */
+                }
+            }
+        }
+        lifecycle.addObserver(listener)
+        return onDispose {
+            lifecycle.removeObserver(listener)
+        }
```
mchowning, line 74: Instead of using a listener, we could use [a service](https://developers.google.com/android/reference/com/google/android/gms/wearable/WearableListenerService) to always listen for changes even when the Pocket Casts wear app is in the background. I don't think we need to listen and respond to these events until the app comes to the foreground, so I went with the listener approach. I do think either approach could work though, so let me know if you have other thoughts.

NoahAndrews:
As a part of [a response](https://github.com/google/horologist/issues/1036#issuecomment-1449261327) to https://github.com/google/horologist/issues/1036, I compared Google's recommendations for syncing credentials in phone apps to the ones for Wear OS, and found that the phone app guidelines clearly discourage syncing credentials to the cloud without end-to-end encryption. This research made me more convinced that the better path here is to keep tokens out of `DataClient`, unless Google documents that that data is end-to-end encrypted somehow (which seems unlikely).

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/WatchSync.kt
```diff
@@ -0,0 +1,135 @@
...
+        withContext(Dispatchers.Default) {
+            try {
+                Timber.i("Updating refresh token in data layer")
+
+                val authData = let {
+                    val email = settings.getSyncEmail()
+                    val password = settings.getSyncPassword()
+                    if (email != null && password != null) {
+                        // FIXME nonononono this makes an api call _every_ time
+                        accountAuth.getTokensWithEmailAndPassword(email, password)
```
NoahAndrews, line 53: Is the intention to stop storing the password in `AccountManager` altogether, and store the refresh token instead? It seems like that would be the most straightforward way to keep track of everything. Only if the user does not have a refresh token stored would you check if there's a password stored. The less state to keep track of, the better.

Let me know if I'm overstepping my bounds at all with this sort of feedback.

mchowning:
Thanks for the great feedback @NoahAndrews 

> As a part of https://github.com/google/horologist/issues/1036#issuecomment-1449261327 to https://github.com/google/horologist/issues/1036, I compared Google's recommendations for syncing credentials in phone apps to the ones for Wear OS, and found that the phone app guidelines clearly discourage syncing credentials to the cloud without end-to-end encryption. This research made me more convinced that the better path here is to keep tokens out of DataClient, unless Google documents that that data is end-to-end encrypted somehow (which seems unlikely).

I was a bit surprised that the response to your issue wasn't a simple "Yes, we should clarify that in the documentation, but of course our recommended approach for sharing authentication tokens is fully encrypted and secure." Based on [their response](https://github.com/google/horologist/issues/1036#issuecomment-1446605228), it sounds like it very well may just be a documentation issue, but we don't know that yet. 😞 

> IANAL, but storing anything on Google's servers without explicit consent looks to me like it would violate the current [Pocket Casts privacy policy](https://support.pocketcasts.com/article/privacy-policy/).

Thanks for bringing this up. If we decide we want to release using the Data Layer approach, we'll make sure we run this by the legal team.

> Let me know if I'm overstepping my bounds at all with this sort of feedback.

Not at all. Definitely appreciate your input. Keep it coming! 😄 

I'm not quite as convinced as you are that we have to use the messages approach. You are raising good points though, and I'm also not convinced that we should use the Data Layer either. My inclination is to stick with the Data Layer approach for now, and to see what more information we get from Google on [the horologist issue](https://github.com/google/horologist/issues/1036) you opened (thanks for opening that btw 🙇 ). Then we can revisit this before we release our wear app to see if we want to switch away from the Data Layer approach.

With that said, if you're interested in prototyping what this would look like using the messages approach, that would be great! Maybe we'll see that it works so well that we want to go that way regardless of Google's answer to the encryption question. If Google does confirm that the Data Layer is end-to-end encrypted though, we might end up sticking with the Data Layer.

Just for the sake of completeness, I'll note that another authentication method on my mind is [custom code authentication](https://developer.android.com/training/wearables/apps/auth-wear#custom-authentication). That feels a bit more complicated than I'd like to use for the initial release, but I think it may be something we want to consider later since that would facilitate authentication of watches paired to iOS devices and watches that are not paired to a phone.

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/WatchSync.kt
mchowning: > Is the intention to stop storing the password in AccountManager altogether, and store the refresh token instead?

Yes, I believe that will be a part of @geekygecko 's work on [Google Single-On work](https://github.com/Automattic/pocket-casts-android/pull/712). The api call on this line to retrieve the refresh token is just a placeholder until that work is done and we can retrieve the token locally.

yschimke:
Some info on data layer security

- All data stored in the data layer is likely to go to the cloud
- All data that goes to the cloud is end to end encrypted
- Outdated devices that do not support E2EE no longer support Cloud Sync at all
- Encryption keys are generated by the devices and exchanged exclusively over local Bluetooth ASAP (generally during device setup). Google never receives keys, or even encrypted messages that contain keys.
- There is no key rotation.
- Messages sent with MessageClient or ChannelClient will also be automatically routed through the cloud when Bluetooth is unavailable (but won’t be if it is)
- Messages sent via either of those are also subject to the same E2EE setup as the DataClient, and cannot be read by Google either.

mchowning:
Thanks so much for confirming the security around the Data Layer @yschimke ! In light of that, I don't think there's any reason for us to avoid using the Data Layer from a security perspective. Do you agree @NoahAndrews ?

mchowning:
> Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/Automattic/pocket-casts-android?pullRequest=801) before we can accept your contribution.

👋 @luizgrp ! I included [your commit](https://github.com/Automattic/pocket-casts-android/pull/801/commits/033afa9ecc382dce4ba7732743e2972aa4079876) in this PR. Can you [sign our CLA](https://github.com/Automattic/pocket-casts-android/pull/801#issuecomment-1518457515)?

luizgrp:
@mchowning done, apologies for the delay!

ashiagr (approved):
I have gone through the changes, and all of them seem good to me. Auto-login works as described in the testing instructions. 👍 

Nice job on showing the gravatar while logging in!

Issue #805, opened 2023-02-27
Full screen search results screen


ashiagr:
The full-screen search result screen will be shown on navigating from inline search results screen using  "SHOW ALL" action.
It will display all podcasts or all episodes search results.

Whenever a user selects any of the search results, the selected item must be saved into the user's search history.

PR #806, opened 2023-02-27
Full screen search results

ashiagr:
Part of #805 

## Description
This adds full-screen search results for podcasts and episodes.

## Testing Instructions

Prerequisites: 
- Enable search improvements feature flag in `base.gradle`
- Select `debug` build variant

#### All Podcasts

1. Go to `Podcasts` or `Discover` tab
2. Search a term that returns podcasts and episodes
3. Tap on "Show all" from the podcasts section
4. ✅ Notice that all podcast search results are shown
5. Tap on plus (subscribe) button
6. ✅ Notice that podcast is subscribed (** To unsubscribe, you'll need to go to the details screen that displays a confirmation dialog before unsubscribing)
9. Tap on a podcast
10. ✅ Notice that podcast details are shown
11. Go back to see the search history
12. ✅ Notice that the podcast is added to search history

#### All Episodes

1. Go to `Podcasts` or `Discover` tab
2. Search a term that returns podcasts and episodes
3. Tap on "Show all" from the episodes section
4. ✅ Notice that all episodes search results are shown
5. Tap on an episode
6. ✅ Notice that episode details are shown
7. Go back to see the search history
8. ✅ Notice that the episode is added to search history

#### Note 
Shared apk for design review: pdeCcb-1iH-p2#comment-1749

## Screenshots or Screencast 
All podcasts | All episodes
---| ----
all_podcasts | all_episodes

## Checklist
 
- [x] with different themes

### DIFF
--- a/modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchResultsFragment.kt
+++ b/modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchResultsFragment.kt
@@ -1,0 +1,77 @@
+package au.com.shiftyjelly.pocketcasts.search
+
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.ViewGroup
+import androidx.compose.ui.platform.ComposeView
+import androidx.compose.ui.platform.ViewCompositionStrategy
+import androidx.fragment.app.activityViewModels
+import au.com.shiftyjelly.pocketcasts.compose.AppThemeWithBackground
+import au.com.shiftyjelly.pocketcasts.views.fragments.BaseFragment
+import dagger.hilt.android.AndroidEntryPoint
+
+private const val ARG_TYPE = "arg_type"
+
+@AndroidEntryPoint
+class SearchResultsFragment : BaseFragment() {
+    private val viewModel by activityViewModels<SearchViewModel>()
+
+    val type: ResultsType
+        get() = ResultsType.fromString(arguments?.getString(ARG_TYPE))
+
+    override fun onCreateView(
+        inflater: LayoutInflater,
+        container: ViewGroup?,
+        savedInstanceState: Bundle?,
+    ) = ComposeView(requireContext()).apply {
+        setContent {
+            AppThemeWithBackground(theme.activeTheme) {
+                setViewCompositionStrategy(ViewCompositionStrategy.DisposeOnViewTreeLifecycleDestroyed)
+                when (type) {
+                    ResultsType.PODCASTS -> {
+                        SearchPodcastResultsPage(
+                            viewModel = viewModel,
+                            onBackClick = ::onBackClick,
+                        )
+                    }
+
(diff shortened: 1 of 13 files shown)


modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchResultsFragment.kt
```diff
@@ -0,0 +1,158 @@
...
+import au.com.shiftyjelly.pocketcasts.views.fragments.BaseFragment
+import au.com.shiftyjelly.pocketcasts.views.helper.UiUtil
+import dagger.hilt.android.AndroidEntryPoint
+
+private const val ARG_ONLY_SEARCH_REMOTE = "arg_only_search_remote"
+private const val ARG_SOURCE = "arg_source"
+private const val ARG_TYPE = "arg_type"
+@AndroidEntryPoint
+class SearchResultsFragment : BaseFragment() {
+    private val viewModel by viewModels<SearchViewModel>({ requireParentFragment() })
```
ashiagr, line 28: I used an activity scope view model earlier to share search results and subscribe actions between search (inline results) fragment and search (full-screen) results fragment but noticed that previous search results were shown in the parent search fragment when it was created. 

I now create a full-screen child search results screen and use a parent fragment scope view model to share results + subscribe actions and not mess with the existing implementation. Let me know if you see any drawbacks.

mchowning: Makes sense to me. 👍

modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/components/PodcastItem.kt
```diff
@@ -32,14 +32,16 @@ fun PodcastItem(
     showSubscribed: Boolean = false,
     showPlusIfUnsubscribed: Boolean = false,
     showDivider: Boolean = true,
+    onSubscribeClick: (() -> Unit)? = null,
```
mchowning, line 35: If we renamed this parameter to be `onPlusClick` the connection between this parameter and the `showPlusIfUnsubscribed` parameter would be more obvious to me. This might just be my personal preference though, so feel free to leave as-is if you prefer.

ashiagr: Good point, I renamed it to `onPlusClick` - 83727aa

ashiagr:
> Tapping on the green check opened the podcast instead of unsubscribing from the podcast. I'm fine with that behavior since the user can unsubscribe from the podcast screen, but I just wanted to call it out since it is different from the behavior expected in the test steps.

My bad, I updated the instructions. I matched the behavior on the discover tab - popular lists which also open the podcast instead of unsubscribing from them. 

Thank you for the careful review!

PR #807, opened 2023-03-01
Search results - Design review and improvements

ashiagr:
Part of #787 and #805

## Description

This 
- Addresses search results design review comments: pdeCcb-1iH-p2#comment-1756 and
- Displays podcast and episode search results as they become available.

Prerequisites: 
- Enable search improvements feature flag in `base.gradle`
- Select `debug` build variant

## Testing Instructions
- Launch the app and connect to the debugger
- Filter for `.pocketcasts.net` in logs
- Go to `Discover` tab
- Search a term
- Notice that podcast and episode search results are shown as they become available ([logs](https://user-images.githubusercontent.com/1405144/222134280-de57ab89-bdb0-4ad8-a845-054cbb671b6b.png)) in search results.

## Screenshots or Screencast 

https://user-images.githubusercontent.com/1405144/222135053-bbbe3122-226d-46ea-8b1e-62d0274b66e6.mp4

#### Updated screenshots

out

### DIFF
--- a/modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchHandler.kt
+++ b/modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchHandler.kt
@@ -128,7 +128,6 @@ class SearchHandler @Inject constructor(
                         globalSearch = globalSearch.copy(podcastSearch = podcastSearch)
                         globalSearch
                     }
-                    .subscribeOn(Schedulers.io())
                     .toObservable()
 
                 if (BuildConfig.SEARCH_IMPROVEMENTS_ENABLED) {
@@ -138,8 +137,6 @@ class SearchHandler @Inject constructor(
                             globalSearch = globalSearch.copy(episodeSearch = episodeSearch)
                             globalSearch
                         }
-                        .subscribeOn(Schedulers.io())
-                        .toObservable()
 
                     podcastServerSearch.mergeWith(episodesServerSearch)
                 } else {
--- a/modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchHandler.kt
+++ b/modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchHandler.kt
@@ -120,28 +120,37 @@ class SearchHandler @Inject constructor(
             } else {
                 analyticsTracker.track(AnalyticsEvent.SEARCH_PERFORMED, AnalyticsProp.sourceMap(source))
                 loadingObservable.accept(true)
+
+                var globalSearch = GlobalServerSearch(searchTerm = it)
                 val podcastServerSearch = serverManager
                     .searchForPodcastsRx(it)
+                    .map { podcastSearch ->
+                        globalSearch = globalSearch.copy(podcastSearch = podcastSearch)
+                        globalSearch
+                    }
                     .subscribeOn(Schedulers.io())
                     .toObservable()
 
                 val episodesServerSearch = cacheServerManager
                     .searchEpisodes(it)
+                    .map { episodeSearch ->
+                        globalSearch = globalSearch.copy(episodeSearch = episodeSearch)
+                        globalSearch
+                    }
                     .subscribeOn(Schedulers.io())
                     .toObservable()
 
                 podcastServerSearch
-                    .zipWith(episodesServerSearch) { podcastsSearch, episodesSearch ->
-                        GlobalServerSearch(
-                            searchTerm = it,
-                            error = podcastsSearch.error,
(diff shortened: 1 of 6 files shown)


modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchHandler.kt
```diff
@@ -120,28 +120,40 @@ class SearchHandler @Inject constructor(
...
+                            globalSearch
+                        }
+                        .subscribeOn(Schedulers.io())
+                        .toObservable()
+
+                    podcastServerSearch.mergeWith(episodesServerSearch)
+                } else {
+                    podcastServerSearch
+                }
                     .subscribeOn(Schedulers.io())
```
mchowning, line 148: With this `.subscribeOn(Schedulers.io()` do we also need the ones on lines 131 and 141? I'm wondering if this might be the only one we need.

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchHandler.kt
```diff
@@ -120,28 +120,40 @@ class SearchHandler @Inject constructor(
...
-                    .searchEpisodes(it)
+                if (BuildConfig.SEARCH_IMPROVEMENTS_ENABLED) {
+                    val episodesServerSearch = cacheServerManager
+                        .searchEpisodes(it)
+                        .map { episodeSearch ->
+                            globalSearch = globalSearch.copy(episodeSearch = episodeSearch)
+                            globalSearch
+                        }
+                        .subscribeOn(Schedulers.io())
+                        .toObservable()
```
mchowning, line 142: It doesn't really do any harm, but I don't think we actually need to convert this `Single` to an `Observable` because there is an `Observable::mergeWith` method that accepts a `Single` as its parameter. Certainly not a blocker (and maybe I'm overlooking something), so feel free to leave it as-is.

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchHandler.kt
```diff
@@ -157,7 +169,7 @@ class SearchHandler @Inject constructor(
             val searchPodcastsResult = (localPodcastsResult + serverPodcastsResult).distinctBy { it.uuid }
             val searchEpisodesResult = serverSearchResults.episodeSearch.episodes
 
-            if (serverSearchResults.searchTerm.isEmpty() || searchPodcastsResult.isNotEmpty() || serverSearchResults.error != null) {
+            if (serverSearchResults.searchTerm.isEmpty() || (searchPodcastsResult.isNotEmpty() || searchEpisodesResult.isNotEmpty()) || serverSearchResults.error != null) {
```
mchowning, line 172: Do you think it would be worth extracting out at variable with some explanation to make this easier to follow?

I came up with this based on my (possibly incorrect) understanding of what we're doing here, but I'm sure it could be improved.

``` kotlin
val hasResults =
    serverSearchResults.searchTerm.isEmpty() // if the search term is empty, we don't have "no results"
        || (searchPodcastsResult.isNotEmpty() || searchEpisodesResult.isNotEmpty()) // check if there are any results
        || serverSearchResults.error != null // an error is a result
```

 Feel free to leave as-is if you prefer.

mchowning (approved):
Looks good, just left a few minor suggestions, but no blockers. Nice work! 👍

adamzelinski:
Looks great! thanks for all your hard work on this @ashiagr no more feedback from me :)

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchHandler.kt
ashiagr: I verified that those were not needed so removed them in 13c6d2f.

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchHandler.kt
ashiagr: You're right, I verified it was not needed so removed it in 13c6d2f.

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchHandler.kt
ashiagr: Your comments make sense to me so I added them in 3ef4879.

ashiagr:
I've addressed your comments,@mchowning. Thank you for your extra eyes on RxJava operators and for help in removing unnecessary ones!

PR #817, opened 2023-03-07
Improve the Automotive skip settings input

geekygecko:
## Description
When entering skip forward and backward times in the Automotive settings area the enter / submit button adds new lines to the skip value, this change fixes that.

## Testing Instructions
1. Load the Automotive app into an Automotive emulator
2. Tap on the settings cog
3. Tap Skip forward time
4. Change the time amount
5. ✅  Verify you can't tap the blue button tick button to add new lines to the amount

## Screenshots

| Before | After |
| --- | --- |
| Screenshot_20230307_155241 | Screenshot_20230307_155454 |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 7.35
 -----
-
+*   Updates
+    *   The Automotive skip forward and backward time settings were improved.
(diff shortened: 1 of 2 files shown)


ashiagr (approved):
Looks good!

PR #821, opened 2023-03-08
Add episode screen and associated functionality to watch app

mchowning:
## Description
Adds the episode screen to the watch app along with the associated functionality.

The `EpisodeViewModel` implementation very closely follows the logic in the `EpisodeFragment` and `EpisodeFragmentViewModel`.

## Testing Instructions

Verify that you can access the episode screen in each of the following ways:

* Tapping on the episode/podcast name in the Now Playing screen
* Tapping on an episode in the Up Next queue
* Tapping on an episode on the podcast screen

From the episode screen, you should see the following UI:

* The Top of the screen has the podcast name, followed by the podcast title
* The download, play, and, queue buttons are tinted based on the podcast
* Below the play button is the episode date, amount of time left, and download size of the episode
* Below that is a list with the options: archive, star, mark as played, and go to podcast

From the episode screen, check the following functionality:

* Tapping the download button should have the following actions (with the button UI updating appropriately for each state):
  * if not downloaded, start a download
  * if a download is in progress, cancel the download
  * if the episode is downloaded, present a confirmation dialog for deleting the download. Confirming on that dialog should delete the download and briefly present a notification screen notifying the user that the download was removed (tapping on the notification screen will dismiss the notification even earlier than that 2 second timeout).
* Tapping the play button should start playing the episode (or pause it if it is already playing. The UI should reflect whether the button is a play or a pause button.
* Tapping the add-to-queue button should:
  * If the episode is in the Up Next queue, remove the episode from the queue and briefly present a notification dialog advising the user that the episode was removed from the queue.
  * If the episode is not in the Up Next queue, and...
    * If the Up Next queue is empty, add the episode to the Up Next queue
    * If the Up Next queue is not empty, present a screen asking where to add the episode to the queue.
 * Tapping the archive/unarchive, star/unstar,  mark as played/unplayed buttons should update the episode state appropriately, and the button's UI should update to reflect the new state.
 * Tapping the go-to-podcast button should take you to the podcast screen

## Screenshots or Screencast 

image

## Checklist
 

wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/episode/EpisodeScreen.kt
```diff
@@ -0,0 +1,222 @@
...
+                textAlign = TextAlign.Center,
+                modifier = Modifier.padding(horizontal = 8.dp)
+            )
+        }
+
+        item {
+            TextP50(
+                text = podcast.title,
+                maxLines = 1,
+                color = Color(0xFFDADCE0),
```
mchowning, line 73: I think I need to create a watch theme [similar to what we have for the automotive app](https://github.com/Automattic/pocket-casts-android/blob/d1502d9bf89fec3e19098f3ec78bc41e1473b447/modules/features/cartheme/src/main/res/values/colors.xml) (but with compose in mind) to handle these colors. I'll do that in a separate PR though.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/download/DownloadManagerImpl.kt
```diff
@@ -470,6 +484,10 @@ class DownloadManagerImpl @Inject constructor(
     }
 
     private fun updateNotification() {
+
+        // Don't show these notifications on wear os
+        if (Util.isWearOs(context)) return
```
mchowning, line 489: I don't have a good feel for how to use notifications on the watch yet. It feels like we shouldn't use them as much as we do in the app, so I've disabled this one for now (also, it's not working on the watch currently), but perhaps we'll want to add it back later. Let me know if you have any thoughts here.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/download/task/DownloadEpisodeTask.kt
```diff
@@ -141,7 +142,9 @@ class DownloadEpisodeTask @AssistedInject constructor(
             }
 
             LogBuffer.i(LogBuffer.TAG_BACKGROUND_TASKS, "Worker Downloading episode ${episode.title} ${episode.uuid}")
-            setForegroundAsync(createForegroundInfo())
+            if (!Util.isWearOs(context)) {
+                setForegroundAsync(createForegroundInfo())
+            }
```
mchowning, line 147: This is throwing an error on the watch because it triest to create a foreground notification, and notifications aren't working currently.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/download/DownloadManagerImpl.kt
ashiagr: Looks like we're not setting up notification channels for the wear os app inside `PocketCastsWearApplication` -> `setupApp()` : `notificationHelper.setupNotificationChannels()`

I can see notification by adding it:

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/download/task/DownloadEpisodeTask.kt
ashiagr: It should be resolved by setting up notification channels, haven't tested it though.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/download/DownloadManagerImpl.kt
```diff
@@ -134,7 +135,15 @@ class DownloadManagerImpl @Inject constructor(
...
-                                    getRequirementsAndSetStatusAsync(episode)
+
+                                    // FIXME this is a hack to avoid an issue where this listener says downloads
+                                    //  on the watch app are enqueued when they are actually still running.
+                                    val queriedState = workManager.getWorkInfoById(workInfo.id).get().state
+                                    if (Util.isWearOs(context) && queriedState == WorkInfo.State.RUNNING) {
+                                        getRequirementsAsync(episode)
+                                    } else {
+                                        getRequirementsAndSetStatusAsync(episode)
+                                    }
```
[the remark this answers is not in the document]
ashiagr: I'm trying to understand this part. Maybe @geekygecko can explain it.

ashiagr (approved):
Great job, @mattchowning! 🥇 

I haven’t matched UI with Figma as it is still evolving. I don't have much feedback at this point, just that app crashes on playing an episode after it is marked played:

```
I/Timber$Forest: Playback: Play now: 4b573437-5881-44b1-813c-3dbe5c8abf7c 194: The Top 6 Ways To Come Up With Your Business Idea
E/AndroidRuntime: FATAL EXCEPTION: main
    Process: au.com.shiftyjelly.pocketcasts.debug, PID: 19340
    java.lang.IllegalStateException: Cannot access database on the main thread since it may potentially lock the UI for a long period of time.
        at androidx.room.RoomDatabase.assertNotMainThread(RoomDatabase.kt:435)
        at androidx.room.SharedSQLiteStatement.assertNotMainThread(SharedSQLiteStatement.kt:52)
        at androidx.room.SharedSQLiteStatement.acquire(SharedSQLiteStatement.kt:74)
        at au.com.shiftyjelly.pocketcasts.models.db.dao.EpisodeDao_Impl.unarchive(EpisodeDao_Impl.java:1169)
        at au.com.shiftyjelly.pocketcasts.repositories.podcast.EpisodeManagerImpl.unarchive(EpisodeManagerImpl.kt:720)
        at au.com.shiftyjelly.pocketcasts.repositories.playback.PlaybackManager.playNowSync(PlaybackManager.kt:373)
        at au.com.shiftyjelly.pocketcasts.repositories.playback.PlaybackManager.playNowSync$default(PlaybackManager.kt:369)
        at au.com.shiftyjelly.pocketcasts.wear.ui.episode.EpisodeViewModel$play$1.invokeSuspend(EpisodeViewModel.kt:197)
```

https://user-images.githubusercontent.com/1405144/223970954-05168efa-8c1e-4df0-97eb-56ef1876db20.mp4

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/download/DownloadManagerImpl.kt
mchowning: 👍 Thanks for that. I'll make this update in a separate PR since this one is already way too big.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/download/task/DownloadEpisodeTask.kt
mchowning: I bet you're right. I'll enable notifications in a separate PR.

mchowning:
> I don't have much feedback at this point, just that app crashes on playing an episode after it is marked played

Well that's because you shouldn't be doing that!!! 😆 

Seriously though, great catch. I fixed it in  [81d847d](https://github.com/Automattic/pocket-casts-android/pull/821/commits/81d847d3425c310d10a19260e30160fe612b3044).

Issue #816, opened 2023-03-07
Automotive: episode duration isn't updated


geekygecko:
### Description

When playing an episode without a duration in the feed XML the seek bar isn't displayed. In the mobile app the media player reports the episode duration length once it has opened the file. The duration from the media player should be updated in the media session `MediaMetadata` but currently it doesn't. 

This is where the duration is set at the start of playback.
https://github.com/Automattic/pocket-casts-android/blob/7afa3d70e9a4bc2370594bcbc09d2c0c7f3dcefc/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/MediaSessionManager.kt#L321

### Step-by-step reproduction instructions

1. Load the Automotive app into an Automotive emulator
2. Search for the podcast 'Global News Podcast' (has an episode durations in the feed XML)
3. Tap play on an episode
4. Notice in the full screen player there is a seek bar
5. Search for the podcast 'Ted Talk Daily' (doesn't have episode durations in the feed XML)
6. FInd an episode that has a dash '-' instead of the duration and tap play on the episode
7. Notice in the full screen player there isn't a seek bar

### Screenshots or screen recording

![Screenshot_20230307_144230](https://user-images.githubusercontent.com/308331/223318953-94d6ae99-de09-408e-83a2-444aea061852.png)

![Screenshot_20230307_144212](https://user-images.githubusercontent.com/308331/223318955-451489f4-aebf-4701-a174-725b28777443.png)

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

Android Automotive

PR #822, opened 2023-03-08
Update the playing episode in the media session if the duration changes.

geekygecko:
## Description
This change fixes the issue when playing an episode without a duration in the feed XML, the seek bar isn't being displayed when using the data from the Media Session. The bug was found in Android Automotive but it's repeatable on mobile too with the notification drawer. The media player reports the episode duration length once it has opened the file and updates the database. This change is now picked up and the media session duration is updated which causes the seek bar to appear.

Fixes https://github.com/Automattic/pocket-casts-android/issues/816

## Testing Instructions

1. Open the mobile app
2. Search for the podcast 'Ted Talk Daily' 
3. Find an episode that has a dash '-' instead of the duration and tap play on the episode
4. ✅ Quickly open the notification drawer. Initially the seek bar won't appear but will once playback starts it will appear.

## Screenshots
Screenshot 2023-03-08 at 3 01 20 pm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
         ([#817](https://github.com/Automattic/pocket-casts-android/pull/817)).
     *   Fixed Automotive Up Next podcast images not loading.
         ([#819](https://github.com/Automattic/pocket-casts-android/pull/819)).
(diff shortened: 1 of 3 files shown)


mchowning (approved):
Works great! 👍

PR #824, opened 2023-03-09
Load carousel sponsored podcasts

ashiagr:
## Description
This PR prepares the Discover Carousel to be able to show sponsored podcasts.

iOS PR: https://github.com/Automattic/pocket-casts-ios/pull/755

**Tracks will be tackled in a separate task.**

## Testing Instructions
1. Switch to `debug` build variant
2. ✅ Make sure that `Discover` tab -> `Featured` carousel appears just fine
3. Go to `ListWebService.kt` and change line 12 to `@GET("/discover/{platform}/content-carousel.json")` (so it will use the mock)
4. Run the app
5. ✅ The carousel should appear. All the items should display the "Featured" label except the first and the fourth, which displays "Sponsored"

### DIFF
--- a/modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/view/DiscoverAdapter.kt
+++ b/modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/view/DiscoverAdapter.kt
@@ -68,11 +68,13 @@ import io.reactivex.android.schedulers.AndroidSchedulers
 import io.reactivex.disposables.Disposable
 import io.reactivex.functions.BiFunction
 import io.reactivex.rxkotlin.subscribeBy
+import io.reactivex.rxkotlin.zipWith
 import io.reactivex.schedulers.Schedulers
 import kotlinx.coroutines.MainScope
 import kotlinx.coroutines.launch
 import timber.log.Timber
 import java.util.Locale
+import kotlin.math.min
 import au.com.shiftyjelly.pocketcasts.localization.R as LR
 import au.com.shiftyjelly.pocketcasts.ui.R as UR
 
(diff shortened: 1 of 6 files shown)


modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/model/DiscoverModel.kt
```diff
@@ -62,7 +62,8 @@ data class DiscoverRow(
     @field:Json(name = "uuid") override val listUuid: String?,
     @field:Json(name = "regions") val regions: List<String>,
     @field:Json(name = "sponsored") val sponsored: Boolean = false,
-    @field:Json(name = "curated") override val curated: Boolean = false
+    @field:Json(name = "curated") override val curated: Boolean = false,
+    @field:Json(name = "sponsored_podcasts") val sponsoredPodcasts: List<SponsoredPodcast> = emptyList(),
```
mchowning, line 66: I like how you added the comma after the last list item so that if/when we add another element the diff will only be one line instead of two. 👍

modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/view/DiscoverAdapter.kt
```diff
@@ -400,18 +405,26 @@ internal class DiscoverAdapter(
...
-                                .take(featuredLimit.toLong())
-                                .concatMap {
+                        .zipWith(loadCarouselSponsoredPodcastList(row.sponsoredPodcasts))
+                        .flatMap {
+                            val (featuredPodcastList, sponsoredPodcastList) = it
+                            val mutableList = featuredPodcastList.podcasts
+                                .take(featuredLimit)
+                                .toMutableList()
+                            sponsoredPodcastList.forEach { sponsoredPodcast ->
+                                mutableList.addSafely(sponsoredPodcast.podcast.copy(isSponsored = true), sponsoredPodcast.position)
```
mchowning, line 417: I wasn't expecting that we grow the size of the carousel with each sponsored podcast (I thought the carousel size would remain the same regardless of the number of sponsored podcasts). What you have also makes sense, I just wanted to confirm that it is the desired behavior.

modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/viewmodel/DiscoverViewModel.kt
```diff
@@ -111,6 +112,31 @@ class DiscoverViewModel @Inject constructor(
             }
     }
 
+    fun loadCarouselSponsoredPodcasts(
+        sponsoredPodcastList: List<SponsoredPodcast>
+    ): Flowable<List<CarouselSponsoredPodcast>> {
+        val sponsoredPodcastsSources = sponsoredPodcastList
+            .filter { it.source != null && it.position != null }
```
mchowning, line 119: If this is something that should never happen, would it be worth logging an error here if it ever does happen?

mchowning (approved):
This works great! I left a few small comments, but no blockers. Nice job!

CookieyedCodes:
I get you guys want to incentivise Plus & make the app viable but please be careful, maby have a policy that no more then 3 sponsored slots viewable on the whole discover page at one time -really 5 in total with the carousel in my opinion- because it can wear on people & make it really feel like they are being used, I prefer the curated stuff & I get you have to fund but try have a balenced page so people don't feel slighted
-and please never a banner ad they are the worst-

modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/view/DiscoverAdapter.kt
ashiagr: I'll keep the existing logic to be consistent with iOS.

modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/viewmodel/DiscoverViewModel.kt
ashiagr: Logged it in 93f0900

Issue #815, opened 2023-03-07
Automotive: Sign out and clear data


geekygecko:
### Description

The Automotive app is often a shared device so it would be good to offer the ability for the user to clear their data when they sign out. This will give users the ability to sign into their own Pocket Casts account and the data such as listening history won't be merged with the previous account.

### Step-by-step reproduction instructions

1. Load the Automotive app into an Automotive emulator
2. Tap the settings cog
3. Sign into the account
4. Go back to the settings back
5. Tap the "Signed in as" option
6. Tap "Sign out"
Here would be a good place to ask if the user wants to clear their data. Maybe it could be a yes or no option rather than always clearing their data.

### Screenshots

![Screenshot_20230307_103649](https://user-images.githubusercontent.com/308331/223285151-60a3819f-4809-4815-92e2-5c1254c02288.png)

![Screenshot_20230307_103657](https://user-images.githubusercontent.com/308331/223285159-e484d83a-5606-427a-a159-53f620195b92.png)

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

Automotive

mchowning:
> Tap "Sign out"
> Here would be a good place to ask if the user wants to clear their data. Maybe it could be a yes or no option rather than always clearing their data.

Do you think giving the pop-up three options instead of two would work @geekygecko ? Something like "Cancel", "Sign out", and "Sign out and clear data"

geekygecko:
As users pressing logout is rare and this is only on Automotive, your two alert boxes idea is fine. Thanks again for your help.

Issue #260, opened 2022-08-23
 Android Auto/Automotive Podcast + Filter Page Played Episode Issue


azollafox:
## What I expected

When an episode is played fully it will disappear from the Podcast or Filter page it's on.

## What happens
If you have the Podcast page opened which lists all of the episodes, the episode which you played fully does not automatically disappear. Going back to the homepage, then opening the podcast page again resolves it.

Reported in https://woothemes.zendesk.com/agent/tickets/5476176
Reproduced https://a8c.slack.com/archives/C02A333D8LQ/p1661253982975559

Debug logs from ticket:
[debug-2.txt](https://github.com/Automattic/pocket-casts-android-private/files/9402462/debug-2.txt)

thabotswana:
Another report in 5766666-zen

>I love using pocket casts both on my phone and my AAOS vehicle infotainment system, but the car never marks episodes complete after finishing them.
>
>I regularly need to go to my phone, select all the episodes I've listened to, mark them played, and then refresh both devices to make them disappear from new releases on my car.
>
>Complicating this, sometimes episodes get stuck in "new releases" in my car that are already marked played on the phone, therefore I have no way to make them disappear on the car. They will eventually go away after some days, but at times I have 10 or more played episodes stuck in the list on the car.

mchowning:
This is also an issue on Android Automotive.

1. Ensure that the app's settings are set to not "Show Played Podcasts"
2. Go to the discover screen
3. Open a podcast page
4. Tap on a podcast to start playing it
5. Complete the episode
6. Tap the back arrow to return to the podcast page
7. Observe that the just-completed podcast is still listed
8. Tap back to go to the Discover page
9. Tap on the podcast that you were just looking at
10. Observe that the just-played podcast is now removed from the list

https://user-images.githubusercontent.com/4656348/205943388-5ba88f71-0fd2-4b1c-ae4e-38d956fe7b25.mov

mchowning:
When [clearing the app data on sign out](https://github.com/Automattic/pocket-casts-android/pull/828), I also see this issue with the Podcasts tab (subscribed podcasts will still show until you switch to another tab and then back to the podcasts tab).

PR #828, opened 2023-03-14
Add option to clear data when logging out of Automotive app

mchowning:
## Description

Fixes https://github.com/Automattic/pocket-casts-android/issues/815 by giving users the option to clear their data when they log out on Android Automotive.

I went with the simplest UI approach here of just presenting a second pop-up when the user signs off, asking them if they want to clear their data. This could certainly be done better, but since this is addressing a bit of an edge case, I didn't want to spend too much time over-optimizing it. With that said, if you have a better idea for the UI, please let me know.

## Testing Instructions
1. Sign into a Plus account on Android Automotive. 
2. Make sure the account has:
    1. An episode that is currently playing
    3. Some episodes in the up next queue
    4. Some podcast subscriptions
    5. Some custom filters
    6. Some cloud files that have been uploaded under Profile → Files
    7. Some starred episodes under Profile → Starred
    8. Some listening history under Profile → Listening 
3. Sign out of the account and clear your data (⚙️  → User Profile View → Sign out → Sign out → Clear data)
4. Go through each of the section in Step 2 and verify that there is no data present. Note that because of https://github.com/Automattic/pocket-casts-android/issues/260 you may still see data on a screen until you reload the tab/screen (for example, the Podcasts tab may still show your subscribed podcasts until you switch away from and back to that tab).
5. Repeat Steps 1 & 2
6. Sign out of the account and **DO NOT* clear your data (⚙️  → User Profile View → Sign out → Sign out → Just sign out)
7. Verify that the data from Step 2 has **not** been removed.

## Screenshots or Screencast

https://user-images.githubusercontent.com/4656348/224863473-6341af73-0534-4ec6-b8fe-60255e02f6d7.mov

## Checklist
 
- [X] with different themes

### DIFF
--- a/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/AccountDetailsFragment.kt
+++ b/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/AccountDetailsFragment.kt
@@ -265,8 +265,8 @@ class AccountDetailsFragment : BaseFragment() {
         val themedContext = if (Util.isAutomotive(context)) ContextThemeWrapper(context, CR.style.Theme_Car_NoActionBar) else context
         val builder = AlertDialog.Builder(themedContext)
         builder.setTitle(getString(LR.string.profile_clear_data_question))
-            .setMessage("Would you also like to clear all data from the app?")
-            .setPositiveButton("Just sign out") { _, _ -> performSignOut() }
+            .setMessage(getString(LR.string.profile_clear_data_would_you_also_like_question))
+            .setPositiveButton(getString(LR.string.profile_just_sign_out)) { _, _ -> performSignOut() }
             .setNegativeButton(getString(LR.string.profile_clear_data)) { _, _ ->
                 signOutAndClearData()
             }
--- a/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/AccountDetailsFragment.kt
+++ b/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/AccountDetailsFragment.kt
@@ -279,7 +279,7 @@ class AccountDetailsFragment : BaseFragment() {
 
         // Block while clearing data so that we don't return to the app until the users data has been cleared
         runBlocking(Dispatchers.IO) {
-            upNextQueue.removeAll()
+            upNextQueue.removeAllIncludingChanges()
 
             playlistManager.resetDb()
             folderManager.deleteAll()
--- a/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/AccountDetailsFragment.kt
+++ b/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/AccountDetailsFragment.kt
@@ -290,7 +290,6 @@ class AccountDetailsFragment : BaseFragment() {
             userEpisodeManager.findUserEpisodes().forEach {
                 userEpisodeManager.delete(it, playbackManager)
             }
-            episodeManager.deleteDownloadedEpisodeFiles()
             episodeManager.deleteAll()
         }
 
--- a/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/AccountDetailsFragment.kt
+++ b/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/AccountDetailsFragment.kt
@@ -11,7 +11,6 @@ import android.widget.Toast
 import androidx.appcompat.app.AlertDialog
 import androidx.appcompat.app.AppCompatActivity
 import androidx.compose.ui.platform.ViewCompositionStrategy
-import androidx.core.content.ContentProviderCompat.requireContext
 import androidx.core.view.isVisible
 import androidx.fragment.app.viewModels
 import au.com.shiftyjelly.pocketcasts.account.ChangeEmailFragment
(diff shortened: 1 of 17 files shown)


modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/AccountDetailsFragment.kt
```diff
@@ -242,11 +255,48 @@ class AccountDetailsFragment : BaseFragment() {
...
+            }
+            .show()
+    }
+
+    private fun signOutAndClearData() {
+        // Sign out first to make sure no data changes get synced
+        userManager.signOut(playbackManager, wasInitiatedByUser = true)
+
+        // Block while clearing data so that we don't return to the app until the users data has been cleared
+        runBlocking(Dispatchers.IO) {
```
mchowning, line 281: I'm using `runBlocking` here because I'm concerned about the problems that might occur if the user starts interacting with the app when we're in the middle of deleting all the data.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/UpNextQueueImpl.kt
```diff
@@ -245,15 +245,6 @@ class UpNextQueueImpl @Inject constructor(
         saveChanges(UpNextAction.ClearAll)
     }
 
-    /**
-     * Removes all episodes in the list
-     */
-    override suspend fun removeAllEpisodes(episodes: List<Playable>) = withContext(coroutineContext) {
```
mchowning, line 251: Just removed this because it wasn't being used anywhere.

modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/AccountDetailsFragment.kt
```diff
@@ -242,11 +255,48 @@ class AccountDetailsFragment : BaseFragment() {
...
+            searchHistoryManager.clearAll()
+
+            podcastManager.deleteAllPodcasts()
+
+            userEpisodeManager.findUserEpisodes().forEach {
+                userEpisodeManager.delete(it, playbackManager)
+            }
+            episodeManager.deleteDownloadedEpisodeFiles()
+            episodeManager.deleteAll()
+        }
```
mchowning, line 295: I could have also called `SearchHistoryManagerImpl::clearAll` here, but it seemed unnecessary because I don't think we are using search history in Android Automotive. Let me know if I'm wrong about that or if you think we should clear the search history just to have everything covered (I waffled back and forth on whether to do that).

mchowning: Another thing I'm not clearing is the `up_next_changes` table because I assume we would still want to sync any pending Up Next changes back to the account, even if the user has signed out.

modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/AccountDetailsFragment.kt
```diff
@@ -242,11 +255,48 @@ class AccountDetailsFragment : BaseFragment() {
...
             .show()
     }
 
+    private fun clearDataAlert() {
+        val context = context ?: return
+        val themedContext = if (Util.isAutomotive(context)) ContextThemeWrapper(context, CR.style.Theme_Car_NoActionBar) else context
+        val builder = AlertDialog.Builder(themedContext)
+        builder.setTitle(getString(LR.string.profile_clear_data_question))
+            .setMessage("Would you also like to clear all data from the app?")
+            .setPositiveButton("Just sign out") { _, _ -> performSignOut() }
```
geekygecko, line 269: Can we localize these strings too?

modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/AccountDetailsFragment.kt
geekygecko: > I could have also called SearchHistoryManagerImpl::clearAll here, but it seemed unnecessary

I don't believe the Automotive UI supports this so it's fine to leave it out.

> I'm not clearing is the up_next_changes

I think we should clear the Up Next changes as the user can't call the server once we have removed their account from the Account Manager.

geekygecko (approved):
Thanks for making this change! 

I did get this exception in the logs during the logout but the logout was successful and my data was cleared.
```
java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.String.equals(java.lang.Object)' on a null object reference
	at au.com.shiftyjelly.pocketcasts.repositories.file.FileStorage.getBaseStorageDirectory(FileStorage.java:144)
	at au.com.shiftyjelly.pocketcasts.repositories.file.FileStorage.getBaseStorageDirectory(FileStorage.java:140)
	at au.com.shiftyjelly.pocketcasts.repositories.file.FileStorage.getStorageDirectory(FileStorage.java:132)
	at au.com.shiftyjelly.pocketcasts.repositories.file.FileStorage.getOrCreateDirectory(FileStorage.java:121)
	at au.com.shiftyjelly.pocketcasts.repositories.file.FileStorage.getPodcastDirectory(FileStorage.java:98)
	at au.com.shiftyjelly.pocketcasts.repositories.podcast.EpisodeManagerImpl.deleteDownloadedEpisodeFiles(EpisodeManagerImpl.kt:807)
	at au.com.shiftyjelly.pocketcasts.profile.AccountDetailsFragment$signOutAndClearData$1.invokeSuspend(AccountDetailsFragment.kt:293)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
```

modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/AccountDetailsFragment.kt
mchowning: 🤦 Yes, thanks for catching that. Done in [062b5ba](https://github.com/Automattic/pocket-casts-android/pull/828/commits/062b5ba2889386aead5afbdf9db99a20e7b30be4).

modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/AccountDetailsFragment.kt
mchowning: > I think we should clear the Up Next changes as the user can't call the server once we have removed their account from the Account Manager.

Makes sense. Did that update in [8cc15b3](https://github.com/Automattic/pocket-casts-android/pull/828/commits/8cc15b3ac4a5b1cd577a1a7364fe650a73b98bc0).

This is making me wonder, do you think we should be forcing a sync before logging out?

mchowning:
> I did get this exception in the logs during the logout but the logout was successful and my data was cleared.

This was happening because I was making the call to delete downloaded files, but I guess that doesn't really make sense due to the fact that we cannot download files in the Automotive app (is that right?). So I went ahead and removed the call to delete downloaded episode files in [c3abc9c](https://github.com/Automattic/pocket-casts-android/pull/828/commits/c3abc9c9c86b8f97e4515c7b003f69befd733010).

modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/AccountDetailsFragment.kt
geekygecko: I reckon in the first version we shouldn't worry, it will add another level of complexity. Thanks for all these changes, I'm going to merge it in.

PR #830, opened 2023-03-15
Search minor updates

ashiagr:
## Description

1. Reduces search history max count to match with iOS (p1678446938002359/1678387434.322809-slack-C03JCLLR9C3)
2. Adds `search_list_shown` track (https://github.com/Automattic/pocket-casts-ios/pull/774#issuecomment-1468243355).

> **Warning**
> Targets release/7.34

#### Change 1
It doesn't require any manual testing, it is covered by `testInsertTooManyItemsKeepsMostRecent` test.

#### Change 2

1. Tap on the search bar in `Discover` or `Podcasts`
2. Search for something
3. Tap "Show all" on Podcasts
4. ✅ search_list_shown should be tracked with correct source (discover or podcast_list) and displaying: podcasts
5. Go back, tap "Show all" for Episodes
6. ✅ search_list_shown should be tracked with correct source (discover or podcast_list) and displaying: episodes


mchowning (approved):
Looks good! 👍

PR #837, opened 2023-03-17
Remove UpNextEpisodeViewModel

mchowning:
The `UpNextEpisodeViewModel` is not being used anywhere, so this PR removes it. The commit `a17164d4bd9300ae93759d45eaca04f54f8e13b4`, which was from before the app was open-sourced, removed the fragment that utilized this view model.

## Testing Instructions
Just make sure you can open the up next queue in the app.

## Checklist


ashiagr (approved):
Good job getting rid of it. 👍
I've verified it isn't used anywhere and also checked by loading the up next queue.

PR #838, opened 2023-03-17
Update lifecycle and core dependencies

mchowning:
> **Note**
> This PR builds on the `update/remove-unused-upnextepisodeviewmodel` branch, so let's merge [that PR](https://github.com/Automattic/pocket-casts-android/pull/837) before this one.

## Description
When updating to the latest version of horologist for the watch app, the `core-ktx` and the `lifecycle` libraries get updated as well. The `core-ktx` update doesn't require any changes, but the `lifecycle` update required a lot of changes, so this is to separate those changes out into its own PR. I'm targeting `main` with this PR because I'm planning to merge `add/wear` into `main` soon.

Most of the changes are pretty safe and automatic. The only change that I think could go wrong is where I replace the `launchWhenX` events. I followed the same pattern [Google used when updating the data store library](https://android-review.googlesource.com/c/platform/frameworks/support/+/2293558/5/datastore/datastore-sampleapp/src/main/java/com/example/datastoresampleapp/SettingsFragment.kt), but it would be good to take a closer look at those changes to make sure I'm not using an incorrect lifecycle owner or something.

Here are the release notes for [lifecycle](https://developer.android.com/jetpack/androidx/releases/lifecycle) and [core](https://developer.android.com/jetpack/androidx/releases/core#1.9.0) if you want to take a look.

## Testing Instructions
Just some general smoke testing.

## Checklist

### DIFF
--- a/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
+++ b/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
@@ -26,10 +26,12 @@ import androidx.core.view.isVisible
 import androidx.core.view.updatePadding
 import androidx.fragment.app.Fragment
 import androidx.fragment.app.commitNow
+import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.LiveData
-import androidx.lifecycle.LiveDataReactiveStreams
 import androidx.lifecycle.Observer
 import androidx.lifecycle.lifecycleScope
+import androidx.lifecycle.repeatOnLifecycle
+import androidx.lifecycle.toLiveData
 import androidx.mediarouter.media.MediaControlIntent
 import androidx.mediarouter.media.MediaRouteSelector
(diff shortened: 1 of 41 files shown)


ashiagr (approved):
I verified that changes matched source conversions for the new version of lifecycle according to the [linked](https://developer.android.com/jetpack/androidx/releases/lifecycle#2.6.0) table.

Also, changes to replace `Lifecycle.launchWhenX` with `Lifecycle.repeatOnLifecycle` look good and follow the recommended approach. 

I had been wanting `collectAsStateWithLifecycle()` to get Compose State in a lifecycle-aware manner which I can have now 🎉  
> Added a new extension on StateFlow and Flow of [collectAsStateWithLifecycle()](https://medium.com/androiddevelopers/consuming-flows-safely-in-jetpack-compose-cde014d0d5a3) that collect from flows and represents its latest value as Compose State in a lifecycle-aware manner.


As discussed, I'll hold off merging it to the main until v7.36 is cut out to give us more time time to test.

PR #842, opened 2023-03-25
add check for user preferences if user prefers unmetered connection before updating 

### LINKED ISSUE #839: Bug: background update ignores download preferences when upnext is set to queue new episodes
### Description

https://github.com/Automattic/pocket-casts-android/blob/b1946883673f38d38d8590fecb5d15d4312d069a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/refresh/RefreshPodcastsTask.kt#L64

In the line above, the constraints specify the background task should be run when `NetworkType.CONNECTED` but it should probably be conditional based on the user preferences for preferring downloads on an UNMETERED connection. The setting should probably not be statically set to `NetworkType.CONNECTED` but should be conditional and alternatively set to `NetworkType.UNMETERED`.

I suspect what is happening is that the task runs every hour in the background and queues the podcast into the upnext queue. The download may/may-not be deferred until the player requests it to be played....which might be what eventually triggers the download.

Having the task mirror the download preferences would prevent the scheduling from triggering the update and enqueing any episodes in the up next queue.

I noticed a high foreground usage of data with daily pocketcasts which led me to look into this.

#### Settings

- Auto Download 
    - only on unmetered wifi (on)
    - episodes added to up next (on)  

### Expected Behavior 

Downloads will be delayed until connected to an unmetered connection and episodes won't be queued into the upnext queue until attached to an unmetered connection.

### Actual Behavior 

Downloads are happening on metered wifi

### Version Observed on

7.33 (9077)

gyund:
## Description

When a user chooses the option to download "only on unmetered wifi", they essentially want to minimize mobile data usage. There's also a setting under "storage & data use" -> "Warn before using data". This warning is designed to warn the user before downloading or streaming episodes.

Currently the task that runs in the background will update podcasts even while on a metered network, which can trigger podcasts to get automatically added to the up next queue _(such as if the setting "auto download episodes added to up next" combined with the "auto add to up next" for specific podcasts)_. This will indirectly cause podcasts to be downloaded on a metered network when added to the up next queue through these settings because a podcast in the up next queue will eventually download when the podcast is played. This is mostly noticable for podcasts that automatically enqueue in the up next queue at the top instead of the bottom.

The fix for this issue hinges on keying off the option to "Warn before using data". This is co-located with the "Background Refresh" option. If the user wants to be warned about data being used, creating conditions that will create data to be used by updating in the background regardless of this setting seems to contradict the user preference. RefreshPodcastTask has been updated to use the setting for warnings (`settings.warnOnMeteredNetwork()`) to update the constraints on the work manager so that it only runs background updates while on an unmetered network. Users can disable this option if they want background updates to use data to pull updates. Since there's already a setting to control downloads "only on unmetered wifi", it has no effect for users that wish to update in the background and download on or off metered networks.

By preventing the podcast from automatically updating in the background on a metered network based on the user's preference to be warned, we help avoid setting the user up for needing a warning by not automatically getting new information about the podcasts that would require a warning in the first place.

This doesn't cover the edge case of updating while leaving an unmetered network, but I think I can live with that 😄 

Fixes #839 

## Testing Instructions

1. Load the app in the debugger
2. Set a breakpoint in the refresh podcast task [here](https://github.com/Automattic/pocket-casts-android/blob/a8bf74fd4924160c9fd6c6cf24a6608123b54c6b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/refresh/RefreshPodcastsTask.kt#L85). Also set a breakpoint [here ](https://github.com/Automattic/pocket-casts-android/blob/a8bf74fd4924160c9fd6c6cf24a6608123b54c6b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/refresh/RefreshPodcastsTask.kt#L89) for the case when we disable the warning.
3. Go to "Settings - > Storage & Data Use"
4. Enable the option "Warn before using data"
5. Observe the breakpoint on line 85 being hit as well as line 89
6. Disable the option "warn before using data"
7. Observe only the breakpoint on line 89 being hit

## Screenshots or Screencast 

228794549-3d3e542b-c28e-49a3-9387-418b839c4434

## Checklist


gyund:
I created two separate commits, [7c9044c](https://github.com/Automattic/pocket-casts-android/pull/842/commits/7c9044c2f656c7b2c2ccbc1ba6699b2f949fccb7) would ensure that when you have the box selecting only on unmetered wifi, you can't possibly refresh or update the feeds to trigger a download. This change reduces data the most.

[561394](https://github.com/Automattic/pocket-casts-android/pull/842/commits/1561394607c7113a35c08112475a5145a48708e4) only affects the scheduling of workmanager but would still allow the user to trigger activity that could cause downloads. It just wouldn't happen on a schedule in the background because work manager wouldn't schedule the task to run when they are on a metered network.

This should make it easy to revert [7c9044c](https://github.com/Automattic/pocket-casts-android/pull/842/commits/7c9044c2f656c7b2c2ccbc1ba6699b2f949fccb7) if it is undesirable.

mchowning:
👋  @gyund ! Thanks so much for reporting this issue and submitting a fix. Really appreciate your explanations for the changes you made and how you broke the changes out into separate, logical commits. 👍 

This change does more than just prevent downloads on metered connections because it also prevents refreshing the podcasts on metered connections, right? I don't think users who update their settings to not automatically download podcasts on metered connections would expect (or necessarily want) to stop getting updates for their podcasts while using a metered connection (for example, new episode notifications). Let me know if I'm just misunderstanding something about the change though.

It does look like our `DownloadManager` [should be respecting the download-on-unmetered setting](https://github.com/Automattic/pocket-casts-android/blob/c965015274d4116583e626ca8b28d512aadbf1e4/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/download/DownloadManagerImpl.kt#L466) and not downloading episodes when on a metered connection if that setting is turned on.

Unrelated minor issue: if you run the `spotlessApply` gradle task, it will fix any formatting changes that are needed for the build pass on CI.

gyund:
@mchowning Thanks, yes, you are correct in your understanding :). I'll revert [7c9044c](https://github.com/Automattic/pocket-casts-android/pull/842/commits/7c9044c2f656c7b2c2ccbc1ba6699b2f949fccb7) and thanks for letting me know about the gradle task. I'll give that a run while I'm there.

gyund:
I pushed the changes. The remaining changes will take care of making sure work manager doesn't schedule refreshing podcasts on a metered network but it will allow the user to manually refresh if they so choose.

gyund:
I pushed the changes. The remaining changes will take care of making sure work manager doesn't schedule refreshing podcasts on a metered network based on the user preferences, but it will allow the user to open the app to trigger a refresh if they so choose. By not refreshing in the background, we can avoid enqueuing podcasts in the top of the up next queue on a metered network and trigger streaming (a download) to occur for those daily podcasts. 

For the notification side effect, I think it if I understand what you're ensuring to consider, it probably depends on their desire to:

- automate playlists
- have control over addition to their lists and podcast orders

My personal thought on the "only on unmetered wifi" setting is a bit more black/white. If I don't want to automatically download/stream it on a metered network, then acquiring any information about it while on a metered network is not necessary. Finding about it later would be fine, like when unwinding after work.

For someone who does want to be notified automatically on any network ASAP while the app is in the background, then I think it might be cleaner for them to disable "only on unmetered wifi" and automatic downloads and instead enable notifications to do things manually when they're notified using the customization options in the notification menu to play next, last, or now, etc. 

I'm hesitant to suggest adding another setting for "update only on wifi" because the current setting gave me this impression (though the wording isn't technically wrong😄). You guys do a good job of keeping things clean, and I have a similar mindset about UI design and avoiding overly complex settings😄. That's why I'm inclined to favor any mention of "only on wifi" to really only do things automatically while on wifi and only have one setting that controls that.

CookieyedCodes:
May I put forward a suggestion because I personally would prefer control then black/white just because some users might be confused with the change (bus wifi for instance), have an option toggle that appears only if unmetered is allready checked on called "including delaying upnext updates" that feels intuitive & won't upset users who don't understand their circumstance

CookieyedCodes:
> May I put forward a suggestion because I personally would prefer control then black/white just because some users might be confused with the change (bus wifi for instance or work wifi restrictions),
It should be their mind you because some people I have read want this but it shouldn't be black & white in my option,
 have an option toggle that appears only if unmetered is allready checked on called "including delaying upnext updates" that feels intuitive & won't upset users who don't  understand the change in setting 😉

gyund:
> > May I put forward a suggestion because I personally would prefer control then black/white just because some users might be confused with the change (bus wifi for instance or work wifi restrictions),
> > It should be their mind you because some people I have read want this but it shouldn't be black & white in my option,
> > have an option toggle that appears only if unmetered is allready checked on called "including delaying upnext updates" that feels intuitive & won't upset users who don't  understand the change in setting 😉

Hmm, like perhaps paired with the background updates setting? 
Screenshot_20230330-053403

Would it make sense to replace or leverage the second option instead? For example, instead of warn, just don't do it on wifi. Or have a completely different option?

CookieyedCodes:
Hmmmmm like that makes sence to an extent tho it would confuse users who want to keep podcasts comeing in even on meetered wifi, (might be a good ui update tho that toggle turns off & goes gray -background refresh- if both "warn before useing data" & "include delaying of upnext updates" like the most important thing is users & PC don't want to block sync tho that is another consideration 🤔



Tbh I think an easier enhancement for this is to also have download upnext limits -custom number tho- like the apple watch, it should consider how many downloads are allready in upnext & should never delete exsisting downloads, this should allow the focus to be on what's been added & allow me to deal with my download bloat 😅 🤔, either that &/or my other suggestion of the ability to change the add to upnext position on the fly to after a spefic podcast (on swipe right to left) this position would follow the podcast until it is played (because then it's at the top again) or until the line is changed the UI would be shown by a line on the upnext feed

gyund:
> Hmmmmm like that makes sence to an extent tho it would confuse users who want to keep podcasts comeing in even on meetered wifi, (might be a good ui update tho that toggle turns off & goes gray -background refresh- if both "warn before useing data" & "include delaying of upnext updates") like the most important thing is users & PC don't want to block sync tho that is another consideration 🤔
> 
> Tbh I think an easier enhancement for this is to also have download upnext limits -custom number tho- like the apple watch, it should consider how many downloads are allready in upnext & should never delete exsisting downloads, this should allow the focus to be on what's been added & allow me to deal with my download bloat 😅 🤔,
> 
> either that &/or my other suggestion of the ability to change the add to upnext position on the fly to after a spefic podcast (on swipe right to left) this position would follow the podcast until it is played (because then it's at the top again) or until the line is changed the UI would be shown by a line on the upnext feed

So I think for users that want to keep podcasts coming even on metered networks, they probably wouldn't have "Warn before using data" checked. Download manager [appears to be](https://github.com/Automattic/pocket-casts-android/blob/c965015274d4116583e626ca8b28d512aadbf1e4/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/download/DownloadManagerImpl.kt#L452) not downloading podcasts immediately already when this setting is checked, though it may have to anyway when the user has settings to increase volume or remove silence, etc. I'm not sure if in the current state it "Warns" but then streams anyway or just stops? It seems a bit silly to warn a user you're about to do something they might not want you to do and then go do it anyway.

I wanted to see how much work it would be to use this alternate approach and it's not much different. Updates have to occur in a few different files. I have a fix locally I've tested this approach that keys off this warning option instead. @mchowning if you think this is a more appropriate spot trigger off, let me know and I'll push it. Otherwise, I'll hold off just to avoid cluttering the change log too much.

I'm not sure if the setting requires a description update and/or how your team typically deals with updating translations.

gyund:
@CookieyedCodes just to clarify, by having work manager key off the warning, it wouldn't prevent you from "keeping podcasts coming". It just wouldn't update in the background. I feel like the co-location of the setting to "background refresh" has an implied relationship. Any episodes you did already know about wouldn't be affected by the work manager change. Work manager is currently scheduled to run once every hour regardless of your connection. This would make it so downloading background updates while on metered would be deferred until you were back on an unmetered connection.

gyund:
> 🤔🤔, I'll leave it to Matt, just was giveing my two cents :)

😆 haha, yah same. I appreciate it though to make sure nothing's being overlooked. UIs are always fun like this 😄

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/refresh/RefreshPodcastsTask.kt
```diff
@@ -59,21 +60,36 @@ class RefreshPodcastsTask @AssistedInject constructor(
...
-            val request = PeriodicWorkRequestBuilder<RefreshPodcastsTask>(REFRESH_EVERY_HOURS, TimeUnit.HOURS)
+            return PeriodicWorkRequestBuilder<RefreshPodcastsTask>(REFRESH_EVERY_HOURS, TimeUnit.HOURS)
                 .addTag(TAG_REFRESH_TASK)
                 .setConstraints(constraints)
                 .setInitialDelay(REFRESH_EVERY_HOURS, TimeUnit.HOURS)
                 .build()
+        }
 
-            workManager.enqueueUniquePeriodicWork(TAG_REFRESH_TASK, ExistingPeriodicWorkPolicy.REPLACE, request)
+        private fun buildPeriodicWorkRequestConstraints(settings: Settings): Constraints {
```
gyund, line 80: Moved the constraint building out of the `scheduleOrCancel` method so it could be tested if desired.

gyund:
I pushed it in, seemed cleaner to just look at the change than describe it. Then at least you have it. I might get busy in the next few days.

gyund:
So I don't have to worry about merge conflicts while I look at #858, which is somewhat related to #839, going to close this PR to take it out of play.

PR #850, opened 2023-03-29
Enable carousel autoscroll

mchowning:
## Description
Turning on autoscroll on the Discover carousel by removing the feature flag.

internal ref: p1680117820760509/1680117264.777019-slack-C02ATC80MUM

> **Warning**
> This PR is targeting the 7.35 release branch. Even though we're close to releasing that, I'm thinking we may want to go ahead and cut a new beta so if there is any issue we have a chance to catch it before release, but let me know what you think.

## Testing Instructions
Review the auto scroll testing instructions from https://github.com/Automattic/pocket-casts-android/pull/818

## Screenshots or Screencast 

https://user-images.githubusercontent.com/4656348/228649029-094478c6-c4e8-4bff-8b9f-ac59eb28d163.mov

## Checklist
 
- [X] with different themes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,14 +5,16 @@
 7.35
 -----
 *   Updates
-    *   The Automotive skip forward and backward time settings were improved.
+    *   The Automotive skip forward and backward time settings were improved
         ([#817](https://github.com/Automattic/pocket-casts-android/pull/817)).
-    *   Fixed Automotive Up Next podcast images not loading.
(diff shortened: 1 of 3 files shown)


ashiagr (approved):

ashiagr:
>  I'm thinking we may want to go ahead and cut a new beta so if there is any issue we have a chance to catch it before release, but let me know what you think.

Sounds good. I included it in `7.35-rc-2` beta release.

PR #854, opened 2023-03-30
Implement watch downloads screen

mchowning:
## Description
This fills in the Downloads screen on the watch app.

In addition, this fixes an issue where after deleting a download, we would navigate back too far instead of just going back to the episode screen that initiated the deletion.

## Testing Instructions

In addition to the following testing steps, the UI should also be in good shape as well.

1. Without any downloads, go to the downloads screen in the watch app.
2. Observe that the screen presents a message indicating you have no downloads.
3. Log into an account and download some episodes (via the download button on the screen for a specific episode, can be accessed via `Podcasts` → Select a podcast → Select an episode → tap on the Download "down arrow" icon).
4. Return to the download screen and observe that the downloaded episodes are listed.
5. Tap on one of the downloaded episodes
6. Observe that you are taken to the episode screen for that episode
7. Delete the download from the episode screen by tapping the download checkmark to the left of the play button.
8. Swipe back to the download screen
9. Observe that the episode with the just-deleted download is no longer included in the download screen.

## Screenshots or Screencast 

https://user-images.githubusercontent.com/4656348/228893993-e85abe60-fbb0-497b-82c4-04632c9f66e5.mov

## Checklist
 
- [X] with different themes


wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/downloads/DownloadsScreen.kt
```diff
@@ -0,0 +1,200 @@
...
+}
+
+@Composable
+fun DownloadsScreen(
+    columnState: ScalingLazyColumnState,
+    onItemClick: (Episode) -> Unit,
+) {
+
+    val viewModel = hiltViewModel<DownloadsScreenViewModel>()
+    val state by viewModel.stateFlow.collectAsState()
```
ashiagr, line 61: Now that `collectAsStateWithLifecycle()` is available with the recent lifecycle version bump, wdyt about replacing `collectAsState()` with it?

wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/downloads/DownloadsScreen.kt
```diff
@@ -0,0 +1,200 @@
...
+                        TextStyle(
+                            platformStyle = PlatformTextStyle(
+                                // So we can align the top of the text as closely as possible to the image
+                                includeFontPadding = false,
+                            ),
+                        )
+                    ),
+                    maxLines = 2,
+                )
+                val shortDate = episode.publishedDate.toLocalizedFormatPattern("dd MMM")
```
ashiagr, line 152: Nitpick: I see that if the duration is less than an hour, mins are displayed slightly differently:

Screenshot 2023-03-31 at 9 08 26 AM

No need to fix it right now, just mentioning it as I noticed it in Figma.

ashiagr (approved):
Looks and works great! I just left minor comments. 

Feel free to merge it even if you decide not to make any changes.

wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/downloads/DownloadsScreen.kt
mchowning: Thanks. I'll confirm with Adam.

PR #855, opened 2023-03-30
Set up notification channels on watch

mchowning:
## Description
This sets up the notification channels on the watch so that notifications will now work. Thanks to @ashiagr for [suggesting this fix](https://github.com/Automattic/pocket-casts-android/pull/821#discussion_r1130613461).

## Testing Instructions
1. Trigger a notification from the watch app. I did it by using adding these changes so that tapping on the Now Playing chip would trigger a notification.
2. Observe that the notification is presented.

## Screenshots or Screencast 

[Uploading Screen Recording 2023-03-30 at 1.26.05 PM.mov…](https://user-images.githubusercontent.com/4656348/228917020-cf01b53e-2346-433e-ba16-dc9a006b980a.mov)

## Checklist


ashiagr (approved):
LGTM 👍

PR #856, opened 2023-03-30
Update scroll position indicator from blue to gray-white

mchowning:
## Description
Updates the scroll position indicator on the watch app from blue to gray. This change was suggested by @geekygecko , and I agree that the blue doesn't really seem to fit. This fix will obviously change more than just the position indicator, but I don't think we necessarily want the blue color anywhere in the watch app currently, so removing it in other places as well seems fine. I'm sure we'll be revisiting how to handle the theming in the watch app again before this ships.

This PR also extracts out some of the raw color declarations we had for wear to a single location (`WearColors`). I don't think moving these colors to a single file is a great solution, but I hope it's at least a (very) small step in the right direction and will, I hope, make it easier to further improve how we handle colors in the watch app in the future.

## Testing Instructions
Verify that when scrolling lists in the app, the scroll indicator is gray instead of blue.

## Screenshots or Screencast 

| Before | After |
| --- | --- |
| Screenshot 2023-03-30 at 2 47 23 PM | Screenshot 2023-03-30 at 2 50 50 PM |

## Checklist
 
- [X] with different themes

### DIFF
--- a/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/episode/ObtainConfirmationScreen.kt
+++ b/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/episode/ObtainConfirmationScreen.kt
@@ -57,7 +57,7 @@ fun ObtainConfirmationScreen(
                 Box(
                     contentAlignment = Alignment.Center,
                     modifier = Modifier
-                        .background(color = WearColors.FF202124) // FIXME
+                        .background(color = WearColors.FF202124)
                         .clip(CircleShape)
                         .fillMaxSize()
                 ) {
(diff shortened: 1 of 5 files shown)


wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/episode/ObtainConfirmationScreen.kt
```diff
@@ -56,7 +57,7 @@ fun ObtainConfirmationScreen(
                 Box(
                     contentAlignment = Alignment.Center,
                     modifier = Modifier
-                        .background(color = Color(0xFF202124)) // FIXME
+                        .background(color = WearColors.FF202124) // FIXME
```
ashiagr, line 60: Do you still need `FIXME` comment?

ashiagr (approved):

wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/episode/ObtainConfirmationScreen.kt
mchowning: Nope, thanks for catching that. I've removed it.

PR #864, opened 2023-04-03
Merge wear work to main branch

mchowning:
## Description
This brings the wear os app into the `main` branch. This does not release the wear OS app, which is a separate application that must be built separately from the phone app and submitted to the Play store on its own.

This PR contains no changes apart from merging `add/wear` and `main`.

## Testing Instructions
1. Smoke test the phone app
2. Smoke test the wear app
3. Verify that installing the phone app does not also install the wear app

I've also confirmed that the apk size for the phone app is substantially the same (if the wear app had been included as a `wearApp` dependency for the phone app, which was required for wear os 1, the wear app would be included in the phone app and installed when the phone app is installed). To be perfectly clear, we are _not_ declaring the wear app module as a `wearApp` dependency of the phone app because we're not currently planning to support Wear OS v1.

## Checklist


ashiagr (approved):
Works as described. Thank you for this, @mchowning! It reduces so much hassle of maintaining a separate wear app branch! 🥳

PR #868, opened 2023-04-04
Show ratings

ashiagr:
## Description

This integrates with the ratings API to display ratings for podcasts.

ℹ️ PR size is big due to the large (1.4K lines) auto-generated `75.json` db schema file. 
ℹ️ There's a lot of scope for UI improvements, the PR is created to build a basic structure to show ratings. 
ℹ️ Since the PR includes DB migration with fields subject to change, it currently targets a feature branch.

#### Test.1 Ratings display

1. Fresh install the app
2. Navigate to a podcast
3. ✅ Verify nothing displays and the app doesn't crash
4. Enable the `SHOW_RATINGS` feature flag in `base.gradle`
5. Navigate to a podcast again, ✅ verify that the app doesn't crash (notice that the star rating view is displayed without any filled stars - UI/ animation for this will get updated in a future PR (we might want to show a disabled view in case of error or show no ratings))
6. Since the API isn't available yet you can apply: [show_ratings_test_endpoint.patch](https://github.com/Automattic/pocket-casts-android/files/11147277/show_ratings_test_endpoint.patch)
9. This will generate the same response but without loading from the server db
10. You can change the total and average numbers to get different values
11. ✅ Verify the star rating appears correctly

#### Test.2 DB Migration

1. Fresh install the app from `main` branch
2. Switch to `task/show-ratings` branch and increment version code in `version.properties`
3. Enable the `SHOW_RATINGS` feature flag in `base.gradle` and  apply: [show_ratings_test_endpoint.patch](https://github.com/Automattic/pocket-casts-android/files/11147277/show_ratings_test_endpoint.patch)
4. Install and launch the app from `task/show-ratings` branch 
5. Navigate to a podcast
6. ✅ Verify the star rating appears correctly (no issues due to db migration)

#### Test.3 Load from local db no network

1. Continue from Test 2 > Step.6
2. Disable the internet
3. Go back and open the same podcast for which the rating was loaded earlier
4. ✅ Verify that the last saved star rating appears

## Screenshots or Screencast 

https://user-images.githubusercontent.com/1405144/229768063-d07f60c2-70ef-49b4-b617-0be674cd1786.mp4

## Checklist
 
- [x] with different themes
- [ ] for accessibility with TalkBack - this is not yet tested

### DIFF
--- a/base.gradle
+++ b/base.gradle
@@ -31,6 +31,7 @@ android {
         // Feature Flags
         buildConfigField "boolean", "END_OF_YEAR_ENABLED", "false"
         buildConfigField "boolean", "SINGLE_SIGN_ON_ENABLED", "false"
+        buildConfigField "boolean", "SHOW_RATINGS", "false"
 
         testInstrumentationRunner project.testInstrumentationRunner
         testApplicationId "au.com.shiftyjelly.pocketcasts.test" + project.name.replace("-", "_")
--- a/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/components/StarRatingView.kt
+++ b/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/components/StarRatingView.kt
@@ -1,0 +1,81 @@
+package au.com.shiftyjelly.pocketcasts.podcasts.view.components
+
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.material.Icon
+import androidx.compose.material.MaterialTheme
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.Star
+import androidx.compose.material.icons.filled.StarBorder
+import androidx.compose.material.icons.filled.StarHalf
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.tooling.preview.PreviewParameter
+import androidx.compose.ui.unit.dp
+import au.com.shiftyjelly.pocketcasts.compose.AppThemeWithBackground
+import au.com.shiftyjelly.pocketcasts.compose.components.TextP50
+import au.com.shiftyjelly.pocketcasts.compose.preview.ThemePreviewParameterProvider
+import au.com.shiftyjelly.pocketcasts.compose.theme
+import au.com.shiftyjelly.pocketcasts.ui.theme.Theme
+import au.com.shiftyjelly.pocketcasts.utils.extensions.abbreviated
+
+private const val MAX_STARS = 5
+@Composable
+fun StarRatingView(averageRating: Double, total: Int? = null) {
+    Row(
+        modifier = Modifier
+            .fillMaxWidth()
+            .padding(horizontal = 14.dp, vertical = 4.dp),
+        horizontalArrangement = Arrangement.Start,
+        verticalAlignment = Alignment.CenterVertically,
+    ) {
+        Stars(rating = averageRating, color = MaterialTheme.theme.colors.filter03)
+
+        total?.let { TextP50(text = it.abbreviated) }
+    }
+}
+
+@Composable
+private fun Stars(
+    rating: Double,
+    color: Color,
+) {
+    // truncate the floating points off without rounding
+    val stars = rating.toInt()
+    // Get the float value
+    val half = rating % 1
+
+    Row(horizontalArrangement = Arrangement.Start) {
+        for (index in 0 until MAX_STARS) {
+            Icon(
+                imageVector = iconFor(index, stars, half),
+                contentDescription = null,
+                tint = color
+            )
+        }
+    }
+}
+
+private fun iconFor(index: Int, stars: Int, half: Double) = when {
+    index < stars -> Icons.Filled.Star
+    (index > stars) || (half < 0.5) -> Icons.Default.StarBorder
+    else -> Icons.Filled.StarHalf
+}
+
+@Preview
+@Composable
+private fun PodcastRatingsPreview(
+    @PreviewParameter(ThemePreviewParameterProvider::class) themeType: Theme.ThemeType,
+) {
+    AppThemeWithBackground(themeType) {
+        StarRatingView(
+            averageRating = 3.75,
+            total = 15071,
+        )
+    }
+}
--- a/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/podcast/PodcastFragment.kt
+++ b/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/podcast/PodcastFragment.kt
@@ -138,8 +138,18 @@ class PodcastFragment : BaseFragment(), Toolbar.OnMenuItemClickListener, Corouti
 
     override var statusBarColor: StatusBarColor = StatusBarColor.Custom(color = 0xFF1E1F1E.toInt(), isWhiteIcons = true)
 
-    private val onHeaderSummaryToggled: (expanded: Boolean) -> Unit = {
-        analyticsTracker.track(AnalyticsEvent.PODCAST_SCREEN_TOGGLE_SUMMARY, mapOf(IS_EXPANDED_KEY to it))
+    private val onHeaderSummaryToggled: (
+        podcastUuid: String,
+        expanded: Boolean,
+        userInitiated: Boolean,
+    ) -> Unit = { podcastUuid, expanded, userInitiated ->
+        if (userInitiated) {
+            analyticsTracker.track(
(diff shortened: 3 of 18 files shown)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/ratings/RatingsManagerImpl.kt
```diff
@@ -0,0 +1,37 @@
...
+    appDatabase: AppDatabase,
+) : RatingsManager, CoroutineScope {
+    private val podcastRatingsDao = appDatabase.podcastRatingsDao()
+
+    override val coroutineContext: CoroutineContext
+        get() = Dispatchers.Default
+
+    override fun observePodcastRatings(podcastUuid: String) =
+        podcastRatingsDao.observePodcastRatings(podcastUuid)
+            .map { it.firstOrNull() ?: PodcastRatings(podcastUuid, 0.0) }
```
ashiagr, line 22: I return a blank rating if there's no rating in the DB for the podcast so that UI is still rendered (without any filled stars). This is subject to change based on final UI decision. It doesn't look like the best way to do it, let me know if you have suggestions. 

Also, it returns a flowable so that it can fit in the [loadPodcasts(...)](https://github.com/Automattic/pocket-casts-android/blob/main/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastViewModel.kt#L92) transformations, where podcast rating it loaded after the podcast is loaded.

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/podcast/PodcastFragment.kt
```diff
@@ -138,8 +138,18 @@ class PodcastFragment : BaseFragment(), Toolbar.OnMenuItemClickListener, Corouti
...
+        expanded: Boolean,
+        userInitiated: Boolean,
+    ) -> Unit = { podcastUuid, expanded, userInitiated ->
+        if (userInitiated) {
+            analyticsTracker.track(
+                AnalyticsEvent.PODCAST_SCREEN_TOGGLE_SUMMARY,
+                mapOf(IS_EXPANDED_KEY to expanded)
+            )
+        }
+        if (BuildConfig.SHOW_RATINGS && expanded) viewModel.refreshPodcastRatings(podcastUuid)
```
ashiagr, line 152: Ratings are refreshed whenever the podcast details view is expanded. We still need to decide on the refresh logic and whether ratings need to be displayed for subscribed podcasts.

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/components/StarRatingView.kt
```diff
@@ -0,0 +1,81 @@
...
+import androidx.compose.ui.tooling.preview.PreviewParameter
+import androidx.compose.ui.unit.dp
+import au.com.shiftyjelly.pocketcasts.compose.AppThemeWithBackground
+import au.com.shiftyjelly.pocketcasts.compose.components.TextP50
+import au.com.shiftyjelly.pocketcasts.compose.preview.ThemePreviewParameterProvider
+import au.com.shiftyjelly.pocketcasts.compose.theme
+import au.com.shiftyjelly.pocketcasts.ui.theme.Theme
+import au.com.shiftyjelly.pocketcasts.utils.extensions.abbreviated
+
+private const val MAX_STARS = 5
```
mchowning, line 27: Since the star values themselves are coming from the backend, I wonder if the maximum possible value should also be coming from the backend. Feels like that makes more sense as the source of truth. I don't feel strongly though since I doubt it makes much practical difference in the end as it's unlikely we would ever change away from 5 stars.

CookieyedCodes:
Are rateings going to tie into search? Also are ratings going to be for your own private database or are they going to itunes or elsewhere just interested 🤔😉:)

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/podcast/PodcastFragment.kt
mchowning: This will refresh the ratings each time the episode details are expanded, right? I'd be inclined to just refresh them once when the screen loads. Admittedly that would refresh the ratings in some cases where they aren't needed (if it's a subscribed podcast and the user never expands the details), but it also avoids any wait while the ratings are loaded if the user does expand the podcast details.

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/podcast/PodcastFragment.kt
```diff
@@ -625,6 +635,13 @@ class PodcastFragment : BaseFragment(), Toolbar.OnMenuItemClickListener, Corouti
             }
         )
 
+        viewModel.ratingsState.observe(viewLifecycleOwner) { state ->
+            when (state) {
+                is PodcastViewModel.RatingState.Loaded -> adapter?.setRatings(state.ratings)
+                is PodcastViewModel.RatingState.Error -> Unit // Do Nothing
```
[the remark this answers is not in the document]
mchowning: Depending on how the UI feature for this plays out, I wonder if it might make sense to include all the ratings retrieval logic ina new dedicated ViewModel inside the `StarRatingView` (or maybe a wrapper composable that handles the `StarRatingView` and whatever loading/error UI we want to present). 

I wouldn't worry about changing this now since depending on the UI requirements that change might not make sense. But it may be worth keeping in mind.

modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/extensions/Int.kt
```diff
@@ -17,3 +18,18 @@ fun Int?.isPositive(): Boolean {
...
     return this != null && this > 0
 }
+
+val Int.abbreviated: String
+    get() {
+        val number = this.toDouble()
+
+        return when {
+            number >= 1_000_000 -> "${scaleDown(value = number, by = 1_000_000.0)}M"
+            number >= 1_000 -> "${scaleDown(value = number, by = 1_000.0)}K"
```
mchowning, line 28: This is really nice! 🤩

mchowning (approved):
Works great and code looks good. Left a few comments, but no blockers.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/ratings/RatingsManagerImpl.kt
ashiagr: Good idea! I applied that in ca1119e

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/podcast/PodcastFragment.kt
ashiagr: > This will refresh the ratings each time the episode details are expanded, right? 

That's right.

>  I'd be inclined to just refresh them once when the screen loads. 

I was thinking of unnecessarily not querying ratings for subscribed podcasts until the user expands the details view. But I see your point. I'll take care of it in a future PR.

base.gradle
```diff
@@ -31,6 +31,7 @@ android {
         // Feature Flags
         buildConfigField "boolean", "END_OF_YEAR_ENABLED", "false"
         buildConfigField "boolean", "SINGLE_SIGN_ON_ENABLED", "false"
+        buildConfigField "boolean", "SHOW_RATINGS", "false"
```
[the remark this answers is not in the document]
ashiagr: I just wanted to have a feature flag in place, without worrying too much about the remote feature flag right now as the feature is still in initial phase. We can discuss it more and standardize the process.

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/components/StarRatingView.kt
ashiagr: I don't feel strongly about it too.

ashiagr:
👋 @CookieyedCodes 

> Are rateings going to tie into search? 

Not as of now 

> Also are ratings going to be for your own private database or are they going to itunes or elsewhere just interested 🤔😉:)

There's a lot of maths going on behind the scene using dynamic weights to calculate ratings. Typically, Pocket Cast ratings will have more weight.

CookieyedCodes:
Coolio that's exactly what I would prefer, maby consider it to help for rankings as that should help prioritise feeds ❤️, my only feedback at this time is I find the stars a tad big (I personally would also expirment with the placement it might be better to put it in the rss information box 🤔, tho it might be getting used to ui change, I'll adapt eventually) 

Also make sure that this yellow dynamically theams with the ui (keep it shades of yellow for distinction) the one thing I don't want is a yellow that blinds me when my phone goes black & white when entering bedtime mode 😅

ashiagr:
Thanks for extra thoughts, @CookieyedCodes 🙌 

> I also assume this is comeing to the webapp ya?

Yep, release timing might be different but it'll be included there as well.

PR #871, opened 2023-04-06
Add expandable show notes to watch episode screen

mchowning:
## Description
Adds an expandable section for the show notes to the episode details screen on the watch.

This is converting the HTML show notes into plain text. This seemed like the best way to deal with this for now since web views are not supported on Wear OS. This does mean that links and other HTML elements are basically not rendered, but I think this implementation is better than nothing and it's something we can address later if it proves to be a problem (I'm skeptical that many will be doing much reading of long show notes on their watch, but maybe I'm wrong).

## Testing Instructions
1. Go to the episode details screen
2. Scroll down and observe that the show notes displays at most three lines and (assuming there is more than 3 lines), there is a gradient fading out the text toward the bottom
3. Tap on the show notes
4. Observe the notes expand
5. Tap on the show notes again
6. Observe that the notes collapse. If you had scrolled down quite a ways in the show notes, when collapsing them the list should scroll to ensure that the collapsed scroll notes are still visible (as opposed to the scroll position remaining at the bottom of the list). I have noticed that sometimes there is a bit of a jump with this on my emulator, but it doesn't seem to occur on my physical device.

## Screenshots or Screencast 

[device-2023-04-06-174133.webm](https://user-images.githubusercontent.com/4656348/230499245-3ec9e3f9-f241-4e46-989a-5d7cb6ea9aef.webm)

## Checklist

### DIFF
--- a/modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/ServerShowNotesManager.kt
+++ b/modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/ServerShowNotesManager.kt
@@ -17,6 +17,8 @@ import timber.log.Timber
 import java.io.IOException
 import javax.inject.Inject
 import javax.inject.Singleton
+import kotlin.coroutines.resume
+import kotlin.coroutines.suspendCoroutine
 
 @Singleton
 class ServerShowNotesManager @Inject constructor(@ShowNotesCache private val httpShowNotesCache: OkHttpClient) {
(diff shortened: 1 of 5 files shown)


ashiagr:
Since [Wear OS does not support using WebView](https://developer.android.com/develop/ui/views/layout/webapps/managing-webview), I agree it is better to show plain text for now. 

The text expands and collapses well in my testing. Nice job getting this added!

ashiagr (approved):

PR #872, opened 2023-04-07
Add header to add to up next screen on watch

mchowning:
## Description
Adds a header with "Add to up next?" to the add to up next screen on the watch to match the Figma designs. In addition, I've made that screen scrollable since there is now more content.

The icons on this screen are different from Figma. I think these are the appropriate icons for "adding" to the up next queue (as opposed to moving something within the up next queue), so I've left them as-is. I've left a question for Adam though confirming whether that is ok, and if they need to be changed I'll do that in a separate PR.

## Testing Instructions
1. Ensure that you have at least 1 episode in your Up Next queue
2. Go to the episode details screen for an episode that is not in your Up Next queue
3. Tap the add to up next button
4. Observe that you get the below screen with the "Add to up next?" header. There are no functionality changes, so no specific testing is needed apart from checking that the screen looks fine.

## Screenshots or Screencast 

image

## Checklist

### DIFF
--- a/modules/services/localization/src/main/res/values/strings.xml
+++ b/modules/services/localization/src/main/res/values/strings.xml
@@ -8,6 +8,7 @@
 
     <string name="account">Account</string>
     <string name="add_to_up_next">Add to Up Next</string>
+    <string name="add_to_up_next_question">Add to up next?</string>
     <string name="archive">Archive</string>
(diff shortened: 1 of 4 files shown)


ashiagr (approved):
Looks good!

PR #873, opened 2023-04-07
Update now playing chip on watch

mchowning:
## Description
This updates the UI of the Now Playing chip.

The current designs only specify that this chip should be showing when an episode is playing, but I'm always presenting it currently because I think otherwise it's too difficult to start playing an episode when you first open the app (I certainly think we need it until I get the pager added).

I'm also using an ellipsis when the text fields are too long instead of clipping them because clipping would cut off the field at the end of the nearest word and, depending on the length and positions of the words, this could result in some absurdly short text on the chip (i.e., "This is spectacular" could get cut so that it would show only "This is" if clipping was used).

## Testing Instructions
1. Open a fresh install of the app
2. Observe the now playing chip has a play icon and says "Now Playing"
3. Log into an account and wait for a bit for it to load (it's pretty slow because it's still doing this over Bluetooth... this is something on my todo list)
4. Once the account has loaded with the currently queued episode, observe the Now Playing chip has a play icon, the podcast artwork as the background, along with the podcast title and episode title.
5. Tap on the Now Playing chip
6. Start playing the podcast
7. Swipe back
8. Observe that the Now Playing chip now says "Now Playing" in place of the podcast name and the play icon has been replaced with an animated equilizer.

## Screenshots or Screencast 

https://user-images.githubusercontent.com/4656348/230661981-4d798cf3-b2e4-44f3-9fca-e55da5ea4e9b.mov

## Checklist


ashiagr:
All looks good! 

Just one finding: My build variant was set to release and it crashed on playing the episode from the chip with [logs](https://github.com/Automattic/pocket-casts-android/files/11196417/crash.txt). It was a fresh install and it happened after I logged in and played from the now-playing chip. It didn't crash when I changed to debug variant. Then again when I switched to the release build variant (fresh install), it crashed.  Again, on the release variant, I played an episode from another podcast and it played fine. I switched back to the episode that caused the crash and it played fine too.

https://user-images.githubusercontent.com/1405144/231055497-9b9f64bb-efac-43c3-a334-528a912864ce.mp4

As the issue cannot be consistently reproduced, I'm approving the PR. We can fix it if it comes up again while testing future PRs.

ashiagr (approved):

CookieyedCodes:
I would recommend a delayed text scroll on those titals especially on watch do minimize it tho 😉 (I would also suggest it on mobile in some places for the same reason:))

mchowning:
> My build variant was set to release and it crashed on playing the episode from the chip...

Thanks for mentioning this, I'll definitely keep an eye out for it going forward. I haven't seen it yet myself, but I also haven't tested much with release builds to this point.

mchowning:
> I would recommend a delayed text scroll on those titals especially on watch

I did consider adding that, but decided to just keep it simple for now. I'll play with it a bit and see how some scrolling would look. Thanks for the suggestion @CookieyedCodes .

PR #874, opened 2023-04-11
Add refresh button to watch

mchowning:
## Description
This adds a refresh button to the watch's settings screen. I'm not sure if this is where it will end up, but I wanted to add the functionality since it will help with development.

> **Note**
> This PR builds on https://github.com/Automattic/pocket-casts-android/pull/873, so you may want to review that PR before this one.

## Testing Instructions
1. Log into an account on the watch.
2. Pick one of the podcasts from the Podcasts screen and remember it
3. Log into the same account on another device
4. Unsubscribe from the podcast you picked on step 2
5. Sync the changes to the server from the other device
6. On the watch, tap the "Refresh now" button on the settings screen
7. Observe that the icon animates while the refresh is in progress
8. Once the refresh is complete, observe that the unsubscribed podcast is no longer listed on the watch's podcasts screen.

## Screenshots or Screencast 

https://user-images.githubusercontent.com/4656348/231039716-d629f065-78ae-403c-9997-bca287bb5e89.mov

## Checklist

### DIFF
--- a/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/SettingsScreen.kt
+++ b/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/SettingsScreen.kt
@@ -1,11 +1,19 @@
 package au.com.shiftyjelly.pocketcasts.wear.ui
 
 import android.content.res.Configuration
+import androidx.compose.animation.core.Animatable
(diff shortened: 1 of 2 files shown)


ashiagr (approved):
LGTM 👍 

I like the rotating animation for the refresh icon! 😄 Let's see what we end up with.

PR #881, opened 2023-04-14
Prefer faster networks for downloads on watch

mchowning:
## Description
This PR uses the horologist network-awareness library to prefer faster networks when we download media. Without this, the watch defaults to using slower network connections (i.e., Bluetooth). More info [here](https://google.github.io/horologist/network-awareness/).

There was some trickiness integrating this because this wraps the `Call.Factory` which we create in the repository module. I initially tried using the library directly there, but the problem is that the network-awareness library (1) uses watch features (or at least claims it does [in its manifest file](https://github.com/google/horologist/blob/main/network-awareness/src/main/AndroidManifest.xml#L22)) and (2) has min SDK of 26. Since the phone app uses our repository module, both of these requirements created issues.

To work around this, I have created an annotation in the repository module to provide the relevant `Call.Factory`, but the 
 `@Provides` methods for those annotations are in the application modules (phone, automotive, and wear). I'm keeping the actual construction of the non-wear dependencies in the repository module—the application module is really only responsible for identifying which dependency to inject.

I had to do something similar with the `Request.Builder` because this library creates an extension method (`.requestType(...)` that needs to be set with an object from the library.

I'm only doing this for downloads currently, but I imagine I'll be doing something similar for some of our other okhttp requests in the future, so let me know any issues you see with this pattern.

This is a bit confusing from the DI perspective, but it completely avoids having any horologist dependency in our repository module, which feels like a nice win. With that said, let me know if you think this is too messy and not worth it.

## Testing Instructions
1. Install the app on a physical watch that is paired to a phone by Bluetooth and connected to a wifi network
2. Go to the episode details screen and download an episode
3. Check the logs for the watch, you should see a message along the lines of `HTTPS request media-download Wifi` reflecting that Wifi was used for the download
4. Alter the code to prefer Bluetooth connections for downloads [like this](https://gist.github.com/mchowning/6be13d811c8346ac849be44dd8806223).
5. Reinstall the app with the altered code
6. Download an episode again
7. Check the logs for the watch, you should see a message along the lines of `HTTPS request media-download BT` reflecting that Bluetooth was used for the download

## Checklist


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/di/RepositoryProviderModule.kt
```diff
@@ -15,4 +18,19 @@ class RepositoryProviderModule {
     @Provides
     @Singleton
     fun provideTokenHandler(syncAccountManager: SyncAccountManager): TokenHandler = syncAccountManager
+
+    @Provides
+    @Singleton
+    @DownloadOkHttpClient
+    fun downloadOkHttpClient(): OkHttpClient {
```
mchowning, line 25: I am providing this in the repository module because the wear app needs the `OkHttpClient` to create a `NetworkSelectingCallFactory` which serves as the `Call.Factory` for downloads (`@DownloadCallFactory`) in the watch app. The phone and automotive apps just return this directly with their `@DownloadCallFactory` providers.

wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/di/WearNetworkModule.kt
```diff
@@ -33,6 +49,65 @@ object NetworkModule {
...
+            coroutineScope = coroutineScope,
+            timeout = 5.seconds,
+        )
+    }
+
+    @Provides
+    @DownloadRequestBuilder
+    fun downloadRequestBuilder(): Request.Builder =
+        Request.Builder()
+            .requestType(RequestType.MediaRequest.DownloadRequest)
```
mchowning, line 107: This `RequestType` comes from the horologist library and setting it on the `Request.Builder` for wear requests is the only reason I need to inject it separately for the wear vs phone/automotive apps.

wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/networking/PocketCastsNetworkingRules.kt
```diff
@@ -0,0 +1,93 @@
...
+import com.google.android.horologist.networks.rules.RequestCheck
+import timber.log.Timber
+
+object PocketCastsNetworkingRules : NetworkingRules {
+    override fun isHighBandwidthRequest(requestType: RequestType): Boolean {
+        if (BuildConfig.DEBUG) {
+            // For testing purposes fail if we get unknown requests
+            check(requestType != RequestType.UnknownRequest) {
+                "Unknown request type. Failing on debug builds."
+            }
```
mchowning, line 24: This seemed like a good way to make sure we don't use these networking rules but forget to set a request type on the request.

automotive/src/main/java/au/com/shiftyjelly/pocketcasts/di/AutomotiveAppModule.kt
```diff
@@ -0,0 +1,29 @@
...
+import dagger.hilt.InstallIn
+import dagger.hilt.components.SingletonComponent
+import okhttp3.Call
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import javax.inject.Singleton
+
+@Module
+@InstallIn(SingletonComponent::class)
+object AutomotiveAppModule {
```
mchowning, line 17: This is a duplicate fo the `AppModule`. We could create a new module shared only by the phone and wear apps to avoid this duplication, but that felt like it would be overcomplicating things. Maybe later something like that may make sense (I kind of doubt it though).

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/download/task/DownloadEpisodeTask.kt
```diff
@@ -64,7 +68,9 @@ class DownloadEpisodeTask @AssistedInject constructor(
     @Assisted params: WorkerParameters,
     var downloadManager: DownloadManager,
     var episodeManager: EpisodeManager,
-    var userEpisodeManager: UserEpisodeManager
+    var userEpisodeManager: UserEpisodeManager,
+    @DownloadCallFactory private val callFactory: Call.Factory,
+    @DownloadRequestBuilder private val requestBuilderProvider: Provider<Request.Builder>
```
mchowning, line 73: Depending on which app is running (phone/automotive/wear), these dependencies will be provided by the `@DownloadCallFactory` and `@DownloadRequestBuilder` provide methods from the relevant app (i.e., from `AppModule`, `AutomotiveAppModule` or `WearNetworkModule`).

mchowning:
>  I still see HTTPS request media-download Wifi in logs even after applying the patch and making sure that watch is paired to the phone using Bluetooth and connected to a WiFi network.

Interesting. I assume you're using a physical watch device paired to a physical phone in your testing. I just retested this, and I'm still seeing the expected behavior on my watch. 🤔 

Perhaps you could try dropping a breakpoint or log message on the changed line and see (1) that the code is actually getting to that point and, if it is (2) what networks are actually available (i.e., maybe the watch is reporting that Bluetooth isn't available for some reason).

ashiagr:
> I assume you're using a physical watch device paired to a physical phone in your testing. 

Yes, I was using a physical watch, Bluetooth paired with a physical mobile device, and connected to wifi.

> Perhaps you could try dropping a breakpoint or log message on the changed line and see (1) that the code is actually getting to that point and, if it is (2) what networks are actually available (i.e., maybe the watch is reporting that Bluetooth isn't available for some reason).

Also added a breakpoint, the code actually hit that point

I'm not sure what was wrong, I factory reset my watch and tried it all over again, 😅 this time it worked 🥳 

Screenshot 2023-04-18 at 12 44 47 PM

Great job! I'm approving the PR now.

ashiagr (approved):

PR #882, opened 2023-04-14
Fix up playback service on wear

mchowning:
## Description
It seems that I left more than a few loose ends when I switched the watch app away from media3 to using the same playback setup as the phone app. The biggest thing I missed was updating the service declaration in wear's manifest file. This meant that the OS wasn't being informed when Pocket Casts was performing playback.

Less critically, I left the old service file and all of the DI setup for it. This PR cleans that up since we'll just switch this to media3 when we do the switch with the phone app.

> **Note**
> This PR builds on top of https://github.com/Automattic/pocket-casts-android/pull/881, so we probably want to merge that one first.

## Testing Instructions
1. Start playing an episode on a pixel watch/emulator
2. Go to the watch's home screen
3. Observe that the playback icon is active (see screencast below)

## Screenshots or Screencast 

https://user-images.githubusercontent.com/4656348/231916963-b3ffc88b-3c5f-46ac-b904-7ebb4fd27039.mov

## Checklist


wear/src/main/AndroidManifest.xml
```diff
@@ -3,17 +3,18 @@
...
-    <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED"/>
-    <uses-permission android:name="android.permission.VIBRATE"/>
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
     <uses-permission android:name="android.permission.BLUETOOTH"/>
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE"/>
-    <uses-permission android:name="com.android.vending.CHECK_LICENSE"/>
-    <uses-permission android:name="com.android.vending.BILLING" />
-    <uses-permission android:name="android.permission.SCHEDULE_EXACT_ALARM" />
+    <uses-permission android:name="android.permission.INTERNET"/>
+    <uses-permission android:name="android.permission.POST_NOTIFICATIONS"/>
```
mchowning, line 10: I added this `POST_NOTIFICATIONS` permission to stay consistent with the phone app. The other changes to the permissions are just reordering them to be alphabetical.

ashiagr (approved):
> The biggest thing I missed was updating the service declaration in wear's manifest file. 

I, too, share the responsibility for not noticing it. 

I confirm that the playback icon is active on watch's home screen when media is playing.

PR #885, opened 2023-04-14
Advanced settings menu

### LINKED ISSUE #839: Bug: background update ignores download preferences when upnext is set to queue new episodes

### LINKED ISSUE #858: Update Interval configuration

A [discussion](https://github.com/Automattic/pocket-casts-android/discussions/738#discussioncomment-4819705) about APIs made reference to control server costs. #839 is making improvements to task scheduling for work manager to reduce data use for users without having to disable background data manually on android app preferences, but it also has the side effect that it will cut the number of requests down for each user that prefers **not** to be updated while on a metered network. Work manager currently runs the update task on an hourly basis.

This of course has the economies of scale. Assuming 100k users current mechanics would trigger 2.4M requests per day.

Additionally some users could be fine with being updated once a day at a specific time or just after they wake up or go to work. There's a huge potential for savings on being more efficient about polling.

The [PeriodicWorkRequest](https://developer.android.com/reference/androidx/work/PeriodicWorkRequest) supports [constraints ](https://developer.android.com/reference/androidx/work/Constraints), both of which get used by current tasking mechanisms. 

### Approach 1 - inexact alarms 

Android also supports [inexact alarms](https://developer.android.com/training/scheduling/alarms#inexact), which would allow for a window for these updates to occur. The idea is that for cases where users typically wake up in the morning, want to get their podcasts updated just before they go to work, inexact alarms would help trigger this. So how do we configure it?

A straight forward approach would be to have a setting the user could specify that the background update should happen "once a day" and give it a time. With android 12+ the window size is 10 minutes. So when provided the time, the inexact alarm could be scheduled for `time - 10 minutes` with a window size of 10 at a bare minimum. This could be made wider using the same approach.

This would eliminate the need for polling all day/night long and could cut down on the API requests the most. Alarms do require an additional permission to be declared in the manifest. There's also a side benefit of being slightly more efficient on the user side, but I think a bigger benefit would be in cutting down on server costs. Generally speaking, I think 90% of the time, the background task to update has nothing to update.

### Approach 2 - push to update

There's also another approach, but it would require server side processing and push notifications to trigger tasks to update on the client side. Given that Automattic knows which podcasts users are subscribed to, when updates to feeds are detected, they send a push notification to subscribers to trigger the app to pull down updates. The client then no longer needs to poll. This also assumes reliable but not necessarily immediate delivery of push notifications. This, however, may increase compute cost for processing and sending push notifications. You also likely want to avoid sending to subscribers that haven't been running the app recently and there's differences between platforms, iOS and Android. The idea though is the less polling is involved in the process and more event driven, the more efficient the use of resources will be.

gyund:
## Description

Advanced settings menu to house various tweaks that might not be mainstream for users. This helps keep them out of the way. Metrics to these options will be added to determine if they warrant a "graduation" down the road. 

Advanced options will always default to a state that does not change the default behavior of the app.

The advanced settings in this PR add functionality that allows the user to disable sync while on a metered network. This prevents podcasts from getting added to the "up-next" queue, which would cause streaming activity to occur for podcasts added to the top of the list. By performing sync on only unmetered networks, a user can generally guarantee that if they configure podcasts to be automatically added to the up-next queue and downloaded when added to the up-next queue, all podcasts in the up-next queue will be fully downloaded and not updated automatically until they return to an unmetered network.

Fixes #839 

## Testing Instructions

Automated testing has been added at the unit and instrumented level. Instrumented tests include verifying the default settings / behavior before the advanced options were added and the effects it has on settings that derive a behavior from these settings.

For manual testing the main thing to observe is:

- When "Sync on Metered" is enabled **_(default)_**
    - No changes from prior releases 
- When "Sync on Metered" is disabled 
    - While connected to wifi **(unmetered)**, there is no change in sync behavior
    - While connected to a metered network:
        - Sync will not happen automatically when you open the app. You'll notice the "Last refresh" on the profile page will not update the time
        - Sync will not happen in the background if you have _Settings -> Storage & Data Use -> Background Refresh_ **(enabled)** 

Background sync is enabled by default. To test this PR:

1. Tap on the Profile tab
2. Note that "Last refreshed" is close to 0
3. Tap on the gear in the upper right
4. Tap on Advanced
5. Tap on "sync on Metered Networks" to disable 
6. Close the app completely
7. Open the app back up a minute or so later 
8. Tap on the Profile tab
9. Observe the Last refresh is not 0
10. Tap Refresh now
11. Observe the refresh occurs

## Screenshots or Screencast 

image
image

https://github.com/Automattic/pocket-casts-android/pull/885#discussion_r1170761086

## Current Development Tasks

- [x] Add [advanced preferences menu](https://github.com/Automattic/pocket-casts-android/issues/858#issuecomment-1506912325) 
- [x] verify preference menu is wired in correctly to the menu
- [x] Add logic to support #839
    - [x] Add unit/instrumented tests to verify settings and expectations
    - [x] Ensure the options are disabled when background sync is disabled (I'm going to just hide it and replacing it with a SettingsRow with primary text. Couldn't figure how I'd do a grey-out yet)
    - [x] manual testing of app open on and off metered network 
    - [x] See if there's other workmanager related settings that warrant separate tweakable options. (See [comment](https://github.com/Automattic/pocket-casts-android/pull/885#issuecomment-1509958157))
- [x] Re-evaluate #858 to see if it warrants the extra work as the implementation requires some INTENT mechanics that WorkManager simplifies
- [x] Search the backlog of issues that might fall under the category of advanced and see if there's any interesting / useful settings to include as well.
- [x] Sanity check any method naming and settings to make sure the meaning is clear.
- [x] Update Advanced Page with Possibly something like in this [review](https://github.com/Automattic/pocket-casts-android/pull/885#discussion_r1170279509)
- [x] Walk around with these advanced options for a week and dogfood them.
 

## Checklist

- [x] with different themes

For the last bullet, I modeled everything after an existing compose fragment and page. Not sure if there's anything to be done there.


CookieyedCodes:
As you are adding this one thing to add is the ability to add an hourly or bi-hourly timer (+time- for instance) just because I find a day a tad exssesive (but I wouldn't mind that option too) -maby another advanced option tho-

One thing to check btw is dose this get rid of the app syncing when not in that period/syncing when a podcast is finished & when the app is entered, because ideally I would like it only to delay stuff being added to upnext not playback sync/upnext sync 🤔

gyund:
@CookieyedCodes I'll have to look. Right now the advanced option to "don't automatically sync on metered" is absolute. You could force a sync on metered by swiping down though to achieve your desired behavior  but it wouldn't happen automatically on open or in the background.

It wouldn't prevent getting new episodes on manual sync. 

I'd have to do some homework to see if they push listening status separately from sync. If so, you should be good. If not, then we'd probably need a different option for "sync listening history on metered" vs "sync podcast feeds on metered" to do what you describe. The way I've structured the settings function in the code should make it easy to take advantage of doze when both are set if we need to make this modification. 

I've added tests for settings to make it explicit the desired outcome. Makes it easier to do code reviews too 😜.

CookieyedCodes:
No bothers then, I've had it setup my way useing the API (broadcast controller refresh or something like that a few months ago -the battery use was why I stopped (mostly dew to it kept trying to refresh until it actually got connected to the internet)-, just noticed a few of those quirks which is why I pointed it out), 100% want more advanced functions 

-dont remind me I want auto archive based on word & the ability to change the add to upnext position on the fly 😉 ect 😂-

gyund:
> @CookieyedCodes I'll have to look. Right now the advanced option to "don't automatically sync on metered" is absolute. You could force a sync on metered by swiping down though to achieve your desired behavior but it wouldn't happen automatically on open or in the background.

I looked 😜 , The [episodeManager ](https://github.com/Automattic/pocket-casts-android/blob/b8e167c1860b9980ba0915dcbc4341409a93adc8/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt#L1024)takes care of this and pushes it to the repository. So this should do what you were hoping without having to introduce any additional options.

CookieyedCodes:
Unfortunately I do suck at code, so I'll trust someone else to do it eventually 😅-what I am good at is understanding code logic-,

 like from my understanding the webapp just takes syncs the que it dosen't add episodes (do test cross platform on mobile too -including apple watch 🤔- tho I only sync the que generally so I only will change add to upnext list when moving to a new phone not generally, so it might be better to put a timer delay on the episode manager spefically that should still allow sync just delay the episodes comeing in if I understand correctly 🤔🤔🤔

gyund:
>  like from my understanding the webapp just takes syncs the que it dosen't add episodes (do test cross platform on mobile too -including apple watch 🤔- tho I only sync the que generally so I only will change add to upnext list when moving to a new phone not generally, so it might be better to put a timer delay on the episode manager spefically that should still allow sync just delay the episodes comeing in if I understand correctly 🤔🤔🤔

I likely won't mess with the backend sync behavior for a variety of reasons. 😱

As far as delays go, you typically want to process events as they happen or avoid triggers that would cause you to wake up and do nothing. Timers can cause event pileups if not done right, hence tags in workmanager. State can also change and generally complicates recovery logic if you don't have one spot to deal with truth and stale data.

Mobile gets more fun because folks can go in airplane mode or through a tunnel at any time. 🚇

CookieyedCodes:
Fair point, might be easier to change the add to upnext location on the fly then 🤔 (ie it "pins" the function + follows to that episode until it enters now playing (& then technically it's back at the top) & basically the episodes would be added after the swiped episode point -adjusted the point the the swipe right to left in my view-

 (add to uplast allways stays the same)
Any thoughts or flaws with my logic?

(Don't remind me I also would like download upnext top "custom number" episodes like the apple watch 😅, I do tend to think of features that would work nicely together 😅)

gyund:
WorkManager has [these ](https://developer.android.com/guide/background/persistent/getting-started/define-work#work-constraints)constraints:

- NetworkType **_(option controlled)_**
- BatteryNotLow  **_(always true)_**
- RequiresCharging
- DeviceIdle
- StorageNotLow

Do we want to tie any of these options in with Work Manager? There's currently an auto-download option to only do it while charging. Do we want an advanced "Sync only while charging" option or StorageNotLow (even just hardcoded like BatteryNotLow while I'm in this area? Otherwise I'll just check this off as covered in my task list.

gyund:
After reevaluating #858, I decided it's not really worth it to have a daily update schedule. I'm having a hard time finding specific details, but it seems that inexactRepeatingAlarms might have fairly large window. So for folks that might want #839, missing the window might mean missing episodes for the day.

gyund:
@mchowning Basically done with this minus testing in the background over the week to see if anything weird pops out. I don't expect anything as I have automated tests for some of the behaviors derived from settings and did manual testing as well. Feel free to piggyback off the instrumented settings tests approach for other things as you see fit.

While I was in this area of the code, if you think there's any want or need to adjust anything related to [other work manager constraints](https://github.com/Automattic/pocket-casts-android/pull/885#issuecomment-1509958157)? I plan on just leaving them as is for now.

mchowning (commented):
Nice work on this @gyund ! The core functionality is working great! 🎉 And thanks for adding in the advanced settings section too! 

I left a few comments with some sugguestions that I think might be worthwhile, but I like the direction this is going. 👍

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/viewmodel/AdvancedSettingsViewModel.kt
```diff
@@ -0,0 +1,100 @@
...
+            }
+        )
+    )
+
+    fun onFragmentResume() {
+    }
+
+    fun doPodcastsRefreshAutomatically(): Boolean {
+        return settings.refreshPodcastsAutomatically()
+    }
```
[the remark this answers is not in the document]
gyund: habits 😄, I jump around languages a lot. I'll convert it to use the State object like with `backgroundSyncOnMeteredState`

modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/SettingsImpl.kt
```diff
@@ -264,6 +265,33 @@ class SettingsImpl @Inject constructor(
         editor.apply()
     }
 
+    override fun syncOnMeteredNetwork(): Boolean {
```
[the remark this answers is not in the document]
gyund: 🤔 The more I think about this one, I'm not sure. All those methods depend on the setting `syncOnMeteredNetwork` to derive the proper response. So if they get split into another class, it seems like it might add unnecessary complexity to accessing settings. But I dunno 🤷 

There's other getter methods like `getPodcastsSortType` that have "derived logic" that might get weird to start breaking out of the settings class...at least as part of this task. It'll start requiring DI of different settings objects or adapters that consume a settings object to provide getters with a bit more logic to them. I think I see your point though and gravitating perhaps towards Single Responsibility Principle, with the Settings currently being a "bag of everything" and needing to trim it down to small bags of settings tailored toward different responsibilities. That would certainly break the interface down into more logical pieces, but perhaps part of another task to refactor.

As far as this function you were mentioning as an alternative:

```kotlin
override fun refreshPodcastsOnResume(context: Context): Boolean =
    settings.syncOnMeteredNetwork() || Network.isUnmeteredConnection(context)
```

My original implementation went this route, but `Network.isUnmeteredConnection(context)` becomes untestable because its type and would require Network to be injected. It can't be stubbed out to a mock from what I could tell, but I could be wrong. Injecting a POD for metered/unmetered made this testable and avoided making changes to the Network object.

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/AdvancedSettingsPage.kt
```diff
@@ -0,0 +1,122 @@
...
+            modifier
+                .fillMaxHeight()
+                .verticalScroll(rememberScrollState())
+        ) {
+            SettingSection(heading = stringResource(LR.string.settings_storage_section_heading_mobile_data)) {
+                if (automaticRefreshEnabled) { SyncOnMeteredRow(state.backgroundSyncOnMeteredState) } else {
+                    SettingRow(
+                        primaryText = stringResource(LR.string.settings_advanced_no_options)
+                    )
+                }
```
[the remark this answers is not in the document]
gyund: Working on this now. Haven't pushed it yet. The new page looks like the following:

image
image

Just need to get the enabled/disabled working correctly so when background sync is disabled, it's greyed out. Pretty sure I found the right spot to do it, but it's late so I'm probably malfunctioning 😄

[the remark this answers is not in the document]
gyund: Got it, now the switch can be shown as disabled: 

image

modules/services/localization/src/main/res/values/strings.xml
```diff
@@ -1302,11 +1302,8 @@
...
     <string name="settings_privacy_crash_link_summary">Allow us to link your Pocket Casts account to crash reports.</string>
 
-    <string name="settings_advanced_no_options">Enable background refresh to see advanced options</string>
     <string name="settings_advanced_sync_on_metered">Sync on Metered Networks</string>
-    <string name="settings_advanced_sync_on_metered_on">While on a metered network, synchronization will occur automatically in the background and when the application is opened.</string>
-    <string name="settings_advanced_sync_on_metered_off">While on a metered network, synchronization will only occur if performed manually.</string>
-
+    <string name="settings_advanced_sync_on_metered_summary">When enabled (default), synchronization will occur automatically in the background and when the application is opened.\n\nWhen disabled, synchronization will only occur if you manually refresh. Disable this setting if you prefer to avoid automatic playlist modifications and data usage while on mobile data.\n\nThis setting is only applicable when background refresh is turned on.</string>
```
mchowning, line 1312: It would be good to shorten this up. What do you think about this?

> Permits automatic background synchronization on metered networks. This is enabled by default and has no effect if background refresh is turned off.

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/viewmodel/AdvancedSettingsViewModel.kt
```diff
@@ -68,13 +61,12 @@ class AdvancedSettingsViewModel
...
         val backgroundSyncOnMeteredState: BackgroundSyncOnMeteredState
     ) {
 
         data class BackgroundSyncOnMeteredState(
-            @StringRes val summary: Int,
             val isChecked: Boolean = true,
+            val isEnabled: Boolean = true,
             val onCheckedChange: (Boolean) -> Unit,
         )
     }
```
mchowning: What you have here is fine, but I would probably collapse the `BackgroundSyncOnMeteredState` fields directly into the `State` object since `State` currently only has that single field:

``` kotlin
    data class State(
        val isChecked: Boolean = true,
        val isEnabled: Boolean = true,
        val onCheckedChange: (Boolean) -> Unit,
    )
```

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/AdvancedSettingsPage.kt
```diff
@@ -81,13 +75,13 @@ private fun SyncOnMeteredRow(
...
         primaryText = stringResource(LR.string.settings_advanced_sync_on_metered),
-        secondaryText = stringResource(state.summary),
-        toggle = SettingRowToggle.Switch(state.isChecked),
+        toggle = SettingRowToggle.Switch(state.isChecked, state.isEnabled),
         modifier = modifier.toggleable(
             value = state.isChecked,
             role = Role.Switch
         ) { state.onCheckedChange(it) }
     )
+    SettingRow(primaryText = "", secondaryText = stringResource(LR.string.settings_advanced_sync_on_metered_summary))
```
mchowning, line 84: Once we shorten up the description text, I think it looks fine just setting this as the secondary text on the main `SettingRow` instead of creating this second one to keep the toggle switch at the top.

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/viewmodel/AdvancedSettingsViewModel.kt
```diff
@@ -68,13 +61,12 @@ class AdvancedSettingsViewModel
...
 
     data class State(
-        val refreshPodcastsAutomatically: Boolean,
         val backgroundSyncOnMeteredState: BackgroundSyncOnMeteredState
     ) {
 
         data class BackgroundSyncOnMeteredState(
-            @StringRes val summary: Int,
             val isChecked: Boolean = true,
```
mchowning, line 68: Can we remove this `isChecked` default value of `true`? This value should always be populated from the user's settings so I think we should force the caller to pass it.

modules/features/settings/src/test/java/au/com/shiftyjelly/pocketcasts/settings/viewmodel/AdvancedSettingsViewModelTest.kt
```diff
@@ -0,0 +1,52 @@
...
+        whenever(settings.syncOnMeteredNetwork()).thenReturn(false)
+        viewModel = AdvancedSettingsViewModel(
+            settings,
+            analyticsTracker,
+            context
+        )
+    }
+
+    @Test
+    fun `verify settings methods gets called when they should`() {
```
mchowning, line 48: This test is failing. I think line 50 needs to be updated.

mchowning (commented):
Thanks for the updates @gyund. I think we're close. 

I think the only two blockers to merging are shortening up the description and fixing the view model test. My other comments are more just suggestions for consideration.

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/viewmodel/AdvancedSettingsViewModel.kt
gyund: True, my only thought with collapsing it inside is that if we decide to create or move other options in here and then it's not the only row state, then we'd have to break it out again. 🤷

gyund:
> Thanks for the updates @gyund. I think we're close.
> 
> I think the only two blockers to merging are shortening up the description and fixing the view model test. My other comments are more just suggestions for consideration.

🤦  can't believe I missed the test failure. I must not have rerun it after refactoring the summary stuff. Should be good now and I updated the code for the other comments. 🤞 - no errors

Issue #887, opened 2023-04-17
Media3 Migration


ashiagr:
## Tasks/ Checklist

- [x] [Migrate Standalone ExoPlayer libs to Media3 ExoPlayer](https://developer.android.com/guide/topics/media/media3/getting-started/migration-guide#exoplayer)  ([PR#886](https://github.com/Automattic/pocket-casts-android/pull/886))
- [x] [Migrate MediaBrowserService to MediaLibraryService](https://developer.android.com/guide/topics/media/media3/getting-started/migration-guide#MediaBrowserService) ([PR#886](https://github.com/Automattic/pocket-casts-android/pull/888))
    - [x] Register both (`MediaBrowserService`, `MediaLibraryService`) service interfaces for backward compatibility
    - [x] Add `media3-session` dependency
    - [x] Change service to inherit from a `MediaLibraryService` instead of `MediaBrowserService`
    - [x] Implement the service life-cycle methods
    - [x] Implement `MediaLibraryService.MediaLibrarySessionCallback`
    - [x] Use `androidx.media3.common.MediaItem` instead of [MediaBrowserCompat.MediaItem](https://developer.android.com/reference/kotlin/android/support/v4/media/MediaBrowserCompat.MediaItem) 
    - [x] Use `Futures` in contrast to the detachable Result approach of the `MediaBrowserServiceCompat`
    - [x] Implement [MediaSession.Callback.onAddMediaItems()](https://github.com/androidx/media/blob/1.0.0/demos/session/src/main/java/androidx/media3/demo/session/PlaybackService.kt#L184)
    - [x] Migrate `AutoPlaybackService`
   
- [ ] Update player for media3 
    - [x] Migrate local player instantiation to `PlaybackService` ([PR#894](https://github.com/Automattic/pocket-casts-android/pull/894))
    - [ ] Migrate switching to cast player
        - [x] Set wrapper player class on the media3 session that switches between the local and cast player ([PR#899](https://github.com/Automattic/pocket-casts-android/pull/899))
        - [x] Testing ([PR#899](https://github.com/Automattic/pocket-casts-android/pull/899))
            - [x] Starting playback on a chromecast from a paused/not-playing state works
            - [x] Transferring playback from the phone to the chromecast (and vice-versa) while playback is in progress, and playback continues
         - [ ] Cleanup

- [ ] [Control playback in the media session](https://developer.android.com/guide/topics/media/media3/getting-started/playing-in-background#controlling-playback)
[...]

PR #888, opened 2023-04-17
Migrate MediaBrowserService to MediaLibraryService

ashiagr:
Parent #887

## Description

This PR 
- Registers both (`MediaBrowserService`, `MediaLibraryService`) service interfaces for backward compatibility - 2040f7b7afed93a6ea4726245fe032418624ae50
- Adds `media3-session` dependency - 5389e811bab6f1c61e4eef89f27a6725c4ea28c3
- Uses `androidx.media3.common.MediaItem` instead of [MediaBrowserCompat.MediaItem](https://developer.android.com/reference/kotlin/android/support/v4/media/MediaBrowserCompat.MediaItem) 
d82b1ff3554e9ca772eaec9202d5ac98b6267a0a
- Implements `MediaLibraryService.MediaLibrarySessionCallback` - [diff](https://github.com/Automattic/pocket-casts-android/commit/d82b1ff3554e9ca772eaec9202d5ac98b6267a0a#diff-9cfa2a8a05b3d2dbd76dd8db7d068a1ea5103a000fbb6b194c2079c544736adbR604-R672)
    - Uses `Futures` in contrast to the detachable `Result` approach of the `MediaBrowserServiceCompat` (Adds [kotlin-coroutines-guava](https://github.com/Automattic/pocket-casts-android/commit/d82b1ff3554e9ca772eaec9202d5ac98b6267a0a#diff-d979b641bd0ea7c9da3fe113f9657636df1002652c75042cfe3b5203da064215R213) dependency to makes the `MediaBrowser` API fit more naturally into a suspend method as recommended [here](https://developer.android.com/guide/topics/media/media3/getting-started/migration-guide#MediaBrowser).)
-  Changes service to inherit from a `MediaLibraryService` instead of `MediaBrowserService` - 
- Implement the service life-cycle methods
   - Overrides `onGetSession(MediaSession.ControllerInfo)` to return the `MediaLibrarySession` that was built in `onCreate`.
- Implements `MediaSession.Callback.onAddMediaItems()`
- Migrates `AutoPlaybackService`

## Testing instructions
- Make sure that intermediate code-changes are as described above
- App compiles and episode can be played

Note that: 
- Lint check is disabled temporarily while changes are in progress 
- Changes to initialize the player correctly, control media playback in the session and more will be added as part of upcoming PRs.


mchowning (approved):
Looks good! 👍

mchowning:
The force push is because I pushed a commit turning warnings as errors on, and then I realized why that was a bad idea at this point 😄  so I force pushed back to the previous commit.

PR #889, opened 2023-04-17
Add google sign in to the automotive app

mchowning:
## Description
This adds the ability to sign in with google to the automotive app.

> **Warning**
> This PR is targeting the `release/7.37` branch because I think we need to roll SSO out at the same time everywhere as much as we can. Let me know if you feel differently.

### 1. Automotive device already connected to a Google Account
1. Go to the sign in screen (⚙️  → "Set Up Account")
2. Observe there is now a "Continue with Google" button
3. Tap "Continue with Google" and select a Google account

### 2. When the automotive device is not already connected to a Google account
1. Go to the sign in screen (⚙️  → "Set Up Account")
2. Observe there is now a "Continue with Google" button
3. Either sign in to a google account on the device or on an Android phone (I didn't see how to test with a connected Android phone using the emulator, maybe you can test that @geekygecko ? )

## Issue when not already connected to a google account

I am seeing an issue when the automotive device is not already signed into a Google account. The OS has you sign into a Google account, but when it returns to the app and shows the OS pop-up screen to select an account it does not show the just-signed-into google account. Closing that pop-up and tapping Continue with Google again fixes the issue though.

Here is a video of that issue.

https://user-images.githubusercontent.com/4656348/232623179-8c114176-f8b7-497c-bb29-5bbbb8db11cc.mov

## Checklist

### DIFF
--- a/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/AccountFragment.kt
+++ b/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/AccountFragment.kt
@@ -5,6 +5,8 @@ import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import androidx.compose.ui.res.dimensionResource
+import androidx.compose.ui.unit.sp
 import androidx.core.view.isVisible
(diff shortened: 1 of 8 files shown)


geekygecko (approved):
That works well! Thanks for making that change.

PR #893, opened 2023-04-20
Renaming Player interface in preparation for media3

mchowning:
This is renaming our `Player` interface to `PocketCastsPlayer` in order to differentiate it from the media3 `Player`. Since that creates a pretty big diff, I thought it would be nice to just separate that out in order to make the other PRs easier to review.

I also took this opportunity to remove the unused `PlaybackManager::isStreaming` method (it's not even used on `main`) in 600d3cb590d3c96a098709a26a5df38c16db7399.

Basically this PR is just an effort to reduce the diffs in future PRs. There are no substantive changes here.

## Testing Instructions
If the code changes look ok and CI is happy, then I think we're good.


mchowning:
The CI failures appear to be because the CI machines can't access the internet at the moment.

ashiagr (approved):
LGTM 👍

PR #895, opened 2023-04-20
Proposal: Rename Playable to Episode

mchowning:
I've found our usage of the `Playable` interface to be a bit confusing in a couple of ways.

First, currently the Playable interface is implemented by both `Episode` and `UserEpisode`. I think it's a bit confusing that a `UserEpisode` is not actually an `Episode`.

Second, "episode" feels like a more natural word, and I suspect that is why there are `Playable` variables that are named "episode" in [many places](https://github.com/Automattic/pocket-casts-android/search?q=%22episode%3A+Playable%22) in the project.

For that reason, I thought it might be nice to rename our `Playable` interface to be `Episode`, and to rename the `Episode` to be a `PodcastEpisode` (certainly open to better suggestions for that name). 

This is certainly not addressing any critical issues, it's just a quick change I thought might be a nice quality-of-life improvement for development. So if for any reason you think this isn't worthwhile, just let me know (particularly you, @geekygecko , since you know the history behind `Playable` much better than me). I'm fine closing this. 

> **Note**
> The diff on this PR is a just a little big, but the changes are almost entirely automated refactorings. The one notable exception that I think is worth taking a closer look at is that I migrated the "episodes" table to be named "podcast_episodes".

## Testing Instructions

1. Install a previous version of the app
2. Log into an account with some episodes on the. up next list and some podcast subscriptions.
4. Smoke test the app a bit to check for breakages.

## Checklist


mchowning:
CI is now green. Let me know whether you think this is worthwhile @ashiagr and @geekygecko .

ashiagr:
What do you think about just renaming `Playable` to `BaseEpisode`? That way, we'll be consistent with iOS:

 [public class Episode: NSObject, BaseEpisode](https://github.com/Automattic/pocket-casts-ios/blob/trunk/Modules/DataModel/Sources/PocketCastsDataModel/Public/Model/Episode.swift#L4)
[public class UserEpisode: NSObject, BaseEpisode](https://github.com/Automattic/pocket-casts-ios/blob/trunk/Modules/DataModel/Sources/PocketCastsDataModel/Public/Model/UserEpisode.swift#L3)

Just a thought, what you have is also fine.

ashiagr:
So, the final renaming will be as follows, right? 

`Playable` -> `BaseEpisode`
`Episode` -> `PodcastEpisode`

and no changes to `UserEpisode`

mchowning:
> So, the final renaming will be as follows, right?
> 
> `Playable` -> `BaseEpisode`
> `Episode` -> `PodcastEpisode`
> 
> and no changes to `UserEpisode`

That's right. I've updated this in [065ecb0](https://github.com/Automattic/pocket-casts-android/pull/895/commits/065ecb006d6116fec6fe1881f0fa09dd4e437e8a), so now the hierarchy looks like this:

image

ashiagr (approved):

PR #896, opened 2023-04-20
Opt-in to UnstableApis instead of creating UnstableApis

mchowning:
## Description
When using `@UnstableApi` when we use an unstable api, anything that uses the annotated method/class will now also get a lint warning. For that reason, I think we should switch to using the not-at-all-pleasant-to-type-or-remember `@androidx.annotation.OptIn(UnstableApi::class)`. Using that does not create lint warnings for callers of the annotated class/method. That's why I was actually able to remove a number of annotations entirely in this PR.

Alternatively, we could just add a rule to ignore this lint warning throughout the project. I'm not at all strongly opposed to that, but since I didn't feel strongly I figured I'd keep the lint check so we're at least aware when we use an unstable API in a new place. Like I said, I don't feel strongly though, so if you'd rather ignore the rule entirely just let me know and I'll make that change.

## Testing Instructions
If CI is happy then we should be good.

### DIFF
--- a/app/src/main/java/au/com/shiftyjelly/pocketcasts/PocketCastsApplication.kt
+++ b/app/src/main/java/au/com/shiftyjelly/pocketcasts/PocketCastsApplication.kt
@@ -264,7 +264,7 @@ class PocketCastsApplication : Application(), Configuration.Provider {
     }
 
     companion object {
-        @UnstableApi
+        @androidx.annotation.OptIn(UnstableApi::class)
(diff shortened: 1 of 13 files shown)


ashiagr:
Looks good! 

Let's keep a watch on the usage of `@androidx.annotation.OptIn(UnstableApi::class)` as we make changes. We should have covered most of the media classes that need it.  Currently, it is at 15, if it grows beyond 25 we can consider ignoring it using a lint rule.

PR #897, opened 2023-04-20
Remove bufferedUpToMs method

mchowning:
## Description
The `PocketCastsPlayer::bufferedUpToMs` method is basically just exposing the `Player::getBufferedPosition` method. Since `PocketCastsPlayer` now implements `Player`, I think we can drop that method. There are other methods that we can do this with, but I limited it to just the one method for this PR so you can see what it looks like without too much noise and let me know if you're on board with moving in this direction.

The benefits of this change are that it removes some code, and gives us a more consistent interface since the `PocketCastsPlayer` would only have one method (via `Player`) for obtaining the bufferedMs position.

The negative of this change is that our implementation of `PocketCastsPlayer::bufferedUpToMs` took care of making sure it was executed on the main thread. We lose that by relying on the (non-suspending) `Player::getBufferedPosition` method. Personally, I think this is ok because essentially the `PlaybackManager` is already our gateway to playback commands and it will just now be the point that is responsible for ensuring that calls to the player are made on the correct thread (and it can do it once for all the `PocketCastsPlayer` implementations instead of each implementation having to handle it separately). In an ideal world I would personally prefer to have the `Player` take care of that itself, but since we can't change that I think having the `PlaybackManager` handle it makes sense.

An alternative approach would be to no longer expose the `Player` interface from `PocketCastsPlayer` and to wrap all the `Player` methods we need in `PocketCastsPlayer`. This doesn't seem like an awful idea, but it will lead to a lot of code most likely and I'm not sure that the benefits of having the `Player` instead of the `PlaybackManager` manage the proper threading for its calls is worth it.

What do you think? If you like this direction, I'll do a follow-up PR removing some more of the now duplicate (imo) `PocketCastsPlayer` methods.

## Testing Instructions
1. Load up a podcast episode and jump to some different positions in the episode.

## Checklist

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
@@ -144,7 +144,7 @@ open class PlaybackManager @Inject constructor(
 
     private var updateCount = 0
     private var resettingPlayer = false
-    private var lastBufferedUpTo: Int = 0
+    private var lastBufferedUpTo: Long = 0
     private var focusWasPlaying: Date? = null
     private var forcePlayerSwitch = false
     private var updateTimerDisposable: Disposable? = null
@@ -316,10 +316,6 @@ open class PlaybackManager @Inject constructor(
         }
 
         return episode.playedUpToMs
-    }
-
-    suspend fun getBufferedUpToMs(): Int {
-        return pocketCastsPlayer?.bufferedUpToMs() ?: 0
     }
 
     fun isPlaybackLocal(): Boolean {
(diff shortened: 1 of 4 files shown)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
```diff
@@ -916,16 +912,16 @@ open class PlaybackManager @Inject constructor(
     suspend fun onBufferingStateChanged() {
         pocketCastsPlayer?.let {
             val isBuffering = it.isBuffering()
-            val bufferedMs = it.bufferedUpToMs()
             withContext(Dispatchers.Main) {
+                val bufferedMs = it.bufferedPosition
```
mchowning, line 916: When the `PlaybackManager` gets the buffered position from the player, we now have to make sure we're on the main thread.

mchowning, line 916: When the `PlaybackManager` gets the buffered position from the player, we now have to make sure we're on the main thread.

ashiagr (approved):
Your changes and explanation make sense to me. 

>  I'll do a follow-up PR removing some more of the now duplicate (imo) PocketCastsPlayer methods.

Sounds good.

PR #898, opened 2023-04-21
Remove bufferedPercentage and handleCurrentPositionMs methods

mchowning:
## Description
Just trying to simplify a bit more by relying directly on the `Player` interface. Again, there's more that could be done here, but I want to keep these PRs small.

## Testing Instructions
Play a podcast and skip around a bit in the episode.

## Checklist


ashiagr (approved):
LGTM 👍

PR #900, opened 2023-04-22
Fix downloading cloud files

ecronin:
3787dbd introduced AccessToken, missed a conversion to token.value

## Description
The latest Android beta can't download cloud files, looking in the debug log the requests being made are:

```
E 21/4 19:03:43 Invalid response returned for episode download. 400  hxxps://api.pocketcasts.com/files/url/62743f70-e11c-11ed-8408-2f2a325ada1f?token=AccessToken(value=eyJ0eX...)
```

when they should be:
```
hxxps://api.pocketcasts.com/files/url/62743f70-e11c-11ed-8408-2f2a325ada1f?token=eyJ0eX...
```

Fixes # 

## Testing Instructions
Use a Plus account with files uploaded to the cloud but not on the local device, try downloading or streaming, confirm that you're no longer getting HTTP/400 responses

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/sync/SyncServerManager.kt
+++ b/modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/sync/SyncServerManager.kt
@@ -228,7 +228,7 @@ open class SyncServerManager @Inject constructor(
         server.deleteFile(addBearer(token), episode.uuid)
 
     fun getPlaybackUrl(episode: UserEpisode, token: AccessToken): Single<String> =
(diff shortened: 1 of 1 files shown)


ecronin:
Found https://www.reddit.com/r/pocketcasts/comments/12tuzqf/cloud_files/ but no open issue in github for this problem

CookieyedCodes:
Ya it will probably be fixed next beta, it is the weekend at least it's in the beta 😉

ashiagr (approved):
Tested, and it works well now. Thank you so much for this PR, @ecronin 🙏

PR #901, opened 2023-04-22
Fix shownote's timestamp not getting parsed when there's a link

### LINKED ISSUE #814: matchIsInUrl catches non url cases
### Description

matchIsInUrl returns true even if it is not in a URL, this causes timeskips to not be parsed correctly

https://github.com/Automattic/pocket-casts-android/blob/0795b5017942d4a31e6805f2c01d54ef52cefb5e/modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/helper/ShowNotesHelper.kt#L17-L20

```html
<p>Example 00:00 <a href="https://example.com">Link</a></p>
```

In the example above the matcher will start at the `00:00` and scan for any pattern like an end tag, despite not being in an `a` tag the first endtag is the `a` endtag so the `matchIsInUrl` function will return true.

_No response_

_No response_

rodrigoborgesdeoliveira:
## Description
"matchIsInUrl" was returning true even when the timestamp was outside of the URL.

- Fixes #814

## Testing Instructions

1. Pass an html like `<p>Example 00:00 <a href="https://example.com">Link</a></p>` to the `convertTimesToLinks` function.
2. It should return `<p>Example <a href="http://localhost/#playerJumpTo=00:00">00:00</a> <a href="https://example.com">Link</a></p>`.

## Checklist

### DIFF
--- a/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/views/helper/ShowNotesHelperTest.kt
+++ b/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/views/helper/ShowNotesHelperTest.kt
@@ -36,5 +36,30 @@ class ShowNotesHelperTest {
         actual = ShowNotesHelper.convertTimesToLinks("<li><a href=\"https://www.theverge.com/2021/12/21/22848957/lg-dualup-32-inch-4k-ultra-fine-monitors-announced-specs\">LG’s new 16:18 monitor looks like a multitasking powerhouse</a></li>")
         expected = "<li><a href=\"https://www.theverge.com/2021/12/21/22848957/lg-dualup-32-inch-4k-ultra-fine-monitors-announced-specs\">LG’s new 16:18 monitor looks like a multitasking powerhouse</a></li>"
         assertEquals(actual, expected)
+
+        actual = ShowNotesHelper.convertTimesToLinks("<p>Example <a>10:30</a></p>")
+        expected = "<p>Example <a>10:30</a></p>"
+        assertEquals(expected, actual)
+
+        // Replace the timestamp if it is outside the <a> tag - https://github.com/Automattic/pocket-casts-android/issues/814
+        actual = ShowNotesHelper.convertTimesToLinks("<p>Example 00:00 <a href=\"https://example.com\">Link</a></p>")
+        expected = "<p>Example <a href=\"http://localhost/#playerJumpTo=00:00\">00:00</a> <a href=\"https://example.com\">Link</a></p>"
+        assertEquals(expected, actual)
+
+        actual = ShowNotesHelper.convertTimesToLinks("<p>Example 00:00 <a>Link</a></p>")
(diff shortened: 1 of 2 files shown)


app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/views/helper/ShowNotesHelperTest.kt
```diff
@@ -36,5 +36,30 @@ class ShowNotesHelperTest {
         actual = ShowNotesHelper.convertTimesToLinks("<li><a href=\"https://www.theverge.com/2021/12/21/22848957/lg-dualup-32-inch-4k-ultra-fine-monitors-announced-specs\">LG’s new 16:18 monitor looks like a multitasking powerhouse</a></li>")
         expected = "<li><a href=\"https://www.theverge.com/2021/12/21/22848957/lg-dualup-32-inch-4k-ultra-fine-monitors-announced-specs\">LG’s new 16:18 monitor looks like a multitasking powerhouse</a></li>"
         assertEquals(actual, expected)
+
+        actual = ShowNotesHelper.convertTimesToLinks("<p>Example <a>10:30</a></p>")
+        expected = "<p>Example <a>10:30</a></p>"
+        assertEquals(expected, actual)
```
rodrigoborgesdeoliveira, line 42: The other tests have the `assertEquals` with parameters as `(actual, expected)`, but according to the docs, the first parameter should be `expected` and the second one should be `actual`.

image

Should I keep it inverted like the others? i.e. `assertEquals(actual, expected)`.

rodrigoborgesdeoliveira:
In the changelog, how should I precede with the header? The latest one is `7.38`. Should I add a `7.39`? `Unreleased`? Something else?

app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/views/helper/ShowNotesHelperTest.kt
ashiagr: What you have is good. For consistency, would you mind adjusting others in this class to conform with the documented guidelines? i.e `assertEquals(actual, expected)`

ashiagr:
This is working well in my testing. Thank you for this contribution and for supporting it with tests, @rodrigoborgesdeoliveira!

> In the changelog, how should I proceed with the header? The latest one is 7.38. Should I add a 7.39? Unreleased? Something else?

Since 7.38 is unreleased, you can add a change log entry to the 7.38 section.

ashiagr (approved):

app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/views/helper/ShowNotesHelperTest.kt
rodrigoborgesdeoliveira: Updated.

PR #907, opened 2023-04-25
Hide custom skip buttons for an Automotive manufacturer

geekygecko:
## Description
After testing, a manufacturer using our Automotive build needs us to hide our custom skip buttons and instead say we respond to the playback events `ACTION_SKIP_TO_PREVIOUS` and `ACTION_SKIP_TO_NEXT`.

## Testing Instructions
1. Add "google" to the list of manufacturers `MANUFACTURERS_TO_HIDE_CUSTOM_SKIP_BUTTONS`
2. Deploy to the Google automotive emulator
3. Play an episode
4. ✅ Verify the skip buttons don't appear

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/MediaSessionManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/MediaSessionManager.kt
@@ -66,6 +66,8 @@ class MediaSessionManager(
     companion object {
         const val EXTRA_TRANSIENT = "pocketcasts_transient_loss"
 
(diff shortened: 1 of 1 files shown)


mchowning (approved):
This looks good 👍 

I added [3d6e652](https://github.com/Automattic/pocket-casts-android/pull/907/commits/3d6e6523b4563953bd41e93762597166843527f5) to this PR to move the comment about why we're not using our buttons on Samsung to the new place where we're specifying Samsung since that seems like the better place for the comment now.

Also, wanted to note that it is only the appearance of the buttons that changes—the buttons still skip forward/backward by the relevant number of seconds instead of skipping forward/back an entire track.

PR #909, opened 2023-04-26
Update watch's Up Next UI

mchowning:
## Description
This PR updates the styling of the Up Next screen on the watch.

One trickiness I ran into here was that it does not seem possible to use horologist's scrollable components on multiple pages of a pager when the pager is inside a single navigation context (more info [here](https://github.com/google/horologist/issues/1181#issuecomment-1522555773)). I'm working around that by just hiding the time text from the Up Next screen and using a scrollable composable from foundation there. That way the time text can still be shown on the pager's "main" screen.

> **Note**
> This PR builds on https://github.com/Automattic/pocket-casts-android/pull/908, so you may want to review that first.

## Testing Instructions
1. Load up the app while logged into an account with episodes in the Up Next queue
2. Swipe to the third page in the pager from any screen where the pager is available
3. Tap on one of the episodes in the up next queue that isn't downloaded
4. From the Episode screen download the episode
5. Return to the up next screen and observe that once the download is complete the download icon is displayed on the episode chip.

Also check that the artwork from user-uploaded files are displayed properly in the Up Next queue. Don't tap on a user-downloaded epsiosde in the queue though, that is crashing currently and I will address that in a separate PR.

Also check the Up Next queue UI when there are no items in the queue.

## Screenshots or Screencast 

https://user-images.githubusercontent.com/4656348/234433386-12332aea-1cdc-4f70-af88-669fa0837982.mov

## Checklist


ashiagr (approved):
Looks good!

PR #910, opened 2023-04-26
Update podcast_episodes table name in hardcoded queries

mchowning:
## Description
When I [updated the table name from `episodes` to `podcast episodes`](https://github.com/Automattic/pocket-casts-android/pull/895), I didn't think to check for hardcoded queries. Turns out I should have. This fixes those.

One place this showed up was that the app would include some database errors in the logs when creating the logs file for support.

## Checklist


ashiagr (approved):
Kudos for catching it while it was unreleased!

PR #912, opened 2023-04-26
Advanced settings screen updates

mchowning:
## Description
* Adds changelog entries for https://github.com/Automattic/pocket-casts-android/issues/858
* Adds a bit of descriptive text to the top of the Advanced settings. My thinking is that we want to caution users a bit about playing with these settings since they can create a worse user experience.
* Removes the indent from the advanced settings to be more like the privacy screen. I think we want to move away from using that on screens like this where we're not using that indented space for any icons. (The privacy settings screen also doesn't have the indent).
* Moves Advanced settings to be above About and Privacy on the Settings screen

## Testing Instructions
* Make sure that the padding on the "General" → "Media notification controls" screen still looks good.
* Check the that the Advanced settings screen looks good.
* Check that the text on the top of the Advanced settings screen sounds right.
* Confirm that the new position for Advanced settings in the settings list (i.e., it's not on the bottom anymore) makes sense.

## Screenshots or Screencast 

image

## Checklist
 
- [X] with different themes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 7.38
 -----
+*   Updates:
+    *   Added Advanced Settings section for experimental settings
+        ([#885](https://github.com/Automattic/pocket-casts-android/pull/885/)).
+    *   Added setting to disable sync on metered networks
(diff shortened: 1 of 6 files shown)


ashiagr (approved):
Looks good!

PR #913, opened 2023-04-26
Use fade instead of ellipsis for email overflow

mchowning:
## Description
The logging in screen on the watch used an ellipsis to indicate overflow before. This updates that to use a gradual fade out only when the email is too long.

I was having some trouble getting this to work right, but I got some great help [in Kotlin's Slack](https://kotlinlang.slack.com/archives/C04TPPEQKEJ/p1682438604723239).

Internal ref: p1682125454156639-slack-C040S9BK7SP

## Testing Instructions
1. While logged in on your phone but not logged into the watch, start the watch app
2. Observe that the "Logging in" screen shows a fade out gradient when the user's email is too long (I manually edited [this line](https://github.com/Automattic/pocket-casts-android/blob/0da3d947fac305fe9c217310e8329fd674be2630/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/LoggingInScreen.kt#L98) to create the email lengths that I needed for testing).

## Screenshots or Screencast 
| Not overflowing | Overflowing |
| --- | --- |
| Screenshot 2023-04-26 at 2 19 54 PM | Screenshot 2023-04-26 at 2 22 23 PM |

## Checklist
 
- [ ] with different themes

### DIFF
--- a/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/LoggingInScreen.kt
+++ b/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/LoggingInScreen.kt
@@ -101,29 +101,38 @@ fun LoggingInScreen(
                     // Without this if "xxxx@gmail.com" ran just a bit long, it would get shortened
                     // to "xxx@gmail".
                     softWrap = false,
-                    modifier = Modifier
-                        .drawWithContent {
-                            drawContent()
-
-                            // Use < instead of <= because text that doesn't fit will have a width that is equal
-                            // to the available space
-                            val textFits = size.width < widthPx
-
-                            if (!textFits) {
-                                // Apply a fade out gradient to the end of the text to indicate overflow
-                                drawRect(
-                                    // Add a bit of extra width so the gradient goes to the end of the screen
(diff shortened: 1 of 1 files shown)


wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/LoggingInScreen.kt
```diff
@@ -80,17 +85,57 @@ fun LoggingInScreen(
...
+    overFlowWidthPx: Int,
+    backgroundColor: Color,
+) = drawWithContent {
+
+    // draw text first
+    drawContent()
+
+    // Use < instead of <= because text that doesn't fit will have a width that is equal
+    // to the available space
+    val textFits = size.width < overFlowWidthPx
```
mchowning, line 124: This does mean that if the text _exactly_ fits on the screen, the overflow gradient will get applied. It seems that no matter how long the text is it will not report a `size.width` that is larger than the `BoxWithConstraint`'s width, so basically if they are equal then I'm assuming overflow. That's correct most of the time, but it does create an edge case for text that is extremely close to overflowing.

ashiagr (approved):
Looking really nice!

PR #914, opened 2023-04-26
Add up next icon to episode chip

mchowning:
## Description
This adds an up-next icon to the `EpisodeChip`.

In addition, this updates the episode chip to vertically center the podcast image when there are no icons present. Unfortunately, you can't test this currently because the only screens using this chip right now are the Up Next screen and the Downloads screen and I certainly hope that any podcast on one of those screens should either be in the Up Next or downloaded. 😄 I've included a screenshot below though so you can see what it would look like.

> **Note**
> This PR builds on https://github.com/Automattic/pocket-casts-android/pull/909.

## Testing Instructions
1. Go to the Up Next queue or the Downloads page
2. Observe that the up-next icon and download icons are presented when appropriate

## Screenshots or Screencast 

| Both icons | Up Next only | Downloaded only | No icons |
| --- | --- | --- | --- |
| image | image | image | image |

## Checklist


ashiagr:
Looks good, just that up-next icon appears a little bit higher 

App



Figma 

Screenshot 2023-04-27 at 1 22 31 PM

mchowning:
> up-next icon appears a little bit higher

Thanks for calling that out. I'm not sure why the up next icon in the project isn't centered vertically, so I updated it from FIgma. The new icon uses a slightly thicker strokes and generally has a slightly different style, so I also updated the download icon so they both have the same style ([a894914](https://github.com/Automattic/pocket-casts-android/pull/914/commits/a894914b4c76bf7d3c9a57de7149e2b237b4673e)). Note that this change also updates the icons in the phone app.

| Before | Now |
| --- | --- |
| image | image |

PR #915, opened 2023-04-26
Handle user episodes

mchowning:
## Description
This updates the watch app to handle user episodes. The first part of that is to update the `EpisodeScreen` so it can handle user episodes. The second part is implementing the FIles screeen.

> **Note**
> This PR builds on https://github.com/Automattic/pocket-casts-android/pull/909.

### Empty File Screen State
1. Open the app with an account that does not have any user episodes
2. Tap on files
3. Observe the UI informs you that you have no files

### Non-Empty FIle Screen State
1. Open the app with an account that has some files uploaded to the cloud
2. Tap on files
3. Observe the UI lists all your files
4. Tap on one of the files to open the episode screen
5. Confirm the UI of the episode screen looks fine (note that files only have the "Mark played" button and there are no show notes for user episodes). 
6. Observe that the episode screen controls are tinted based on the appropriate "color" for the episode. For podcast episodes, this comes back from our server (this logic should be unchanged in this PR). For user episodes, if the user has set a color for the episode, that color is used; if the user has set custom artwork for the episode, we're using the [palette api](https://developer.android.com/develop/ui/views/graphics/palette-colors) to extract a light vibrant color from the artwork.
7. Tap the download button to download the episode
8. Tap the play button to play the episode

## Screenshots or Screencast 

| No Files | Files list | Episode screen for file |
| --- | --- | --- |
| Screenshot 2023-04-26 at 5 17 06 PM | Screenshot 2023-04-26 at 5 17 34 PM | Screenshot 2023-04-26 at 5 34 43 PM |

## Checklist

### DIFF
--- a/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/cloud/AddFileActivity.kt
+++ b/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/cloud/AddFileActivity.kt
@@ -104,6 +104,29 @@ class AddFileActivity :
             intent.putExtra(EXTRA_EXISTING_EPISODE_UUID, fileUuid)
             return intent
         }
+
+        fun darkThemeColors() = listOf(
+            AddFileColourAdapter.Item.Colour(1, ThemeColor.primaryText02Dark, false),
+            AddFileColourAdapter.Item.Colour(2, ThemeColor.filter01Dark, true),
+            AddFileColourAdapter.Item.Colour(3, ThemeColor.filter05Dark, true),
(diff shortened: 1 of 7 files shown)


ashiagr (approved):

mchowning:
> I see a delay after pressing the play icon for a file before it changes to pause. The change is immediate for a podcast episode.

Oh nice catch! I think this issue comes from somewhere in the `PlaybackManager` because the `EpisodeScreen` is not doing any different handling for UserEpisodes compared to PodcastEpisodes.

I tried immediately updating the UI when taping the button to make it at least feel more responsive, but this caused the button to flicker back and forth between play and pause. 

I have filed https://github.com/Automattic/pocket-casts-android/issues/917 to track this.

PR #920, opened 2023-04-27
Hide up next icons on up next screen

mchowning:
## Description
This hides the up next icons on the up next screen on the watch since the icon would be on every episode on that screen. This is also consistent with the phone's behavior.

internal discussion: p1682536583755259-slack-C040S9BK7SP

## Testing Instructions
1. Go to the Up Next screen
2. ✅ Confirm that none of the episodes show the Up Next icon
3. Download one of the episodes in the Up Next queue
4. Go to the Downloads screen
5. ✅  Confirm that the episode you just downloaded is there with both the up-next icon and the downloaded icon.

## Screenshots or Screencast 
| Up Next Screen | Downloads Screen|
| --- | --- |
| image | image |

## Checklist


ashiagr (approved):
Looks good

PR #923, opened 2023-05-01
Patron: Upgrade screen UI adjustments

ashiagr:
## Description

This updates the new upgrade screen to keep the upgrade button position fixed (pdeCcb-2vS-p2#comment-1989), updates the pager height to the tallest page, and fixes jumping of the pager due to the variable header at the top in large font sizes.

## Testing Instructions
1. Apply [this patch](https://github.com/Automattic/pocket-casts-android/files/11343238/upgrade-testing-patch-with-patron.patch)
2. Increase device font size
3. Run the app being logged out
4. Subscribe to any podcast
5. Go to Podcasts
6. Tap the folder icon
7. Notice that 
    - the upgrade screen is scrollable with the upgrade button fixed
    - switching between the pages does not create a jumping effect
    - page sizes are the same

## Screenshots or Screencast 

https://user-images.githubusercontent.com/1405144/235459006-38300d60-3dbb-4d1a-aaab-3eac42c87e95.mp4

## Checklist
 
- [x] with different themes


mchowning (approved):
This is looking good. Great work!

The only issue I see is that with a big enough font size, the subscribe button's text can expand to two lines. This reveals two issues: the text isn't center aligned and the button can now cover some of the cards even when you scroll all the way to the bottom. I'd probably either not scroll the cards behind the button (so the scrollable column would be above the button), or maybe use your fancy `AutoResizeText` component to keep the button text to one line. 


<details>
<summary>Video</summary>

https://user-images.githubusercontent.com/4656348/235465108-76480474-6e8a-4dd0-a089-21418ae56093.mov

</details>

ashiagr:
> The only issue I see is that with a big enough font size, the subscribe button's text can expand to two lines

I updated the upgrade button title to fit one line in e1606a6. Thanks for noticing!

PR #928, opened 2023-05-03
Update sign_in and create_account events

mchowning:
## Description
This updates the sign-in and create-account events. I think that the SSO changes caused these to stop working the way they should.

### 1. Create account with SSO
1. Fresh install of the app
2. Create an account using SSO
3. Observe the `user_account_created` event
4. Sign out
5. Sign in using SSO
6. Observe the `user_signed_in` event

### 2. Create account with email/pass
1. Delete the account you created in the previous test
2. Create an account using that email, but with a password instead of using SSO
3. Observe the `user_account_created` event
4. Sign out
5. Sign in using email/pass
6. Observe the `user_signed_in` event
7. Sign out
8. Sign in using SSO
9. Observe the `user_signed_in` event

## Checklist


ashiagr (approved):
Looks good 👍

Issue #927, opened 2023-05-03
`user_signed_out` event fired each time app starts if user isn't logged in


mchowning:
### Description

Currently, we fire a `user_signed_out` event each time the app starts because [we sign the user out](https://github.com/Automattic/pocket-casts-android/blob/7e2c82d8e55a74527887cdab1ff7648ded832c55/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/user/UserManager.kt#L54) if the `AccountManager` reports that the user is not logged in. We do this because the user can remove their Pocket Casts account view Android's "Manage Accounts" settings, and we need to do additional cleanup within the app if that happens.

### Step-by-step reproduction instructions

1. Start up a fresh install of the app
2. Observe that the logs include "Tracked: user_signed_out..."
3. Swipe the app away from recents
4. Start the app again
5. Again, observe that the logs include "Tracked: user_signed_out..."

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

Samsung A13, Android 13, Pocket Casts 7.38 pre-release build

mchowning:
It seems like we do need to do this check because if we don't the app ends up in a weird state. The way to address this is probably to not only check if the user is signed out, but to also check for some state that shouldn't exist if the user is fully signed out but does exist if the user was signed out from outside the app. Only in that case would we do the automatic sign out of the user.

mchowning:
I spent some time thinking about this, and the best I've come up with so far is to only send the event if `synManager.loggedIn()` was `true` before the sign out. This works pretty well, but it does mean that this event won't get sent if a user signs out of their Pocket Casts account from the Android settings, so it's not perfect. 🤔

PR #931, opened 2023-05-04
Only send sign out analytics when user was signed in

mchowning:
> **Warning**
> I am targetting the `release/7.37.1` hotfix because this seems like a completely safe change and it helps our analytics be more useful in identifying possible authentication issues. This is certainly not a critical fix, so just let me know if you think we're better off not including it in the hotfix.

## Description

This improves the situation in #927, but it does not fully fix the issue because although we no longer send the sign-out analytics event every time a logged-out user starts the app, we now can miss sending the sign-out analytics event when the user signs out of their Pocket Casts account from the account management options in the Android OS. I think this is probably not a very common thing for people to do, so this feels like a worthwhile tradeoff, but it would be good to find a better way to handle this in the future that also captures those events (let me know if you have any ideas).

## Testing Instructions
1. Freshly install the app
2. Start the app
3. ✅  Observe that there is not a `user_signed_out` event

## Checklist

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/user/UserManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/user/UserManager.kt
@@ -91,6 +91,8 @@ class UserManagerImpl @Inject constructor(
 
     @OptIn(DelicateCoroutinesApi::class)
     override fun signOut(playbackManager: PlaybackManager, wasInitiatedByUser: Boolean) {
+        val wasSignedIn = syncManager.isLoggedIn()
+
         LogBuffer.i(LogBuffer.TAG_BACKGROUND_TASKS, "Signing out")
         subscriptionManager.clearCachedStatus()
         syncManager.signOut {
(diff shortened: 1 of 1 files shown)


ashiagr:
>  we now can miss sending the sign-out analytics event when the user signs out of their Pocket Casts account from the account management options in the Android OS. I think this is probably not a very common thing for people to do, so this feels like a worthwhile tradeoff

We can try to improve it by setting a last login boolean in preference on login. [`wasSignedIn`](https://github.com/Automattic/pocket-casts-android/pull/931/commits/832406107bfef0e5e7a23e9e974b8ac8ee49efae#diff-cafdaad2a8d667fa6007b0ff9a204594899bb58723ba7c8e1c9378a6558c2eeeR94) can be set from this boolean and reset on sign out. If this last login boolean is combined with `syncManager.isLoggedIn()`, we can also know if the sign-out was done out of their Pocket Casts account from the account management options in the Android OS.

prefs.isLoggedIn + syncManager.isLoggedIn() -> signing out from the app
prefs.isLoggedIn + !syncManager.isLoggedIn() -> already signed out outside of Pocket Casts app from the account management options in the Android OS

For the hotfix, I agree, it is a worthwhile tradeoff. Great investigation!

ashiagr (approved):

mchowning:
> We can try to improve it by setting a last login boolean in preference on login.

I like this idea. I'll play with it today and try to open a PR (targeting `main`) assuming you're not already looking into this.

PR #932, opened 2023-05-04
Update sign-in event handling for 7.37.1

mchowning:
> **Warning**
> This PR is bringing some of the analytics changes from https://github.com/Automattic/pocket-casts-android/pull/928 to the 7.37.1 hotfix. I believe these changes are safe for the hotfix because they only update how the analytics are sent, so there is very little risk of any regressions our users would notice. It is certainly not critical that this be included in the hotfix though, so let me know if you have any hesitation about including it.

See https://github.com/Automattic/pocket-casts-android/pull/928 for description and testing steps. That PR has two commits, one of which involved a minor refactoring, which I have not included in this PR in order to minimize the hotfix changes.

The reason I think this is worth including this in the hotfix is because with our SSO changes we now call `downloadTokens` in many circumstances where a user is not signing in, and removing this noise from our analytics will help us identify potential authentication issues.

## Checklist

### DIFF
--- a/modules/services/analytics/src/main/java/au/com/shiftyjelly/pocketcasts/analytics/AnalyticsEvent.kt
+++ b/modules/services/analytics/src/main/java/au/com/shiftyjelly/pocketcasts/analytics/AnalyticsEvent.kt
@@ -10,7 +10,9 @@ enum class AnalyticsEvent(val key: String) {
 
     /* User lifecycle events */
     USER_SIGNED_IN("user_signed_in"),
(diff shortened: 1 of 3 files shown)


mchowning:
There have been enough underlying changes between 7.37 and the `main` branch that bringing this code over, is not as straightforward as I expected. As a result, I don't think this makes sense to include in a hotfix, and we're probably just better off waiting for this code to make it to prod from `main`.

PR #937, opened 2023-05-05
Only sign out user if they have not already been fully signed out

### LINKED ISSUE #927: `user_signed_out` event fired each time app starts if user isn't logged in

mchowning:
## Description
Follow-up PR to https://github.com/Automattic/pocket-casts-android/pull/931 implementing @ashiagr 's suggested improvement. 🙇 

I'm taking this a bit further and avoiding all sign out work when the user is already fully signed out unless this signout was manually triggered by the user. In theory, it shouldn't be possible for a fully signed out user to manually trigger a sign out, but I figured I'd leave that in as a safety hatch so if a user did somehow get into a weird state they could still possibly recover by manually triggering a sign out.

Fixes #927

## Testing Instructions
1. Fresh install of the app
2. Start the app
3. ✅  Verify there is no `user_signed_out` event
4. Sign into the app
5. Swipe the app away from recents
6. Restart the app
7.  ✅  Verify there is no `user_signed_out` event
8. Sign out of the app from within the app
9. ✅  Verify there IS a `user_signed_out` event
10. Sign back into the app
11. Swipe the app away from recents
12. Sign out of your Pocket Casts account from the Android OS's settings for managing your accounts
13. ✅ Verify there IS a `user_signed_out` event. I see this event in the logs without even needing to open the app, but I think it's also fine if the event doesn't show up until the app is opened.

## Checklist

### DIFF
--- a/modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/Settings.kt
+++ b/modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/Settings.kt
@@ -613,4 +613,12 @@ interface Settings {
 
     fun setLastSelectedSubscriptionFrequency(frequency: SubscriptionFrequency)
     fun getLastSelectedSubscriptionFrequency(): SubscriptionFrequency?
+
+    // This boolean should be update to false when a user signs in and should be set to
+    // true once a user signs out and that sign out has been fully handled
+    // by the app. This field helps make sure the app fully handles signing a user
+    // out even if they sign out from outside the app (i.e., from the Android OS's
+    // account management settings).
+    fun setFullySignedOut(boolean: Boolean)
+    fun getFullySignedOut(): Boolean
 }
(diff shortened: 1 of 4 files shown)


ashiagr (approved):
Looks good 👍

Issue #359, opened 2022-10-03
Reduce frequency of discover_list_impression tracking


ashiagr:
`discover_list_episode_play` event gets fired multiple times when it shouldn't be. 

## Example 1
This is what happens on tapping the play button on a single episode feature in the discover feed, without any scrolling interaction:

```
11:08:34.741 🔵 Tracked: discover_list_episode_play
11:08:34.853 🔵 Tracked: discover_list_impression
11:08:34.860 🔵 Tracked: discover_list_impression
11:08:34.885 🔵 Tracked: discover_list_impression
11:08:34.899 🔵 Tracked: discover_list_impression
11:08:34.906 🔵 Tracked: discover_list_impression
11:08:34.911 🔵 Tracked: discover_list_impression
11:08:34.967 🔵 Tracked: playback_play, Properties: {"source":"discover",...}
```

## Example 2 

After opening up one of the featured podcasts from the discover screen, tapping the subscribe button fires off the following events even though the discover screen isn't visible either before or immediately after tapping the button:

```
11:13:33.413  🔵 Tracked: podcast_screen_toggle_summary
11:13:33.418  🔵 Tracked: podcast_screen_subscribe_tapped
11:13:33.580  🔵 Tracked: discover_list_impression
11:13:33.586  🔵 Tracked: discover_list_impression
11:13:33.592  🔵 Tracked: discover_list_impression
11:13:33.599  🔵 Tracked: discover_list_impression
```

Noticed in https://github.com/Automattic/pocket-casts-android/pull/345#discussion_r984697055

Note that 

- This is how we were handling the event for Firebase. 
- Only the unique impressions is the main value shown to the user. (p1664575119091089-slack-C02ATC80MUM)

mchowning:
In doing some testing just now I noticed extra events when I:

1. Went to the discover screen
2. Subscribed to a podcast in the trending list by tapping the plus symbol
3. Went to the podcast's tab
4. Tapped on the podcast I just subscribed to
5. Tapped on the play icon next to one of the episodes
6. 🔴  Two `discover_list_impression` events were fired

mchowning:
> I just got extra events when I was just playing episodes from a podcast that I had added via the discover tab even though I accessed the podcast via the Podcasts tab...

I get even more extra events if I scroll a bit on the discover page first. After that, taping play and pause both fire off a few `discover_list_impression` events.

mebarbosa:
[...]

PR #945, opened 2023-05-09
Only rebind discover views when there are changes

mchowning:
## Description
This filters the podcast lists used for the discover views so that the list is resent only when changes occur. This avoids rebinding the views unnecessarily and appears to significantly improve the situation reported in https://github.com/Automattic/pocket-casts-android/issues/359  where we were sending off too many `discover_list_impression` events.

"Example 1" in that issue is _not_ entirely fixed by the change in this PR because we still send off an extra `discover_list_impression` event for the specific list episode item that is being played. We don't send off any events for any other of the list items though, so even that example is significantly improved.

## Testing Instructions

In addition to the specific tests below, also do some smoke testing of the Discover feed. In particular, try subscribing to some podcasts and playing episodes from the discover feed and make sure the UI updates appropriately.

### 1. Subscribing/Unsubscribing from featured podcast
1. Open one of the featured podcasts from the discover screen
2. Tap the subscribe button
3. ✅  Observe that no `discover_list_impression` events are sent
4. Unsubscribe from the podcast
5. ✅  Observe that no `discover_list_impression` events are sent

### 2. Playing/Pausing an episode
1. Go to the discover screen
2. Observe some `discover_list_impression` events are sent
3. Go to the Podcasts tab (or any tab other than the Discover tab)
4. Tap play on the miniplayer
5. ✅ Observe that no `discover_list_impression` events are sent
6. Tap pause on the miniplayer
7. ✅ Observe that no `discover_list_impression` events are sent

## Screenshots or Screencast

https://user-images.githubusercontent.com/4656348/236903060-05d16e74-2c01-410d-ab6d-2147a887ed09.mov

## Checklist

### DIFF
--- a/modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/view/DiscoverAdapter.kt
+++ b/modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/view/DiscoverAdapter.kt
@@ -92,7 +92,7 @@ internal class DiscoverAdapter(
     val staticServerManager: StaticServerManagerImpl,
     val listener: Listener,
     val theme: Theme,
-    val loadPodcastList: (String) -> Flowable<PodcastList>,
+    loadPodcastList: (String) -> Flowable<PodcastList>,
     val loadCarouselSponsoredPodcastList: (List<SponsoredPodcast>) -> Flowable<List<CarouselSponsoredPodcast>>,
     private val analyticsTracker: AnalyticsTrackerWrapper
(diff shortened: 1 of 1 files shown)


ashiagr (approved):
LGTM 👍 It reduces the frequency of the event when it was clearly not needed.

I just want to remind us that there is an existing limitation when the event is not fired on returning to the discover tab from different tabs or from a nested screen https://github.com/Automattic/pocket-casts-android/pull/776#discussion_r1110245420. This, combined with the decreased frequency of the event in this PR, may lead to the event being triggered less often compared to iOS. Also, only unique impressions are captured (p1664575119091089-slack-C02ATC80MUM), so there might not be any effect, but let's keep a watch.

PR #947, opened 2023-05-10
Remove the web page QR code from Automotive

geekygecko:
## Description

The Automotive partner that required us to show the QR code no longer needs this. I thought about leaving the logic in the app but I haven't heard of another manufacturer that requires it so thought it would be cleaner and less maintenance to remove it. The emulator doesn't have a browser included, but it shows its own QR code fragment so I assume other manufacturers handle a URL intent in a similar way if they have to.

## Testing Instructions
1. Install the Automotive build
2. Tap the settings cog
3. Tap 'About'
4. Tap 'Terms of service'
5. ✅ Verify it opens the built-in QR code viewer

## Screenshot
Screenshot_20230510_143929


ashiagr (approved):
Looks good, thanks for thinking about maintainability!

PR #949, opened 2023-05-10
Patron: Update account details header

ashiagr:
## Description

This updates "Account Details" screen profile header to include a subscription badge.

#### Free account

1. Enable feature flag for Patron
2. Install release build
3. Login with an account without any plan 
4. Go to Profile -> Account screen 
5. Notice there's no subscription badge

#### Plus - Gift 

1. Logout 
2. Login with an account with a Plus - Gift account
3. Go to Profile -> Account screen 
4. Notice there's a Plus subscription badge

#### Plus - Subscription

1. Logout 
3. Login with a license test account with a Plus - Gift expiring account
4. Go to Profile -> Account screen 
5. Notice there's a Plus subscription badge
6. Buy Plus plan using upgrade banner
7. Notice there's a Plus subscription badge after the plan is bought

#### Patron  - Subscription

I tested it by setting `isSignedInAsPatron` to true in `SignInState`
```
val isSignedInAsPatron: Boolean
        get() = true
```
You can skip testing this part until the API is functional to support Patron.

## Screenshots or Screencast 
out

## Checklist
 
- [x] with different themes

### DIFF
--- a/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/components/SubscriptionTierPill.kt
+++ b/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/components/SubscriptionTierPill.kt
@@ -20,8 +20,7 @@ import androidx.compose.ui.semantics.semantics
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import au.com.shiftyjelly.pocketcasts.compose.components.TextH50
-import au.com.shiftyjelly.pocketcasts.images.R as IR
-import au.com.shiftyjelly.pocketcasts.localization.R as LR
+import au.com.shiftyjelly.pocketcasts.images.R
 
 private val iconSizeInDp = 14.dp
 private val pillCornerRadiusInDp = 800.dp
@@ -40,6 +39,7 @@ fun SubscriptionTierPill(
     Card(
         shape = RoundedCornerShape(pillCornerRadiusInDp),
         backgroundColor = backgroundColor ?: bgColor,
+        modifier = modifier,
     ) {
         Row(
             modifier = Modifier
@@ -50,7 +50,7 @@ fun SubscriptionTierPill(
             Icon(
                 painter = painterResource(iconRes),
                 contentDescription = null,
-                modifier = modifier
+                modifier = Modifier
                     .size(iconSizeInDp)
                     .background(bgColor),
                 tint = iconColor,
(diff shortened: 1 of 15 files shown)


modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/SignInState.kt
```diff
@@ -17,14 +19,14 @@ sealed class SignInState {
         get() = this is SignedIn && this.subscriptionStatus is SubscriptionStatus.Free
 
     val isSignedInAsPlus: Boolean
-        get() = this is SignedIn && this.subscriptionStatus is SubscriptionStatus.Plus
+        get() = this is SignedIn && this.subscriptionStatus is SubscriptionStatus.Plus && this.subscriptionStatus.type == SubscriptionType.PLUS
+
+    val isSignedInAsPatron: Boolean
+        get() = BuildConfig.ADD_PATRON_ENABLED && this is SignedIn && this.subscriptionStatus is SubscriptionStatus.Plus && this.subscriptionStatus.type == SubscriptionType.PATRON
```
ashiagr, line 25: I should either rename `SubscriptionStatus.Plus` to `SubscriptionStatus.PlusOrPatron` differentiated by type or include another data class  `SubscriptionStatus.Patron` with `SubscriptionType.Patron` and keep `SubscriptionStatus.Plus` only for `SubscriptionType.Plus`.

https://github.com/Automattic/pocket-casts-android/blob/0c4e15cc507b62020cf4df43b4f8a02a4646fa80/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/SubscriptionStatus.kt#L23

Lmk if you have a preference or I'll defer this decision until corresponding endpoint is updated.

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/components/SubscriptionTierPill.kt
```diff
@@ -69,7 +69,8 @@ fun SubscriptionTierPill(
 @Composable
 private fun SubscriptionTierPillPreview() {
     SubscriptionTierPill(
-        iconRes = IR.drawable.ic_patron,
-        shortNameRes = LR.string.pocket_casts_patron_short,
+        iconRes = R.drawable.ic_patron,
+        shortNameRes = au.com.shiftyjelly.pocketcasts.localization.R.string.pocket_casts_patron_short,
```
mchowning, line 73: Might want to use `LR.string` here

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/SignInState.kt
mchowning: I would tend to favor separate types for Plus and Patron, but I don't feel strongly and if that's fighting against what the endpoint is doing it may not be worth it. Basically, I'm fine with whatever feels better to you.

mchowning (approved):
Looking good! 👍

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/components/SubscriptionTierPill.kt
ashiagr: Fixed in 8ed61e9

In addition, I renamed the class to `SubscriptionBadge` and displayed the badge only for Patron on the profile header.

Issue #958, opened 2023-05-12
Crash on logout race condition


mchowning:
### Description

In the watch app I have observed that occasionally logging out will cause a crash because it tries to get an access token [here](https://github.com/Automattic/pocket-casts-android/blob/5049b0109fa90c0074d5428a77ca990041cf39e8/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/SyncManagerImpl.kt#L522-L521), but fails because the refresh token is null. This appears to be a race condition since there is an earlier guard on this code confirming that the user is logged in.

I have only reproduced this on the watch app, but this is using the same logic as the phone app, so the same problem might be possible there. I have not been able to reproduce this on the phone app though.

### Step-by-step reproduction instructions

I cannot reproduce this consistently, but the steps I've been following taht seems to cause it the most consistently are (note that all of the details in these steps probably aren't necessary, I'm just reporting what I've been doing):
3. Log into the watch app using google sign in
4. Log out of the watch app
5. If that doesn't crash, then log back into the google account and sign out again. On the watch app I get a crash within the first 2-3 signouts pretty consistently.

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

Watch emulator with a 7.39 pre-release build.

mchowning:
I have confirmed that this is a race condition and that it can also occur on the phone app. It can be recreated consistently by applying a delay inside a `Single.fromCallable` call [like this](https://gist.github.com/mchowning/3f567e4754bd67f5acb5e466e48479a4). 

The problem seems to be that while the server call is being made inside a Single.fromCallable [here](https://github.com/Automattic/pocket-casts-android/blob/5049b0109fa90c0074d5428a77ca990041cf39e8/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/SyncManagerImpl.kt#L521-L524) the downstream disposes of the Single source. When that happens any Throwable that is thrown bypasses any error handling in the stream and instead goes straight to RxJava's global error handler which, [on Android, crashes the app](https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#error-handling:~:text=However%2C%20Android%20is%20more%20strict%20and%20terminates%20the%20application%20in%20such%20uncaught%20exception%20cases.).

mchowning:
[...]

PR #959, opened 2023-05-12
Add/google sign in to watch

mchowning:
## Description
Adds login with google to the watch.

h/t to @luizgrp for his help getting this set up. 🙇 

> **Note**
> There is currently an issue where once you log in with one google account, you cannot then select a different google account to log in with later (it just automatically logs in with the previously selected account). It looks like [this](https://github.com/Automattic/pocket-casts-android/blob/64bdb4d2c1e222d7e86937dc512fa2d1188abceb/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/viewmodel/GoogleSignInButtonViewModel.kt#L77) is where we handle that on the phone app. I'll address this separately although it feels like a bit of an edge case since I doubt (1) many users will be switching accounts on their watch and (2) many users will have more than one google account on their watch. Still, it would be nice to address this.

> **Warning**
> There appears to be a bug where the app will occasionally crash when logging out because it tries to get an access token [here](https://github.com/Automattic/pocket-casts-android/blob/5049b0109fa90c0074d5428a77ca990041cf39e8/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/SyncManagerImpl.kt#L522-L521), but fails because the refresh token is null. I'm looking into this and I've filed an issue: https://github.com/Automattic/pocket-casts-android/issues/958

## Testing Instructions
1. While logged in on a connected phone
2. Open a fresh install of the watch app
3. Observe that the watch automatically logs into the account from the phone
4. The logging in screen should respect the minimum display duration of [5 seconds](https://github.com/Automattic/pocket-casts-android/blob/6160ab2262145169cf5b0f670d4e93c08b11b8b2/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/LoggingInScreenViewModel.kt#L28-L27) (i.e., the logging in screen should display for at least 5 seconds even if the refresh is done before then).
5. Go to "Settings" and log out on the watch
6. Tap on "Log in"
7. Tap on "Log in with Google"
8. Select a google account that has been set up on the watch
9. Observe the logging in screen is presented
10. This screen should dismiss as soon as the refresh is finished, even if it finishes before the minimum display duration of 5 seconds
11. You should automatically be navigated to the top of the initial list screen in the watch app.

## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/4656348/3ba5a953-9dfa-41c8-bd3c-62baadff78b4

## Checklist


ashiagr (approved):
Looks good 👍

PR #960, opened 2023-05-15
Fix the listening stats

geekygecko:
## Description
The device user listening stats require a device id to keep track of the listening stats on different devices. As part of the Google single sign-on change, I removed this. 🤦‍♂️ This change puts it back. 

## Testing Instructions
1. Open the Android Studio panel called 'App Inspection'
2. Tap the 'Network Inspector' tab
3. In the app go to the Profile tab and tap 'Refresh Now'
4. In the 'Network Inspector' find the 'https://api.pocketcasts.net/sync/update' call
5. ✅  Verify the request contains the device parameter with a value

## Screenshot
Screenshot 2023-05-15 at 3 31 05 pm


ashiagr (approved):
I can see the params in the inspector 👍

PR #971, opened 2023-05-17
Avoiding crashing on RxJava errors in production

mchowning:
## TLDR

This updates the app so that in release builds, when there are uncaught errors in RxJava streams, we log the error but no longer crash the app. On debug builds, however, the current default behavior of crashing the app remains.

## But Why?

I was digging into [this crash](https://github.com/Automattic/pocket-casts-android/issues/958) on the watch app, and it appears to be a fairly tricky race condition. In particular, the problem seems to happen if while the server call is being made inside a `Single.fromCallable` [here](https://github.com/Automattic/pocket-casts-android/blob/5049b0109fa90c0074d5428a77ca990041cf39e8/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/SyncManagerImpl.kt#L521-L524) the downstream disposes of the `Single` source. When that happens any `Throwable` that is thrown [bypasses any error handling in the stream and instead goes straight to RxJava's global error handler](https://github.com/ReactiveX/RxJava/blob/06e52db5356c7cda93480e5019959ff5c651938e/src/main/java/io/reactivex/rxjava3/core/Single.java#L965-L969) which, [on Android, crashes the app](https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#error-handling:~:text=However%2C%20Android%20is%20more%20strict%20and%20terminates%20the%20application%20in%20such%20uncaught%20exception%20cases.).

The best solution would be to find and fix the race condition, but it appears to be quite tricky, and I'm not sure how much time we want to devote to improving our RxJava handling since our long-term goal is to remove RxJava from the app. For that reason, I thought that this might be a "good enough" solution.

I feel conflicted about this, so let me know if you think this isn't a good idea. The reasons I decided to propose this is that we're planning to eventually remove RxJava anyway and, as long as we're still receiving these exceptions in Sentry, I don't see why we would want the app to crash in production. If you have any other ideas for how to handle this, though, definitely let me know.

The reason I applied this change to all 3 of our apps is that:
1. It is possible for the crash to happen on the phone and automotive apps (although it appears _much_ less likely)
2. Applying it to all 3 apps makes development easier because we have consistent behavior on all the apps.

### A. Debug build
1. Install a debug build
2. Attempt to recreate https://github.com/Automattic/pocket-casts-android/issues/958, and observe that it still crashes. Alternatively, apply [this patch](https://gist.github.com/mchowning/3f567e4754bd67f5acb5e466e48479a4), and the crash will happen pretty consistently on both the watch and phone apps.

### B. Release build
1. Install a release build
2. Observe that you cannot crash the app by attempting to recreate https://github.com/Automattic/pocket-casts-android/issues/958 or applying [the patch](https://gist.github.com/mchowning/3f567e4754bd67f5acb5e466e48479a4), but the exception does show up in Sentry [like this](https://a8c.sentry.io/issues/4114268937/events/7eee9cc683b644c4855c6f42c430a99f/).

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
         ([#960](https://github.com/Automattic/pocket-casts-android/pull/960)).
     *   Fixed crash on grouping by season
         ([#962](https://github.com/Automattic/pocket-casts-android/pull/962)).
+    *   Fixed crash that could occur when signing out of the app
+        ([#971](https://github.com/Automattic/pocket-casts-android/pull/971)).
 *   Updates
     *   Link users to support forum from within the app
         ([#950](https://github.com/Automattic/pocket-casts-android/pull/950)).
--- a/app/src/main/java/au/com/shiftyjelly/pocketcasts/PocketCastsApplication.kt
+++ b/app/src/main/java/au/com/shiftyjelly/pocketcasts/PocketCastsApplication.kt
@@ -29,6 +29,7 @@ import au.com.shiftyjelly.pocketcasts.ui.helper.AppIcon
 import au.com.shiftyjelly.pocketcasts.utils.TimberDebugTree
 import au.com.shiftyjelly.pocketcasts.utils.log.LogBuffer
 import au.com.shiftyjelly.pocketcasts.utils.log.LogBufferUncaughtExceptionHandler
+import au.com.shiftyjelly.pocketcasts.utils.log.RxJavaUncaughtExceptionHandling
 import coil.Coil
 import coil.ImageLoader
 import com.google.firebase.FirebaseApp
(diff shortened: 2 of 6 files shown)


ashiagr (approved):
> I feel conflicted about this, so let me know if you think this isn't a good idea. The reasons I decided to propose this is that we're planning to eventually remove RxJava anyway and, as long as we're still receiving these exceptions in Sentry, I don't see why we would want the app to crash in production. If you have any other ideas for how to handle this, though, definitely let me know.

Looks fine to me as long as we're capturing this exception in production and until we figure out a proper fix 👍

mchowning:
Somehow, I completely overlooked that we were already doing something similar to this on the phone app [here](https://github.com/Automattic/pocket-casts-android/blob/eed52a5124af33ab8be37cb66f291dcdd867d128/app/src/main/java/au/com/shiftyjelly/pocketcasts/PocketCastsApplication.kt#L254-L265). 🤦  We should revisit this and handle this in only one place, making sure it is consistent between the phone, watch, and automotive apps.

Issue #968, opened 2023-05-15
New profile header uses low resolution Avatar


b-aaron:
### Description

The new profile header now shows the Profile Picture / Avatar in a bigger size, while still using the low resolution image.

### Step-by-step reproduction instructions

1. Log in with account that has an Avatar set
2. Go to Profile tab
3. Observe that the Avatar has a low resolution image

### Screenshots or screen recording

![Screenshot_20230515_203113_Pocket Casts](https://github.com/Automattic/pocket-casts-android/assets/72910336/3e6265b5-59b3-4207-ad27-cb5336f85009)

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

Samsung Galaxy S20 FE 5G
Android 13
OneUI 5.1
Pocket Casts 7.39-rc-1

PR #973, opened 2023-05-17
Improve the resolution of the Gravatar image

geekygecko:
## Description
The default image size for a Gravatar image is 80px. This increases it to 400px. The sizes I could find it being used were 104dp, 100dp, 64dp, and 56dp so I thought 400px would be big enough.
https://en.gravatar.com/site/implement/images/

Fixes https://github.com/Automattic/pocket-casts-android/issues/968

## Testing Instructions
1. Tap on the Profile tab
2. Sign in if you aren't already
3. ✅  Verify the quality of the Gravatar image

## Screenshots 

| Before | After |
| --- | --- |
| Screenshot_20230517-211941 | Screenshot_20230517-212229 |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 7.40
 -----
-
+*   Bug Fixes:
+    *   Improved the resolution of the Gravatar image
+        ([#973](https://github.com/Automattic/pocket-casts-android/pull/973)).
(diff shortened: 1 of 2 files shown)


mchowning (approved):
That looks much better! 😄  Thanks for the fix!

PR #911, opened 2023-04-26
Add screen to more easily access logs

mchowning:
## Description
This adds a screen to enable the user to more easily access their logs within the app.

> **Note**
> This PR builds on https://github.com/Automattic/pocket-casts-android/pull/910, so you may want to review that one first.

Fixes #833 

## Testing Instructions
1. Go to the Profile tab
2. Tap on the ⚙️ 
3. Tap on "Help & feedback"
4. Tap on the list icon at the top of the screen
5. ✅ Observe that the logs are displayed
6. Tap on the copy icon at the top of the screen
7. ✅  Observe that the logs are copied to the clipboard
8. Tap on the Share icon
9. ✅ Observe that you can share a file with the logs (similar to what happens currently when someone contacts support through the app).

## Screenshots or Screencast 

https://user-images.githubusercontent.com/4656348/234631340-cc39fce0-e5be-44a9-8e4d-9e4ea21d391a.mov

## Checklist
- [X] If this is a user-facing change, I have added an entry in CHANGELOG.md
- [X] I have considered whether it makes sense to add tests for my changes
- [X] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml`
- [X] Any jetpack compose components I added or changed are covered by compose previews
 
#### I have tested any UI changes...

- [X] with different themes
- [X] with a landscape orientation
- [X] with the device set to have a large display and font size
- [X] for accessibility with TalkBack

PR #974, opened 2023-05-17
Add LogBuffer output to device logs screen

mchowning:
## Description
When I initially implemented the device logs screen in https://github.com/Automattic/pocket-casts-android/pull/911, I failed to append the `LogBuffer` logs to the output. This PR fixes that.

Before this fix, the logs would end with the "Episode Issues" and would contain none of the timestamped logs like "I 16/5 17:31:44 App started. 7.38 (9094)".

> **Note**
> I am targeting `release/7.39` with this because it seems like a pretty safe change and it would be good to get this out quickly since it fixes a new feature. I don't think we need to do a hotfix for this because emailing the logs (either from the logs display screen or if the user tries to contact support) will still include the full logs, so there is a workaround. I don't think we need to worry about immediately cutting a new beta with this—it's not urgent imo.

## Testing Instructions
1. Go to "Profile" → ⚙️ → "Help & feedback"
2. Tap on the logs icon at the top fo the screen
3. Verify that the screen includes the LogBuffer logs at the bottom.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@
         ([#960](https://github.com/Automattic/pocket-casts-android/pull/960)).
     *   Fixed crash on grouping by season
         ([#962](https://github.com/Automattic/pocket-casts-android/pull/962)).
+    *   Fixed bug that prevented full logs from being displayed within the app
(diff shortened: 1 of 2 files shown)


ashiagr (approved):
Verified, full logs are appended now. Thanks for fixing it!

PR #976, opened 2023-05-18
Fix custom skip buttons not appearing

geekygecko:
## Description
This change fixes the issue that the media session isn't using our custom skip buttons.

## Testing Instructions
1. Play an episode
2. Swipe down the notification shade
3. ✅ Verify that our custom skip buttons are used in the notification player view, not the next / previous track buttons.

## Screenshots 
| Before | After |
| --- | --- |
| Screenshot_20230518_104156 | Screenshot_20230518_104429 |


ashiagr (approved):
LGTM 👍

PR #977, opened 2023-05-18
Improving the Automotive buffering and seeking

geekygecko:
## Description
With Automotive, we are getting reports of issues with the seeking not working, buffering issues and out-of-memory crashes. A fix for this was to add the `largeHeap` configuration to the Automotive manifest file (already in the app manifest) and reduce the player buffer sizes. This is only with certain manufacturers so I have limited the buffer size for those.

## Testing Instructions
1. Deploy to Automotive
2. ✅ Verify you can play an episode, seek to a new position, and use the skip buttons.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 7.40
 -----
 *   Bug Fixes:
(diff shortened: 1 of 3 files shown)


ashiagr (approved):
Looks good 👍

PR #978, opened 2023-05-18
On Automotive add the option to clear your data before logging in

geekygecko:
## Description
For the Automotive app, we have been asked if we can add the option to clear your data before you can log in or create an account. If you have an Up Next, any podcasts or episodes then it will ask you, otherwise the dialog won't be shown to you.

## Testing Instructions
1. Deploy the Automotive version
2. Tap on the Discover tab
3. Tap a podcast
4. Tap on an episode to play it
5. Close the full screen player with the arrow at the top left
6. Tap the settings cog
7. Tap 'Set up account'
8. ✅ Verify a dialog is shown asking if you want to clear your data
9. Tap 'Clear data'
10. Tap close and back to get back to the app
11. ✅ Verify the mini player has closed
12. Tap the settings cog
13. Tap 'Set up account'
14. ✅ Verify it goes straight to the "Sign in or create account" page and doesn't show the dialog

## Screenshots
Screenshot_20230518_170246

### DIFF
--- a/automotive/src/main/java/au/com/shiftyjelly/pocketcasts/AutomotiveSettingsFragment.kt
+++ b/automotive/src/main/java/au/com/shiftyjelly/pocketcasts/AutomotiveSettingsFragment.kt
@@ -21,7 +21,6 @@ import au.com.shiftyjelly.pocketcasts.repositories.podcast.PodcastManager
 import au.com.shiftyjelly.pocketcasts.repositories.searchhistory.SearchHistoryManager
 import au.com.shiftyjelly.pocketcasts.repositories.sync.SyncManager
 import au.com.shiftyjelly.pocketcasts.repositories.user.UserManager
-import au.com.shiftyjelly.pocketcasts.utils.Util
 import dagger.hilt.android.AndroidEntryPoint
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
@@ -89,7 +88,7 @@ class AutomotiveSettingsFragment : Fragment(), CoroutineScope {
         // ask the user if they want to clear their data before they sign in
         if (isAccountUsed()) {
             val context = context ?: return
-            val themedContext = if (Util.isAutomotive(context)) ContextThemeWrapper(context, CR.style.Theme_Car_NoActionBar) else context
+            val themedContext = ContextThemeWrapper(context, CR.style.Theme_Car_NoActionBar)
             val builder = AlertDialog.Builder(themedContext)
             builder.setTitle(getString(LR.string.profile_clear_data_question))
                 .setMessage(getString(LR.string.profile_clear_data_would_you_also_like_question))
@@ -102,7 +101,7 @@ class AutomotiveSettingsFragment : Fragment(), CoroutineScope {
                         folderManager = folderManager,
                         searchHistoryManager = searchHistoryManager,
                         episodeManager = episodeManager,
-                        wasInitiatedByUser = true
+                        wasInitiatedByUser = false
                     )
                     openSignInActivity()
                 }
--- a/automotive/src/main/java/au/com/shiftyjelly/pocketcasts/AutomotiveSettingsFragment.kt
+++ b/automotive/src/main/java/au/com/shiftyjelly/pocketcasts/AutomotiveSettingsFragment.kt
@@ -2,23 +2,51 @@ package au.com.shiftyjelly.pocketcasts
 
 import android.content.Intent
 import android.os.Bundle
+import android.view.ContextThemeWrapper
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import androidx.appcompat.app.AlertDialog
 import androidx.fragment.app.Fragment
 import androidx.lifecycle.toLiveData
 import au.com.shiftyjelly.pocketcasts.account.AccountActivity
 import au.com.shiftyjelly.pocketcasts.databinding.FragmentAutomotiveSettingsBinding
 import au.com.shiftyjelly.pocketcasts.profile.AccountDetailsFragment
+import au.com.shiftyjelly.pocketcasts.repositories.playback.PlaybackManager
+import au.com.shiftyjelly.pocketcasts.repositories.playback.UpNextQueue
+import au.com.shiftyjelly.pocketcasts.repositories.podcast.EpisodeManager
+import au.com.shiftyjelly.pocketcasts.repositories.podcast.FolderManager
+import au.com.shiftyjelly.pocketcasts.repositories.podcast.PlaylistManager
+import au.com.shiftyjelly.pocketcasts.repositories.podcast.PodcastManager
+import au.com.shiftyjelly.pocketcasts.repositories.searchhistory.SearchHistoryManager
+import au.com.shiftyjelly.pocketcasts.repositories.sync.SyncManager
 import au.com.shiftyjelly.pocketcasts.repositories.user.UserManager
+import au.com.shiftyjelly.pocketcasts.utils.Util
 import dagger.hilt.android.AndroidEntryPoint
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
 import javax.inject.Inject
+import kotlin.coroutines.CoroutineContext
+import au.com.shiftyjelly.pocketcasts.cartheme.R as CR
+import au.com.shiftyjelly.pocketcasts.localization.R as LR
 
 @AndroidEntryPoint
-class AutomotiveSettingsFragment : Fragment() {
+class AutomotiveSettingsFragment : Fragment(), CoroutineScope {
     @Inject lateinit var userManager: UserManager
+    @Inject lateinit var syncManager: SyncManager
+    @Inject lateinit var podcastManager: PodcastManager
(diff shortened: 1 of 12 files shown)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/UpNextQueue.kt
```diff
@@ -68,17 +68,20 @@ interface UpNextQueue {
...
                         .combineLatest(podcastManager.observePodcastByUuid(state.podcast.uuid).distinctUntilChanged { t1, t2 -> t1.isUsingEffects == t2.isUsingEffects })
-                        .map { State.Loaded(it.first, it.second, state.queue) }
+                        .map<State> { State.Loaded(it.first, it.second, state.queue) }
+                        .onErrorReturn { State.Empty }
                         .toObservable()
                 } else {
                     episodeManager.observeEpisodeByUuidRx(state.episode.uuid)
-                        .map { State.Loaded(it, state.podcast, state.queue) }
+                        .map<State> { State.Loaded(it, state.podcast, state.queue) }
+                        .onErrorReturn { State.Empty }
```
geekygecko, line 84: As part of clearing the data when the episode was deleted, it caused the Up Next and playback event monitoring to crash. This meant that after clearing the data it wasn't possible to play another episode. This change fixes that.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/PodcastManagerImpl.kt
```diff
@@ -741,17 +741,6 @@ class PodcastManagerImpl @Inject constructor(
         )
     }
 
-    /**
-     * Will return true if all the podcasts are have auto download turned on.
-     */
-    private fun isAutoDownloadingAllPodcasts(): Boolean {
-        return if (countPodcasts() == 0) false else podcastDao.countDownloadStatus(Podcast.AUTO_DOWNLOAD_NEW_EPISODES) == podcastDao.countSubscribed()
-    }
```
geekygecko, line 749: These methods weren't been used

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/MediaSessionManager.kt
```diff
@@ -285,7 +285,16 @@ class MediaSessionManager(
...
+                    if (state.isEmpty) {
+                        Observable.just(Optional.empty())
+                    } else {
+                        episodeManager.observeEpisodeByUuidRx(state.episodeUuid)
+                            .distinctUntilChanged(BaseEpisode.isMediaSessionEqual)
+                            .map { Optional.of(it) }
+                            // if the episode is deleted from the database while playing catch the error and just return an empty state
+                            .onErrorReturn { Optional.empty() }
+                            .toObservable()
+                    }
```
geekygecko, line 297: As part of clearing the data when the episode was deleted, it caused the Up Next and playback event monitoring to crash. This meant that after clearing the data it wasn't possible to play another episode. This change fixes that.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/user/UserManager.kt
```diff
@@ -115,4 +129,42 @@ class UserManagerImpl @Inject constructor(
...
+
+        // Need to stop playback before we start clearing data
+        playbackManager.removeEpisode(
+            episodeToRemove = playbackManager.getCurrentEpisode(),
+            // Unknown is fine here because we don't send analytics when the user did not initiate the action
+            source = AnalyticsSource.UNKNOWN,
+            userInitiated = false,
+        )
+
+        // Block while clearing data so that users cannot interact with the until we're done clearing data
```
mchowning, line 153:
```suggestion
        // Block while clearing data so that users cannot interact with the app until we're done clearing data
```

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/user/UserManager.kt
```diff
@@ -115,4 +129,42 @@ class UserManagerImpl @Inject constructor(
...
+        playbackManager: PlaybackManager,
+        upNextQueue: UpNextQueue,
+        playlistManager: PlaylistManager,
+        folderManager: FolderManager,
+        searchHistoryManager: SearchHistoryManager,
+        episodeManager: EpisodeManager,
+        wasInitiatedByUser: Boolean,
+    ) {
+        // Sign out first to make sure no data changes get synced
+        signOut(playbackManager, wasInitiatedByUser = true)
```
mchowning, line 143:
```suggestion
        signOut(playbackManager, wasInitiatedByUser = wasInitiatedByUser)
```

automotive/src/main/java/au/com/shiftyjelly/pocketcasts/AutomotiveSettingsFragment.kt
```diff
@@ -49,15 +77,55 @@ class AutomotiveSettingsFragment : Fragment() {
...
+                .setPositiveButton(getString(LR.string.sign_in)) { _, _ -> openSignInActivity() }
+                .setNegativeButton(getString(LR.string.profile_clear_data)) { _, _ ->
+                    userManager.signOutAndClearData(
+                        playbackManager = playbackManager,
+                        upNextQueue = upNextQueue,
+                        playlistManager = playlistManager,
+                        folderManager = folderManager,
+                        searchHistoryManager = searchHistoryManager,
+                        episodeManager = episodeManager,
+                        wasInitiatedByUser = true
```
mchowning, line 105: I wonder if `wasInitiatedByUser` should be `false` here. The user is initiating the action here, but they're not trying to sign out. One of the things this `wasInitiatedByUser` variable is used for is [to identify in our analytics whether the user was attempting to sign out](https://github.com/Automattic/pocket-casts-android/blob/5c6a321c9115b6290e979bbccc50fb0c30a1398f/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/user/UserManager.kt#L123). 

Admittedly this is a bit of a moot point at this time since we're not sending any analytics to Tracks from Android Automotive yet.

Also, if you change `wasInitiatedByUser` to be `false`, you may need to do something like add a call to `settings.setFullySignedOut(false)` to get past [this if-check](https://github.com/Automattic/pocket-casts-android/blob/5c6a321c9115b6290e979bbccc50fb0c30a1398f/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/user/UserManager.kt#L108) in the `signOut` method.

The `settings.getFullySIgnedOut` value is a relatively recent addition I did largely to avoid extra signout calls to analytics (we were making a call to signout every time a user who was not signed in started the app). Maybe you have some ideas for how to better accomplish that because it does seem to be creating some undesirable complexity here. 🤔

mchowning (approved):
This is looking good. 👍

automotive/src/main/java/au/com/shiftyjelly/pocketcasts/AutomotiveSettingsFragment.kt
geekygecko: I wonder if it would be easier to understand if we named the parameter something like `userAttemptingToSignOut`. 🤔 

I was trying to reuse the same logic but I think I have made it more complicated. Thanks for picking this up, I have committed a fix to set `wasInitiatedByUser` to `false`.

mchowning: > I wonder if it would be easier to understand if we named the parameter something like `userAttemptingToSignOut`. 🤔

If that seems like a better name to you I'm certainly not opposed to switching to that.

mchowning: Just wanted to note explicitly that with this change none of [this code](https://github.com/Automattic/pocket-casts-android/blob/7a9fbad50b762563536f5b0f659117087d232832/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/user/UserManager.kt#L108-L129) is getting run when the user clears data on sign-in:

``` kotlin
if (wasInitiatedByUser || !settings.getFullySignedOut()) {
    LogBuffer.i(LogBuffer.TAG_BACKGROUND_TASKS, "Signing out")
    subscriptionManager.clearCachedStatus()
    syncManager.signOut {
        settings.clearPlusPreferences()
        GlobalScope.launch {
            userEpisodeManager.removeCloudStatusFromFiles(playbackManager)
        }

        settings.setMarketingOptIn(false)
        settings.setMarketingOptInNeedsSync(false)
        settings.setEndOfYearModalHasBeenShown(false)

        analyticsTracker.track(
            AnalyticsEvent.USER_SIGNED_OUT,
            mapOf(KEY_USER_INITIATED to wasInitiatedByUser)
        )
        analyticsTracker.flush()
        analyticsTracker.clearAllData()
        analyticsTracker.refreshMetadata()
    }
}

```

I think that's probably fine because all those things seem to relate to accounts and obviously the user isn't signed-in in this scenario, but it's worth keeping in mind if we see any issues.

PR #980, opened 2023-05-18
Update player background and add volume button

ashiagr:
## Description

This PR updates the Wear app's "Now Playing" screen to include
- background based on the currently playing podcast
- volume/source options

## Testing Instructions

- Play an episode from any podcast
- Navigate to the player screen in the pager
- ✅ Notice that the background has a tint based on the currently playing podcast
- Notice that there is a volume button at the bottom
- Tap on the volume button
- ✅ Notice that the volume screen opens with a volume stepper and there's an option to select the volume source
- Select a volume source 
- Control volume using the volume stepper
- ✅ Notice that the volume changes for the selected source

ℹ️ The PR uses the horologist-provided volume screen, the screen needs to be updated based on Figma design. 

## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/1405144/146ff201-148c-4922-93cd-67cef5fdb727

## Checklist
 
- [ ] with different themes


mchowning (approved):
This looks good! Great work!

PR #981, opened 2023-05-19
Update UI for email sign in on watch

mchowning:
## Description
This improves the UI for signing in by email on the watch.

Text entry on the watch seems to not be particularly well-supported (probably because it is strongly discouraged by Google). As you'll see in this PR, it's much more complicated than just using a `TextField` composable on the phone, and the experience is not ideal. 

Fortunately, the vast majority of users should never need to sign in with the app on their connected phone or by using Sign in with Google. The email sign in is only included as a last resort to make sure no users are blocked from signing in. Also, we plan to remove email sign-in in the future once we add the ability for users to sign in on the web and get a short code they can enter on their watch to sign in.

Because entering text on the watch requires launching a new activity, it is not easy to exercise much control over the flow. The flow I have here seems pretty good, but it certainly isn't ideal. Given that this should be a rarely used method of signing in, and that we want to release this soon, my thinking is that this is good enough. Let me know if you disagree though.

You'll notice that there is an option for voice input. I don't think it makes a lot of sense for users to use voice input to enter their email or password, but it's not clear to me how to remove that option and to just go straight to the keyboard input for either the email or the password.

### 1. Successful login
1. Go to "Settings" → "Log in" → "Log in with email"
2. Enter a PC account email address and password
3. Observe the logging in screen
4. After the log in is complete, you are taken to the top of the Watch List screen.

### 2. Failed login
1. Go to "Settings" → "Log in" → "Log in with email"
2. Enter an email/password combination that does not match a PC account
3. Observe the error screen is shown with the relevant error message. If the error message is longer than the screen then the screen should be scrollable so that the entire error message can be viewed.
4. Swipe away the error message and yous should be returned to the screen with the log in options

## Screenshots or Screencast 

|  Successful Sign In | Sign-in error |
| --- | --- |
| <video src="https://github.com/Automattic/pocket-casts-android/assets/4656348/f7eb3b1c-875d-4bb9-b687-fe2e16f1b1b8"/> | <video src="https://github.com/Automattic/pocket-casts-android/assets/4656348/8faeaf6d-3c1a-4be0-933b-2e0ca757c8e4"/> |

## Checklist


geekygecko (approved):
The login works well! It's a shame about it offering the voice input and not being about to set an input type of email but I guess that's just the limitations of Wear OS.

I'm not sure if I did something wrong but for an invalid email, I get an error that looks like a server issue.

https://github.com/Automattic/pocket-casts-android/assets/308331/f5fb4687-4fdc-433d-b676-153241e119bc

mchowning:
> I'm not sure if I did something wrong but for an invalid email, I get an error that looks like a server issue.

Thanks for calling this out. I forgot to mention that this isn't doing any validation on the email and password currently. I have created [an issue](https://github.com/Automattic/pocket-casts-android/issues/983) for addressing that. Hopefully, very few users will actually need to use email login, so I don't think this is a blocker for v1. Let me know if you disagree though.

PR #986, opened 2023-05-21
translation fix

apollo2k4:
## Description
removing html tags from `<string name="podcasts_unarchive">`

## Screenshots or Screencast 
Screenshots

## Checklist
- [-] If this is a user-facing change, I have added an entry in CHANGELOG.md
- [-] I have considered whether it makes sense to add tests for my changes
- [+] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml`
- [-] Any jetpack compose components I added or changed are covered by compose previews


vladd-g (changes requested):

sryze:
I noticed the same bug just a couple of days ago.

How long does it usually take to validate a translation? It seems that someone suggsted a new translation (without the `&nbsp;`) back in June but is still in Waiting status.

mchowning:
> How long does it usually take to validate a translation? It seems that someone suggsted a new translation (without the &nbsp;) back in June but is still in Waiting status.

Thanks for raising this. 🙇  Those have been approved, and we're looking into why they were waiting for so long.

PR #990, opened 2023-05-23
Open watch app's Now Playing screen when tapping on episode screen's Play icon

ashiagr:
## Description
This opens watch app's Now Playing screen when tapping on episode screen's Play icon

## Testing Instructions

1. Install the app
2. Login with an account having podcasts
3. Open `Podcasts` -> `Podcast` -> `Episode` screen
4. Tap play on the `Episode` screen
5. ✅ Notice that `Now playing` screen is shown
6.  Comment out lines 243, 245-247 inside `EpisodeViewModel`'s `fun onPlayClicked(showStreamingConfirmation: () -> Unit)` to show the streaming dialog confirmation
7. Reinstall the app and repeat steps 3-4
8. Notice that Streaming confirmation dialog is shown
9. Tap X
10. Notice that Streaming confirmation dialog is dismissed and you're not navigated to the `Now playing` screen
11. Re-tap play on the Episode screen
12. Notice that Streaming confirmation dialog is shown
13. Tap ✔️ 
14. ✅ Notice that `Now playing` screen is shown

## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/1405144/a1427666-22cc-4712-9aa8-996dbaa89061

https://github.com/Automattic/pocket-casts-android/assets/1405144/a6e243be-5d32-4230-85b7-ebec4d584584

## Checklist
 
- [ ] with different themes

### DIFF
--- a/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/episode/EpisodeViewModel.kt
+++ b/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/episode/EpisodeViewModel.kt
@@ -36,6 +36,7 @@ import coil.request.SuccessResult
 import dagger.hilt.android.lifecycle.HiltViewModel
 import dagger.hilt.android.qualifiers.ApplicationContext
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.SharingStarted
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.combine
@@ -103,6 +104,8 @@ class EpisodeViewModel @Inject constructor(
     )
 
     val stateFlow: StateFlow<State>
+    var showNowPlaying = MutableSharedFlow<Boolean>()
+        private set
 
     init {
         val episodeUuid = savedStateHandle.get<String>(EpisodeScreenFlow.episodeUuidArgument)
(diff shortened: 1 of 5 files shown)


wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/episode/EpisodeViewModel.kt
```diff
@@ -103,6 +104,8 @@ class EpisodeViewModel @Inject constructor(
     )
 
     val stateFlow: StateFlow<State>
+    var showNowPlaying = MutableSharedFlow<Boolean>()
+        private set
```
mchowning, line 108: Do you think it might be worth adding a comment here noting that we're using a shared flow here because we need the event to only be sent one time and then discarded? Just to make it clear why we're not using a state flow like we normally do.

Also, the `private set var` on the mutable flow reads a bit odd to me because I read it as implying that we might want to switch out this flow for an entirely different flow. Maybe something like this would work:

``` kotlin
val _showNowPlaying = MutableSharedFlow<Boolean>()
val showNowPlaying = _showNowPlaying.asSharedFlow()
```

mchowning (approved):
Looks great. This definitely ended up being pretty tricky, and I like what you did with it. Just left one minor comment.

wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/episode/EpisodeViewModel.kt
ashiagr: Makes sense 👍 
Addressed in bd5a809

PR #992, opened 2023-05-24
Improving the Wear OS podcasts page

geekygecko:
## Description

This change improves the podcasts page in the following ways.

- Adds a podcast color gradient to the chip.
- Adds a page header. 

It also includes:

- Changing to the extra dark theme. This fixes the issue that you can see Pixel watch bezel if the background isn't black.
- Changing the theme typography to match the design.

## Testing Instructions
1. Tap "Podcasts"
2. ✅  Scroll down the list and verify the tint colours look fine.

## Screenshots
Screenshot_20230524_203452

### DIFF
--- a/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/theme/WearColors.kt
+++ b/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/theme/WearColors.kt
@@ -3,8 +3,11 @@ package au.com.shiftyjelly.pocketcasts.wear.theme
 import androidx.compose.ui.graphics.Color
 
 object WearColors {
-    val FF202124 = Color(0xFF202124)
-    val FFA1E7B0 = Color(0xFFA1E7B0)
-    val FFBDC1C6 = Color(0xFFBDC1C6)
-    val FFDADCE0 = Color(0xFFDADCE0)
+    val highlight = Color(0xFFF43E37)
+    val success = Color(0xFFA1E7B0)
+    val surface = Color(0xFF202124)
+    val primaryText = Color.White
+    val secondaryText = Color(0xFFBDC1C6)
+    val upNextIcon = Color(0xFF4BA1D6)
+    val downloadedIcon = Color(0xFF54C483)
 }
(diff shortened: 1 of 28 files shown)


CookieyedCodes:
Love this any chance it can be added to apple watch?

wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/podcasts/PodcastsScreen.kt
```diff
@@ -31,9 +40,17 @@ fun PodcastsScreen(
 
     ScalingLazyColumn(
         modifier = modifier.fillMaxWidth(),
-        state = listState,
+        state = listState
```
mchowning, line 43: Curious how you feel about trailing commas. I like them because they make the diffs smaller when adding/removing items, but it looks like you may not be a fan? Obviously this is a SUPER CRITICAL ISSUE! :trollface: 

I'll actually look into seeing if we can get spotless to enforce this one way or the other (although I like trailing commas, I'm fine having spotless enforce either approach).

mchowning (approved):
Looks good! Thanks so much for improving the theming! 🙇

mchowning:
> Love this any chance it can be added to apple watch?

We'll do some UI updates to the apple watch app at some point, but we don't have any immediate plans.

CookieyedCodes:
Ya fair enough, might as well wait for wwdc and see what they bring too watch for you guys :)

wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/podcasts/PodcastsScreen.kt
geekygecko: It would be good to enforce this. I don't mind either way which might show by the way I do both. 😁

wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/theme/WearColors.kt
```diff
@@ -3,8 +3,11 @@ package au.com.shiftyjelly.pocketcasts.wear.theme
...
-    val FFA1E7B0 = Color(0xFFA1E7B0)
-    val FFBDC1C6 = Color(0xFFBDC1C6)
-    val FFDADCE0 = Color(0xFFDADCE0)
+    val highlight = Color(0xFFF43E37)
+    val success = Color(0xFFA1E7B0)
+    val surface = Color(0xFF202124)
+    val primaryText = Color.White
+    val secondaryText = Color(0xFFBDC1C6)
+    val upNextIcon = Color(0xFF4BA1D6)
+    val downloadedIcon = Color(0xFF54C483)
```
geekygecko, line 12: I thought it might be nicer if we named these colors.

Issue #993, opened 2023-05-24
Watch automatically selects previously selected Google accounts when signing in


mchowning:
### Description

When signing into the watch app using a Google account, if you log out and try to log in with Google again, the previously selected Google account is automatically used. This makes it impossible to select a different Google account without uninstalling the app.

We allow selecting a different google account on the phone app, and it looks like [this](https://github.com/Automattic/pocket-casts-android/blob/64bdb4d2c1e222d7e86937dc512fa2d1188abceb/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/viewmodel/GoogleSignInButtonViewModel.kt#L77) may be where we enable that on the phone app.

### Step-by-step reproduction instructions

1. Log into the watch app using a google account on a watch with multiple google accounts
2. Verfiy that you are presented with the option of selecting which account to use when you log in
3. Log out of the watch app
4. Attempt to log back into the watch app using a google account
5. 💥 Note that you are NOT presented with the option of selecting which account to use when you log in.

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

Wear os emulator API 30, running a pre-release 7.29 build.

mchowning:
I believe this will be fixed when https://github.com/Automattic/pocket-casts-android/pull/994 is merged with https://github.com/Automattic/pocket-casts-android/pull/994/commits/519a1898593fc890f31ea4dcbb40112d17da6f03.

PR #994, opened 2023-05-24
Restrict independent watch functionality to Plus users

mchowning:
## Description
This prevents users from accessing the wear app if they are not Plus users since v1 of the wear app will not have control-the-phone-app-from-the-watch-app functionality for free users.

Getting this working well turned out to be more difficult than I expected. In particular, getting the "logging in" screen to show and block the user until the refresh was complete was the tricky part. I'm mentioning this just to call out that this PR probably merits some extra testing.

A couple of issues you may see:
1. Because the button for proceeding to log in is off screen, I also made the text on that screen clickable to reduce the chance of a user get hung up on that screen.
1. I've noticed that occasionally when installing the app from Android Studio on top of a previous install, then logging out, and then logging back in with google, the app could get stuck on a logging in screen. I've only seen this a couple of times, and I have never reproduced it on a clean install.
3. Because of [this issue](https://github.com/Automattic/pocket-casts-android/issues/958) you may see crashes when logging out (but only on debug builds). For testing purposes, I just [updated the RxJavaUncaughtExceptionHandling object](https://github.com/Automattic/pocket-casts-android/blob/b06c9721e256f74ea31299131929330e16d0c99d/modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/log/RxJavaUncaughtExceptionHandling.kt#L20-L22) to not throw any exceptions because anytime someone logs into the watch app with a free account, they get immediately logged out (which can result in a crash on debug builds). I'm starting to wonder if maybe this [crash-only-in-debug idea of mine](https://github.com/Automattic/pocket-casts-android/pull/971) wasn't such a great idea.
4.  I had it happen a couple of times where the watch app would log in automatically even though the phone app was not logged in. I haven't been able to track down the cause of this (and it doesn't seem like a big problem since the user can just log out if they want to switch accounts).
5. Note that, currently, once you log in using a google account, you [cannot then switch to a different google account](https://github.com/Automattic/pocket-casts-android/issues/993) without reinstalling the app.

## Testing Instructions

For each of the following scenarios:

1. Logging in automatically because your watch-connected phone is logged in
2. Logging in using a Google account
3. Logging in using email and password

Verify that:

1. if it is a free account, the logging in screen will be displayed, and then the user is returned to the UI prompting the user to log in.
2. If it is a plus account, the logging in screen is displayed, and that you are taken into the app once the refresh is complete

In addition, also test that we are handling accounts that change from Plus to Free appropriately:
1. Sign in with a Plus account
2. Change the account to being Free
3. Perform a manual refresh from the Settings screen
4. Verify that you are returned to the `RequirePlusScreen`

### Phone app

Because I made a small change to how we report a user's subscription status, it would be good to do some smoke-testing around logging into and out of some Plus and Free accounts on the phone app.

## Screenshots or Screencast 

| Plus | Free |
| --- | --- |
| <video src="https://github.com/Automattic/pocket-casts-android/assets/4656348/a3710eb7-af98-4119-a0b2-132201476f30"/> | <video src="https://github.com/Automattic/pocket-casts-android/assets/4656348/479d7cab-04d4-4949-accf-af58315137e0"/> |

## Checklist
 
- [ ] with different themes


mchowning:
> I've noticed that occasionally when installing the app from Android Studio on top of a previous install, then logging out, and then logging back in with google, the app could get stuck on a logging in screen. I've only seen this a couple of times, and I have never reproduced it on a clean install. 

I havne't managed to reproduce this issue, but I improved the logic a bit in [f4ac9b8](https://github.com/Automattic/pocket-casts-android/pull/994/commits/f4ac9b821639aa04531224779838df5738ef1343), and I think that might help avoid the issue from happening.

ashiagr:
Great job in managing the tricky navigation required for this task!

I tested the app for each of the listed scenarios and found it working well:

Scenario | Free Account (logging in screen displayed, and user is returned to the UI prompting the user to log in) | Plus Account ( logging in screen displayed, and user is taken to the app once the refresh is complete)
-----|---|----
Logging in automatically because your watch-connected phone is logged in | ✅ | ✅ 
Logging in using a Google account | ✅ | ✅ 
Logging in using email and password | ✅ | ✅ 

> I've noticed that occasionally when installing the app from Android Studio on top of a previous install, then logging out, and then logging back in with google, the app could get stuck on a logging in screen. I've only seen this a couple of times, and I have never reproduced it on a clean install.

I was stuck at the login screen at times, I will continue to test and share the steps to reproduce once I find a pattern. It might have happened because the account might have many podcasts and I see we already have a task to address it:

> Add some kind of animation spinner to the logging in screen when an image is shown so the user knows something is happening if the refresh can take a really long time on long accounts (need to do some testing here).

but I think something else is also going on here.

I'm not approving the PR yet as I'd like to give some more time for testing it and also review the code changes.

mchowning:
This is working fine for me, but an additional scenario worth testing is:

1. Sign in with a Plus account
2. Change the account to being Free
3. Perform a manual refresh from the Settings screen
4. Verify that you are returned to the `RequirePlusScreen`

This is working fine for me, but I just wanted to call it out as an additional test scenario, so I'll add this to the PR description.

mchowning:
I found that there is an issue where when a user signs in with a google account, the app still uses the same google account and idToken on subsequent sign in attempts even if the sign in attempt occurs after the idToken has expired (1 hour expiration). This isn't something we're doing but is happening within the google sign in libraries and horologist. 

I think that's the cause of some of the issues I've seen (and why I found them so hard to reproduce). I've done an update in [519a189](https://github.com/Automattic/pocket-casts-android/pull/994/commits/519a1898593fc890f31ea4dcbb40112d17da6f03) that avoids that issue (and also fixes https://github.com/Automattic/pocket-casts-android/issues/993 by allowing the user to select a different account if they want when they sign back in). When the user selects an account, we get a fresh idToken.

ashiagr (approved):
> Sign in with a Plus account
> Change the account to being Free ...

This worked as expected, thanks for listing it as an additional testing scenario 👍 

> I found that there is an issue where when a user signs in with a google account, the app still uses the same google account and idToken on subsequent sign in attempts even if the sign in attempt occurs after the idToken has expired (1 hour expiration). This isn't something we're doing but is happening within the google sign in libraries and horologist.

> I think that's the cause of some of the issues I've seen (and why I found them so hard to reproduce). I've done an update in [519a189](https://github.com/Automattic/pocket-casts-android/pull/994/commits/519a1898593fc890f31ea4dcbb40112d17da6f03) that avoids that issue (and also fixes https://github.com/Automattic/pocket-casts-android/issues/993 by allowing the user to select a different account if they want when they sign back in). When the user selects an account, we get a fresh idToken.

You nailed it! 🌟 🌟 🌟 I was able to switch between Google accounts today. 

> I've noticed that when I change my emulator to have a slow internet connection, the refresh sometimes takes extremely long (as in, multiple minutes). It's weird because as I slow down the emulator, everything seems to work fine, and then suddenly there's a huge drop off in the speed once I slow it down past a certain point. I'm not sure what's going on there either, but just sharing in case it gives you any hints as you test further.

I think that might have been the issue. So that I got a better load and refresh performance, I tested the PR on a release apk and things were so much smoother. 

By now, I have tested it multiple times and it behaved correctly each time. I also smoke-tested phone app with different account types and it also worked correctly. Code changes look good as well. With this, I'm approving the PR.

PR #997, opened 2023-05-26
Pride: Add new default icon

ashiagr:
## Description

This adds pride icon as the default icon for the pride month. 

## Testing Instructions

1. Install `release/7.39` release build on a device
2. Change the default icon
3. Checkout this branch
4. Run the app on the device
5. ✅ The custom icon changed in step 2 should be displayed
6. Go to Settings > Appearance
7. ✅ The custom icon changed in step 2 should be selected
8. ✅ The Pride icon should be displayed after "Red Velvet" with the "Pride 2023" label
9. Tap on it
10. ✅ The icon should change
11. Fresh install the app from this branch
12. ✅ The pride icon should be displayed
6. Go to Settings > Appearance
7. ✅ The Pride icon should be the selected icon

## Instructions for 7.41

We'll revert back to the original icon in 7.41. In order to do so all that needs to be done is:

```
git revert 78f641f0d972af2b6215682639f498260570e00f
```

We'll keep the icon there in Settings though! 🏳️‍🌈

## Checklist
 
- [ ] with different themes


CookieyedCodes:
Any chance of a pride cat pockects icon? 😅

emilylaguna (commented):
Nice work, I tested:

- ✅ The default icon was changed when I upgraded from the previous version and I didn't have a custom icon chosen.
- ✅ When I upgraded and DID have a custom icon chosen my selection was maintained
- ✅ After reverting back to the old icon I am still able to choose Pride 2023
- ✅ I can switch between all the other app icons

I wasn't able to test automotive but I think once that's ✅ we'll be good to go.

mchowning:
> I wasn't able to test automotive but I think once that's ✅ we'll be good to go.

Confirmed that the icon is unchanged on Automotive, so I think this is good to go!

mchowning (approved):

gyund:
> > I wasn't able to test automotive but I think once that's ✅ we'll be good to go.
> 
> Confirmed that the icon is unchanged on Automotive, so I think this is good to go!

@mchowning FYI, in automotive the icon is changed. The default icon is changed to the new pride icon and the user's default/choice is ignored for automotive. Here's a screenshot from maps on mobile android 13 on a pixel 7 with the default icon selected as the default.



There's chatter in Reddit about confusion with the icon change and having user choice ignored. I'm guessing it's tied to the default icon being changed.

mchowning:
Thanks for the heads up @gyund ! 

Just to be perfectly clear, what you're talking about is Android Auto, [which is different from Android Automotive](https://source.android.com/docs/devices/automotive/start/what_automotive) (who came up with these names, right?!?!). I believe the red icon is still showing on Android Automotive, but you're right that the pride icon is showing for Auto now.

Pentaphon:
@ashiagr there is a bug in the implementation on android:

- you can change your launcher icon in Android but the pride icon will always be there. This differs from how it normally works. Basically makes pride the main icon and default secondary.
- in Android automotive, it defaults to the pride icon and ignores your preferences. It is slightly confusing if you're looking for red typically and takes a few seconds to adjust.
- you can also select the pride icon manually, giving you a single icon back

CookieyedCodes:
@Pentaphon this has been the case since icons were introduced tho so might be a limitation of android from my understanding, but if it can be fixed then great (I'll try get an example of another open source app that dose icons in app😉) some types of people also find the icon controversial unfortunately tho

 -to the devs sorry if you lose some some subs over this but please don't bow to them, do try and fix it tho :)-

gyund:
> - in Android automotive, it defaults to the pride icon and ignores your preferences. It is slightly confusing if you're looking for red typically and takes a few seconds to adjust.


As @mchowning mentioned above, this default applies to auto not automotive. Silly naming 😂

PR #1009, opened 2023-05-26
Audio ducking now supported during external notifications

### LINKED ISSUE #529: Audio ducking not working
### Description

I would expect Pocket Casts to duck its audio as in lowering its own volume while a notification sound is playing. I know this from Spotify and it's working great there so I always wondered why Pocket Casts doesn't have it.

There's a switch in settings to control what happens on notifications. Turning it off will make Pocket Casts pause playback while a notification sound is playing. Turning it will make Pocket Casts just continue playback while the notification plays but not lower its volume.

There seems to be code in place that by the looks of it should be handing that but somehow it's not working as it should. 

Turn on the switch in the settings
Start a podcast
Receive a notification

_No response_

Android 13 but was the same on Android 10 and 11
Pocket Casts 7.24

jamie23:
## Description

Added functionality for audio ducking when external notifications come in. There was already some logic there around it hence I've labeled this a fix.

Fixes #529

## Testing Instructions
1. Fresh install of the app.
2. Start the app.
3. Navigate to Profile, then Settings, then Notifications
4. Notice that there is no longer a toggle for 'Play over notifications' and it now launches a dialog
5. Select Never
6. Play a podcast then launch an external audible notification
     - To launch an external audible notification, I used Google maps and asked for turn by turn directions whilst a podcast was playing.
7. Notice that the apps audio stops playing while the audible notification is active and resumes after the notification stops.
8. Select the toggle for 'Play over notifications' again.
9. Select 'Duck volume' and repeat step 6.
10. Notice that the podcast volumes dips to half volume while the external notification plays and resumes full volume after the notification ends.
11. Select the toggle for 'Play over notifications' again.
12. Select 'Always' and repeat step 6.
13. Notice that the podcast audio plays throughout the external notification plays at full volume

## Screenshots or Screencast 
DuckAudio1
DuckAudio2
DuckAudio3

Updated the logs file to show the setting chosen:
DuckAudio4

## Checklist
 
- [x] with different themes


modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/NotificationsSettingsFragment.kt
```diff
@@ -140,6 +135,13 @@ class NotificationsSettingsFragment :
             )
             true
         }
+        playOverNotificationPreference?.setOnPreferenceChangeListener { _, newValue ->
+            analyticsTracker.track(
+                AnalyticsEvent.SETTINGS_NOTIFICATIONS_PLAY_OVER_NOTIFICATIONS_TOGGLED,
+                mapOf("enabled" to (newValue != "2"))
```
jamie23, line 141: One thing to note around this, in the old form the tracking is a boolean since the setting is a switch and is either on or off:
```
     AnalyticsEvent.SETTINGS_NOTIFICATIONS_PLAY_OVER_NOTIFICATIONS_TOGGLED,
     mapOf("enabled" to newValue as Boolean)
```

I assume you're using this behind the scenes to do analytics and so I can't just create a new event, so to keep it useful, I kept the event the same, so true is when the setting is duck or always and false is when the setting is never.

If you want me to improve this in some way please let me know!

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/FocusManager.kt
```diff
@@ -150,17 +150,24 @@ open class FocusManager(private val settings: Settings, context: Context?) : Aud
...
     }
 
-    protected open fun hasUserAllowedDucking(): Boolean {
-        return settings.canDuckAudioWithNotifications()
+    private fun canDuck(): PlayOverNotificationSetting {
+        if (audioFocus != AUDIO_NO_FOCUS_CAN_DUCK_TRANSIENT) return PlayOverNotificationSetting.NEVER
+        return playOverNotification()
+    }
+
+    protected open fun playOverNotification() = when (settings.getPlayOverNotification()) {
```
mchowning, line 162: I don't think this needs to be protected or open.

modules/services/localization/src/main/res/values/strings.xml
```diff
@@ -1090,8 +1090,10 @@
     <string name="settings_notification_choose_podcasts">Choose podcasts</string>
     <string name="settings_notification_hide_on_pause">Hide playback notification on pause</string>
     <string name="settings_notification_notify_me">Notify me</string>
-    <string name="settings_notification_play_over">Play over notifications</string>
-    <string name="settings_notification_play_over_summary">Keep playing even when other apps, like notifications or navigation, play sounds.</string>
+    <string name="settings_notification_play_over">Play over external notifications</string>
```
mchowning, line 1093: I noticed that you added "external" to this string. Do you find "notifications" on it's own to be unclear? Just asking because I'm curious.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
         ([#973](https://github.com/Automattic/pocket-casts-android/pull/973)).
     *   Improved the Automotive user profile view
         ([#975](https://github.com/Automattic/pocket-casts-android/pull/975)).
+    *   Added audio ducking as an option when playing over external notifications
+        ([#1009](https://github.com/Automattic/pocket-casts-android/pull/1009)).
```
mchowning, line 14: We'll need to merge in `main` and move this entry to the next release before we merge this.

mchowning (commented):
This is looking really good @jamie23 ! Thanks so much for the contribution! 🙇

I left a few comments, but there are no major issues. This works very well.

One thing I did notice as I was looking through the code is that it seemed like the logic around handling the different settings values was getting spread around to a lot of places (there are a lot of classes that need to know what `0`, `1`, and `2` mean. I really liked how you created a `PlayOverNotificationSetting` enum and thought we could maybe expand on that to consolidate the logic around those settings a bit and generally make the code a bit more readable. 

I put together a quick POC [here](https://github.com/Automattic/pocket-casts-android/commit/1b0a05b14b6573966d98251cdc04d5c4b1f387fa) to show what I had in mind. Don't feel like you have to use that though, feel free to do it in your own style or decline this suggestion if you'd rather leave it as-is.

larena1:
@jamie23 are you still working on this? It'd be great to have this 😊

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/NotificationsSettingsFragment.kt
jamie23: > we can also add a new property with the form of "value" to "never"/"duck"/"always". wdyt?

Yea that sounds good! Would you like me to add that now in this PR and you guys can add it behind the scenes when you get some time or would you rather me leave it out of this PR and you can add it in on both sides at the same time?

mchowning: Let's include it in this PR.

modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/SettingsImpl.kt
```diff
@@ -513,8 +513,12 @@ class SettingsImpl @Inject constructor(
...
     }
 
-    override fun canDuckAudioWithNotifications(): Boolean {
-        return sharedPreferences.getBoolean(Settings.PREFERENCE_OVERRIDE_AUDIO, false)
+    override fun getPlayOverNotification(): Int {
+        val value = sharedPreferences.getString(
+            Settings.PREFERENCE_NOTIFICATION_AUDIO,
+            Settings.PREFERENCE_NOTIFICATION_AUDIO_DEFAULT
+        ) ?: Settings.PREFERENCE_NOTIFICATION_AUDIO_DEFAULT
+        return Integer.parseInt(value)
```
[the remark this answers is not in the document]
jamie23: Yea great pickup and much more user friendly than a user losing their previous preference, I made an adjustment like so:

```
  override fun getPlayOverNotification(): PlayOverNotificationSetting {
        val value = sharedPreferences.getString(Settings.PREFERENCE_NOTIFICATION_AUDIO, null) ?: legacyPlayOverNotification()
        return PlayOverNotificationSetting.fromPreferenceString(value)
    }

    private fun legacyPlayOverNotification(): String {
        if (sharedPreferences.getBoolean(Settings.PREFERENCE_OVERRIDE_AUDIO, false)) {
            return PlayOverNotificationSetting.ALWAYS.preferenceInt.toString()
        }

        return PlayOverNotificationSetting.NEVER.preferenceInt.toString()
    }
```

So basically, check if they have a preference set, if not check then check if the legacy preference is set to true then return `Always` otherwise return `Never`.

I gave it a quick test on my device and it worked well for your test. Happy to modify if though if it can be improved further.

jamie23:
> This is looking really good @jamie23 ! Thanks so much for the contribution! 🙇
> 
> I left a few comments, but there are no major issues. This works very well.
> 
> One thing I did notice as I was looking through the code is that it seemed like the logic around handling the different settings values was getting spread around to a lot of places (there are a lot of classes that need to know what `0`, `1`, and `2` mean. I really liked how you created a `PlayOverNotificationSetting` enum and thought we could maybe expand on that to consolidate the logic around those settings a bit and generally make the code a bit more readable.
> 
> I put together a quick POC [here](https://github.com/Automattic/pocket-casts-android/commit/1b0a05b14b6573966d98251cdc04d5c4b1f387fa) to show what I had in mind. Don't feel like you have to use that though, feel free to do it in your own style or decline this suggestion if you'd rather leave it as-is.

Yea I really like what you've done with `PlayOverNotificationSetting`, really nice improvement.

modules/services/localization/src/main/res/values/strings.xml
jamie23: I think when I worked on it pre vacation I must have thought so, after some time away and re-reading it though I'm happy to leave "external" out of it.

mchowning (approved):
This looks good! Thanks!

PR #1024, opened 2023-05-31
Send Sentry additional data as tags

ashiagr:
## Description

This sends additional data for an exception "separately" to Sentry using [local scope](https://docs.sentry.io/platforms/python/guides/logging/enriching-events/scopes/#local-scopes). This helps in grouping the exception in Sentry. 
I've used `scope.tags` instead of `scope.extra` as [tags help in searching and grouping](https://docs.sentry.io/platforms/python/guides/logging/enriching-events/tags/?original_referrer=https%3A%2F%2Fduckduckgo.com%2F).

> Tags are key/value string pairs that are both indexed and searchable.

## Screenshots or Screencast 

Events with different episode uuids are grouped together:

Screenshot 2023-05-31 at 12 19 29 PM

Screenshot 2023-05-31 at 12 21 27 PM

We can filter using tags 

image

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
@@ -893,10 +893,8 @@ open class PlaybackManager @Inject constructor(
                 } else {
                     event.message
                 }
-                val isAutomotive = Util.isAutomotive(application)
                 Sentry.withScope { scope ->
                     episode?.uuid?.let { scope.setTag("episodeUuid", it) }
-                    scope.setTag("isAutomotive", isAutomotive.toString())
                     SentryHelper.recordException(
                         message = "Illegal playback state encountered",
                         throwable = event.error ?: IllegalStateException(event.message)
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
@@ -63,6 +63,7 @@ import io.reactivex.android.schedulers.AndroidSchedulers
 import io.reactivex.disposables.Disposable
 import io.reactivex.rxkotlin.subscribeBy
 import io.reactivex.schedulers.Schedulers
+import io.sentry.Sentry
 import kotlinx.coroutines.CoroutineScope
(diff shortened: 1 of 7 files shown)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
```diff
@@ -893,10 +894,14 @@ open class PlaybackManager @Inject constructor(
                     event.message
                 }
                 val isAutomotive = Util.isAutomotive(application)
-                SentryHelper.recordException(
-                    message = "Illegal playback state encountered for episode uuid ${episode?.uuid}, isAutomotive $isAutomotive}: ",
-                    throwable = event.error ?: IllegalStateException(event.message)
-                )
+                Sentry.withScope { scope ->
+                    episode?.uuid?.let { scope.setTag("episodeUuid", it) }
+                    scope.setTag("isAutomotive", isAutomotive.toString())
```
mchowning, line 899: Do you think it would be worth having this cover wear as well with something like "appPlatform" and give it a value of either "phone", "auto", or "wear"? I'm still not completely clear, but could it be worth setting this tag at a top-level so that every event would have the "appPlatform" tag?

mchowning (approved):
This looks great!

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
ashiagr: Good point, I added it in 68eefa557c19a3d441ee356907c2cf64c2430f3f.

I was thinking to add Sentry for Wear app in a separate PR but I added it in this PR itself in [ea1c4a5](https://github.com/Automattic/pocket-casts-android/pull/1024/commits/ea1c4a57d546f613b36a2dde0ed4c8cc1bc80f9f).

ashiagr: I tested on mobile and wear apps.

`app.platform` added as a global tag:

mobile

mobile

wear/build.gradle
```diff
@@ -43,8 +44,11 @@ android {
         }
 
         release {
-            minifyEnabled true
-            shrinkResources true
```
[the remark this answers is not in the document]
mchowning: >  I'll open a separate PR...

[PR opened](https://github.com/Automattic/pocket-casts-android/pull/1031)

PR #1025, opened 2023-05-31
Display errors on episode screen in watch app

ashiagr:
## Description

This displays episode (playback/download) errors on the episode details screen.

Task #1026

## Testing Instructions

1. Disable the internet on your watch device
2. Launch that app and go to the episode screen
3. Tap on the play icon (you're navigated to the Now Playing screen)
4. Wait for a while
5. ✅ Notice that the play icon switches to pause icon
6. Slide to the previous page in the pager to get back to the Episode screen
7. ✅ Notice that you see playback error message
8. Tap on download icon
9. ✅ Notice that the error is cleared
10. Wait for a while
11. ✅ Notice that you see a download error message and the download icon changes to retry icon

## Screenshots or Screencast 

Playback | Download 
--- | ---
image | image

## Checklist
 
- [ ] with different themes

### DIFF
--- a/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/episode/EpisodeScreen.kt
+++ b/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/episode/EpisodeScreen.kt
@@ -1,8 +1,10 @@
 package au.com.shiftyjelly.pocketcasts.wear.ui.episode
 
 import androidx.annotation.StringRes
+import androidx.compose.foundation.background
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
@@ -11,12 +13,14 @@ import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.font.FontWeight.Companion.W700
 import androidx.compose.ui.text.style.TextAlign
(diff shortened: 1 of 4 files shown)


wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/episode/EpisodeScreen.kt
```diff
@@ -283,6 +294,46 @@ private fun EpisodeListChip(episodeScreenItem: EpisodeScreenItem) {
...
+                modifier = Modifier.padding(padding)
+            )
+            errorData.errorDescription?.let {
+                Text(
+                    text = it,
+                    color = MaterialTheme.colors.onSecondary,
+                    style = MaterialTheme.typography.caption3,
+                    modifier = Modifier.padding(start = padding, end = padding, bottom = padding)
+                )
+            }
```
mchowning, line 332: It doesn't seem like the title adds much info. I feel like we might be better off just displaying the description without the title and only using the title if the description is null.

mchowning (commented):
The error message looks nice, but just having this on the episode details screen seems like it would be very hard to find when it happens as a result of playback error. I think it would be better to either present a toast or a full-screen error message. wdyt? Obviously that could be in a different PR.

Also, on my devices the download was never failing, it would just sit in the pending state forever. I don't think that has anything to do with your changes, but I wanted to see if there was anything unusual you might have done to get it to fail since obviously you're getting the failure state.

wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/episode/EpisodeScreen.kt
ashiagr: Changed in 53854a0

Screenshot_20230601_152750

ashiagr:
> The error message looks nice, but just having this on the episode details screen seems like it would be very hard to find when it happens as a result of playback error. I think it would be better to either present a toast or a full-screen error message. wdyt? 


Listeners might not be looking at the screen while playing the song, an auto-dismissible toast wouldn't serve the purpose I think. Also, a blocking full-screen error message doesn't look like a good idea. I can replicate mobile app behavior and display the error on the episodes list too. It will remain there until the error is cleared when the episode is next played. Here's the commit with the changes: [6ed4fb3](https://github.com/Automattic/pocket-casts-android/pull/1025/commits/6ed4fb3931041b21c7991d609d25698f7abc0203). Lmkwyt.

Wear app

Episodes List | Episode Details
----|----
 | 


Mobile app

Episodes List | Episode Details
----| ----
 |

mchowning:
> Listeners might not be looking at the screen while playing the song, an auto-dismissible toast wouldn't serve the purpose I think. Also, a blocking full-screen error message doesn't look like a good idea. I can replicate mobile app behavior and display the error on the episodes list too. It will remain there until the error is cleared when the episode is next played. Here's the commit with the changes: [6ed4fb3](https://github.com/Automattic/pocket-casts-android/pull/1025/commits/6ed4fb3931041b21c7991d609d25698f7abc0203). Lmkwyt.

I get those points, but I still think that it would be good if a listener on the Now Playing screen got some kind of information about why the play button isn't working for them. That doesn't need to block this PR though which is a great improvement 🙇, so I created [an issue](https://github.com/Automattic/pocket-casts-android/issues/1030) to track this. Really nice work here!

mchowning (approved):

PR #1031, opened 2023-06-01
Shrink resources in release builds

mchowning:
## Description
This adds resource shrinking to the release builds for all of our apps. This doesn't make a huge difference, but it helps a bit, and it is [encouraged by Google](https://developer.android.com/build/shrink-code). Let me know if you know of a reason we don't want to do this. 

Here are the sizes in megabytes of the release apks I created with and without shrinking on my local machine.

| | Before | With Shrinking | Difference |
| --- |  --- | --- | --- |
| phone |  29.12 | 29.07 | -0.05 |
| automotive | 36.61 | 36.56 | -0.05 |
| wear |  28.87 | 28.78 | -0.09 |

I checked the private commit history to make sure there wasn't a reason we removed this at some point, and I didn't find anything.

## Testing Instructions
Ssmoke test the release builds for all 3 of our apps.

## Checklist
 
- [ ] with different themes


mchowning:
This change should probably go in the `base.gradle` file. I'll make this as a draft until we make that update.

mchowning:
> This change should probably go in the base.gradle file. I'll make this as a draft until we make that update.

Tried to do this and was reminded why I didn't do it initially: `shrinkResources` cannot be used on libraries, so the app will fail to build if I do that in `base.gradle`. I'll reopen this for review.

geekygecko (approved):
I assembled release versions and tried them on Automotive, Wear OS, and the mobile app. They all worked well. Thanks for making this change.

Issue #1008, opened 2023-05-26
Add background syncing to watch app


mchowning:
Currently the watch app only syncs when a user logs in or manually refreshes from the settings screen. Let's set up some background syncing like we have on the phone. For the initial iteration of this, it might be able to be as simple as just refreshing when the watch app is opened.

PR #1036, opened 2023-06-02
Refresh podcasts in background in watch app 

ashiagr:
## Description
This adds a background refresh toggle in settings and refreshes podcasts in the background if the toggle is set to on. Initially, the toggle is set to off to prevent unnecessary background processing for the watch app.

Task https://github.com/Automattic/pocket-casts-android/issues/1008

## Testing Instructions

1. Apply [this patch](https://github.com/Automattic/pocket-casts-android/files/11634560/Reduce_background_refresh_time.patch) to reduce refesh initial and periodic interval. [Note that [periodic delay lesser than 15 minutes](https://android.googlesource.com/platform/frameworks/support/+/androidx-1.0-dev/work/workmanager/src/main/java/androidx/work/impl/model/WorkSpec.java#163) is not allowed.]
3. Fresh install the app 
4. Login with a plus account having podcasts
5. Go to Settings
6. Notice that you see "Background refresh" toggle which is set to off
7. Clear logs and go to recent apps
8.  Search `BgTask: RefreshPodcastsTask` in logs
9. Notice that you do not see `BgTask: RefreshPodcastsTask - runNow - Start` in logs 
10. Reopen the app
11. Go to Settings 
12. Toggle on "Background refresh"
13. Clear logs and go to recent apps
14. Notice that you see `BgTask: RefreshPodcastsTask - runNow - Start` in logs 
15. (Optional) Wait for the periodic interval set in the patch in Step 1. [Step 15, 16 are optional, you can see results in the screen shot included below.]
16. (Optional) Notice that you again see `BgTask: RefreshPodcastsTask - runNow - Start` in logs

## Screenshots or Screencast 

logs

## Checklist
 
- [ ] with different themes


mchowning (approved):
These changes look good.

It seems like the text is not entirely accurate. For example, it seems to me that with background refresh turned on, the app refreshes every time the app comes from the background, but if it is turned off it will wait at least 5 minutes before refreshing again.

With that said, this is the same setting and text that we use in the phone app, so for now I think it's fine to leave it as-is and we can iterate on it.

Similarly, the length of the text isn't ideal for a watch, but it feels like having some explanatory text is good. Maybe there's a different UI pattern (opening a full screen toggle or using a toast message) that we could use here in the future (cc: @david-gonzalez-a8c ).

Issue #1006, opened 2023-05-26
Improve syncing/refresh performance of watch app


mchowning:
Syncing the watch app can be very slow with large accounts.

We might just need to use the [horologist network awareness](https://google.github.io/horologist/network-awareness/) library to prefer wifi when it is available [like we do for downloads](https://github.com/Automattic/pocket-casts-android/pull/881).

PR #1037, opened 2023-06-02
Update/incorporate network awareness library

mchowning:
## Description
This expands the usage of the network awareness library to be used for (almost) all calls made from the watch app to [improve the app's syncing performance](https://github.com/Automattic/pocket-casts-android/issues/1006). Before this PR we were only using network-awarenss [for downloads](https://github.com/Automattic/pocket-casts-android/pull/881). 

> **Note**
> This PR does **not** include updates to our ExoPlayer usage, so streaming episodes is not using network-awareness. That's probably ok because using the default (and low-bandwidth) Bluetooth connection for streaming when it is available is better for the watch's battery, but I'll do some testing of this.

## Testing Instructions
Test various flows in the app making sure that the logs show that an appropriate network is being used for each call (filtering your logs for "https request" can help here).
- logging in
- performing a refresh
- download an episode
- loading show notes by visiting an episode page

## Checklist


mchowning:
Closing this since we don't have any immediate plans to make this switch.

PR #1038, opened 2023-06-02
Switching to use a library catalog

geekygecko:
## Description

Fixes # 

## Checklist
 
- [ ] with different themes


mchowning:
I believe this has been superseded, so I'll close this. Feel free to re-open if I'm wrong about that though.

Issue #1007, opened 2023-05-26
Add Tracks for initial watch release


mchowning:
We want to be able to see how the watch app is being used so we can make it better. That means we need to get some data.

- [ ] Add tracks for watch screens (screen shown and direct user interactions)
- [ ] Validate that the tracking events we've already implemented for the phone work on the watch
- [ ] Add a setting/toggle allowing users to disable analytics

PR #1040, opened 2023-06-02
Tracks for watch

mchowning:
## Description
This turns on Tracks in the watch app (issue https://github.com/Automattic/pocket-casts-android/issues/1007).

In addition, I am adding a global `platform` property to identify if a given event is coming from a watch, phone, or automotive app. I thought this was better than adding duplicate events/sources for everything that happens on the phone as well as the watch. Let me know if you think this isn't the right way to go though.

Lastly, this PR includes a fix for the user signed in events. Before this PR, the `user_signed_in_watch_from_phone` event was sent for all sign-ins on the watch. This updates that so that event is only sent when the user is automatically signed into the watch from the phone. All user-initiated sign-ins on the watch now send the `user_signed_in` event.

This PR does **not** add any new events to the watch app.

## Testing Instructions

1. Check and make sure that events from the watch is firing correctly and has an `platform` property of "watch" for each event. Here are the events I see now working in the watch app (there may be more that I missed):

    1. episode_added_to_up_next
    1. episode_removed_from_up_next
    1. episode_archived
    1. episode_unarchived
    1. episode_download_queued
    1. episode_download_cancelled
    1. episode_download_finished
    1. episode_marked_as_played
    1. episode_marked_as_unplayed
    1. episode_starred
    1. episode_unstarred
    1. playback_play
    1. playback_pause
    1. playback_skip_back
    1. playback_skip_forward
    1. user_signed_in
    1. user_signed_out

2. In addition, make sure these events are getting logged in the Tracks backend. I did this by creating my own unique event that got sent off from a release build, and then looking for that event.

3. Check that the `user_signed_in` event is sent if a user logs in with google or email and that the `user_signed_in_watch_from_phone` event is sent if the user is signed in automatically from the phone.

4. Spot-check a few events on the phone to make sure they have the `platform` property of "phone".

## Checklist
 
- [ ] with different themes

### DIFF
--- a/app/src/main/java/au/com/shiftyjelly/pocketcasts/PocketCastsApplication.kt
+++ b/app/src/main/java/au/com/shiftyjelly/pocketcasts/PocketCastsApplication.kt
@@ -45,7 +45,6 @@ import io.sentry.protocol.User
 import kotlinx.coroutines.DelicateCoroutinesApi
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.GlobalScope
-import kotlinx.coroutines.MainScope
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.withContext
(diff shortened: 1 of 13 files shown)


ashiagr (approved):
Nice job in adding global `platform` property! Checked all events are fired as described ✅

PR #1041, opened 2023-06-02
Track application lifecycle on watch

mchowning:
## Description
This PR enables the application lifecycle tracking of the watch app (issue https://github.com/Automattic/pocket-casts-android/issues/1007).

As part of this PR, I have added a "shared" module for code that is shared between the automotive, phone, and wear apps. I've had a few times come up where I thought this would be helpful, so I think it will be worthwhile even apart from the changes in this PR, but let me know if you disagree.

## Testing Instructions
1. Install the watch app
2. Launch the watch app
3. ✅  Observe the `application_installed` and `application_opened` events are fired
4. Background the wear app
5. ✅ Observe the `application_closed` event is fired

## Checklist
 
- [ ] with different themes


ashiagr:
> As part of this PR, I have added a "shared" module for code that is shared between the automotive, phone, and wear apps.

Now that we have three different platforms to support, I am on board on having a shared module. 👍 

> Observe the application_installed and application_opened events are fired

If I force-stop the application and restart it, I see the `application_installed` event getting fired. I think it's because we're not calling `VersionMigrationsJob` for wear app.

ashiagr (approved):
`application_installed` event is properly tracked after adding the version migrations job in c0323e5.
I'm approving the PR and merging it. Feel free to reopen it if you see any issues, @mchowning

PR #1042, opened 2023-06-02
Adding tracks for the watch's main list screen

mchowning:
## Description
This PR adds tracks to the main list screen on the watch (issue https://github.com/Automattic/pocket-casts-android/issues/1007).

## Testing Instructions
1. Install the watch app and log in so that you are brought to the main list screen
2. ✅ Observe that the `wear_main_list_shown` event is sent
3. Tap on each of the buttons on the main list screen
4. ✅ Observe that for each button the appropriate `wear_main_list_[button]_tapped` event is sent.

## Checklist


ashiagr (approved):
Looks good!

PR #1043, opened 2023-06-03
Add privacy settings for watch

mchowning:
## Description
This adds user privacy options for the watch (issue https://github.com/Automattic/pocket-casts-android/issues/1007). For now I'm just going with the same settings we have on the phone.

## Testing Instructions

The privacy settings are available under `Settings` → `Privacy`. 
- Verify that the default settings have "Analytics" and "Crash Reports" toggled on, with "Link your account" toggled off
- Verify that Tracks are only sent off if "Analytics" is toggled on
- Verify that crash reports are only sent off if "Crash reports" is toggled on
- Verify that the crash reports only includes the users email address if "Link your account" is toggled on

## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/4656348/d965a1e6-d09d-4974-b7ee-1018200114cd

#### Crash report
| Account unlinked | Account linked |
| --- | --- |
| image | image |

## Checklist

### DIFF
--- a/modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/privacy/PrivacyFragment.kt
+++ b/modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/privacy/PrivacyFragment.kt
@@ -62,7 +62,7 @@ class PrivacyFragment : BaseFragment() {
                             viewModel.updateAnalyticsSetting(it)
                         },
                         onCrashReportsClick = {
-                            viewModel.updateCrashReportsSetting(context, it)
+                            viewModel.updateCrashReportsSetting(it)
                         },
                         onLinkAccountClick = {
                             viewModel.updateLinkAccountSetting(it)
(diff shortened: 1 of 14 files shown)


ashiagr (approved):
Looks good, I just guarded the initial Sentry dsn setting with the crash log setting in 1439b07.

mchowning:
> Looks good, I just guarded the initial Sentry dsn setting with the crash log setting in https://github.com/Automattic/pocket-casts-android/commit/1439b076d293cbce80c03b861bb4c6c9c929abb7.
> 
> I also made changes in the Sentry setup on app create to use email settings value in [2fd7f48](https://github.com/Automattic/pocket-casts-android/pull/1043/commits/2fd7f48908d5eaf39d369d17ff7d0bbffc75a511).

Thanks for the updates! 🙇

mchowning:
I've added the call to `refreshMetadata` in https://github.com/Automattic/pocket-casts-android/pull/1043/commits/955918566477f97d6ae07fc17d4faab0835b376e, and I also moved the `Firebase.initializeApp` call so that it is in the same place it is in the phone's Application class. There isn't any functional difference with the , but it's nice to have it consistent so it's easier to spot differences between the two application classes). Let me know if those changes look good to you @ashiagr.

PR #1044, opened 2023-06-03
Use radioactive icons for debug builds on wear and automotive apps

mchowning:
## Description
This uses the radioactive icon by default for debug builds of the automotive and wear apps like we do for the phone app currently. In addition, this uses the "Pocket Debug" app name for the debug builds like the phone app does. It's so much easier to differentiate the apps on your device during development with these changes.

## Testing Instructions
1. Verify that debug builds of the wear and automotive apps have the green radioactive icon by default.
2. Verify that release builds of the wear and automotive apps still have the regular icon by default

## Screenshots or Screencast 

image

image

## Checklist

### DIFF
--- a/automotive/build.gradle
+++ b/automotive/build.gradle
@@ -25,6 +25,7 @@ android {
             applicationIdSuffix '.debug'
 
             manifestPlaceholders = [
+                    appIcon: "@mipmap/ic_launcher_radioactive",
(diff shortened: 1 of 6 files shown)


ashiagr (approved):
>  It's so much easier to differentiate the apps on your device during development with these changes.

I agree, thanks!

Issue #1045, opened 2023-06-03
Rotary spinner does not scroll lists on watch app


mchowning:
### Description

Turning the physical rotary spinner on the watch app does not scroll lists like the main list. Seems like it should.

### Step-by-step reproduction instructions

1. Launch the watch app on a Pixel watch
2. Go to the main list screen on the app (or pretty much any other screen with a list)
3. Turn the rotary spinner
4. Observe that nothing happens

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

Pixel Watch, 7.40 pre-release build

mchowning:
This has been addressed by https://github.com/Automattic/pocket-casts-android/pull/1046

PR #1046, opened 2023-06-03
Support scrolling with rotary input on watch list

mchowning:
## Description
This PR is adding [support for the rotary input to scroll the screen](https://github.com/Automattic/pocket-casts-android/issues/1045) on the main watch list screen.

I think we just need to import/use the horologist scrollable component instead of the foundation compose one to get this. I'm only making this change on a single screen in this PR so it can be reviewed before we make the same change to all th other screens. I figured we can do that in a follow-up PR, or by updating this PR, but I wanted to get some eyes on this first before investing the additional time.

## Testing Instructions
1. Go to the main watch list screen
2. Verify that you can scroll the list using the rotary spinner

## Checklist

### DIFF
--- a/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/MainActivity.kt
+++ b/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/MainActivity.kt
@@ -150,7 +150,7 @@ fun WearApp(
                 ScrollToTop.handle(navController, it.scrollableState)
 
                 WatchListScreen(
-                    scrollState = it.columnState,
+                    columnState = it.columnState,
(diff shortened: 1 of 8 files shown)


ashiagr (approved):
This looks good to me. With the foundation compose component, we will need to add a modifier with rotary explicitly which comes inbuilt with the horologist one.

ashiagr:
I went ahead and replaced foundation compose scrollable component with horologist scrollable component on other screens: podcast, podcasts, upnext, filter, filters. 

Download, files, episode screens are already using the horologist component.

ashiagr:
Overall it looks good, I'm merging it @mchowning.

PR #1056, opened 2023-06-05
Add tracks for sign in screens on watch

mchowning:
## Description
This adds tracks for the require plus screen and the login screen on the watch (relevant issue: https://github.com/Automattic/pocket-casts-android/issues/1007).

## Testing Instructions
1. Install the watch app (make sure you're not logged into PC on a connected phone)
2. ✅  Observe that when the require plus screen is shown the `wear_require_plus_shown` event is tracked
3. Proceed to the login screen
4. ✅ Observe that the `wear_signin_shown` event is tracked
5. Tap on each of the signin options and observe that the following events are tracked when appropriate:
    - `wear_signin_google_tapped`
    - `wear_signin_phone_tapped`
    - `wear_signin_email_tapped`

## Checklist


ashiagr (approved):
All events were fired correctly 👍

PR #1057, opened 2023-06-06
Improve error message during login when no network is present

### LINKED ISSUE #1027: Improve error message when google sign on fails because of no internet connection
### Description

When there is no network connection available (such as with airplane mode on), signing in with either email or a google account displays a generic error message. 

In the case of email, the message is "Unable to login to your sync account, sorry. Please try again later". In the case of google sign in, the message is "Unable to launch Google sign in."

It would be nice if we checked if a network connection was available and displayed a more specific and actionable message along the lines of, "No network connection detected. Please try again later."

1. Open the app
2. Turn on airplane mode
3. Attempt to sign into the app
4. Observe that a generic error message is provided.

_No response_

Samsung A13 running a 7.39 pre-release build.

atharv028:
## Description

When there is no network connection available, the app used to display a generic message.

In the case of email, the message is "Unable to login to your sync account, sorry. Please try again later". In the case of google sign in, the message is "Unable to launch Google sign in."

This PR just modifies that message to a more specific message "No network connection detected. Please try again later.".

Fixes #1027 

## Testing Instructions

1. Open the app
2. Turn on Airplane Mode
3. Attempt to Sign in using either google or email.
4. Observe that a specific message telling that no network connection detected is displayed.

## Screenshots

Screenshot_20230606_125011

### DIFF
--- a/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingLoginPage.kt
+++ b/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingLoginPage.kt
@@ -113,13 +113,13 @@ internal fun OnboardingLoginPage(
                 modifier = Modifier.padding(16.dp),
             )
 
-            if(state.showNetworkError){
+            if (state.showNetworkError) {
                 TextP40(
                     text = stringResource(id = LR.string.log_in_no_network),
                     color = MaterialTheme.theme.colors.support05,
                     modifier = Modifier.padding(horizontal = 16.dp)
                 )
-            }else if(state.serverErrorMessage != null){
+            } else if (state.serverErrorMessage != null) {
                 TextP40(
                     text = state.serverErrorMessage!!,
                     color = MaterialTheme.theme.colors.support05,
--- a/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingLoginPage.kt
+++ b/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingLoginPage.kt
@@ -23,7 +23,6 @@ import androidx.compose.runtime.getValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
-import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.platform.LocalView
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.tooling.preview.Preview
@@ -41,7 +40,6 @@ import au.com.shiftyjelly.pocketcasts.compose.components.TextP40
 import au.com.shiftyjelly.pocketcasts.compose.preview.ThemePreviewParameterProvider
 import au.com.shiftyjelly.pocketcasts.compose.theme
 import au.com.shiftyjelly.pocketcasts.ui.theme.Theme
-import au.com.shiftyjelly.pocketcasts.utils.Network
 import au.com.shiftyjelly.pocketcasts.views.helper.UiUtil
 import com.google.accompanist.systemuicontroller.rememberSystemUiController
 import au.com.shiftyjelly.pocketcasts.localization.R as LR
@@ -58,9 +56,6 @@ internal fun OnboardingLoginPage(
 
     val viewModel = hiltViewModel<OnboardingLogInViewModel>()
     val state by viewModel.state.collectAsState()
-    val context = LocalContext.current
-
-    val networkErrorMessage = stringResource(id = LR.string.log_in_no_network)
 
     CallOnce {
         viewModel.onShown()
@@ -111,21 +106,22 @@ internal fun OnboardingLoginPage(
                 showEmailError = state.showEmailError,
                 showPasswordError = state.showPasswordError,
                 enabled = state.enableSubmissionFields,
-                onDone = {
-                    viewModel.updateServerErrorMessage(
-                        if (Network.isConnected(context)) null else networkErrorMessage
-                    )
-                    viewModel.logIn(onLoginComplete)
-                },
+                onDone = { viewModel.logIn(onLoginComplete) },
                 onUpdateEmail = viewModel::updateEmail,
                 onUpdatePassword = viewModel::updatePassword,
                 isCreatingAccount = false,
                 modifier = Modifier.padding(16.dp),
             )
 
-            state.serverErrorMessage?.let {
+            if(state.showNetworkError){
                 TextP40(
-                    text = it,
+                    text = stringResource(id = LR.string.log_in_no_network),
+                    color = MaterialTheme.theme.colors.support05,
+                    modifier = Modifier.padding(horizontal = 16.dp)
+                )
+            }else if(state.serverErrorMessage != null){
+                TextP40(
+                    text = state.serverErrorMessage!!,
                     color = MaterialTheme.theme.colors.support05,
                     modifier = Modifier.padding(horizontal = 16.dp)
                 )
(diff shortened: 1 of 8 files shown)


mchowning (commented):
Thanks for the contribution @atharv028 ! These changes look good.

The only thing I noticed is that with your changes the error message has been updated when logging in using Google on the phone and when logging in using email/password on the watch.

| Login type | Google | Email |
| --- | --- | --- |
| Phone app | ✅  | |
| Watch app | | ✅  |

It would be great if we could also update the message that is displayed for email/password on the phone (I think [this](https://github.com/Automattic/pocket-casts-android/blob/09c8136193d1852d222a8453a0dff26e2d472c68/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/viewmodel/OnboardingLogInViewModel.kt#L77) may be the spot that needs an update), and when using Google on the watch (I think we just need a slightly more intelligent error screen [here](https://github.com/Automattic/pocket-casts-android/blob/519a1898593fc890f31ea4dcbb40112d17da6f03/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/authentication/LoginWithGoogleScreen.kt#L32)).

With that said, these changes are already a nice improvement, so if you'd rather just merge this as-is let me know and we can just keep https://github.com/Automattic/pocket-casts-android/issues/1027 open.

atharv028:
Hey @mchowning, thanks for pointing out the bug.

I have fixed this with my latest commit, so now all the screens of Wear and Phone app with email/google sign in will show a error screen/message when the user tries to login without an active network connection.

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingLoginPage.kt
```diff
@@ -140,7 +150,12 @@ internal fun OnboardingLoginPage(
             RowButton(
                 text = stringResource(LR.string.onboarding_log_in),
                 enabled = state.enableSubmissionFields,
-                onClick = { viewModel.logIn(onLoginComplete) },
+                onClick = {
+                    viewModel.updateServerErrorMessage(
+                        if (Network.isConnected(context)) null else networkErrorMessage
+                    )
+                    viewModel.logIn(onLoginComplete)
```
mchowning, line 157: Checking the network availability here and then having the view model's login method check if the error message is already null or not feels like it has a bit too many moving parts than would be ideal (for example, you have to call the methods in the right order for everything to work). I prefer to drop the `viewModel.updateServerErrorMessage` method and have the `logIn` method check (and handle) the lack of a network connection directly. We could either add a parameter to that method indicating whether the network is available or update the view model to have access to context so we can check it in there with something along these lines:

``` diff
class OnboardingLogInViewModel @Inject constructor(
     private val podcastManager: PodcastManager,
     private val subscriptionManager: SubscriptionManager,
     private val syncManager: SyncManager,
-) : ViewModel(), CoroutineScope {
+    @ApplicationContext appContext: Context,
+) : AndroidViewModel(appContext as Application), CoroutineScope {
 
         get() = Dispatchers.Default

```

What do you think?

wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/authentication/LoginWithGoogleScreen.kt
```diff
@@ -29,7 +35,13 @@ fun LoginWithGoogleScreen(
             Timber.i("Google sign in cancelled")
             onAuthCanceled()
         },
-        failedContent = { AuthErrorScreen() },
+        failedContent = {
+            if(!Network.isConnected(context)){
+                ErrorScreen(stringResource(LR.string.log_in_no_network))
+            }else {
+                AuthErrorScreen()
+            }
```
mchowning, line 43: Instead of using different error screens here, what would you think about always using the `ErrorScreen` you created and instead just having the text vary?

``` kotlin
import com.google.android.horologist.auth.composables.R as HR

//...

failedContent = {
    val message = if (Network.isConnected(context)) {
        HR.string.horologist_auth_error_message
    } else {
        LR.string.log_in_no_network
    }
    ErrorScreen(stringResource(message))
},
```

mchowning (commented):
This is looking good @atharv028, and everything works great. I have a couple of suggestions for the code, but they're nothing major.

> If we take this issue one step further, then we can also add a more personalized error message in case of network failure during sign up in phone app (I think we need to modify the code [here](https://github.com/Automattic/pocket-casts-android/blob/3f4485c2cb9dbac75ee92932dca87b9333cc2e04/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/CreateEmailFragment.kt#LL113C22-L113C22))

👍 That would be great!

mchowning:
There are some formatting violations causing CI to fail. Running `./gradlew spotlessCheck` or `./gradlew spotlessApply` will take care of those.

wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/authentication/LoginWithGoogleScreen.kt
atharv028: yes, I think that would be an optimal solution for maintaining consistency throughout the app.

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingLoginPage.kt
mchowning: > One suggestion that I had in mind is instead of using Network.isConnected() and passing the context everywhere, we could just change the Network object into a class and inject it where ever we want to use it. This way we can directly use the constructor injection for the viewmodel and field injection if we require that in any composable file. We could name the class something along the lines of "NetworkManager".
>
> Sounds good?

👍 Sounds great!

atharv028:
I have improved the code as per your suggestions.

As far as converting `Network.kt` file to `NetworkManager.kt`, it would be better to create a new issue.

atharv028:
Also added same message when the user tries to create an account with no network connection.

mchowning:
👋 @atharv028 ! This is looking pretty good. I pushed two commits to your branch. 

- https://github.com/atharv028/pocket-casts-android/commit/8a9b4303e47db3a5e04835737af11209a9c016aa moves the error message logic out of the composable.
- https://github.com/atharv028/pocket-casts-android/commit/aaa3d7e49f810e33975b763c832554d765372ab6 removes the errorMessage check in the login method. This avoided a bug where a server error message would never get cleared (because it could only be cleared after the point in the method where we exited the method early if it was not null.

Take a look at them and let me know what you think. If they look good to you I think we can merge this.

mchowning:
Weird, I pushed my commits to [your branch](https://github.com/atharv028/pocket-casts-android/commits/fix/no_network_login), but I'm not seeing them show up in this PR for some reason. I thought that pushing to the branch alone would be sufficient, but maybe there's something you need to do on your end for the PR to get updated? 🤔

mchowning:
> I pushed my commits to [your branch](https://github.com/atharv028/pocket-casts-android/commits/fix/no_network_login), but I'm not seeing them show up in this PR for some reason.

Looks like GitHub is having an incident relating to pull requests right now ([status page](https://www.githubstatus.com/)). I bet that's the problem.

image

mchowning:
> Currently we are only checking for validEmail and validPassword format and not checking for an active network connection. I think this can easily be fixed by just adding the following condition [here](https://github.com/atharv028/pocket-casts-android/blob/aaa3d7e49f810e33975b763c832554d765372ab6/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/viewmodel/OnboardingCreateAccountViewModel.kt#L64) and making isNetworkAvailable public.

Makes sense. You're right that we don't need to be attempting the call when there isn't a network connection.

mchowning:
When you have a chance, if you can fix the merge conflict, I'll merge this.

mchowning (approved):
Looks good! 👍

PR #1059, opened 2023-06-06
Added validations for email/password wear app

### LINKED ISSUE #983: Add validation to email/password in wear app
### Description

Currently, the wear app does not do any validation of the email and password—it just sends off whatever the use enters to the server. If there is an error, it presents the server error or [defaults to](https://github.com/Automattic/pocket-casts-android/blob/989c3237d4e83be8b9cdb2d56b031ab309dd41d9/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/authentication/LoginWithEmailScreen.kt#L82) "Unable to login to you sync account, sorry. Try again later."

It would be better to validate that we have a valid email and a password that is long enough and present more relevant error messages like we do on the phone app.

image

h/t to @geekygecko for [noting this](https://github.com/Automattic/pocket-casts-android/pull/981#pullrequestreview-1433681964)

1. Log into the watch app with an invalid email or password
2. Observe the generic error message

_No response_

Watch emulator running pre-release 7.30 pre-release build

atharv028:
## Description

The validations for email and password were present but the message was not displaying due to some bug in the logic. This PR fixes that logic and allows the Wear app to show error screens for invalid email/password.

Fixes #983 

## Testing Instructions

1. Open the app.
2. Click on Login with email.
3. Observe that whenever you enter invalid email or password less than 6 characters, you get the error message accordingly.

## Screenshots

|                                                               Credentials                                                              |                                                                 Screens                                                                |
|:--------------------------------------------------------------------------------------------------------------------------------------:|:--------------------------------------------------------------------------------------------------------------------------------------:|
| Screenshot_20230606_190607 | Screenshot_20230606_190618 |
| Screenshot_20230606_190707 | Screenshot_20230606_190714 |

## Checklist

### DIFF
--- a/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/authentication/LoginWithEmailScreen.kt
+++ b/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/authentication/LoginWithEmailScreen.kt
@@ -54,7 +54,7 @@ fun LoginWithEmailScreen(
                 LaunchedEffect(Unit) {
                     launchRemoteInput(label, launcher)
                 }
-            } else if(password.isNullOrEmpty()) {
+            } else if (password.isNullOrEmpty()) {
                 val label = stringResource(LR.string.enter_password)
                 val launcher = getLauncher {
                     viewModel.updatePassword(it)
@@ -62,7 +62,7 @@ fun LoginWithEmailScreen(
                 LaunchedEffect(Unit) {
                     launchRemoteInput(label, launcher)
                 }
-            }else{
+            } else {
                 viewModel.signIn()
             }
         }
@@ -80,7 +80,11 @@ fun LoginWithEmailScreen(
         is SignInState.Failure -> {
             loading = false
             val currentState = signInState as? SignInState.Failure
-            val message = (currentState?.message ?: currentState?.errors?.last()?.message?.let { stringResource(id = it) }) ?: stringResource(id = LR.string.error_login_failed)
+            val message = currentState?.message
+                ?: stringResource(
+                    currentState?.errors?.last()?.message
+                        ?: LR.string.error_login_failed
+                )
             ErrorScreen(message)
         }
     }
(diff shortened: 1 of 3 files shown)


wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/authentication/LoginWithEmailScreen.kt
```diff
@@ -77,9 +79,8 @@ fun LoginWithEmailScreen(
 
         is SignInState.Failure -> {
             loading = false
-            val message = (signInState as? SignInState.Failure)
-                ?.message
-                ?: stringResource(LR.string.error_login_failed)
+            val currentState = signInState as? SignInState.Failure
+            val message = (currentState?.message ?: currentState?.errors?.last()?.message?.let { stringResource(id = it) }) ?: stringResource(id = LR.string.error_login_failed)
```
mchowning, line 83: Doing all of this on a single line makes it very hard to follow the logic. Can we break this line up a bit?

Also, correct me if I'm wrong, but I dont' think we need the round brackets around `(currentState?.message ?: currentState?.errors?.last()?.message?.let { stringResource(id = it) })`

If I were writing this, I'd probably do it like this, but this certainly not the only way to go (just sharing as an example, I'm not saying it has to be like this):

```kotlin
val message = currentState?.message
    ?: currentState?.errors?.last()?.message?.let { stringResource(id = it) }
    ?: stringResource(id = LR.string.error_login_failed)
```

atharv028: I did feel like that doing this in one line could make it confusing.

Anyways, considering your approach it looks much better than mine. I tried to make it more concise by writing it up like this:

```kotlin
val message = currentState?.message
                ?: stringResource(
                    currentState?.errors?.last()?.message 
                    ?: LR.string.error_login_failed
                )
```
Does this look good?

mchowning (approved):
Looks good! Thanks for the contribution!

Issue #1047, opened 2023-06-03
Logging in to watch takes a long time


mchowning:
### Description

There are two issues here:
1. When logging in using a Google account, the watch app takes a very long time to get the token from Google (I've seen this take 4 minutes).
2. Performing all of our API calls to download the account data after we get the Google account can easily take 1-2 minutes (not surprisingly, it seems to take longer with larger accounts).

Some things I've noticed:
- This is only really an issue on physical devices, emulators are much, much faster, to the point where it is not a noticeable issue on emulators
- If you log out and log back into the same account, the second log in goes _much_ more quickly, to the point where it is not a noticeable issue.
- Release builds are faster debug builds, but I'm still see significant slowness on release builds

### Step-by-step reproduction instructions

1. Do a fresh install of the app on a physical wear os device
2. Log in using a google account
3. Observe that that it takes multiple minutes to get to the main watch list screen.

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

Pixel watch running Wear OS 3, 7.40 pre-release PC build

mchowning:
### Google token slowness might only be a debug build issue

In my testing on debug builds, I am seeing it taking anywhere from 1 minute to 4 minutes, just to get the token from Google and start making our API calls to login/refresh. In release builds, however, this seems to be around 20 seconds or so. Not great, but nothing like the crazy times I was seeing on debug builds.

Another way of looking at this, on debug builds, I saw login with a Google account taking from 2m30s to 6 minutes. If I logged in to the same account with email/password, however the times were between 1 and 2 minutes.

Something weird must be going on here, but I haven't dug into it since we're short on time and it seems like it may only be an issue with debug builds.

### Release builds still take >1 minute to log in

Testing email/password login on release builds, I see times of between 1 and 2 minutes (roughly the same as on the debug builds). It would be really good to get this time down. If we can't get this time down, perhaps we need to update the text on the logging in screen to give some indication about how it can take a long time (i.e., "Please be patient" or "This may take a while'). A set of funny, rotating messages might be even better.

### Network Awareness Library

I have also done some testing with the network-awareness library from [this PR](https://github.com/Automattic/pocket-casts-android/pull/1037), and I have not noticed any significant difference.

mchowning:
This is closely related to https://github.com/Automattic/pocket-casts-android/issues/1006

mchowning:
[...]

PR #1062, opened 2023-06-08
Download smaller images on the watch

mchowning:
## Description
Instead of downloading 960 pixel images like we do on the phone, let's just download 480 pixel images on the watch. This will reduce the watch's data usage and improve [the initial login and sync time](https://github.com/Automattic/pocket-casts-android/issues/1047). Thanks for the suggestion @geekygecko . 🙇 

## Testing Instructions
1. Log into a fresh install of the app
2. Use the network inspector to verify that 480 pixel images are being instead of 960 pixel images
3. Check the UI in the app to make sure the images still look good. I believe the larges image we display in the watch app is the podcast artwork we display at the top of the screen for a particular podcast.

## Screenshots or Screencast 

image

## Checklist


ashiagr (approved):
Good call 👍

PR #1063, opened 2023-06-08
Avoid widget setup on wear

mchowning:
## Description
This avoids widget configuration on wear os. `AppWidgetManager.getInstance` returns null on wear os, so this was throwing a null pointer exception before. That wasn't a problem really because we immediately caught the exception, but it clutters up the logs, so it seems better to avoid that unnecessary work on the watch.

## Testing Instructions
* Run the watch app and verify that when you play a podcast without this PR you see a null pointer exception from `WidgetManagerImpl`, but that with this PR you do not see that exception.
* Run the phone app and verify that the pocket casts home screen widget still works as expected.

## Checklist

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/widget/WidgetManagerImpl.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/widget/WidgetManagerImpl.kt
@@ -19,6 +19,7 @@ import au.com.shiftyjelly.pocketcasts.repositories.R
 import au.com.shiftyjelly.pocketcasts.repositories.images.PodcastImageLoader
 import au.com.shiftyjelly.pocketcasts.repositories.playback.PlaybackManager
 import au.com.shiftyjelly.pocketcasts.repositories.podcast.PodcastManager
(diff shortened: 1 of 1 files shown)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/widget/WidgetManagerImpl.kt
```diff
@@ -33,27 +34,29 @@ class WidgetManagerImpl @Inject constructor(
...
-            val views = RemoteViews(context.packageName, R.layout.widget)
-            val widgetName = ComponentName(context, PodcastWidget::class.java)
-            if (playingEpisode == null) {
-                showPlayingControls(false, views)
-            } else {
-                showPlayingControls(true, views)
-                updateArtWork(podcast, playingEpisode, views, widgetName, context)
-                showPlayButton(playing, views)
-                updateSkipAmounts(views, settings)
+        when (Util.getAppPlatform(context)) {
```
mchowning, line 37: I'm suing a when here instead of just adding to the `if` statement we had before so that if/when we add more app platforms the compiler will force us to make a conscious decision here about whether we want to perform the widget setup instead of just defaulting to set up the widget like we were before.

ashiagr (approved):

PR #1068, opened 2023-06-12
Add Sentry proguard setting for watch app

ashiagr:
## Description
This adds `Sentry` proguard settings

## Testing Instructions
Ensure that exceptions from the release build are readable in Sentry.

## Checklist
 
- [ ] with different themes


mchowning (approved):
Working great! Thanks for this!

PR #1077, opened 2023-06-14
Fixing Automotive playback state

geekygecko:
### Issue 1

When playing audio in Automotive the following error is in the logs:

```
isServiceRunningInForeground found no matching service
```

The issue is that the playback service with Automotive is AutoPlaybackService and not PlaybackService. The AutoPlaybackService extends the PlaybackService so changing the condition to `this` stops the error message.

### Issue 2

The Automotive player is having trouble indicating the correct playback state. For example if you tap on the seek bar sometimes it won't show it is buffering and the playback time with continue to change without any audio playback. I believe there is a few issues happening but the main one is that we have a `seeking` flag to stop events when we think the user is seeking, the issue is that this blocks the event that says playback in buffering. There is also another issue that the switchMap doesn't give us all the events, only the latest, so we can't rely on getting all the events from `lastChangeFrom`.

## Issue 3

The error message "Could not load podcast. Please check your internet connection" isn't translated so I have added it to the strings file.

## Issue 4

We received an email asking why we include the feature flag `<uses-feature android:name="android.software.car.templates_host"android:required="true"/>`. I can't find any documentation on this and removing it doesn't affect the app. So I have removed it for now.

### Test the error fix
1. Install the Automotive app
2. Play an episode
3. ✅  Verify the error message "isServiceRunningInForeground found no matching service" is no longer in the logs
4. Install the mobile app
5. Play an episode
3. ✅  Verify the error message "isServiceRunningInForeground found no matching service" is not in the logs

### Test the seek bar
1. Open the Automotive app 
2. Play an episode
3. Open the full screen player
4. Tap a new position on the seek bar 
5. ✅ Verify audio starts playing from the new position and that the buffering animation is shown.
6. Repeat this process a few times

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/MediaSessionManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/MediaSessionManager.kt
@@ -279,12 +279,18 @@ class MediaSessionManager(
     }
 
     private fun observePlaybackState() {
-        val ignoreStates = listOf(
-            // ignore the current position update because the media session progress the time without it and it causes the position to jump when seeking
-            "updateCurrentPosition",
+        val ignoreStates = mutableListOf(
             // ignore buffer position because it isn't displayed in the media session
             "updateBufferPosition"
         )
+
+        val isAutomotive = Util.isAutomotive(context)
+        // listen to the playback progress every second on Automotive as it can get out of sync
+        if (!isAutomotive) {
+            // ignore the playback progress updates as the media session can calculate this without being sent it every second
+            ignoreStates.add("updateCurrentPosition")
+        }
+
         var previousEpisode: BaseEpisode? = null
 
         playbackManager.playbackStateRelay
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/MediaSessionManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/MediaSessionManager.kt
@@ -89,7 +89,6 @@ class MediaSessionManager(
 
     val mediaSession = MediaSessionCompat(context, "PocketCastsMediaSession")
     val disposables = CompositeDisposable()
-    var seeking = false
     private val source = AnalyticsSource.MEDIA_BUTTON_BROADCAST_ACTION
 
     override val coroutineContext: CoroutineContext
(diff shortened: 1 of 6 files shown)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/MediaSessionManager.kt
```diff
@@ -89,7 +89,6 @@ class MediaSessionManager(
 
     val mediaSession = MediaSessionCompat(context, "PocketCastsMediaSession")
     val disposables = CompositeDisposable()
-    var seeking = false
```
mchowning, line 92: I like that we're able to get rid of this. 👍

mchowning (approved):
This seems like a really tricky one to track down. Nice work investigating it and doing some cleanup along the way.

## Issue 2
> The Automotive player is having trouble indicating the correct playback state. For example if you tap on the seek bar sometimes it won't show it is buffering and the playback time with continue to change without any audio playback...

I wasn't able to reproduce this issue even when building from `main`. In my testing both `main` and this branch were working pretty much identically. Maybe it was because I'm just using an emulator which runs faster than an actual headunit? If there are some steps that might help me reproduce the issue, just let me know, and I'd be happy to test it again.

One thing I was able to recreate was that on some occasions seeking twice _in quick succession_ would cause the seek indicator to:
1. Go the the location where I tapped first
2. Go to the location where I tapped second
3. Then it would pop back to the location where I tapped first. It would sometimes flicker between the two positions a few times too.

I was able to recreate this both on `main` and with the changes from your PR, so I certainly don't think this is a blocker for this PR. We could just open an issue for this on GitHub.

https://github.com/Automattic/pocket-casts-android/assets/4656348/0db5b1a8-bdfb-4450-ab49-2d8185a18105

## Issue 4
> We received an email asking why we include the feature flag <uses-feature android:name="android.software.car.templates_host"android:required="true"/>. I can't find any documentation on this and removing it doesn't affect the app. So I have removed it for now.

Did you maybe forget to commit this? I'm not seeing it.

I went ahead and tested the app with this removed, and I also didn't observe any issues, so feel free to add that change in.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/MediaSessionManager.kt
geekygecko: Yes, it did seem like a bit of a hack.

geekygecko:
> One thing I was able to recreate was that on some occasions seeking twice in quick succession

Thanks for mentioning this, I will look at it as part of a separate PR.

> Did you maybe forget to commit this? I'm not seeing it.

Oops, I forgot to push. Thanks for picking this up.

PR #1080, opened 2023-06-14
Update/gravatar use sha256

mchowning:
## Description
This updates the app to use sha256 hashing for gravatar.

I am also consolidating our hashing code, so this is touching the `sha1()` function we use for generating share lists.

One thing worth taking a closer look at is how the sha1 function was using `ISO_8859_1` charset encoding before. I'm sure @geekygecko remembers this very well since he added this... 11 years ago. 😮 In my testing, sharing still worked fine both with this charset specified and with it not specified (which means utf-8 is used). I went with using the default of utf-8 since that is simpler, as best I can tell works fine, and we generally use utf-8 in the app as best I can tell. It does make me a bit nervous though since I don't understand why we were specifying iso-8859-1 before. If you think we should stick with iso-8859-1, just let me know (will only require reverting 42946fd7decdab095272f929b9f7e3e9db5bfea2).

## Testing Instructions
1. Fresh install of the app
2. Log into an account using an email that has a gravatar image
3. Observe that the gravatar image is loaded on the profile tab
4. Go to the podcasts tab and share a list of podcasts from the overflow menu
5. Observe that a valid share link is generated.

## Screenshots or Screencast 

image

## Checklist
 
- [ ] with different themes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 7.42
 -----
 
+*   Updates:
+    *   Use sha256 to encode email for gravatar
+        ([#1080](https://github.com/Automattic/pocket-casts-android/pull/1080)).
 
 7.41
 -----


ashiagr:
Works well 👍

ashiagr (approved):

PR #1083, opened 2023-06-15
Retrieve show notes from JSON cache

geekygecko:
## Description

This change switches getting the show notes from the JSON cache.

## Testing Instructions
1. Open a podcast page
2. Tap on an episode row
3. ✅ Verify the show notes are displayed
4. Close the episode fragment
5. Turn on airplane mode
6. Tap another episode row
7. ✅ Verify the show notes are displayed 
8. Turn off airplane mode
9. Go to a different podcast
10. Tap play on an episode row without opening the episode fragment
11. Tap on the mini player
12. Tap on the Notes tab
13. ✅ Verify the show notes are displayed

### Test show notes are cached with an episode download
1. Tap the Profile tab
2. Tap the settings cog
3. Tap General
4. Tap Row action
5. Change it to Download
6. Go to discover and find a podcast that you haven't opened
7. Tap download on the episode row and wait for the download to finish
8. Turn on airplane mode
9. Tap on an episode row
10. ✅ Verify the show notes are displayed

### DIFF
--- a/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/episode/EpisodeFragment.kt
+++ b/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/episode/EpisodeFragment.kt
@@ -413,6 +413,8 @@ class EpisodeFragment : BaseDialogFragment() {
                 val showNotes = showNotesState.showNotes
                 formattedNotes = showNotesFormatter.format(showNotes) ?: showNotes
                 loadShowNotes(formattedNotes ?: "")
+            } else if (showNotesState is ShowNotesState.Error || showNotesState is ShowNotesState.NotFound) {
+                loadShowNotes("")
             }
         }
 
--- a/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/episode/EpisodeFragment.kt
+++ b/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/episode/EpisodeFragment.kt
@@ -409,13 +409,19 @@ class EpisodeFragment : BaseDialogFragment() {
         )
 
         viewModel.showNotesState.observe(viewLifecycleOwner) { showNotesState ->
-            if (showNotesState is ShowNotesState.Loaded) {
-                val showNotes = showNotesState.showNotes
-                formattedNotes = showNotesFormatter.format(showNotes) ?: showNotes
-                loadShowNotes(formattedNotes ?: "")
-            } else if (showNotesState is ShowNotesState.Error || showNotesState is ShowNotesState.NotFound) {
-                formattedNotes = ""
-                loadShowNotes("")
+            when (showNotesState) {
+                is ShowNotesState.Loaded -> {
+                    val showNotes = showNotesState.showNotes
+                    formattedNotes = showNotesFormatter.format(showNotes) ?: showNotes
+                    loadShowNotes(formattedNotes ?: "")
+                }
+                is ShowNotesState.Error, is ShowNotesState.NotFound -> {
+                    formattedNotes = ""
+                    loadShowNotes("")
+                }
+                is ShowNotesState.Loading -> {
+                    // Do nothing as the starting state is loading
+                }
             }
         }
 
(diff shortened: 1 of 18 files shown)


modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/episode/EpisodeFragmentViewModel.kt
```diff
@@ -71,23 +71,23 @@ class EpisodeFragmentViewModel @Inject constructor(
     var episode: PodcastEpisode? = null
     var isFragmentChangingConfigurations: Boolean = false
 
-    fun setup(episodeUUID: String, podcastUUID: String?, forceDark: Boolean) {
+    fun setup(episodeUuid: String, podcastUuid: String?, forceDark: Boolean) {
```
geekygecko, line 74: In other places we use `Uuid` so I changed it here to match.

modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/ServerShowNotesManager.kt
```diff
@@ -1,165 +1,85 @@
...
-                        cont.resume(data)
-                    }
-
-                    override fun notFound() {
-                        cont.resume(null)
+                // download or update cache
+                val showNotesDownloaded = downloadShowNotes(podcastUuid = podcastUuid, episodeUuid = episodeUuid)
+                if (showNotesDownloaded != null) {
+                    if (showNotesDownloaded != showNotesCached) {
+                        emit(ShowNotesState.Loaded(showNotesDownloaded))
```
mchowning, line 34: I doubt it makes any practical difference since we're never going to throw an exception after this point, but in theory should we be setting `loaded = true` here?

mchowning (approved):
This works great! 👍 I left a couple of minor comments, but nothing major and none of them are blockers.

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/episode/EpisodeFragment.kt
```diff
@@ -407,10 +408,15 @@ class EpisodeFragment : BaseDialogFragment() {
...
-            loadShowNotes(formattedNotes ?: "")
+        viewModel.showNotesState.observe(viewLifecycleOwner) { showNotesState ->
+            if (showNotesState is ShowNotesState.Loaded) {
+                val showNotes = showNotesState.showNotes
+                formattedNotes = showNotesFormatter.format(showNotes) ?: showNotes
+                loadShowNotes(formattedNotes ?: "")
+            } else if (showNotesState is ShowNotesState.Error || showNotesState is ShowNotesState.NotFound) {
+                formattedNotes = ""
+                loadShowNotes("")
+            }
```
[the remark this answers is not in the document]
geekygecko: Thanks, that's a great suggestion. It looks much better now.

modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/ServerShowNotesManager.kt
```diff
@@ -1,165 +1,85 @@
...
-                        cont.resume(data)
-                    }
-
-                    override fun networkDataFound(data: String) {
-                        cont.resume(data)
-                    }
-
-                    override fun notFound() {
-                        cont.resume(null)
+                // download or update cache
```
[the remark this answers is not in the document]
geekygecko: After it has checked the cache it tries to download the show notes from the server again. The okhttp library will respect the server five minute expiry and not call again if it's within that time. If it's been longer it will call the server. The server will either respond back with a 'not modified' or the new content. This will update the local cache and show the user the updated show notes if they have changed. Any idea how to make that into a better comment? 🤔

modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/ServerShowNotesManager.kt
geekygecko: Thanks, I will do that.

PR #1085, opened 2023-06-15
Add remote feature flags provider (in release build only)

ashiagr:
## Description

This adds Firebase remote feature flags provider (in release build only). Values from this provider override those of the default provider. A feature should be enabled in this provider only when it has to be controlled through `Firebase` `RemoteConfig`.

## Testing Instructions

Added `FEATURE_FLAG_TEST` in `Firebase` `RemoteConfig` with the value set to true for the release build. 

The following test checks that the value from `FirebaseRemoteFeatureFlagProvider` overrides that of the default provider for the release build (`StoreFeatureFlagProvider`) where it is set to false.

1. Apply [this patch](https://github.com/Automattic/pocket-casts-android/files/11757161/Test_firebase_remote_feature_flag.patch)
2. Fresh install `release` build
3. Go to `About` page
4. Notice in logs "test feature flag enabled"

## Checklist
 
- [ ] with different themes


modules/services/featureflag/src/main/java/au/com/shiftyjelly/pocketcasts/featureflag/FeatureFlagProvider.kt
```diff
@@ -1,9 +1,25 @@
...
 interface ModifiableFeatureFlagProvider : FeatureFlagProvider {
     fun setFeatureEnabled(feature: Feature, enabled: Boolean)
 }
+
+interface RemoteFeatureFlagProvider : FeatureFlagProvider {
+    fun refreshFeatureFlags()
+}
+
+const val MAX_PRIORITY = 1
+const val MIN_PRIORITY = 2
```
mchowning, line 25: I like how you made this so much more explicit than plain numbers. 👍

modules/features/shared/src/main/java/au/com/shiftyjelly/pocketcasts/shared/AppLifecycleObserver.kt
```diff
@@ -23,6 +25,7 @@ class AppLifecycleObserver @Inject constructor(
     override fun onResume(owner: LifecycleOwner) {
         super.onResume(owner)
         appLifecycleAnalytics.onApplicationEnterForeground()
+        FeatureFlagManager.refreshFeatureFlags()
```
mchowning, line 28: Might it be worth adding some kind of debounce here just to make sure we don't make this api call too much? Hopefully the remote feature flags won't be changing too often. 🤔

modules/services/featureflag/src/main/java/au/com/shiftyjelly/pocketcasts/featureflag/FeatureFlagManager.kt
```diff
@@ -17,23 +17,25 @@ object FeatureFlagManager {
...
+        providers.filter { it.hasFeature(feature) }
+            .sortedBy(FeatureFlagProvider::priority)
+            .firstOrNull()
             ?.isFeatureEnabled(feature)
             ?: feature.defaultValue
-    }
 
     fun setFeatureEnabled(feature: Feature, enabled: Boolean) =
-        providers.filterIsInstance(ModifiableFeatureFlagProvider::class.java)
+        providers.filterIsInstance<ModifiableFeatureFlagProvider>()
```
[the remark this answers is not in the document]
ashiagr: Good point, I added in 36d40af.

modules/features/shared/src/main/java/au/com/shiftyjelly/pocketcasts/shared/AppLifecycleObserver.kt
ashiagr: I reduced the fetch interval to 1 sec in the test patch. We have a minimum fetch interval set to 2 hours [here](https://github.com/Automattic/pocket-casts-android/pull/1085/files#diff-68eeaa8c8ded9e538e37b89d78e81c210fe6a139510e555647ea1c10934c1f36R21). 

I think fetch on refresh should respect this interval. But let me check.

ashiagr: It respects this interval on refresh. But I also had to activate remote config after fetch in 42fe398.

PR #1088, opened 2023-06-15
Enable sharing logs from watch

### LINKED ISSUE #1069: Watch: Add a way to access logs and contact support

There is no way for a user of the watch app to contact support and send their logs.

We should add an ability to "share" the logs with support via email.

mchowning:
## Description
This adds a "Help & feedback" section to the settings and gives users a way to either (1) send an email to support with their watch logs attached or (2) send their watch logs to the phone.

Fixes #1069

> **Warning**
> I am targeting `release/7.41`. I think it is important that we get this in with the initial Wear OS release since it will be critical to helping us debug issues. Let me know if you disagree though.

Since it is not possible to send an email from the watch or send an email-triggering intent directly from the watch to the phone, I'm taking the approach of sending a message from the watch to the phone with the watch's logs. The phone app then takes that message and constructs an appropriate intent with it. This does mean that this functionality will only work if the user has the watch paired and actively connected to a phone with the Pocket Casts Android app.

### Challenges detecting if the phone is available

In order to detect whether the phone is available, I am listening for changes to the availability of the phone (it is called a "node"), and displaying a prompt to the user instead of the contact buttons if the phone is not available.

Unfortunately, this listener seems to frequently not get triggered when the phone becomes unavailable. For that reason, I am also checking to ensure that the message is queued, and if it is not queued (which is what happens when a node is unavailable), I display a toast message indicating that they need to connect their watch to their phone.

Unfortunately, I have found that sometimes a message reports as queued, but will never get delivered. This seems to happen most commonly when the message is sent while the phone is in the process of connecting/disconnecting from the watch. For that reason, when a message is successfully queued I am starting the confirmation Toast that gets displayed the text "If your phone is currently paired..." as a way to give the user a hint as to what the problem is if the message does not get sent and received.

### A. Happy path
1. Install the app from this PR on both a phone and a currently paired watch
2. On the watch go to `Settings` → `Help & Feedback`
3. Tap on "Contact Support"
4. That you are shown a toast message directing you to your phone
5. Observe that you get a share prompt on your phone
6. Tap an email client and observe you have an email prepared that will send the user's logs to support (the resulting ticket from sending this email would look like this: 6413683-zen)
7. From your watch, tap "Send Logs To Phone"
8. Observe that  you are shown a toast message directing you to your phone
9. Observe that you get a share prompt on your phone and that you can share your logs with an application on your phone (this is very similar to "Contact Support" with the main difference being that if you share to an email client you don't get our support email address pre-filled or the short little introductory blurb in the email text).

### B. Sad path

Unfortunately I have not found a way to consistently create each of these sad paths in the code as I discussed earlier in this PR description. Here are some situations to try when contacting support or sending logs to phone:

1. Switching the phone between having airplane mode on and off while periodically trying to contact support or send logs to your phone from your watch.
2. Turning off the paired phone before installing the app on the watch.

You should expect to see either that:
1. The "Contact Support" and "Send Logs To Phone" buttons are hidden and the user is directed to connect to their phone
2. Tapping either button brings up a toast error message asking the user to connect to their phone; or
3. Unfortunately, sometimes the message queues successfully and silently fails. In this case, the watch will respond in the same way as the happy path, but the phone will never receive the logs and open the share dialog. I think/hope this is a bit of an edge case that only occurs when the phone is in the middle of connecting/disconnecting.

## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/4656348/8d556560-dbd7-4e72-8ed0-f4729d6a73c1

## Checklist

### DIFF
--- a/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/settings/HelpScreen.kt
+++ b/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/settings/HelpScreen.kt
@@ -11,6 +11,7 @@ import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.unit.dp
 import androidx.hilt.navigation.compose.hiltViewModel
 import androidx.wear.compose.foundation.lazy.ScalingLazyListScope
+import androidx.wear.compose.material.MaterialTheme
 import androidx.wear.compose.material.Text
 import au.com.shiftyjelly.pocketcasts.wear.ui.component.ScreenHeaderChip
 import au.com.shiftyjelly.pocketcasts.wear.ui.component.WatchListChip
@@ -51,6 +52,19 @@ private fun ScalingLazyListScope.phoneAvailableContent(
     onEmailLogsToSupport: () -> Unit,
     onSendLogsToPhone: () -> Unit,
 ) {
+    item {
+        Text(
+            text = stringResource(id = LR.string.settings_help_contact_support_wear_requires_nearby_phone),
+            textAlign = TextAlign.Center,
+            style = MaterialTheme.typography.caption3,
+            color = MaterialTheme.colors.onSecondary,
+        )
+    }
+
+    item {
+        Spacer(Modifier.height(8.dp))
+    }
+
     item {
         WatchListChip(
             title = stringResource(LR.string.settings_help_contact_support),
--- a/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/settings/HelpScreen.kt
+++ b/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/settings/HelpScreen.kt
@@ -5,6 +5,7 @@ import androidx.compose.foundation.layout.height
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.collectAsState
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.res.stringResource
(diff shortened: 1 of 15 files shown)


wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/settings/HelpScreen.kt
```diff
@@ -0,0 +1,95 @@
...
+            title = stringResource(LR.string.settings_help_contact_support),
+            onClick = onEmailLogsToSupport,
+        )
+    }
+
+    item {
+        WatchListChip(
+            title = stringResource(LR.string.settings_help_send_logs_to_phone),
+            onClick = onSendLogsToPhone,
+        )
```
geekygecko, line 79: I don't reckon we need both options of "Contact Support" and "Send Logs To Phone", it seems a bit confusing to have both. But I will leave that decision with you.

geekygecko (approved):
Good job getting that working! It worked well for me once I have paired my emulator together correctly.

wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/settings/HelpScreen.kt
mchowning: I went back and forth on this, but you're probably right that there's no point in having both, so I removed the send logs to phone option in [3658c43](https://github.com/Automattic/pocket-casts-android/pull/1088/commits/3658c4333d04d7bc05418f16ca05241cd87b01db).

PR #1090, opened 2023-06-16
Upgrade to Android Gradle Plugin 8.0.2

geekygecko:
## Description

This upgrades the project to Android Gradle Plugin 8.0.2. The stable build of Android Studio "Android Studio Flamingo | 2022.2.1 Patch 2" kept prompting me to take this upgrade, so I thought it would be a good idea.

The only one area that I wasn't sure about was the following:

Screenshot 2023-06-16 at 1 45 48 pm

By default it leaves it as the old configuration by adding the following to the `gradle.properties` file. But I have removed it as it seems good for build performance. I installed on mobile, Wear, and Automotive without issue but there could be something that I have missed.
```
android.nonFinalResIds=false
```

> Use non-constant R classes
> Use non-constant [R class](https://developer.android.com/reference/android/R) fields in apps and tests to improve the incrementality of Java compilation and allow for more precise resource shrinking. R class fields are always not constant for libraries, as the resources are numbered when packaging the APK for the app or test that depends on that library. This is the default behavior in Android Gradle Plugin 8.0.0 and higher.

https://developer.android.com/build/optimize-your-build#use-non-constant-r-classes
https://web.archive.org/web/20230203152426/http://tools.android.com/tips/non-constant-fields

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,23 +4,11 @@ buildscript {
         kotlin_version = '1.7.10'
         hilt_version = '2.43.2'
     }
-    repositories {
-        google()
-        mavenLocal()
-
-        maven {
-            url 'https://a8c-libs.s3.amazonaws.com/android' 
-            content {
-                includeGroup "com.automattic.android"
-                includeGroup "com.automattic.android.publish-to-s3"
-            }
-        }
-    }
 
     // Gradle Plugins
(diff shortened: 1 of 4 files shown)


geekygecko:
I'm closing this PR as I failed to get it working.

PR #1095, opened 2023-06-16
Update horologist to 0.4.9

luizgrp:
## Description

Update Horologist version to `0.4.9` and update usages of components from `base-ui` to `compose-material` library.

## Testing Instructions
Check visual of chip components across all the screens of the app if they didn't get affected.

## Checklist
- [N/A] If this is a user-facing change, I have added an entry in CHANGELOG.md
- [N/A] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml`
- [N/A] Any jetpack compose components I added or changed are covered by compose previews
 
⚠️ not fully tested!

- [ ] with different themes


mchowning (approved):
Looks good! Thanks for contributing @luizgrp ! 🙇

PR #1097, opened 2023-06-17
Updated empty state in now playing screen

### LINKED ISSUE #1094: Watch Now Playing Screen does not have a proper empty state
### Description

On the watch, the Now Playing screen does not have a good empty state. Some of the issues are:
1. There is no text prompting the user that they need to select an episode to play
2. 💥  If you tap on the area where the episode name would be, the app crashes
3. The fast-forward/rewind/play buttons are not disabled, so they provide a ripple effect if you tap them, but nothing happens.

1. Open the app with an account that has no podcasts currently playing or in the up next queue. 
2. Swipe left to get to the Now Playing screen

Screenshot 2023-06-16 at 11 50 29 AM

7.41-rc-2

atharv028:
## Description

Updates the `NowPlayingPager.kt` to handle empty state of the player. Also avoids the 💥crash that happens when user previously clicked on the area where the episode name would be.

Fixes #1094

## Testing Instructions
1. Open the app with an account that has no podcasts currently playing or in the up next queue.
2. Swipe left to get to the Now Playing screen
3. ✅ Verify that the player shows the message "Pick an episode to begin" in the title.
4. ✅ Verify that the play/rewind/forward buttons are disabled and greyed out.
5. ✅ Verify that clicking anywhere on the player doesn't crash the app.

## Checklist
 

ashiagr (approved):
Looking great! Code changes look good as well. 

Thanks for working on it 🙇‍♀️

PR #1101, opened 2023-06-19
When a token refresh fails don't show the sign in notification.

geekygecko:
## Description

When there was an issue refreshing the user's token such as a network or server error, a notification would be shown to the user and a sign-out event would be recorded despite the user still being signed in. This PR fixes this issue but I also cleans up some of the associated code.

#### Make sure the user is signed out when the refresh token expires.
1. Log in to an account on staging (I have reduced the expiry to 15 minutes on staging)
2. In Profile -> Settings -> Storage & data use -> Turn off Background refresh
3. Close the app and wait 15 minutes
4. Open the app
5. ✅ Verify the sign in notification is shown and the user has been signed out.

#### Make sure an internal server error doesn't display the sign in notification.
1. Log in to an account
2. Change the method `SyncServerManager.loginToken` to the following to simulate a server error. 
```
    suspend fun loginToken(refreshToken: RefreshToken): LoginTokenResponse {
        val request = LoginTokenRequest(refreshToken = refreshToken, scope = SCOPE_MOBILE)
        server.loginToken(request)
        throw HttpException(Response.error<LoginTokenResponse>(500, "".toResponseBody()))
    }
```
3. On the Profile tab tap the "Refresh now" button.
4. ✅ Verify the error message "Refresh failed" is shown, the account is still signed in and the sign in notification isn't shown.

Tap to see the error message.

#### Make sure the user sync'ing still works
1. Sign in with an account that has data and a Plus subscription
2. ✅ Verify the account is setup with the correct podcasts by comparing them to the Web Player
3. Add an episode to the Up Next by playing it
4. Manually refresh
5. ✅ Verify the episode is in the Up Next on the Web Player


modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/sync/SyncServer.kt
```diff
@@ -143,7 +138,4 @@ interface SyncServer {
 
     @POST("/subscription/promo/validate")
     fun validatePromoCode(@Body request: PromoCodeRequest): Single<PromoCodeResponse>
-
-    @POST("/subscription/cancel/web")
-    fun cancelSubscription(@Header("Authorization") authorization: String, @Body request: SupporterCancelRequest): Single<Response<Void>>
```
geekygecko, line 148: This wasn't being used.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/SyncAccountManager.kt
```diff
@@ -77,14 +75,13 @@ class SyncAccountManager @Inject constructor(
...
                 // Token failed to refresh
                 if (token == null) {
-                    val intent = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
-                        bundle.getParcelable(AccountManager.KEY_INTENT, Intent::class.java)
+                    val intent = BundleCompat.getParcelable(bundle, AccountManager.KEY_INTENT, Intent::class.java)
+                    if (intent == null) {
+                        throw NetworkErrorException()
                     } else {
-                        @Suppress("DEPRECATION")
-                        bundle.getParcelable(AccountManager.KEY_INTENT) as? Intent
```
geekygecko, line 84: There is an Android KTX `BundleCompat` to reduce this logic. I will add it to other areas of the code in a separate PR. https://developer.android.com/reference/kotlin/androidx/core/os/BundleCompat

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/SyncManagerImpl.kt
```diff
@@ -577,6 +583,7 @@ class SyncManagerImpl @Inject constructor(
         )
         isLoggedInObservable.accept(true)
 
+        settings.setFullySignedOut(false)
```
geekygecko, line 586: This wasn't being set when you created an account so the next time my refresh token failed I wasn't being signed out.

modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/sync/SyncServerManager.kt
```diff
@@ -61,17 +61,21 @@ open class SyncServerManager @Inject constructor(
...
         return server.loginGoogle(request)
     }
 
+    /**
+     * Update the access token using the refresh token.
+     * If any 4xx is returned the user should be logged out and asked to login.
+     */
     suspend fun loginToken(refreshToken: RefreshToken): LoginTokenResponse {
-        val request = LoginTokenRequest(refreshToken = refreshToken)
+        val request = LoginTokenRequest(refreshToken = refreshToken, scope = SCOPE_MOBILE)
```
geekygecko, line 78: Simon mentioned we should be setting the scope to mobile.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/SyncManagerImpl.kt
```diff
@@ -188,9 +193,6 @@ class SyncManagerImpl @Inject constructor(
         val loginResult = try {
             val response = loginFunction()
             val result = handleTokenResponse(loginIdentity = loginIdentity, response = response)
-
-            settings.setFullySignedOut(false)
```
geekygecko, line 192: I moved this into the common logic for sign in and create.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/SyncManagerImpl.kt
```diff
@@ -546,10 +547,8 @@ class SyncManagerImpl @Inject constructor(
 
     private fun refreshToken(): Single<AccessToken> {
         syncAccountManager.invalidateAccessToken()
-        return Single.fromCallable {
-            runBlocking {
-                syncAccountManager.getAccessToken()
-            } ?: throw RuntimeException("Failed to get token")
+        return rxSingle {
+            syncAccountManager.getAccessToken() ?: throw RuntimeException("Failed to get token")
```
geekygecko, line 551: This isn't related to the issue but removing runBlocking seems to be a good idea.

mchowning (commented):
Earlier today when I did the steps for "Make sure an internal server error doesn't display the sign in notification" it worked as expected, but now it's actually refreshing successfully for me. I'm probably just doing something wrong with my setup, but it's late in my day so I'd like to do a bit more testing tomorrow.

The only thing I'm concerned about so far is that we're now catching the exception when we fail to send a subscription to the server earlier than we used to. It seems like that may short-circuit some of the logic we have around purchases.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/subscription/SubscriptionManagerImpl.kt
```diff
@@ -281,11 +281,15 @@ class SubscriptionManagerImpl @Inject constructor(
...
-        purchaseEvents.accept(PurchaseEvent.Success)
+        try {
+            val response = syncManager.subscriptionPurchase(SubscriptionPurchaseRequest(purchase.purchaseToken, purchase.products.first())).await()
+            val newStatus = response.toStatus()
+            cachedSubscriptionStatus = newStatus
+            subscriptionStatus.accept(Optional.of(newStatus))
+            purchaseEvents.accept(PurchaseEvent.Success)
+        } catch (ex: Exception) {
+            LogBuffer.e(LogBuffer.TAG_BACKGROUND_TASKS, ex, "Failed to send purchase to server.")
+        }
```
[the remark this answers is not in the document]
geekygecko: Good point. I was trying to figure out why the app was crashing with the error 'UndeliverableException: The exception could not be delivered to the consumer because it has already canceled/disposed...' and left that in there. I will remove the try catch.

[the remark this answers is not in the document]
mchowning: > I was trying to figure out why the app was crashing with the error 'UndeliverableException: The exception could not be delivered to the consumer because it has already canceled/disposed...'

The crashing part of that was because I accidentally messed up how we were catching this exception before on teh phone app. I've put up [a PR](https://github.com/Automattic/pocket-casts-android/pull/1104) with a fix and some more context on that (does not address the cause of the exception though).

mchowning (approved):
This is working well for me.

I've followed your test steps and also tried airplane mode and editing the code to simulate other possible error conditions. One additional specific test I did was to allow the access token to expire (5 minutes on staging), and then seeing the effect of a 400 error (logged out the user and presented the notification) versus a 500 error (left the user logged in).

I also tested the purchase flow to verify that is still working.

One issue I noticed that is not connected to your changes is that tapping on the sign on notification. I filed an issue for that here: https://github.com/Automattic/pocket-casts-android/issues/1106.

PR #1102, opened 2023-06-19
Update horologist to 0.4.10

luizgrp:
## Description
Update Horologist version to `0.4.10` and update `ToggleChip` component in `SettingsScreen` to use the one from Horologist. These updates address issues with RTL mode.

## Testing Instructions
1. Activate option "Force RTL layout" in the OS Settings > Developer options;
2. Launch the app and observe the screens with toggle controls and chips;

## Screenshots or Screencast 

Screenshots were taken from app ([7.41-rc-2](https://github.com/Automattic/pocket-casts-android/releases/tag/7.41-rc-2)) without these changes, and from the compose previews after these changes.

| Before (from app) | After (from preview) |
|--------|--------|
|  |  |
|  |  |
|  |  |
|  | |

## Checklist
- [N/A] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml`
 
⚠️ not fully tested!

- [ ] with different themes

### DIFF
--- a/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/AppDatabaseTest.kt
+++ b/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/AppDatabaseTest.kt
@@ -150,7 +150,7 @@ class AppDatabaseTest {
     }
 
     private fun countRows(db: SupportSQLiteDatabase?, tableName: String): Int {
-        return db?.query("select count(*) from $tableName", null).use { cursor ->
+        return db?.query("select count(*) from $tableName").use { cursor ->
             cursor?.let {
                 it.moveToFirst()
                 return@use it.getInt(0)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/refresh/RefreshPodcastsTask.kt
```diff
@@ -72,7 +72,7 @@ class RefreshPodcastsTask @AssistedInject constructor(
                 .setInitialDelay(REFRESH_EVERY_HOURS, TimeUnit.HOURS)
                 .build()
 
-            workManager.enqueueUniquePeriodicWork(TAG_REFRESH_TASK, ExistingPeriodicWorkPolicy.REPLACE, request)
+            workManager.enqueueUniquePeriodicWork(TAG_REFRESH_TASK, ExistingPeriodicWorkPolicy.CANCEL_AND_REENQUEUE, request)
```
mchowning, line 75: `CANCEL_AND_REENQUEUE` maintains the behavior of the deprecated `REPLACE`

mchowning (commented):
Thanks for the contribution @luizgrp ! 🙇 I pushed an update to our work manager and room dependencies so the app would compile, and everything is working well for me.

This looks good to me now, but since I pushed some nontrivial updates to the work manager and room dependencies, I'd like @ashiagr or @geekygecko to look at this before we merge it. Some of the things I tested to try and catch any breakages:
- Smoke-testing the app both with and without RTL layout
- Upgrading from a very old version of the app to make sure the migrations still all work fine with the update to the `getColumnNames` method
- Syncing repeatedly to fire off the `RefreshPodcastsTask`

ashiagr:
👋 @mchowning 

I smoke-tested the app for upgrades to room and work manager dependencies. I also checked their release notes and found no breaking changes. 

:shipit:

PR #1103, opened 2023-06-21
Replacing the backwards compatible code with Android KTX methods.

geekygecko:
## Description

In my last pull request, I found Android KTX could reduce some of our code to handle older versions of Android, this makes the change in other areas of the code. It felt like a nice clean up change.

It also added a changelog entry I missed for the Automotive change.

## Testing Instructions

1. Logout of the Pocket Casts account and clear the data
2. Open the app
3. ✅ Verify you are shown the the onboarding pages
4. Tap the Profile tab
5. Tap settings
6. Tap Auto download
7. Turn on "New episodes"
8. Tap Choose podcasts
9. ✅ Verify the page opens and the app doesn't crash
10. Start playing an episode
11. Simulate the headphone key next being pressed
```
adb shell input keyevent 87
```
12. ✅ Verify the episode skips forward
13. Go to the discover tab
14. Scroll to the bottom
15. Tap the region
16. ✅ Verify the region list is displayed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
         ([#1101](https://github.com/Automattic/pocket-casts-android/pull/1101)).
     *   Fixed Extra Dark theme not applying proper background on some settings screens
         ([#987](https://github.com/Automattic/pocket-casts-android/pull/987)).
+    *   Fixed Automotive seek bar and playback state issues.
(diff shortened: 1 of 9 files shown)


ashiagr (approved):
Thanks for this nice code improvement!

PR #1104, opened 2023-06-21
Consolidate rxjava undeliverable exception handling

mchowning:
## Description

This is a follow-up to #971 . In that PR I began catching raxjava's `UndeliverableException`s because I noticed that they were getting thrown pretty frequently on logout due to a race condition. I wasn't seeing these on the phone app though so I was concerned that I had something set up wrong on the watch app, so I only caught the exceptions on release builds, allowing them to still crash the app on debug builds.

I later noticed that now the phone app was crashing with these exceptions. Turns out I overlooked that the phone app had the same issue and was already catching the `UndeliverableException`s.

This PR consolidates that handling and, since I now understand that this is a longstanding issue and not unique to the wear app, I'm using the phone app's approach of always catching the exception on all of our apps. This wasn't being done on the automotive app before but I don't see any reason that this exception would not occur on the automotive app, so I think it's better to just handle this consistently on all of our apps. Let me know if you disagree.

I have also started logging the undeliverable exception causes to the log buffer because I wouldn't be surprised if these might be the cause of some issues for our users, but we're not able to help them because we never see that they're occurring. Or, we may see that these are happening so often that they're not helpful at all, in which case we can remove the logging.

## Testing Instructions

1. Create a phone/wear/auto build that forces an `UndeliverableException` (see below)
2. Run the app and cause the exception
3. Observe the exception appears in the logs but does not crash the app.

You can force an `UndeliverableException` by adding this somewhere:

``` kotlin
Observable.create { o ->
    o.onNext(1)
    o.onError(Exception("test"))
}.take(1).subscribe { Timber.d("$it") }
```

## Checklist

### DIFF
--- a/app/src/main/java/au/com/shiftyjelly/pocketcasts/PocketCastsApplication.kt
+++ b/app/src/main/java/au/com/shiftyjelly/pocketcasts/PocketCastsApplication.kt
@@ -38,8 +38,6 @@ import coil.ImageLoader
 import com.google.firebase.FirebaseApp
 import com.google.firebase.analytics.FirebaseAnalytics
 import dagger.hilt.android.HiltAndroidApp
-import io.reactivex.exceptions.UndeliverableException
-import io.reactivex.plugins.RxJavaPlugins
 import io.sentry.Sentry
 import io.sentry.android.core.SentryAndroid
 import io.sentry.protocol.User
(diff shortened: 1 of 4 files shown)


ashiagr (approved):
> I'm using the phone app's approach of always catching the exception on all of our apps. This wasn't being done on the automotive app before but I don't see any reason that this exception would not occur on the automotive app, so I think it's better to just handle this consistently on all of our apps. Let me know if you disagree.

Makes sense to me 👍

PR #1111, opened 2023-06-22
Remove create account for an Automotive build

geekygecko:
## Description

For an version of Android less than 11 it is locking up for an Automotive build after a create account. 

#### Confirm create account exists for a regular build

1. Deploy the Automotive app
2. Tap the settings cog
3. Tap Set up account
4. On the popup dialog tap sign in
5. ✅ Verify you are taken to the Sign in or create account page
6. Tap Create account
7. Enter an email and password. 
8. Tap Create account
9. ✅ Verify you can successfully create an account

#### Confirm create account doesn't exists for certain builds

1. In AccountActivity.kt change it to the following line
```
val navigateToSignIn = true
```
2. Deploy the Automotive app
3. Tap the settings cog
4. Tap Set up account
5. On the popup dialog tap sign in
6. ✅ Verify you are taken to the sign in form with an input box for the email and password
7. Enter existing account details and login
8. ✅ Verify you are logged in successfully

## Screenshots or Screencast 

[Screen_recording_20230622_223645.webm](https://github.com/Automattic/pocket-casts-android/assets/308331/91920f10-c6eb-4dfb-a3cf-c8b9ade7b0bb)

### DIFF
--- a/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/AccountActivity.kt
+++ b/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/AccountActivity.kt
@@ -4,6 +4,7 @@ import android.accounts.AccountAuthenticatorResponse
 import android.accounts.AccountManager.KEY_ACCOUNT_AUTHENTICATOR_RESPONSE
 import android.content.Context
 import android.content.Intent
+import android.os.Build
 import android.os.Bundle
 import androidx.activity.viewModels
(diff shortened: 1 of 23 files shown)


mchowning (approved):
Looks good and works well. Cleaning out some of the supporter code is a nice bonus! 😄 

I realize that this is a very short-term workaround, so I understand that this is probably not worth addressing, but ideally, we should probably be updating the "Set up account" button to something like "Sign in" when we don't allow account creation.

image

Issue #1113, opened 2023-06-22
Audio stutters randomly during playback


notamaiar:
### Description

I've found two main issues with the Wear OS app.

1. Often, though my earbuds are paired to the watch per the Bluetooth settings, I'll hit Play on the Pocket Casts app but no sound comes through. Sometimes this is persistent (I'll wait up to 30 seconds and I hear no sound), and sometimes it resolves itself after 10 seconds or so.
2. During playback, the audio will suddenly start skipping and stuttering - or more accurately, it seems like it's pausing without being told to pause, and despite the controls showing that it's still in "Play." I'm sure it's not actually just skipping audio, because it will go silent in the middle of a sentence and then start up again in the same spot 10-30 seconds later. Sometimes it will do this a dozen times in a row, sometimes I'll go all day and have no problems. It's hard to predict, but this doesn't happen with other on-watch audio apps like YouTube Music, so I don't think it's solely a Bluetooth issue.

### Step-by-step reproduction instructions

1. hit "play" on a podcast episode
2. get asked to pair with Bluetooth (or not, if Bluetooth headphones are already paired)
3. no sound

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

Google Pixel 6 with Android 13 TQ3A.230605.010, Pocket Casts for Wear OS 7.41-rc-3w.

notamaiar:
Unfortunately, yes. The first play of a downloaded episode was fine, but when paused or skipped I got a big delay. I tried swapping my earbuds pair to my phone and back again, as I do that regularly, and when I swapped the pair back to the watch, hitting play gave me no sound. I had to manually unpair and restart the app to get sound again.

Also, after downloading the episode I saw two new weird behaviours:

- When I play the downloaded episode I intermittently get a message warning me that streaming will use data, even if I start the episode from Downloads.
- Hitting play simply reloads the same screen without actually starting play (with this last one I had to Force Stop the app to stop the behaviour).

mchowning:
Would you be willing to contact support with your watch app logs @notamaiar ? You do that by going to `Settings` → `Help & feedback` on the watch (it's important that you do this from your watch and not using the phone app so that the logs from the watch and not the phone get sent).

notamaiar:
No problem: done.

mchowning:
Thanks! 🙇 

Internal ref: 6456300-zen

mchowning:
[...]

PR #1114, opened 2023-06-23
Enable watch to use bluetooth to download episodes

mchowning:
## Description

This fixes a bug where episodes would not be downloaded over Bluetooth. This meant that if you started downloading an episode on wifi, then lost your wifi connection, the download would stop until the wifi connection was regained.

I believe this will improve the situation reported in https://github.com/Automattic/pocket-casts-android/issues/1113, but there may be more to do here so I'm leaving that issue open.

> **Warning**
> This PR is targeting the `release/7.41` branch

## Testing Instructions
Verify that downloads follow these rules:
- If a wifi network is available, it uses that, otherwise
- If a cellular network is available, it uses that; otherwise
- It uses BT if available

The last bullet is the most important. Verify that if you only have a Bluetooth connection you can download an episode.

Note that these changes only actually affect downloads because we have not moved forward yet with having the network awareness library control more of our calls from the watch (see https://github.com/Automattic/pocket-casts-android/pull/1037). The code relating to streaming isn't actually being used at this point, but it will be used if/when we decide to merge #1037.

## Checklist

### DIFF
--- a/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/networking/PocketCastsNetworkingRules.kt
+++ b/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/networking/PocketCastsNetworkingRules.kt
@@ -2,14 +2,11 @@ package au.com.shiftyjelly.pocketcasts.wear.networking
 
 import androidx.annotation.VisibleForTesting
 import au.com.shiftyjelly.pocketcasts.BuildConfig
-import com.google.android.horologist.networks.ExperimentalHorologistNetworksApi
 import com.google.android.horologist.networks.data.NetworkInfo
 import com.google.android.horologist.networks.data.NetworkStatus
 import com.google.android.horologist.networks.data.NetworkType
 import com.google.android.horologist.networks.data.Networks
 import com.google.android.horologist.networks.data.RequestType
-import com.google.android.horologist.networks.data.RequestType.MediaRequest.Companion.DownloadRequest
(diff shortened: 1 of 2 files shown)


wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/networking/PocketCastsNetworkingRules.kt
```diff
@@ -55,27 +52,22 @@ object PocketCastsNetworkingRules : NetworkingRules {
...
-        networks.networks.firstOrNull { it.networkInfo is NetworkInfo.Wifi }?.let {
-            return it
-        }
-
-        return when (requestType) {
-            DownloadRequest -> {
-                // For downloads force LTE as the backup to Wifi, to avoid slow downloads.
-                networks.networks.firstOrNull {
-                    it.networkInfo.type == NetworkType.Cell
-                }
```
mchowning, line 69: This was the problem before. If there was not a cell network it was returning null, which meant that the call [would fail](https://github.com/google/horologist/blob/bb2deb64dd3bdcd8eb8ef291c1339819ef81a558/network-awareness/okhttp/src/main/java/com/google/android/horologist/networks/okhttp/NetworkSelectingCallFactory.kt#L94-L101).

ashiagr (approved):
Changes look good to me. I was able to download an episode on a Bluetooth connection. 

Great detective work! 🕵️

PR #1115, opened 2023-06-23
Use reduced buffer for watch app

ashiagr:
## Description

This reduces buffer time for the watch app.

## Testing Instructions

1. Try playing an episode on a slow network in watch app
2. Notice that it starts playing sooner


mchowning (approved):
These changes look good to me. I'm not sure it will actually allow podcasts to play more quickly though because there is a different parameter in the [setBufferDurationMs](https://developer.android.com/reference/androidx/media3/exoplayer/DefaultLoadControl.Builder#setBufferDurationsMs(int,int,int,int)) that affects how many seconds much be buffered before playback will start. I did some testing with changing that parameter and it didn't seem to help much.

With that said, reducing the app's memory usage still makes complete sense, so I'm still completely on board with this change. Thanks for the update!

PR #1124, opened 2023-06-27
Add auto play up next setting

mchowning:
## Description
This adds a setting for auto playing up next to the phone app. No functionality beyond just adding a setting for this is added in this PR—that will be handled in follow-up PRs.

## Testing Instructions
- On debug builds, verify that when the "Auto Play When Up Next Empty Setting" beta toggle is off, the "Continuous playback" setting is not present at the bottom of the "General" settings, and that the setting is present when the toggle is on.
    - Verify that both settings persist after the app has been killed and restarted
- Verify that flipping the continuous playback toggle fires off the `settings_general_continuous_playback_toggled` event with the appropriate "enabled" boolean property.
- On a release build, verify that the "Continuous playback" setting is not present at the bottom of the "General" settings
- On an automotive build, verify that auto play continues to default to `true` by checking the settings in the automotive app.

## Checklist

### DIFF
--- a/automotive/src/main/java/au/com/shiftyjelly/pocketcasts/AutomotiveApplication.kt
+++ b/automotive/src/main/java/au/com/shiftyjelly/pocketcasts/AutomotiveApplication.kt
@@ -102,9 +102,6 @@ class AutomotiveApplication : Application(), Configuration.Provider {
         if (!settings.contains(Settings.PREFERENCE_AUTO_SUBSCRIBE_ON_PLAY)) {
             settings.setBooleanForKey(Settings.PREFERENCE_AUTO_SUBSCRIBE_ON_PLAY, true)
         }
-        if (!settings.contains(Settings.PREFERENCE_AUTO_PLAY_ON_EMPTY)) {
-            settings.setBooleanForKey(Settings.PREFERENCE_AUTO_PLAY_ON_EMPTY, true)
-        }
     }
 
     private fun setupSentry() {
--- a/modules/services/featureflag/src/main/java/au/com/shiftyjelly/pocketcasts/featureflag/Feature.kt
+++ b/modules/services/featureflag/src/main/java/au/com/shiftyjelly/pocketcasts/featureflag/Feature.kt
@@ -20,4 +20,9 @@ enum class Feature(
         title = "Patron",
         defaultValue = false
     ),
+    AUTO_PLAY_UP_NEXT_SETTING(
+        key = "auto_play_when_up_next_empty",
+        title = "Auto Play When Up Next Empty Setting",
+        defaultValue = BuildConfig.DEBUG,
+    ),
 }
--- a/modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/SettingsImpl.kt
+++ b/modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/SettingsImpl.kt
@@ -25,6 +25,7 @@ import au.com.shiftyjelly.pocketcasts.preferences.Settings.Companion.SETTINGS_EN
 import au.com.shiftyjelly.pocketcasts.preferences.Settings.MediaNotificationControls
 import au.com.shiftyjelly.pocketcasts.preferences.di.PrivateSharedPreferences
 import au.com.shiftyjelly.pocketcasts.preferences.di.PublicSharedPreferences
+import au.com.shiftyjelly.pocketcasts.utils.AppPlatform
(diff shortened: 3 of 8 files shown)


automotive/src/main/java/au/com/shiftyjelly/pocketcasts/AutomotiveApplication.kt
```diff
@@ -102,9 +102,6 @@ class AutomotiveApplication : Application(), Configuration.Provider {
         if (!settings.contains(Settings.PREFERENCE_AUTO_SUBSCRIBE_ON_PLAY)) {
             settings.setBooleanForKey(Settings.PREFERENCE_AUTO_SUBSCRIBE_ON_PLAY, true)
         }
-        if (!settings.contains(Settings.PREFERENCE_AUTO_PLAY_ON_EMPTY)) {
-            settings.setBooleanForKey(Settings.PREFERENCE_AUTO_PLAY_ON_EMPTY, true)
-        }
```
mchowning, line 107: I'm letting the default value we use when we pull from shared preferences control this now.

modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/SettingsImpl.kt
```diff
@@ -999,7 +1001,17 @@ class SettingsImpl @Inject constructor(
     }
 
     override fun getAutoPlayNextEpisodeOnEmpty(): Boolean {
-        return getBoolean(Settings.PREFERENCE_AUTO_PLAY_ON_EMPTY, false)
+        val defaultValue = when (Util.getAppPlatform(context)) {
+            AppPlatform.Automotive -> true
+            AppPlatform.Phone -> true
+            AppPlatform.WearOs -> false
```
mchowning, line 1007: Leaving this feature off in wear os because I haven't added the setting on wear os. I'm not sure how many settings we want to add to wear os, so I'm holding off on this for now and leaving the feature off, but I'm sure this is something we'll revisit later.

modules/services/featureflag/src/main/java/au/com/shiftyjelly/pocketcasts/featureflag/Feature.kt
```diff
@@ -20,4 +20,9 @@ enum class Feature(
         title = "Patron",
         defaultValue = false
     ),
+    AUTO_PLAY_UP_NEXT_SETTING(
+        key = "auto_play_when_up_next_empty",
+        title = "Auto Play When Up Next Empty Setting",
+        defaultValue = BuildConfig.DEBUG,
```
ashiagr, line 26: The default value is true for debug build, right? Asking because of a contradictory test instruction:

> On debug builds, verify that when the "Auto Play When Up Next Empty Setting" beta toggle is off,

modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/SettingsImpl.kt
```diff
@@ -999,7 +1001,17 @@ class SettingsImpl @Inject constructor(
     }
 
     override fun getAutoPlayNextEpisodeOnEmpty(): Boolean {
-        return getBoolean(Settings.PREFERENCE_AUTO_PLAY_ON_EMPTY, false)
+        val defaultValue = when (Util.getAppPlatform(context)) {
+            AppPlatform.Automotive -> true
+            AppPlatform.Phone -> true
```
ashiagr, line 1006: The default value for phone was `false` earlier. Can you reconfirm if we're making it `true` now?

ashiagr (approved):
Looks fine. I asked a couple of questions just for my understanding.

modules/services/featureflag/src/main/java/au/com/shiftyjelly/pocketcasts/featureflag/Feature.kt
mchowning: That's right, I set it to default to `true` for debug builds. I didn't see any reason to have it off, and that makes it easier to test/catch-issues.

modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/SettingsImpl.kt
mchowning: In this PR it is true, but this is more of a placeholder. Apparently the end goal is for it to be true for new installations, and false for users upgrading the app. That will be handled in a separate PR though.

PR #1125, opened 2023-06-27
Upgrade subscription from Plus to Patron

ashiagr:
## Description

This applies recommended proration rate on upgrading to a more expensive plan through Play Store.

## Testing Instructions

Prerequisites: 
- In-app billing sandbox setup
- Release build variant
- License test account
- Feature flag enabled for the release build variant

`Plus` monthly plan
1. Fresh install the release build
2. Login with a `Free` account and purchase a `Plus` monthly plan
3. Tap `Upgrade to Patron` from `Profile` -> `Account` details screen
4. Choose a `Patron` plan (monthly or yearly)
5. Complete payment flow
6. Notice that recommended proration rate is applied 

`Plus` yearly plan
1. Repeat above steps for the `Plus` yearly plan
2. Notice that recommended proration rate is applied 

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/subscription/SubscriptionManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/subscription/SubscriptionManager.kt
@@ -29,7 +29,7 @@ interface SubscriptionManager {
     suspend fun sendPurchaseToServer(purchase: Purchase)
     fun refreshPurchases()
     suspend fun getPurchases(): PurchasesResult?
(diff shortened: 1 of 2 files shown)


mchowning:
Went through and tested this and everything is working well. 👍 

Only thing I noticed is that the dialog Google presents when upgrading from Plus yearly to Patron monthly isn't super-clear imo. 

image

Basically, it is billing you for the remainder of your Plus "yearly" subscription (30 minutes in this case) in Patron "monthly" increments (5 minutes in this case), but taking into account the amount already paid for the Plus yearly subscription. This makes sense, but doesn't seem to be communicated well in Google's dialog menu. I don't think there's anything we can do about this, but I wanted to mention it.

mchowning (approved):

PR #1129, opened 2023-06-29
Add intent for opml import from share sheet

ashiagr:
## Description

This adds intent for opml import from share sheet

## Testing Instructions

1. Save an opml to your device 
2. Open files app
3. Select opml file and tap share menu
4. Notice that Pocket Casts app is shown in the share sheet
5. Select the app
6. Notice that podcasts are imported from the opml into the app

## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/1405144/189ce08c-8fb6-4e8a-b46e-df92b2f8c3f4

## Checklist
 
- [ ] with different themes

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -85,6 +85,13 @@
                         android:mimeType="text/xml"
                         android:scheme="content"/>
             </intent-filter>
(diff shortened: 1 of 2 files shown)


mchowning (approved):
This works great! Thanks for updating this!

PR #1130, opened 2023-06-29
Patron: Update tracks

ashiagr:
## Description

This
- Adds global property for the subscription tier and reuses `plus_upsell` analytic value in the "Plus Upsell" flow (Internal ref: p1684957446866019/1684952142.413199-slack-C02ATC80MUM)
- Returns full product id for patron subscriptions in `purchase_`* events

## Testing Instructions

Prerequisites: 
- In-app billing sandbox setup
- Debug build variant
- License test account
- Patron feature flag enabled

1. Login with a `Free`
2. Purchase a `Patron` plan
3. Notice that `subscription_tier` is shown as a global property in logs
4. Notice that the purchase fails and the `purchase_failed` event is shown in logs with the full product id for the `Patron` plan

## Checklist
 
- [ ] with different themes


ashiagr:
I'm guessing you tested it on a debug build as per instructions. If yes, it is expected.

PR #1131, opened 2023-06-29
Patron: Update feature items

ashiagr:
## Description

With the (soonish) launch of Wear OS app v1,  this adds it to the new layout of "Plus" feature card.

## Testing Instructions

- Make sure that
    - new `Plus` features card includes "Watch Playback"  feature item
    - `Patron` features card includes "Early access to features"  feature item for upcoming features
   
** Also updated app themes and icons title to match with that on iOS

## Checklist
 
- [ ] with different themes


mchowning (approved):
This looks good. Thanks for the update!

Should we also do some updates to the `Profile` → ⚙️ → `Pocket Casts Plus` screen (only shown for free accounts) as well? It looks like it is still using the old icons.

ashiagr:
> Should we also do some updates to the `Profile` → ⚙️ → `Pocket Casts Plus` screen (only shown for free accounts) as well? It looks like it is still using the old icons.

Good point. I matched iOS app to show plus upsell screen from `Pocket Casts Plus` menu in https://github.com/Automattic/pocket-casts-android/pull/1135

PR #1132, opened 2023-06-29
Add autoplay functionality

mchowning:
# Description
This adds an initial version of autoplay for users who have no episodes in their up next queue. This feature is still a bit in flux as we figure out exactly what we want this to look like.

I made a few different attempts at doing this by passing the information up from the relevant views, but that just didn't seem to work well at all. One particular difficulty with that approach is that the Episode Details screen can be accessed from so many places and it would need to know how it was created to pass that information up. We could certainly do it, but it seemed overly complicated and required a lot of changes.

The solution I'm going with here uses a global variable—so it feels a bit hacky—but it seems to work well and require much less change. For each screen that should change what is auto played, we set a global variable each time the user visits that screen ([for example](https://github.com/Automattic/pocket-casts-android/blob/dd7804324e62aaa62c8cd2fa457e54490877b697/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/podcast/PodcastFragment.kt#L452-L455)). Then when playback happens, the `PlaybackManager` uses the `AnalyticsSource` [to determine if it should record an autoplay list/episode or not and, if it should, it fetches the most recently persisted autoplay source (or, if appropriate, the currently playing episode)](https://github.com/Automattic/pocket-casts-android/blob/dd7804324e62aaa62c8cd2fa457e54490877b697/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt#L414-L453). 

# Testing Instructions

So many things to test... 😅 

The expected order of playback for autoplay currently is that the "next" episode lower in the list will be selected (taking account of the user's sort option). If there are no episodes "below" the current episode, then the nearest episode "above" the episode will be selected. In both cases, episodes that have been played or archived are skipped.

## 1. Phone

In all of these scenarios, autoplay should never happen if either 
1. the feature flag is disabled,
2. the settings is turned off in the settings, or
3. The up next queue was not empty either when playback started or at any point during playback

In all of these scenarios, autoplay should not behave differently regardless of whether you:
1. Allow an episode to complete on its own,
2. Skip forward to the end of the episode, or
3. Mark the episode as played

<details><summary>i. Playing from Podcasts screen</summary>

When playing an episode from the screen for a podcast, autoplay should continue playing episodes from that podcast. 
* This happens regardless of where the podcast screen is opened from (i.e., the podcasts tab, the search results, the discover screen, the now playing screen).
* This also happens regardless of whether the playback is started from the Podcast screen itself or if the episode details screen is opened from there and play is tapped on that screen.
</details> 

<details><summary>ii. Playing from Filter, Starred, Downloads, or Files</summary>
When playing an episode from a list on any of these screens (or by opening the episode details from one of these screens and tapping play), should allow autoplay after the episode is complete. 
</details>

<details><summary>iii. Autoplay does not happen if an episode is added to the up next queue</summary>
Test 1:
1. Start playing a single episode in a way that auto play would happen.
4. Add an episode to the up next queue
5. Mark both episodes as played
6. Confirm that playback stops

Test 2:
1. Start playing a single episode in a way that auto play would happen.
2. Add an episode to the up next queue
3. Remove the episode you just added to your up next queue from the up next queue
4. Mark the currently playing episode as played
5. Confirm that playback stops
</details>

<details><summary>iv. Playing from new episode notification</summary>
1. With an empty up next queue, trigger a new episode notification from the developer settings
2. Tap "Play" on the notification
3. Complete the episode
4. Verify that playback continues with episodes from the current podcast
</details>

<details><summary>v. Takes account of sort order</summary>

1. Play an episode from the middle of a podcasts list of episodes
2. Sort the list from oldest to newest, and note the name of the episode below the currently playing episode (so the next newest episode) that has not already been played or archived
3. Complete the currently playing episode
4. Observe that the next newest episode is autoplayed
5. Change the sort order to be newest to oldest, and now the name of the episode that is now below the currently playing episode (so the next oldest episode) that has not been played or achived.
6. Complete the currently playing episode
7. Observe that the next oldest episode is autoplayed
</details>

<details><summary>vi. Stream warning permits autoplay</summary>

1. Use a metered network and have the stream warning setting enabled
2. Begin streaming an episode without an up next queue
3. Complete the episode
4. Observe that an episode is autoplayed such that the stream warning is presented and that choosing to go ahead and stream the episode will begin playback.
</details>

<details><summary>vii. Does not autoplay from Listening History</summary>
If you start playback from the Listening History screen (or by opening the episode details from the screen and then tapping play), there should not be any autoplay after the episode is completed.
</details>

<details><summary>viii. Playback continues when started from Google Assistant</summary>
1. Using a release build (or a debug build with the release package name should work too)
2. On your device with an empty up next queue, trigger the Google Assistant and ask to play a podcast on Pocket Casts (i.e., "Play This American Life on Pocket Casts").
3. Playback should start in Pocket Casts
4. Complete the episode
6. Observe that playback continues
</details>

<details><summary>ix. Autoplay stops after all episodes are played on a list on phone (Automotive behaves differently—see below)</summary>
1. Start playing an episode from a short list
2. Complete playing each episode until every episode in the list has been played
3. Confirm that playback stops
</details>

## 2. Confirm no Automotive Regressions

<details><summary>i. Autoplay toggle defaults to on and works on automotive</summary>
1. Fresh install of Android Automotive
2. Play an episode
3. Mark it as played
4. Observe another episode starts playing
5. Go to settings and turn off autoplay
8. Mark the currently playing episode as played
9. Observe that playback stops
</details>

<details><summary>ii. Autoplay continues after all episodes in a list are played on Automotive</summary>

1. Subscribe to a podcast, but create create a filter with only a few episodes in it
2. Log into this account on Android automotive
3. Make sure autoplay is turned on
4. Start playing a an episode from teh screen for that filter
5. Mark the episodes as played until you've listened to all the episodes on that filter
6. Observe that playback on continues with other episodes (this is different from the behavior we want on the phone).
</details>

# Checklist

### DIFF
--- a/modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/filters/FilterEpisodeListFragment.kt
+++ b/modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/filters/FilterEpisodeListFragment.kt
@@ -32,6 +32,7 @@ import au.com.shiftyjelly.pocketcasts.preferences.Settings
 import au.com.shiftyjelly.pocketcasts.repositories.chromecast.CastManager
 import au.com.shiftyjelly.pocketcasts.repositories.download.DownloadManager
 import au.com.shiftyjelly.pocketcasts.repositories.images.PodcastImageLoader
+import au.com.shiftyjelly.pocketcasts.repositories.playback.AutomaticUpNextSource
 import au.com.shiftyjelly.pocketcasts.repositories.playback.PlaybackManager
 import au.com.shiftyjelly.pocketcasts.repositories.playback.UpNextQueue
 import au.com.shiftyjelly.pocketcasts.ui.extensions.getColor
@@ -140,6 +141,7 @@ class FilterEpisodeListFragment : BaseFragment() {
         recyclerView.adapter = adapter
         listSavedState?.let { recyclerView.layoutManager?.onRestoreInstanceState(it) }
         setShowFilterOptions(showingFilterOptionsBeforeModal)
+        AutomaticUpNextSource.mostRecentList = viewModel.playlistUUID
     }
 
     override fun onDestroyView() {
--- a/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/ProfileEpisodeListFragment.kt
+++ b/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/ProfileEpisodeListFragment.kt
@@ -28,6 +28,7 @@ import au.com.shiftyjelly.pocketcasts.podcasts.view.podcast.EpisodeListAdapter
 import au.com.shiftyjelly.pocketcasts.preferences.Settings
 import au.com.shiftyjelly.pocketcasts.repositories.download.DownloadManager
 import au.com.shiftyjelly.pocketcasts.repositories.images.PodcastImageLoader
+import au.com.shiftyjelly.pocketcasts.repositories.playback.AutomaticUpNextSource
 import au.com.shiftyjelly.pocketcasts.repositories.playback.PlaybackManager
 import au.com.shiftyjelly.pocketcasts.repositories.playback.UpNextQueue
 import au.com.shiftyjelly.pocketcasts.repositories.podcast.EpisodeManager
@@ -139,6 +140,11 @@ class ProfileEpisodeListFragment : BaseFragment(), Toolbar.OnMenuItemClickListen
     override fun onResume() {
         super.onResume()
         binding?.recyclerView?.adapter = adapter
+        when (mode) {
+            Mode.Downloaded -> AutomaticUpNextSource.Companion.Predefined.downloads
+            Mode.History -> null
+            Mode.Starred -> AutomaticUpNextSource.Companion.Predefined.starred
+        }?.let { AutomaticUpNextSource.mostRecentList = it }
     }
 
     override fun onDestroyView() {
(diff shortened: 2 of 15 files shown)


ashiagr:
Great work, @mchowning! I just have few questions:

### Archived Episodes

Archived episodes behavior doesn't seem to match what is described in the iOS PR https://github.com/Automattic/pocket-casts-ios/pull/923#issuecomment-1613586728

Can we make sure that both platforms are in sync?

### Grouped Episodes

Autoplay only takes into account sort order and ignores any group setting (like group by downloaded), right?

### Automotive Filters

I created a filter in which I selected random starred episodes from a podcast.
Automotive app played another episode (next in the order from the same podcast) which was not in the filter. Phone app played episodes only from the filter. This behavior also exists in the production app so I guess we're maintaining that, correct? 

https://github.com/Automattic/pocket-casts-android/assets/1405144/181a234b-0126-4fc2-bcdc-6bbb810a147b

https://github.com/Automattic/pocket-casts-android/assets/1405144/4cd25414-9d18-4542-8283-b0f5efa5e84d

ashiagr (approved):

mchowning:
> Automotive Filters
> I created a filter in which I selected random starred episodes from a podcast.
> Automotive app played another episode (next in the order from the same podcast) which was not in the filter. Phone app played episodes only from the filter. This behavior also exists in the production app so I guess we're maintaining that, correct?

That's interesting, when I tested this it played the next episode from the filter. Did you maybe not start playing the episode from within the filter on Automotive when you tested it?

https://github.com/Automattic/pocket-casts-android/assets/4656348/d12520ad-453a-489b-9c84-4e3f1cd939f9

> Autoplay only takes into account sort order and ignores any group setting (like group by downloaded), right?

I didn't think about episode grouping. I'll sync with iOS and update this in a separate PR if needed. Thanks for catching this. 👍 

> Archived episodes behavior doesn't seem to match what is described in the iOS PR https://github.com/Automattic/pocket-casts-ios/pull/923#issuecomment-1613586728
> 
> Can we make sure that both platforms are in sync?

I think this is a bit of an edge case (since iOS only behaves differently if the user is displaying archived episodes) so it's not critical that the platforms be in sync, but I have talked to @leandroalonso  about this, and he may be updating iOS with this behavior. Skipping played/archived episodes does feel like the more natural behavior imo and this keeps the behavior consistent between the phone and Automotive apps.

PR #1133, opened 2023-06-29
Autoplay: Rename AnalyticsSource

mchowning:
## Description
This builds on https://github.com/Automattic/pocket-casts-android/pull/1132 by renaming the `AnalyticsSource` class since we're no longer just using it for analytics. There are no other changes in this PR. This is really a part of that PR, but I wanted to separate it out so to make that PR more reviewable, so you probably want to review that before merging this.

I thought about also moving it out of the analytics package, but I didn't see a great other place to put it at this point (utils would be ok, but doesn't seem like a perfect fit), so I figured I'd leave it in analytics for now, and we can move it later if we want.

I also don't love the name `SourceView` so let me know if you have any better ideas.

## Testing Instructions
If CI is happy, this should be good.


ashiagr (approved):
CI is happy and `SourceView` seems like a good name. 

I thought we could move it to the `Shared` module but it might cause issues due to the module hierarchy. So I agree let's keep it in the `Analytics` module for the time being.

mchowning:
> I thought we could move it to the Shared module but it might cause issues due to the module hierarchy. 

Right, the analytics module needs access to this. We could get around this by having the `analytics` module contain the interface it needs and add a sealed class implementing the interface at a "higher" module, but that felt like overkill at this point.

PR #1134, opened 2023-06-29
Default autoplay on for new installations only

mchowning:
## Description
This updates the auto play setting to only default to true for new installations.

## Testing Instructions

For all these tests, the up next queue needs to be empty.

### 1. Upgraded from having setting enabled
1. Install a build from this PR with the [version code](https://github.com/Automattic/pocket-casts-android/blob/dc7084f631248d0c65d5d37c1ad1b57d539c9ccc/base.gradle#L23) set to 1
2. Verify that the auto play setting is enabled
3. Upgrade the installation by Installing a build from this PR with the [version code](https://github.com/Automattic/pocket-casts-android/blob/dc7084f631248d0c65d5d37c1ad1b57d539c9ccc/base.gradle#L23) set to 2
4. Verify that the auto play setting is still enabled

### 2. Upgraded from having setting disabled
1. Install a build from this PR with the [version code](https://github.com/Automattic/pocket-casts-android/blob/dc7084f631248d0c65d5d37c1ad1b57d539c9ccc/base.gradle#L23) set to 1
2. Verify that the auto play setting is enabled
3. Change the setting to be disabled
4. Upgrade the installation by Installing a build from this PR with the [version code](https://github.com/Automattic/pocket-casts-android/blob/dc7084f631248d0c65d5d37c1ad1b57d539c9ccc/base.gradle#L23) set to 2
5. Verify that the auto play setting is still disabled

### 3. Upgraded from previous version of the app that did not have an auto play setting
1. Install a build from the `release/7.42` branch with the [version code](https://github.com/Automattic/pocket-casts-android/blob/dc7084f631248d0c65d5d37c1ad1b57d539c9ccc/base.gradle#L23) set to 1
2. Verify that there isn't an auto play setting in the settings
3. Upgrade the installation by Installing a build from this PR with the [version code](https://github.com/Automattic/pocket-casts-android/blob/dc7084f631248d0c65d5d37c1ad1b57d539c9ccc/base.gradle#L23) set to 2
5. Verify that the auto play setting is disabled

### 4. Watch
1. Install the app on a watch
2. Start playing an episode from a podcast with more than one podcast
3. Complete the episode
4. Verify that playback stops

### 5. Automotive
1. Install an Automtive build
2. Go to the app's settings
3. Verify that the auto play setting is on by default

## Checklist

### DIFF
--- a/modules/features/shared/src/test/java/au/com/shiftyjelly/pocketcasts/shared/AppLifecycleObserverTest.kt
+++ b/modules/features/shared/src/test/java/au/com/shiftyjelly/pocketcasts/shared/AppLifecycleObserverTest.kt
@@ -1,0 +1,122 @@
+package au.com.shiftyjelly.pocketcasts.shared
+
+import android.content.Context
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.LifecycleOwner
+import au.com.shiftyjelly.pocketcasts.analytics.AppLifecycleAnalytics
+import au.com.shiftyjelly.pocketcasts.featureflag.providers.DefaultReleaseFeatureProvider
+import au.com.shiftyjelly.pocketcasts.featureflag.providers.FirebaseRemoteFeatureProvider
+import au.com.shiftyjelly.pocketcasts.featureflag.providers.PreferencesFeatureProvider
+import au.com.shiftyjelly.pocketcasts.preferences.Settings
+import au.com.shiftyjelly.pocketcasts.utils.AppPlatform
+import au.com.shiftyjelly.pocketcasts.utils.PackageUtil
+import dagger.hilt.android.qualifiers.ApplicationContext
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.Mock
+import org.mockito.junit.MockitoJUnitRunner
+import org.mockito.kotlin.any
+import org.mockito.kotlin.doReturn
+import org.mockito.kotlin.never
+import org.mockito.kotlin.spy
+import org.mockito.kotlin.verify
+import org.mockito.kotlin.whenever
+
+private const val VERSION_CODE_DEFAULT = 0
+private const val VERSION_CODE_AFTER_FIRST_INSTALL = 1
+private const val VERSION_CODE_AFTER_SECOND_INSTALL = 2
+
+@RunWith(MockitoJUnitRunner::class)
+class AppLifecycleObserverTest {
+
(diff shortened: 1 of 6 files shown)


modules/services/analytics/src/main/java/au/com/shiftyjelly/pocketcasts/analytics/AppLifecycleAnalytics.kt
```diff
@@ -18,18 +18,18 @@ class AppLifecycleAnalytics @Inject constructor(
     /* The date the app was last opened, used for calculating time in app */
     private var applicationOpenedDate: Date? = null
 
-    fun onApplicationInstalledOrUpgraded() {
-        // Track app upgrade and install
-        val versionCode = packageUtil.getVersionCode(appContext)
-        val oldVersionCode = settings.getMigratedVersionCode()
```
ashiagr, line 24: These are unused in the class constructor now and can be removed:

```
appContext: Context,
settings: Settings,
packageUtil: PackageUtil,
```

modules/features/shared/src/test/java/au/com/shiftyjelly/pocketcasts/shared/AppLifecycleObserverTest.kt
```diff
@@ -0,0 +1,122 @@
...
+import org.mockito.kotlin.spy
+import org.mockito.kotlin.verify
+import org.mockito.kotlin.whenever
+
+private const val VERSION_CODE_DEFAULT = 0
+private const val VERSION_CODE_AFTER_FIRST_INSTALL = 1
+private const val VERSION_CODE_AFTER_SECOND_INSTALL = 2
+
+@RunWith(MockitoJUnitRunner::class)
+class AppLifecycleObserverTest {
```
ashiagr, line 31: Kudos for adding these tests 🙌

ashiagr (approved):
Everything works great! Just left a nitpick.

modules/services/analytics/src/main/java/au/com/shiftyjelly/pocketcasts/analytics/AppLifecycleAnalytics.kt
mchowning: Thanks for catching that. Fixed in [68daf84](https://github.com/Automattic/pocket-casts-android/pull/1134/commits/68daf84546bb01469e93bf4ea9d5d03122d3c6b6)

PR #1138, opened 2023-06-30
Track when an episode is autoplayed

mchowning:
## Description
Tracking when an episode is autoplayed.

## Testing Instructions

When referring to the autoplay event in the test steps below, I am referring to the event "playback_episode_autoplayed" with  an "episode_uuid" property reflecting the uuid of the episode being autoplayed.

1. With the autplay feature enabled and an empty up next queue
2. Begin playing an episode
3. Observe that the autoplay event is NOT sent
4. Complete the episode
5. Observe that the autoplay event IS sent
7. Add an episode to your up next queue
8. Complete playing the current episode and the episode in the up next queue
9. Observe that no new autoplay events are sent (should only have 1 autoplay event from following all these test steps)

## Checklist


ashiagr (approved):
Looks good 👍

PR #1139, opened 2023-06-30
Adding +1 Minute Timer

Mzazi25:
## Description
This PR adds +1 Minute to sleep timer.  Instead of the minimum set time as 5 min, a user can now set up 1 min as the least time.
The layout on the +1 and +5 was changed so as to avoid more thumb reach. 
Fixes # N/A

## Testing Instructions

1. Tap on the Discover tab 
2. Tap on any random podcast and hit play 
3. Set Sleep timer. 
4. Click on 5 minute timer. 
5.  Click on the timer again. 
6. Add 1 Minute  

## Screenshots or Screencast 
splitminute

## Checklist
 
- [ ] with different themes


ashiagr (approved):
Nice work @Mzazi25!

Can you add a [changelog](https://github.com/Automattic/pocket-casts-android/blob/main/CHANGELOG.md) for it?

Mzazi25:
@ashiagr Done

PR #1141, opened 2023-07-03
Bookmarks: Adds the feature flag

geekygecko:
## Description

Adds the bookmarks feature flag and only enables it in debug while we work on the feature.

## Testing Instructions

1. Launch the app
3. Go to Profile > Settings cog > Beta features
4. ✅ Verify the feature flag is listed there

### DIFF
--- a/modules/services/localization/src/main/res/values/strings.xml
+++ b/modules/services/localization/src/main/res/values/strings.xml
@@ -1033,7 +1033,7 @@
     <string name="settings_auto_download_up_next">Episodes added to Up Next</string>
     <string name="settings_auto_play">Auto play</string>
     <string name="settings_auto_play_summary">When there\'s nothing left to play, automatically pick an episode for me</string>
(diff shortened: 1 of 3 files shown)


modules/services/localization/src/main/res/values/strings.xml
```diff
@@ -1033,7 +1033,7 @@
     <string name="settings_auto_download_up_next">Episodes added to Up Next</string>
     <string name="settings_auto_play">Auto play</string>
     <string name="settings_auto_play_summary">When there\'s nothing left to play, automatically pick an episode for me</string>
-    <string name="settings_beta_features">Beta Features</string>
+    <string name="settings_beta_features">Beta features</string>
```
geekygecko, line 1036: I believe this should be lowercase to match the other menu titles and the design spec.

ashiagr (approved):
LGTM 👍

PR #1142, opened 2023-07-03
Bookmarks: Adds database table and DAO

geekygecko:
## Description

This change creates the new bookmarks table and adds a basic DAO.

I don't know if we usually hold back the migration until later in case it changes early on in the project. 

iOS has the following structure.
```
CREATE TABLE IF NOT EXISTS \(Self.tableName) (
    \(Column.uuid) varchar(40) NOT NULL,
    \(Column.episode) varchar(40) NOT NULL,
    \(Column.podcast) varchar(40),
    \(Column.time) real NOT NULL,
    \(Column.createdDate) INTEGER NOT NULL,
    \(Column.deleted) int NOT NULL DEFAULT 0,
    PRIMARY KEY (\(Column.uuid))
);
```
but the Android version is slightly different.
```
CREATE TABLE IF NOT EXISTS `bookmarks` (
    `uuid` TEXT NOT NULL,
    `podcast_uuid` TEXT NOT NULL,
    `episode_uuid` TEXT NOT NULL,
    `time` INTEGER NOT NULL, 
    `created_at` INTEGER NOT NULL,
    `deleted` INTEGER NOT NULL,
    `sync_status` INTEGER NOT NULL,
    PRIMARY KEY(`uuid`)
);
```
I have gone with the not null for everything. With user file episodes we have a `UserEpisodePodcastSubstitute.substituteUuid` that we send to the server, so I wondered if we should wrap that logic in a BookmarkManager and keep it the same. I'm open to suggestions for this.

## Testing Instructions
1. Install the previous version.
2. Install this version over the top
3. ✅ Verify this is no Room migration issues

### DIFF
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/BookmarkDao.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/BookmarkDao.kt
@@ -1,0 +1,25 @@
+package au.com.shiftyjelly.pocketcasts.models.db.dao
+
+import androidx.room.Dao
+import androidx.room.Delete
+import androidx.room.Insert
+import androidx.room.OnConflictStrategy
+import androidx.room.Query
+import androidx.room.Update
+import au.com.shiftyjelly.pocketcasts.models.entity.Bookmark
+
+@Dao
+abstract class BookmarkDao {
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    abstract suspend fun insert(bookmark: Bookmark)
+
+    @Update
+    abstract suspend fun update(bookmark: Bookmark)
+
+    @Delete
+    abstract suspend fun delete(bookmark: Bookmark)
+
+    @Query("SELECT * FROM bookmarks WHERE uuid = :uuid")
+    abstract suspend fun findByUuid(uuid: String): Bookmark?
+}
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/Bookmark.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/Bookmark.kt
@@ -10,7 +10,7 @@ import java.util.Date
 @Entity(
     tableName = "bookmarks",
     indices = [
-        Index(name = "podcast_uuid", value = arrayOf("podcast_uuid"))
+        Index(name = "bookmarks_podcast_uuid", value = arrayOf("podcast_uuid"))
     ]
 )
 data class Bookmark(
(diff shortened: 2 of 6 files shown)


ashiagr:
This is looking good 👍 

I just have a question: iOS restricts adding a bookmark to an episode at the same timestamp (checked [here](https://github.com/Automattic/pocket-casts-ios/blob/967d99a4d8548c58e4792ef8c5a56e1284f01bde/Modules/DataModel/Tests/PocketCastsDataModelTests/BookmarkDataManagerTests.swift#L44)). Looks like there is no such restriction in Android at the database level. Is that correct?

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/BookmarkDao.kt
```diff
@@ -0,0 +1,25 @@
...
+import androidx.room.Insert
+import androidx.room.OnConflictStrategy
+import androidx.room.Query
+import androidx.room.Update
+import au.com.shiftyjelly.pocketcasts.models.entity.Bookmark
+
+@Dao
+abstract class BookmarkDao {
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
```
ashiagr, line 14: Now that we're using Room `2.5.1` can we use the new shortcut [@Upsert](https://developer.android.com/jetpack/androidx/releases/room#2.5.0) now?

> Added a new shortcut annotation, @Upsert, which attempts to insert an entity when there is no uniqueness conflict or update the entity if there is a conflict.

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/Bookmark.kt
```diff
@@ -0,0 +1,30 @@
...
+import java.io.Serializable
+import java.util.Date
+
+@Entity(
+    tableName = "bookmarks",
+    indices = [
+        Index(name = "bookmarks_podcast_uuid", value = arrayOf("podcast_uuid"))
+    ]
+)
+data class Bookmark(
```
ashiagr, line 16: Should we add a bookmark title field?

geekygecko: Oh, I missed that one. I have added this column.

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/BookmarkDao.kt
geekygecko: Good pick up. Do you mind if I add this in a separate PR so we can merge the multiple PRs we have chained?

ashiagr (approved):
Looks good 👍

PR #1143, opened 2023-07-03
Bookmarks: Add a data manager

geekygecko:
## Description

This adds a bookmarks data manager to help access the bookmark DAO. I have added some methods but we might need more as we add the feature.

## Testing Instructions
There isn't much to test here. If it passes the tests it should be good to start using.


ashiagr (approved):
Looks great! Thanks for laying the foundation 🙌

PR #1144, opened 2023-07-03
+ - 1 min Custom Increments on the Sleep timer

Mzazi25:
## Description
This PR adds +-1 increments if the custom timer is <5. 
Fixes # N/A

## Testing Instructions

1. Tap on the Discover tab
2. Tap on any random podcast and hit play
3. Click on  Sleep timer.
4. Click on - +.
5. If the Custom timer is below 5, it reduces /increases by 1 min

## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/95022986/6ea0ba14-2f93-4157-873c-9f38ea1c536d

## Checklist
 
- [ ] with different themes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,10 @@
 -----
 *   New Feature:
     *   Add capability to add +1 minute on the sleep timer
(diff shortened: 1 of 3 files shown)


ashiagr (approved):
Works like a charm ❤️

PR #1145, opened 2023-07-03
Bookmarks: Add new tab on player

ashiagr:
## Description
This adds a placeholder `Bookmarks` tab to the player.

## Testing Instructions
1. Open the debug build with the bookmarks `Feature Flag`enabled
2. Play any episode having chapters (https://pca.st/upgrade)
3. Open the full player
4. ✅ Verify the `Bookmarks` tab appears at the correct position
5. ✅ Verify tapping on it opens a placeholder view
6. ✅ Verify swiping between tabs works and opens the current view. 
7. Close the player
8. Disable the `Bookmarks` flag and restart the app
9. Open the full player
10. ✅ Verify you don't see the `Bookmarks` tab

## Screenshots or Screencast 

Without Feature Flag | With Feature Flag
-----|-----
 | 

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/BookmarksFragment.kt
+++ b/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/BookmarksFragment.kt
@@ -44,7 +44,7 @@ class BookmarksFragment : BaseFragment() {
 }
 
 @Composable
-fun BookmarksPage(
+private fun BookmarksPage(
     playerViewModel: PlayerViewModel,
 ) {
     val listData = playerViewModel.listDataLive.asFlow().collectAsState(initial = null)
--- a/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/BookmarksFragment.kt
+++ b/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/BookmarksFragment.kt
@@ -1,0 +1,64 @@
+package au.com.shiftyjelly.pocketcasts.player.view
+
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.ViewGroup
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.material.MaterialTheme
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.collectAsState
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.platform.ComposeView
+import androidx.compose.ui.platform.ViewCompositionStrategy
+import androidx.fragment.app.activityViewModels
+import androidx.lifecycle.asFlow
+import au.com.shiftyjelly.pocketcasts.compose.AppThemeWithBackground
+import au.com.shiftyjelly.pocketcasts.compose.components.TextH30
+import au.com.shiftyjelly.pocketcasts.compose.theme
+import au.com.shiftyjelly.pocketcasts.player.viewmodel.PlayerViewModel
+import au.com.shiftyjelly.pocketcasts.views.fragments.BaseFragment
+import dagger.hilt.android.AndroidEntryPoint
+
+@AndroidEntryPoint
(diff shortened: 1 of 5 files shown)


CookieyedCodes:
Can I ask is this bookmarks feature focused on the current episode or is it global? Just interested :)

ashiagr:
Here it will show bookmarks for the selected episode.

modules/services/localization/src/main/res/values/strings.xml
```diff
@@ -276,6 +276,7 @@
     <string name="player_sleep_timer_start_failed">Unable to start timer. Please allow \'Alarms &amp; reminders\' permission.</string>
     <string name="player_sleep_when_episode_ends">Sleeping when episode ends</string>
     <string name="player_star" translatable="false">@string/star</string>
+    <string name="player_tab_bookmarks">Bookmarks</string>
```
geekygecko, line 279: As we need Bookmarks in multiple places maybe we should either just have `bookmarks` or maybe link to a common string. 
```
<string name="player_tab_bookmarks" translatable="false">@string/bookmarks</string>
```

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/BookmarksFragment.kt
```diff
@@ -0,0 +1,64 @@
...
+    private val playerViewModel: PlayerViewModel by activityViewModels()
+
+    override fun onCreateView(
+        inflater: LayoutInflater,
+        container: ViewGroup?,
+        savedInstanceState: Bundle?,
+    ) = ComposeView(requireContext()).apply {
+        setContent {
+            AppThemeWithBackground(theme.activeTheme) {
+                setViewCompositionStrategy(ViewCompositionStrategy.DisposeOnViewTreeLifecycleDestroyed)
```
geekygecko, line 37: I didn't know about this call. Thanks for the lesson!

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/BookmarksFragment.kt
```diff
@@ -0,0 +1,64 @@
...
+                BookmarksPage(
+                    playerViewModel = playerViewModel
+                )
+            }
+        }
+    }
+}
+
+@Composable
+fun BookmarksPage(
```
geekygecko, line 47: Should this be private?

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/PlayerContainerFragment.kt
```diff
@@ -120,17 +122,20 @@ class PlayerContainerFragment : BaseFragment(), HasBackstack {
 
             override fun onPageSelected(position: Int) {
                 super.onPageSelected(position)
-                when (position) {
-                    0 -> {
+                when (adapter.pageTitle(position)) {
+                    LR.string.player_tab_playing -> {
```
geekygecko, line 126: This seems to link back to `VR.string.player_tab_playing` and has different values, either `Now Playing` or `Playing` depending on the screen width. 

I wonder if it would be more explicit with something like:

```
when {
    adapter.isPlayerTab(position) -> {
```

```
fun isPlayerTab(position: Int): Boolean {
    return sections[position] is Section.Player
}
```

Just an idea, feel free to ignore it.

geekygecko (approved):

modules/services/localization/src/main/res/values/strings.xml
ashiagr: Makes sense, updated in 5658c6f

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/BookmarksFragment.kt
ashiagr: Yep, in 079c171

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/PlayerContainerFragment.kt
ashiagr: Suggestion applied to make it more explicit in 2342c47

PR #1146, opened 2023-07-03
Make user aware of playback error on now playing screen

### LINKED ISSUE #1030: Watch's now playing screen does not inform the user of playback errors
### Description

When initiating playback, the Now Playing screen does not show any errors to the user. Instead, it just flips the pause icon back to a play icon when the error occurs. We do show errors [on the episode details and list screens](https://github.com/Automattic/pocket-casts-android/pull/1025). 

It would be good to display some kind of error message that users don't have to find if they are on the Now Playing screen. Some possibilities include a toast (but the user would have to be looking at the watch), a blocking dialog (maybe too heavy handed), or a notification.

1. Install the watch app on a device without an internet connection
2. Attempt to play an episode that is not downloaded from the Episode details or Now Playing screens
3. Observe that once the playback errors, it just reverts the play/pause icon to be a play icon, and no error information is displayed to the user on this screen
4. Observe that there is error information on the Episode details and episode list screens from #1025 

Note that I trimmed the middle of this clip to keep it shorter. I actually had to wait 20 seconds or so for the playback to fail.

https://github.com/Automattic/pocket-casts-android/assets/4656348/32cd50a5-1aca-4a37-ac7f-63c2cbd5b8f1

Wear 3 emulator running a 7.40 pre-release build.

atharv028:
## Description

Adds an icon to Now Playing screen as suggested by @david-gonzalez-a8c. Upon clicking on icon or on the title/subtitle, it redirects to episode detail screen where the actual error is displayed. 

Fixes #1030 

## Testing Instructions
1. Turn off wifi and open the app.
2. Play any podcast.
3. ✅Verify that the error icon is displayed on now playing screen.

## Checklist


atharv028:
@mchowning creating a draft PR, as having some issues purchasing pocket cast plus subscription.

CookieyedCodes:
Care to share a screenshot, I'm interested what it looks like :)

mchowning:
Thanks for the contribution @atharv028 ! 🙇 

Your changes make sense, but the icon doesn't seem to be showing up correctly for me. 



Android can be picky about the svgs it likes, so we might have to use pngs for this. Here are some png assets in case we need them.

1x
bang-circle-fill@1x


1.5x
bang-circle-fill@1_5x

2x
bang-circle-fill@2x

3x
bang-circle-fill@3x

4x
bang-circle-fill@4x

mchowning:
I also think that it maybe good to add a bit of padding between the icon and the text so they're not so close together.

atharv028:
@mchowning thanks for testing it out, the icon was same in the jetpack preview but I thought that it would be different in real device. Turns out that wasn't the case.

I tried various things to make the svg icon show as it is but couldn't find anything. So for now, I have replaced the svg to the png set of icons. Also added a margin on 4dp between the icon and text.

Hopefully this resolves the issue

mchowning:
> I tried various things to make the svg icon show as it is but couldn't find anything. So for now, I have replaced the svg to the png set of icons. Also added a margin on 4dp between the icon and text.

Looks good! 👍 

I noticed that the icon size itself seemed a bit off to my eye, so I updated it and made it dynamically adjust to be the same as the text.

| | large font | small font |
| --- | ---- | ---- |
| static icon size | static-small | static-large |
| dynamic icon size | dynamic-small | dynamic-large |

I went ahead and pushed this change to make the icon size dynamic to your branch. Let me know if that looks good, or if you want to make any additional changes. If you're good with this though, I think we can merge this.

atharv028:
Seems perfect to me!! 🫶🏻

 I'll mark this PR as ready to review

mchowning (approved):

PR #1149, opened 2023-07-04
Bookmarks: Display bookmarks list

ashiagr:
## Description

This adds bookmarks list on full screen player > `Bookmarks` tab

## Testing Instructions

1. Apply the [patch](https://github.com/Automattic/pocket-casts-android/files/11949580/bookmarks.patch)
2. Install debug build
3. Go to `Discover`
4. Tap a podcast and play an episode
5. Open full player
6. Scroll to the `Bookmarks` tab
7. Notice that the UI matches the `Figma` design (Io7PCz1H1hmOgEE9eljoYW-fi-2033_32210)

## Checklist
 
- [x] with different themes

### DIFF
--- a/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/BookmarksFragment.kt
+++ b/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/BookmarksFragment.kt
@@ -1,0 +1,159 @@
+package au.com.shiftyjelly.pocketcasts.player.view
+
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.ViewGroup
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.lazy.LazyColumn
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material.Icon
+import androidx.compose.material.MaterialTheme
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.collectAsState
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.platform.ComposeView
+import androidx.compose.ui.platform.ViewCompositionStrategy
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.unit.dp
+import androidx.fragment.app.activityViewModels
+import androidx.lifecycle.asFlow
+import au.com.shiftyjelly.pocketcasts.compose.AppThemeWithBackground
+import au.com.shiftyjelly.pocketcasts.compose.components.TextH40
+import au.com.shiftyjelly.pocketcasts.compose.components.TextH60
+import au.com.shiftyjelly.pocketcasts.compose.theme
+import au.com.shiftyjelly.pocketcasts.player.viewmodel.PlayerViewModel
+import au.com.shiftyjelly.pocketcasts.views.fragments.BaseFragment
+import dagger.hilt.android.AndroidEntryPoint
+import au.com.shiftyjelly.pocketcasts.images.R as IR
+import au.com.shiftyjelly.pocketcasts.localization.R as LR
+
+@AndroidEntryPoint
+class BookmarksFragment : BaseFragment() {
+    private val playerViewModel: PlayerViewModel by activityViewModels()
+
+    override fun onCreateView(
+        inflater: LayoutInflater,
+        container: ViewGroup?,
+        savedInstanceState: Bundle?,
+    ) = ComposeView(requireContext()).apply {
+        setContent {
+            AppThemeWithBackground(theme.activeTheme) {
+                setViewCompositionStrategy(ViewCompositionStrategy.DisposeOnViewTreeLifecycleDestroyed)
+                BookmarksPage(
+                    playerViewModel = playerViewModel
+                )
+            }
+        }
+    }
+}
+
+@Composable
+fun BookmarksPage(
+    playerViewModel: PlayerViewModel,
+) {
+    val listData = playerViewModel.listDataLive.asFlow().collectAsState(initial = null)
+    listData.value?.let {
+        val backgroundColor = Color(it.podcastHeader.backgroundColor)
+        LazyColumn(
+            modifier = Modifier
+                .fillMaxSize()
+                .background(color = backgroundColor)
+        ) {
+            item {
+                HeaderItem()
+            }
+            item {
+                BookmarkItem(backgroundColor)
+            }
+        }
+    }
+}
+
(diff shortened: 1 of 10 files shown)


modules/features/player/src/test/java/au/com/shiftyjelly/pocketcasts/player/util/MainCoroutineRule.kt
```diff
@@ -0,0 +1,24 @@
...
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.TestDispatcher
+import kotlinx.coroutines.test.UnconfinedTestDispatcher
+import kotlinx.coroutines.test.resetMain
+import kotlinx.coroutines.test.setMain
+import org.junit.rules.TestWatcher
+import org.junit.runner.Description
+
+@ExperimentalCoroutinesApi
+class MainCoroutineRule(
```
ashiagr, line 13: This is duplicated across modules and needs to be extracted to a common test module or a test fixture if possible.

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/BookmarksFragment.kt
```diff
@@ -44,21 +81,198 @@ class BookmarksFragment : BaseFragment() {
 }
 
 @Composable
-private fun BookmarksPage(
+fun BookmarksPage(
```
geekygecko, line 84:
```suggestion
private fun BookmarksPage(
```

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/viewmodel/BookmarksViewModel.kt
```diff
@@ -0,0 +1,52 @@
...
+import kotlinx.coroutines.flow.stateIn
+import kotlinx.coroutines.launch
+import javax.inject.Inject
+
+@HiltViewModel
+class BookmarksViewModel
+@Inject constructor(
+    private val bookmarkManager: BookmarkManager,
+    private val episodeManager: EpisodeManager,
+    @IoDispatcher private val ioDispatcher: CoroutineDispatcher,
```
geekygecko, line 22: I'm learning something new with ever pull request!

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/BookmarksFragment.kt
```diff
@@ -44,21 +81,198 @@ class BookmarksFragment : BaseFragment() {
...
+
+@Composable
+private fun Content(
+    bookmarks: List<Bookmark>,
+    backgroundColor: Color,
+) {
+    LazyColumn(
+        modifier = Modifier
+            .fillMaxSize()
+            .background(color = backgroundColor)
```
geekygecko, line 118: Should we set this earlier, maybe as the background to the fragment as I'm seeing a white page flash while the content is loading.

Screenshot 2023-07-05 at 10 41 19 am

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/BookmarksFragment.kt
```diff
@@ -44,21 +81,198 @@ class BookmarksFragment : BaseFragment() {
...
+                .fillMaxWidth()
+                .padding(16.dp)
         ) {
-            TextH30(
-                text = "Placeholder Text",
-                color = MaterialTheme.theme.colors.playerContrast01
+            val createdAtText by remember {
+                mutableStateOf(
+                    bookmark.createdAt.toLocalizedFormatPattern(
+                        pattern = "MMM d, YYYY 'at' h:mm a"
```
geekygecko, line 190: I wonder if it would be nicer if we drop the yea if it's the current year. Probably more a question for our designer.

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/BookmarksFragment.kt
```diff
@@ -44,21 +81,198 @@ class BookmarksFragment : BaseFragment() {
...
+                    )
+                )
+            }
+            Column(
+                Modifier.weight(1f, fill = true)
```
geekygecko, line 195:
```suggestion
                Modifier.weight(1f)
```
The default is true so we might just need the weight. It seems strange to use a weight that wouldn't fill.

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/BookmarksFragment.kt
```diff
@@ -44,21 +81,198 @@ class BookmarksFragment : BaseFragment() {
...
+            ) {
+                TextH40(
+                    text = bookmark.title,
+                    color = MaterialTheme.theme.colors.playerContrast01,
+                    maxLines = 2,
+                )
```
geekygecko, line 201: It handled long titles well! 
Screenshot 2023-07-05 at 10 50 23 am

modules/services/images/src/main/res/drawable/ic_play.xml
```diff
@@ -0,0 +1,4 @@
+<vector android:height="24dp" android:viewportHeight="13"
+    android:viewportWidth="10" android:width="18.461538dp" xmlns:android="http://schemas.android.com/apk/res/android">
```
geekygecko, line 2: Is it worth trying to export the image with the size at whole numbers?

I just tried the following:
```
<vector xmlns:android="http://schemas.android.com/apk/res/android"
    android:width="10dp"
    android:height="13dp"
    android:viewportWidth="10"
    android:viewportHeight="13">
  <path
      android:pathData="M0,1.615C0,1.061 0.38,0.84 0.853,1.124L9.147,6.101C9.618,6.383 9.62,6.84 9.147,7.124L0.853,12.101C0.382,12.383 0,12.156 0,11.61V1.615Z"
      android:fillColor="#000000"/>
</vector>
```

Instead of setting the size to `.size(10.dp, 16.dp)` we can then set it to `.size(10.dp, 13.dp)` so they exactly match.

It's very hard to tell the difference so maybe it doesn't matter.
Screenshot 2023-07-05 at 11 07 40 am

geekygecko (approved):
It looks great, I only had a few minor comments.

modules/services/images/src/main/res/drawable/ic_play.xml
ashiagr: Good catch. I imported at standard 24dp size and it created fractional viewport sizes. I've updated it in 7d7d1922

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/BookmarksFragment.kt
ashiagr: It takes time for the view to load. I think it should be resolved when I add the layout for `Loading` and `Empty` states.

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/viewmodel/BookmarksViewModel.kt
ashiagr: So do I! Teamwork.

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/BookmarksFragment.kt
ashiagr: I dropped it in 5b7edf55. I can revisit it during the design review.

PR #1150, opened 2023-07-04
Long Press Playback Effect Turns Custom Effect On

CalebKL:
## Description
This PR sets up custom playback effects when you long press the Playback effect Icon.  

Fixes # @CookiedCodes niggles

## Testing Instructions

1. Tap on the Discover tab
2. Tap on any random podcast
3. Long Press on the Playback effect Icon
## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/95022986/0b5afa10-2939-4a31-9920-3f9106e99457

## Checklist
- [X ] If this is a user-facing change, I have added an entry in CHANGELOG.md
 
- [ ] with different themes

### DIFF
--- a/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/viewmodel/PlayerViewModel.kt
+++ b/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/viewmodel/PlayerViewModel.kt
@@ -2,6 +2,7 @@ package au.com.shiftyjelly.pocketcasts.player.viewmodel
 
 import android.content.Context
 import android.content.res.Resources
+import android.widget.Toast
 import androidx.lifecycle.LiveData
 import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.ViewModel
@@ -393,6 +394,28 @@ class PlayerViewModel @Inject constructor(
         playbackManager.upNextQueue.currentEpisode?.let {
             markAsPlayedConfirmed(it)
         }
+    }
+
+    fun toggleNotifications(context: Context) {
+        Toast.makeText(context, LR.string.player_effects_custom_effects_applied, Toast.LENGTH_SHORT).show()
+    }
+
+    fun updateOverrideGlobalEffects(override: Boolean) {
+        val podcast = this.podcast ?: return
+        launch {
+            podcastManager.updateOverrideGlobalEffects(podcast, override)
+            if (shouldUpdatePlaybackManager()) {
+                val effects = if (override) podcast.playbackEffects else settings.getGlobalPlaybackEffects()
+                playbackManager.updatePlayerEffects(effects)
+            }
+        }
+    }
+
+    private fun shouldUpdatePlaybackManager(): Boolean {
+        val podcast = this.podcast ?: return false
+        val currentEpisode = playbackManager.upNextQueue.currentEpisode
+        val podcastUuid = if (currentEpisode is PodcastEpisode) currentEpisode.podcastUuid else null
+        return podcastUuid == podcast.uuid
     }
 
     fun hasNextEpisode(): Boolean {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
         ([#1139](https://github.com/Automattic/pocket-casts-android/pull/1139))
     * Adds +- 1 increments to the sleep timer if the custom timer is less than 5 Min
         ([#1144](https://github.com/Automattic/pocket-casts-android/pull/1144)).
+    * Adds Long Press Click listener to the playback effect icon, which turns custom effects on.
+        ([#1150](https://github.com/Automattic/pocket-casts-android/pull/1150)).
 
 
 7.42
--- a/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/PlayerClickListener.kt
+++ b/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/PlayerClickListener.kt
@@ -9,6 +9,7 @@ interface PlayerClickListener {
     fun onSkipForward()
     fun onPlayClicked()
     fun onSkipForwardLongPress()
+    fun onEnableLocalPlayBackEffects()
     fun onClosePlayer()
     fun onEffectsClick()
     fun onSleepClick()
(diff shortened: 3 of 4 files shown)


ashiagr:
👋 

I just gave it a brief run and the override does seem like a useful feature! 

The shortcut to enable custom effects can remain but we were thinking if we could replace the snack bar like view with a toggle to switch between local and global podcast settings. Maybe it can remain visible all the time and be set to off by default. 

/cc @david-gonzalez-a8c

CookieyedCodes:
@ashiagr I was debating this as it was being mocked up and kind of came to the conclusion that a switch with a global/home icon in the top right would add too much visual clutter to this UI popup in my view but I'm not against it if it's just better 😅





Regarding what was implemented, I did find it a bit weird that it not popping up would not be possible but I suppose that's a problem with dealing with long press, after a day I was thinking why wouldn't a NAND gate not work tho,
As in IF info notification is turned on && the playback fx panel is not visible THEN popup the playback fx dialogue 

That should make it quite elegant in my view, if that's possible correct me if I am wrong/it's not possible tho 🤔


Regarding your suggestion just to clarify are you suggesting a link or that the notification has the toggle 🤔

ashiagr:
Sorry that it was not clear. I meant this view on the playback effect dialog and if we can convert it to a switch that will allow us to toggle between local/podcast effects:



> Regarding your suggestion just to clarify are you suggesting a link (that would popup the playback fx) or that the notification has the toggle 🤔

Oh no, notification is fine. No changes to that.

ashiagr:
> Regarding what was implemented, I did find it a bit weird that it not popping up would not be possible but I suppose that's a problem with dealing with long press, after a day I was thinking why wouldn't a NAND gate not work tho,
As in IF info notification is turned on && the playback fx panel is not visible THEN popup the playback fx dialogue

> That should make it quite elegant in my view, if that's possible correct me if I am wrong/it's not possible tho 🤔

I didn't notice this behavior earlier. I think what you suggested should work but I haven't tried it. @Mzazi25 have you tried it already and come across any issues?

CookieyedCodes:
> Sorry that it was not clear. I meant this view on the playback effect dialog and if we can convert it to a switch that will allow us to toggle between local/podcast effects:
> 
> 
> 
> > Regarding your suggestion just to clarify are you suggesting a link (that would popup the playback fx) or that the notification has the toggle 🤔
> 
> Oh no, notification is fine. No changes to that. 

Ah I got you, I think it's not a bad solution I think the only problem is that people would have to read to distinguish between local vs Global and might cause some mistap issues hence my recent thought on the icons 🤔🤔

ashiagr:
> people would have to read to distinguish between local vs Global and might cause some mistap issues hence my recent thought on the icons

Oh, I see. Let's try out toggling through icons then 👍

Mzazi25:
@ashiagr I tried  but since we are using a snackbar to toast the message, we can't adjust the position of the snackbar to the screen, hence it would not pop up above the "Custom Effects Applied" 
I guess that was the only challenge

Mzazi25:
> 👋 @Mzazi25 and @ashiagr
> 
> Just checking in on the status of this PR. Is there anything specific that needs to be done to move this forward?

I believe for this feature, there were some discussions going on, if it were a priority then.  Give me a green light and I will work on it ASAP
@ashiagr @mchowning lmk what you suggest

CookieyedCodes:
I was going to do a design like somewhat soonish -maby tomorrow- because I like the dial what Spotify did, I was also going to include a toggle for local & global playback FX 😉

CookieyedCodes:
A new design for a speed settings, its a dial UI with preset manager -long press the UI number- as you swipe left and right the numbers to indicate location, the ability to toggle between global and local settings is also added to this UI, Lightly inspired by Spotify's new dial UI & minimalist radio tuning UI.  Also unlocks 5x Speed because some people wants that as an option and would be less cumbersome with this UI (Local FX under podcast settings doesn't change UI)                                                                                                                                   image @Mzazi25 @ashiagr @mchowning @geekygecko @david-gonzalez-a8

david-gonzalez-a8c:
Hey there, @CookieyedCodes 

Thank you for your work on this, nice job! Here are my thoughts on the designs:

I personally think that the dial UI adds a lot of complexity to this screen, and I kinda like how in the current version we have a balance in the three settings with the icon, then the name of the setting, then the setting itself. I think I'd like to keep that so I'd consider a couple alternatives:

1. Use the dial UI in a subsequent screen, keeping the current screen more simple, but I'm not sure about the double modal:

image

2. Change the behaviour of the current screen so tapping on the selected speed displays a menu (I'm showing iOS here, as my design files are for iOS, but an equivalent could be used for Android, like [this component](https://m3.material.io/components/menus/overview)). This might be cleaner and more practical than the slider, and you still have the -/+ buttons to fine tune the speed if necessary.

image

What do you think?

CookieyedCodes:
@david-gonzalez-a8c

The main reason I put it above was to prevent accidental miss touches but I wouldn't be against a seperate UI popup that could accessed from a long press of the number 🤔
also I designed it a tad funky to look not exactly like Spotify 😅(just to be clear long pressing on the diall saves the current time to the preset but my preference would be 1)

Any thoughts on the local/global toggle 🤔

CookieyedCodes:
Gave a whole redesign of this screen more thought 
image

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/viewmodel/PlayerViewModel.kt
```diff
@@ -395,6 +396,28 @@ class PlayerViewModel @Inject constructor(
         }
     }
 
+    fun toggleNotifications(context: Context) {
+        Toast.makeText(context, LR.string.player_effects_custom_effects_applied, Toast.LENGTH_SHORT).show()
+    }
+
+    fun updateOverrideGlobalEffects(override: Boolean) {
```
mukul-scogo, line 403: Here, `override` looks like a pre-defined naming convention. Can we use self-explanatory name?

MiSikora:
@CalebKL We can proceed with the long press to enable podcast-specific effects without any switches or anything else. Could you sync you branch with `main`? I'll review the code once you do it.

Internal ref: p1726815809727549-slack-C05RR9P9RAT

MiSikora:
Thank you for this contribution. I'll close the ticket though because we'll address this issue ourselves in the near future.

PR #1151, opened 2023-07-05
Bookmarks: Add empty/ loading/ upsell views

ashiagr:
## Description
This adds bookmarks empty/ loading/ upsell views

## Testing Instructions

1. Install debug build without login or login with a free account
3. Go to `Discover`
4. Tap a podcast and play an episode
5. Open full player
6. ✅ Notice Plus upsell view matches `Figma` design (Io7PCz1H1hmOgEE9eljoYW-fi-2100%3A34078)
7. Dismiss the player 
8. Logout
9. Login with a Plus account
10. Repeat steps 3-5
11. ✅  Notice no bookmarks view matches `Figma` design (Io7PCz1H1hmOgEE9eljoYW-fi-2032%3A45279)
12. ✅  Notice that content is scrollable in landscape mode with large font

## Screenshots or Screencast 

Plus Upsell | No bookmarks | Large font portrait
-----|-----|---
|  | 

|Large font landscape |
| ----|

## Checklist
 
- [x] with different themes


modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/BookmarksFragment.kt
```diff
@@ -255,6 +287,112 @@ private fun PlayButton(
...
+@Composable
+private fun TextButton(
+    @StringRes buttonTitleRes: Int,
+    buttonAction: () -> Unit = {},
+    modifier: Modifier,
+) {
+    Box(
+        contentAlignment = Alignment.Center,
+        modifier = modifier
+            .minimumInteractiveComponentSize()
```
geekygecko, line 369: Thanks again for the Compose lesson! I didn't know about this method.

It would be great to make this component reusable. Maybe not as part of this pull request, but maybe we can try [Showkase](https://github.com/airbnb/Showkase) to make it easier to find and share components.

geekygecko (approved):
This worked well for me.

PR #1157, opened 2023-07-06
Have autoplay account for podcast grouping

mchowning:
## Description
This has autoplay take the episode grouping setting for a podcast into account when it selects what episode to play next. This is a follow-up to [this comment](https://github.com/Automattic/pocket-casts-android/pull/1132#issuecomment-1614279942).

## Testing Instructions

1. With the autoplay beta flag turned on (it is on by default), and the autoplay setting turned on (it is on by default for new installs), and with an empty up next queue
2. Start playing an episode from a podcast screen, then complete the episode so that another episode is autoplayed
3. Observe that the next unplayed, unarchived episode below the current episode (based on the user's selected sort order), is played. If there are no episodes below the currently playing episode available, the next nearest episode "above" the current episode will be selected. Note that this means that if you play the last episode in the "top" group, then the next episode to be autoplayed will be the first episode in the second group.
4. Repeat steps 2 and 3 with all the possible group options.

## Checklist

### DIFF
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/PodcastGrouping.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/PodcastGrouping.kt
@@ -3,8 +3,10 @@ package au.com.shiftyjelly.pocketcasts.models.to
 import android.content.Context
 import android.content.res.Resources
 import androidx.annotation.StringRes
+import au.com.shiftyjelly.pocketcasts.models.entity.Podcast
 import au.com.shiftyjelly.pocketcasts.models.entity.PodcastEpisode
 import au.com.shiftyjelly.pocketcasts.models.to.PodcastGrouping.Season.getSeasonGroupId
+import au.com.shiftyjelly.pocketcasts.models.type.EpisodesSortType
 import au.com.shiftyjelly.pocketcasts.localization.R as LR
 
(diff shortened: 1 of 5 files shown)


modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/PodcastGrouping.kt
```diff
@@ -81,7 +83,9 @@ sealed class PodcastGrouping(@StringRes val groupName: Int, val sortFunction: ((
      * @param episodes A sorted list of episodes
      * @return A pair of episodes and their group indexes
      */
-    open fun formGroups(episodes: List<PodcastEpisode>, reversedSort: Boolean = false, resources: Resources): List<List<PodcastEpisode>> {
+    open fun formGroups(episodes: List<PodcastEpisode>, podcast: Podcast, resources: Resources): List<List<PodcastEpisode>> {
+        val reversedSort = podcast.podcastGrouping is Season &&
+            podcast.episodesSortType == EpisodesSortType.EPISODES_SORT_BY_DATE_DESC
```
mchowning, line 88: I just moved this reverseSort logic into this function because (1) we were using the same logic everywhere we called this function and (2) it seemed like not super-intuitive logic so at least now we only have this not-intuitive-logic in a single place instead of multiple places.

ashiagr (approved):
This is working well in my testing. Nice work @mchowning!

PR #1164, opened 2023-07-10
Converted from kapt to KSP

geekygecko:
## Description

Google recommends migrating from kapt to KSP. "[it’s up to 2x faster](https://android-developers.googleblog.com/2021/09/accelerated-kotlin-build-times-with.html)". And it helps with my next PR of introducing Showkase, as it doesn't work well with kapt.

I followed [the guide](https://developer.android.com/build/migrate-to-ksp) to migrate the code. There is an issue that Daggger / Hilt don't support KSP yet but it seems fine to run them along side each other. Room, Moshi, and Glide will be using KSP.

### Build a production APK
1. Assemble all the builds `./gradlew assemble`.
2. Install the release build to an emulator `app/build/outputs/apk/release/app-release.apk`.
3. ✅ Verify the release installs and runs on the emulator.

### DIFF
--- a/base.gradle
+++ b/base.gradle
@@ -2,6 +2,7 @@ apply plugin: 'kotlin-android'
 apply plugin: 'kotlin-parcelize'
 apply plugin: 'kotlin-kapt'
 apply plugin: 'dagger.hilt.android.plugin'
+apply plugin: 'com.google.devtools.ksp'
 
 def SERVER_MAIN_URL_PROD = "\"https://refresh.pocketcasts.com\""
 def SERVER_API_URL_PROD = "\"https://api.pocketcasts.com\""
@@ -243,8 +244,9 @@ dependencies {
     implementation libs.media3Extractor
     implementation libs.media3Ui
 
-    kapt libs.moshiKotlinCompile
-    kapt libs.glideCompile
+    ksp libs.moshiKotlinCompile
+    ksp libs.glideCompile
+    // Dagger doesn't support KSP yet https://github.com/google/dagger/issues/2349
     kapt libs.hiltCompiler
     kapt libs.hiltDaggerCompiler
 
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,7 @@ buildscript {
     ext {
         kotlin_version = '1.7.10'
         // KSP releases need to match the Kotlin version - https://github.com/google/ksp/releases
-        ksp_version = '1.7.10-1.0.6'
+        ksp_version = "$kotlin_version-1.0.6"
         hilt_version = '2.43.2'
     }
     repositories {
--- a/build.gradle
+++ b/build.gradle
@@ -2,6 +2,8 @@
 buildscript {
     ext {
         kotlin_version = '1.7.10'
+        // KSP releases need to match the Kotlin version - https://github.com/google/ksp/releases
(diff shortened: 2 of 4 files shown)


modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/type/PodcastsSortType.kt
```diff
@@ -28,8 +33,8 @@ enum class PodcastsSortType(
         clientId = 2,
         serverId = 1,
         labelId = R.string.name,
-        podcastComparator = compareBy { cleanStringForSort(it.title) },
-        folderComparator = compareBy { cleanStringForSort(it.title) },
+        podcastComparator = compareBy { cleanStringForSortInternal(it.title) },
+        folderComparator = compareBy { cleanStringForSortInternal(it.title) },
```
geekygecko, line 37: This was to fix the following compile time warning.
```
[build file 'build.gradle': 48: argument list must be exactly 1 literal String or String with property replacement](w: modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/type/PodcastsSortType.kt:31:41 Companion object of enum class 'PodcastsSortType' is uninitialized here. This warning will become an error in future releases. See https://youtrack.jetbrains.com/issue/KT-49110 for details)
```

modules/services/model/build.gradle
```diff
@@ -5,10 +5,8 @@ apply plugin: 'kotlin-parcelize'
 android {
     namespace 'au.com.shiftyjelly.pocketcasts.model'
     defaultConfig {
-        javaCompileOptions {
-            annotationProcessorOptions {
-                arguments = ["room.schemaLocation": "$projectDir/schemas".toString()]
-            }
+        ksp {
+            arg("room.schemaLocation", "$projectDir/schemas")
```
geekygecko, line 9: This was to fix the following warning:
```
w: [ksp] modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/AppDatabase.kt:80: 
Schema export directory is not provided to the annotation processor so we cannot export the schema. 
You can either provide `room.schemaLocation` annotation processor argument OR set exportSchema to false.
```

build.gradle
```diff
@@ -2,6 +2,8 @@
 buildscript {
     ext {
         kotlin_version = '1.7.10'
+        // KSP releases need to match the Kotlin version - https://github.com/google/ksp/releases
+        ksp_version = '1.7.10-1.0.6'
```
mchowning, line 6: Since the versions need to match, should we do this?

```suggestion
        ksp_version = "$kotlin_version-1.0.6"
```

base.gradle
```diff
@@ -243,8 +244,9 @@ dependencies {
     implementation libs.media3Extractor
     implementation libs.media3Ui
 
-    kapt libs.moshiKotlinCompile
-    kapt libs.glideCompile
+    ksp libs.moshiKotlinCompile
+    ksp libs.glideCompile
+    // Dagger doesn't support KSP yet https://github.com/google/dagger/issues/2349
```
mchowning, line 249: Thanks for adding this comment. 👍

mchowning (approved):

PR #1165, opened 2023-07-10
Adding Showkase to manage our Compose components

geekygecko:
## Description

This introduces the Showkase tool to help us find and reuse existing Compose components.
https://github.com/airbnb/Showkase

I have yet to add all the components and instead marked the previews as private to get it set up with only a few initially. 

It would be good to write new Compose pages using components but if they don't exist then create new components. Having the `@Preview` annotation adds the component to Showkase but I think the best way might be to do the following. Adding a light and dark preview.

```
@ShowkaseComposable(name = "RowButton", group = "Button", styleName = "Light", defaultStyle = true)
@Preview(name = "Light")
@Composable
fun RowButtonsLightPreview() {
    AppThemeWithBackground(Theme.ThemeType.LIGHT) {
        RowButton(text = "Accept", onClick = {})
    }
}

@ShowkaseComposable(name = "RowButton", group = "Button", styleName = "Dark")
@Preview(name = "Dark")
@Composable
fun RowButtonsDarkPreview() {
    AppThemeWithBackground(Theme.ThemeType.DARK) {
        RowButton(text = "Accept", onClick = {})
    }
}
```

The colour of the developer icon isn't right but I might rewrite the page in Compose.

## Testing Instructions
1. Tap the Profile tab
2. Tap the settings cog
3. Tap the Developer option
4. Tap Showkase
5. ✅ Verify you can see the components


modules/services/images/src/main/res/drawable-hdpi/ic_heart_outline.xml
```diff
@@ -5,6 +5,6 @@
     android:viewportHeight="24">
   <path
       android:pathData="M3,10C3,6.6184 5.8218,4 9.1482,4C10.301,4 11.2686,4.4383 12,4.9569C12.7314,4.4383 13.699,4 14.8519,4C18.1782,4 21,6.6184 21,10C21,11.2923 20.504,12.4657 19.9147,13.4219C19.3172,14.3916 18.5359,15.2804 17.758,16.0432C16.2024,17.5686 14.4573,18.7853 13.6091,19.3447C12.63,19.9903 11.37,19.9903 10.3909,19.3447C9.5427,18.7853 7.7976,17.5686 6.242,16.0432C5.4641,15.2804 4.6828,14.3916 4.0853,13.4219C3.496,12.4657 3,11.2923 3,10ZM11.4919,17.675C11.8031,17.8802 12.1969,17.8802 12.5081,17.675C14.1179,16.6134 19,13.1354 19,10C19,7.7909 17.1428,6 14.8519,6C13.9074,6 13.0911,6.5478 12.5523,7.111C12.2781,7.3977 11.7219,7.3977 11.4477,7.111C10.9089,6.5478 10.0926,6 9.1482,6C6.8572,6 5,7.7909 5,10C5,13.1354 9.8821,16.6134 11.4919,17.675Z"
-      android:fillColor="#ffffff"
+      android:fillColor="#000000"
```
geekygecko, line 8: This icon isn't used anywhere else so thought it would be ok to change the base colour.

mchowning: This doesn't look great with a dark theme. 

image

Using the themed color seems to work.

```suggestion
      android:fillColor="?attr/primary_icon_01"
```

This is obviously a super important and critical mega blocker!!!!

mchowning (approved):
This looks nice. I just wonder if it's really going to be something that will  be useful and help us do do a better of reusing...

_< notices a component I should have used in my PR today >_
_< goes and [updates PR](https://github.com/Automattic/pocket-casts-android/pull/1170/commits/8b3666bdf0235547b61a0940f5158e3cfdf6a4dd) >_
_< clears throat >_

So... this seems really handy. 😄  Thanks so much for adding it!

modules/services/images/src/main/res/drawable-hdpi/ic_heart_outline.xml
geekygecko: It was happy locally but as the images module doesn't import any other the release failed.
```
ERROR: /pocket-casts-android/modules/services/images/build/intermediates/packaged_res/debug/drawable-hdpi-v4/ic_heart_outline.xml:9: 
AAPT: error: resource attr/primary_icon_01 (aka au.com.shiftyjelly.pocketcasts.testshared:attr/primary_icon_01) not found.
```
I'm going to change it back and upgrade the page to Compose. It's the only way to fix it. 😉

PR #1170, opened 2023-07-11
Whats new: Autoplay

mchowning:
## Description
This PR:
1. Adds a what's new dialog announcing the autoplay feature with new styling for the dialog. As part of this the old what's new dialog code is removed.
2. Removes the autoplay feature flag because we've decided to release this with `7.43`. **That means this PR is targeting `release/7.43`.**

The figma designs requested a pop-up dialog for this, but we've been using a bottom sheet everywhere else in the app. In this PR, I'm making a bottom sheet look like a pop-up dialog by fading out the scrim before the "dialog" is dismissed. If we stick with using dialogs, we may want to do a more complete switch, but for now I thought that this worked pretty well and helped to minimize the number of changes we need.

The language we're using in the What's new" dialog ("will autoplay episodes from your current list, if your Up Next queue is empty") seems misleading to me because it indicates that autoplay will play a new episode if I listen to all the episodes in my up next queue. In fact, autoplay only works if the Up Next queue is empty when you start playing the episode _and_ the Up Next queue stays empty the entire time you listen to the episode. 

I don't have any great ideas for how to communicate that in a short blurb though. If you have any suggestions for how to improve that, let me know because I am concerned that our users will be confused when they finish their up next queue and autoplay doesn't do anything. internal ref: p1689101648157169/1689091788.481629-slack-C02ATC80MUM

Related iOS PR: https://github.com/Automattic/pocket-casts-ios/pull/956

### 1. Verify What's New is shown when appropriate
1. Download a 7.19.2 build from the Google Play Store (basically any build before folders was released in 7.20), install it, and start the app. Observe there is no announcement pop-up/dialog
2. Download a 7.42 build from the Google Play Store, install it, and start the app. Dismiss the onboarding flow and observe the old "What's New" dialog is displayed.
4. Change the code in this PR [to force the build number to be 9115](https://gist.github.com/mchowning/2ce380ac8f04c5085b89f7376ea849a8), and install a release build.
5. Observe that the new What's New dialog is shown announcing the autoplay feature. Note that the animation on this is smooth on release builds, but really choppy on debug builds.

### 2. Test What's New dialog
- Verify that tapping on the "Enable it now" button (1) closes the dialog, (2) Opens the general settings, (3) scrolls to the bottom of that screen, and (4) flashes the autoplay setting
- Verify that tapping either "Maybe later" or somewhere outside the dialog closes the dialog.

### 3. Confirm removal of feature flag
- Verify that on debug builds there is no longer a feature flag option in the settings and the autoplay setting is visible.
- Verify that on release builds the autoplay setting is visible

## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/4656348/30277b60-dc30-4ab2-b477-08f638e1c68c

||||||
| :---: | :---: | :---: | :---: | :---: |
| Light | Dark | Rosé | Indigo | Extra Dark |
| Screenshot 2023-07-11 at 2 18 14 PM | Screenshot 2023-07-11 at 2 19 32 PM | Screenshot 2023-07-11 at 2 19 48 PM | Screenshot 2023-07-11 at 2 20 03 PM | Screenshot 2023-07-11 at 2 20 21 PM |
| Dark Contrast | Light Contrast | Electricity | Classic | Radioactivity |
| Screenshot 2023-07-11 at 2 21 18 PM | Screenshot 2023-07-11 at 2 21 30 PM | Screenshot 2023-07-11 at 2 21 43 PM | Screenshot 2023-07-11 at 2 22 05 PM | Screenshot 2023-07-11 at 2 22 17 PM |

## Checklist
 
- [X] with different themes


CookieyedCodes:
@mchowning thanks for adding this & getting it to ignore archived podcasts & respect grouping ❤️❤️.

Any thoughts on these following things:
1. It might be a nice feature add that the next upcomeing episode downloads in the background as the current episode is playing 🤔 (no gaps 🤔 😉)

2. As the upnext is empty it might be nice to remind people of the autoplay function & tell them what is upcoming (also good for  download symbolism as mentioned in 1)🤔🤔🤔

3.With a feature like this it might be smart to re-add sorting A-Z & Z-A on filters for people who want to host their own audio books on rss -yes their are a few people who would like to do this as folder files would also be appreciated & enhance this feature (also another ++ for folder pills with folder sort order btw)-

4. If you use add to upnext for keeping on top of podcasts will this feature stop or continue on reading the auto play after catching up on the podcasts that were added to upnext (might be a good toggle too as I would prefer it to pickup & continue but maby not everyone)

mchowning:
👋 @CookieyedCodes 

> 1. It might be a nice feature add that the next upcomeing episode downloads in the background as the current episode is playing 🤔 (got to avoid gaps 🤔 😉❤️)
> 
> 2. As the upnext is empty it might be nice to remind people of the autoplay function & tell them what is upcoming (also good for download symbolism as mentioned in 1)🤔🤔🤔

> 3. With a feature like this it might be smart to re-add sorting A-Z & Z-A on filters for people who want to host their own audio books on rss -yes their are a few people who would like to do this as folder files would also be appreciated & enhance this feature (also another ++ for folder pills with folder sort order btw)-

Thanks for the suggestions, we'll keep them in mind. This is very much a first iteration for this feature, and there are a lot of things I think we can do to build on it and make it better.

> If you use add to upnext for keeping on top of podcasts will this feature stop or continue on reading the auto play after catching up on the podcasts that were added to upnext (might be a good toggle too as I would prefer it to pickup & continue but maby not everyone)

The feature currently only works if you are not using your up next queue at all. Basically autoplay will only happen if your up next queue is empty when you start playback, and the queue stays empty until you are finished with playback. It's almost an alternative to the up next queue in this iteration. The focus was to help out people who migrate from other podcast apps since they tend to (a) expect playback to continue after an episode is finished and (b) are not familiar with Pocket Casts and the Up Next queue system we have. This is also possibly a first very small step toward playlists.

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/whatsnew/WhatsNewFragment.kt
```diff
@@ -1,107 +1,260 @@
...
+            return Settings.WHATS_NEW_VERSION_CODE > (versionCode ?: 0)
+        }
     }
+}
 
-    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
-        super.onViewCreated(view, savedInstanceState)
+@OptIn(ExperimentalAnimationApi::class)
+@Composable
+private fun WhatsNewComposable(
```
geekygecko, line 114: Is it weird to use Composable in the component name? Would Page be better. This is a super important comment, feel free to ignore it. 😄

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/whatsnew/WhatsNewFragment.kt
```diff
@@ -1,107 +1,260 @@
...
 
-        val pages = WhatsNew.pages
-        binding.pageIndicatorView.count = pages.size
-        binding.pageIndicatorView.isVisible = pages.size > 1
+                    // Using the same spring animation for the scaleIn and rotation ensures that they
+                    // finish at the same time.
+                    val springAnimation = spring<Float>(
+                        stiffness = Spring.StiffnessVeryLow,
+                        dampingRatio = Spring.DampingRatioLowBouncy,
+                    )
```
geekygecko, line 168: This is awesome to see! 🏆

geekygecko (approved):
It worked well for me.

This probably isn't part of this PR but it seems strange to take the user to the settings when they tap enable. Instead of showing them were it is hidden in settings maybe we should have some UI in the player or Up Next for them to easily toggle it on or off.

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/whatsnew/WhatsNewFragment.kt
mchowning: > Is it weird to use Composable in the component name? Would Page be better. 

I did mean to update that name and just forgot. Thanks for catching it. Updated to Page in [e84b9c7](https://github.com/Automattic/pocket-casts-android/pull/1170/commits/e84b9c7a7f87ec7fe7ba0ac7a03c7904cc215df0).

mchowning:
> it it seems strange to take the user to the settings when they tap enable. Instead of showing them were it is hidden in settings maybe we should have some UI in the player or Up Next for them to easily toggle it on or off.

I do see your point, but I also think it is nice how this gives the user a hint about where the setting exists in case they want to update it in the future.

PR #1171, opened 2023-07-12
Developer settings clean up

geekygecko:
## Description

This pull request makes the following changes:
- Rewrites the Developer page in Compose. 
- Adds a preview to Showkase for the RowButton disabled and without padding.
- Fixes the Showkase preview for the UserPasswordFields in the dark theme.
- Adds the Text components to Showkase.
- Changes the SettingRow component to accept a Painter rather than a GradientIconData. It felt a bit strange always passing a GradientIconData and it means we can use Compose ImageVectors at a later date.

As part of the Developer page rewrite I removed the actions "Sync up next" and "Launch billing dialog" as they were no longer being used.

## Testing Instructions
1. Tap the Profile tab
2. Tap the settings cog
3. Tap the Developer section
4. ✅ Verify the Developer looks correct

## Screenshots

| Before | After |
| --- | --- |
| Screenshot_20230712_115845 | Screenshot_20230712_161700 |


mchowning (approved):
Looks good!

PR #1175, opened 2023-07-13
added left and right padding to toolbar

### LINKED ISSUE #63: Spacing off on Add File screen header
The save button on the add file header is too close to the right side of the screen. 

Also, the header leaves room for the top system bar, but it covers up that bar, so there is too much blank space at the top of the screen.

P-Burk:
## Description

I just want to preface this with this is my first open source contribution. I use the Pocket Casts app almost everyday and am an aspiring software engineer, so I decided to give a few of the issues tagged as "good first issue" a shot. I figured moving the _Save_ button on the Add File screen would be a good start. As it is now, the _Save_ button is a bit too far to the right of the screen as you can see in the before screenshot below.

Fixes #63 

## Testing Instructions

1. Go to the user account page.
2. Click on the _Files_ button.
3. Click the _Add_ button (Floating Action Button). 
4. Select a file to upload.
5. You will automatically be taken to the _Add File_ screen.

#### Before:
image

#### After:
image

## Checklist
- It's just a minor UI tweak.
- I don't think this change needs any tests updated.
- No strings added.
- No compose components added or changed.
 
- [x] with different themes
- Changing the orientation on the _Add File_ screen actually introduces a new bug for me where the entire top toolbar and the audio file's art go away.
- I'm not sure how to change the font.
- I don't know how to test this.

## Notes:
I wanted to add that I am getting a warning in Android studio that I should use `paddingStart` and `paddingEnd` rather than `paddingLeft`/`paddingRight`. I tried this, but no change occurred in the position of the _Save_ button. For what it's worth, my emulated device is a Pixel 6 on API 33.

### DIFF
--- a/modules/features/profile/src/main/res/layout/activity_add_file.xml
+++ b/modules/features/profile/src/main/res/layout/activity_add_file.xml
@@ -18,7 +18,9 @@
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
             android:background="?attr/secondary_ui_01"
-            android:minHeight="?android:attr/actionBarSize" />
+            android:minHeight="?android:attr/actionBarSize"
+            android:paddingLeft="0dp"
+            android:paddingRight="20dp"/>
     </com.google.android.material.appbar.AppBarLayout>
 
     <ScrollView
--- a/modules/features/profile/src/main/res/layout/activity_add_file.xml
+++ b/modules/features/profile/src/main/res/layout/activity_add_file.xml
@@ -20,7 +20,9 @@
             android:background="?attr/secondary_ui_01"
             android:minHeight="?android:attr/actionBarSize"
             android:paddingLeft="0dp"
-            android:paddingRight="20dp"/>
+            android:paddingStart="0dp"
+            android:paddingRight="20dp"
+            android:paddingEnd="20dp"/>
     </com.google.android.material.appbar.AppBarLayout>
 
     <ScrollView


ashiagr:
Hey @P-Burk!

Thank you for this contribution! And great job finding an existing issue in the code while answering PR questions so well! 🥇 

On that note, can you report the bug in the `Issues` section? You can also take it up as your next task, but feel free to choose something else. 🙂
 
> I wanted to add that I am getting a warning in Android studio that I should use paddingStart and paddingEnd rather than paddingLeft/paddingRight. I tried this, but no change occurred in the position of the Save button. For what it's worth, my emulated device is a Pixel 6 on API 33.

Android [recommends](https://android-developers.googleblog.com/2013/03/native-rtl-support-in-android-42.html#:~:text=Change%20all%20of%20your%20app%27s%20%22left/right%22%20layout%20properties%20to%20new%20%22start/end%22%20equivalents.) using `paddingStart` / `paddingEnd` for RTL mirroring, you do not see any changes by using them as the app doesn't support RTL well and has `android:supportsRtl=false` in the AndroidManifest. It might be good to fix the lint warning and test it by  `android:supportsRtl=true` in the AndroidManifest temporarily. We should be able to merge the PR after this.

> I'm not sure how to change the font.
> I don't know how to test <talkback>.

While you don't need to test them out for these changes, here are some helpful links:

[Display size](https://support.google.com/accessibility/android/answer/11183305?hl=en)
[Talkback](https://support.google.com/accessibility/android/answer/6007100?hl%3Den)

P-Burk:
Hello @ashiagr. Sorry for not getting back sooner, but thank you for all the help and details!!

> It might be good to fix the lint warning and test it by android:supportsRtl=true in the AndroidManifest temporarily. We should be able to merge the PR after this.

I tested this out and it works as intended. I left both sets of padding, which still generates a lint warning, but it will fix the issue for when/if RTL support is more fully fleshed out and the issue as it is currently.

> On that note, can you report the bug in the Issues section? You can also take it up as your next task, but feel free to choose something else.

I reported this bug. It is issue #1192.

ashiagr (approved):

PR #1177, opened 2023-07-16
Watch: Updated empty state for filter screen

### LINKED ISSUE #1154: Watch: Filters screens are empty when there are no episodes in the filter
### Description

The filters screens just present an empty screen when there are no episodes. We should present some kind of empty state (even if it is minimal like we have [for the files screen](https://github.com/Automattic/pocket-casts-android/pull/915)) or for the podcasts screen when there are no subscribers.

1. Log into a new account with no podcast subscriptions
2. Check the filters screens for New Releases and In Progress and notice they are empty

image

By contrast, this is what the podcasts screen shows when there are no subscriptions:

image

Wear OS 3 emulator, 7.42 pre-release build

atharv028:
## Description

Adds a message in case there are no episodes in the selected filter.

Fixes #1154 

## Testing Instructions
1. Log into a new account with no podcast subscriptions
2. Check the filters screens for New Releases and In Progress.
3. ✅ Verify that the user is displayed a message telling that no results are found.

## Screenshots or Screencast 
Screenshot_20230716_152508

## Checklist

### DIFF
--- a/modules/services/localization/src/main/res/values/strings.xml
+++ b/modules/services/localization/src/main/res/values/strings.xml
@@ -580,6 +580,8 @@
 
     <string name="filters">Filters</string>
     <string name="episode_filters">Episode Filters</string>
+    <string name="no_results_for_filters">No results found for filter %s</string>
+    <string name="no_result_found">No results found</string>
     <string name="filters_all_your_podcasts">All your podcasts</string>
     <string name="filters_auto_download_limit" translatable="false">@string/episodes_plural</string>
     <string name="filters_chip_episode_status">Episode Status</string>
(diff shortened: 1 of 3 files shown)


modules/services/localization/src/main/res/values/strings.xml
```diff
@@ -580,6 +580,8 @@
 
     <string name="filters">Filters</string>
     <string name="episode_filters">Episode Filters</string>
+    <string name="no_results_for_filters">No results found for filter %s</string>
```
ashiagr, line 583: Can we reuse the "filters_no_episodes" string used for the phone app? Since the filter title is already shown at the top, we can skip adding it here again.

modules/services/localization/src/main/res/values/strings.xml
```diff
@@ -580,6 +580,8 @@
...
+    <string name="no_result_found">No results found</string>
```
ashiagr, line 584: Np: This can changed to "No filters found".

ashiagr (approved):

PR #1179, opened 2023-07-17
Linter Check on the PR Template 

Mzazi25:
## Description

Added  `./gradlew :spotlessApply` linter check on the PR template

Fixes # 

## Checklist
 
- [ ] with different themes


mchowning (commented):
Thanks for this! Adding the reminder to the PR checklist is a good idea. Just curious, why did you choose to create a shell script and add an instruction for running that to the PR description instead of giving the gradle shell command in the PR description?

Mzazi25:
> Thanks for this! Adding the reminder to the PR checklist is a good idea. Just curious, why did you choose to create a shell script and add an instruction for running that to the PR description instead of giving the gradle shell command in the PR description?

@mchowning My bad. You are right, I should have written the the shell command on the PR description. 
should I change it?

mchowning:
Using the shell script is fine, but I would lean toward not using a separate script if there's not a clear benefit to doing so. The fewer things we have to maintain the better.

mchowning:
👋 @Mzazi25 ! Would you mind going ahead and removing the shell script here? The more I think about it, the more I think it would be better to not have something else to maintain. Let me know if you disagree though.

Mzazi25:
> would

Okay, Its fine. I agree with you, the less things to maintain, the better. So should I close this PR?

mchowning:
What about updating the PR to remove the shell script and instead directly referencing the gradle spotless command in the PR template? I do think that this is something that trips up a lot of contributors, so it seems like a good idea to point out the linter like you're doing.

Mzazi25:
I have update the PR. 
Removed the shell script and added the linter check on the PR template.

mchowning (approved):
I did a small tweak to remove the `:` from the command. Thanks for contributing this change—I think it will help out a lot of people! 🙇

PR #1182, opened 2023-07-18
Bookmarks: Add delete action

ashiagr:
## Description

This adds a delete action on the bookmarks list screen

## Testing Instructions

1. Add a few bookmarks for an episode to the bookmark table (with delete = 0) 
2. Install the debug build and login with a Plus account
3. Open the Full Screen Player for the episode selected in step 1
4. Swipe over to Bookmarks
5. Long press on a row
6. From the multiselect toolbar, tap delete icon
7. Notice that a confirmation dialog is shown
8. Confirm "Delete bookmark"
9. ✅ Notice that the bookmark is deleted
10. Repeat it for multiple bookmarks
11. ✅ Notice that bookmarks are deleted

** Bookmarks are marked deleted locally, they will be synced with the server in a separate sync request.

## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/1405144/bdf443b1-4a70-4508-ad3f-63085380e86f

## Checklist
 
- [x] with different themes

### DIFF
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/BookmarkDao.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/BookmarkDao.kt
@@ -31,8 +31,8 @@ abstract class BookmarkDao {
     @Query("SELECT * FROM bookmarks WHERE podcast_uuid = :podcastUuid AND episode_uuid = :episodeUuid AND time = :timeSecs LIMIT 1")
     abstract suspend fun findByEpisodeTime(podcastUuid: String, episodeUuid: String, timeSecs: Int): Bookmark?
 
-    @Query("SELECT * FROM bookmarks WHERE podcast_uuid = :podcastUuid AND episode_uuid = :episodeUuid AND deleted = 0")
-    abstract fun findByEpisodeFlow(podcastUuid: String, episodeUuid: String): Flow<List<Bookmark>>
+    @Query("SELECT * FROM bookmarks WHERE podcast_uuid = :podcastUuid AND episode_uuid = :episodeUuid AND deleted = :deleted")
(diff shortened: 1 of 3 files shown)


modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/BookmarkDao.kt
```diff
@@ -31,7 +31,7 @@ abstract class BookmarkDao {
     @Query("SELECT * FROM bookmarks WHERE podcast_uuid = :podcastUuid AND episode_uuid = :episodeUuid AND time = :timeSecs LIMIT 1")
     abstract suspend fun findByEpisodeTime(podcastUuid: String, episodeUuid: String, timeSecs: Int): Bookmark?
 
-    @Query("SELECT * FROM bookmarks WHERE podcast_uuid = :podcastUuid AND episode_uuid = :episodeUuid")
+    @Query("SELECT * FROM bookmarks WHERE podcast_uuid = :podcastUuid AND episode_uuid = :episodeUuid AND deleted = 0")
     abstract fun findByEpisodeFlow(podcastUuid: String, episodeUuid: String): Flow<List<Bookmark>>
```
geekygecko, line 35:
```suggestion
    @Query("SELECT * FROM bookmarks WHERE podcast_uuid = :podcastUuid AND episode_uuid = :episodeUuid AND deleted = :deleted")
    abstract fun findByEpisodeFlow(podcastUuid: String, episodeUuid: String, deleted: Boolean = false): Flow<List<Bookmark>>
```
As the method doesn't say it only returns bookmarks which haven't been deleted, how about passing it as a parameter with a default value? Feel free to ignore this idea.

geekygecko (approved):
This worked really well.

PR #1185, opened 2023-07-20
Migrate share fragment to Jetpack Compose

CalebKL:
## Description

This PR migrates the share Fragment views to Jetpack Compose.
The Compose version is the exact replica from the views. 
In case of any padding values I missed, I would be glad to counter check
Fixes # 

## Checklist
 
- [ ] with different themes


modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/ShareFragment.kt
```diff
@@ -4,124 +4,264 @@ import android.os.Bundle
...
-        viewModel.playingEpisodeLive.observe(viewLifecycleOwner) { (_, backgroundColor) ->
-            applyColor(theme, backgroundColor)
-        }
+private val PillSize = DpSize(width = 48.dp, height = 4.dp)
+private val PillCornerRadius = 10.dp
 
-        (dialog as? BottomSheetDialog)?.behavior?.state = BottomSheetBehavior.STATE_EXPANDED
-    }
+@Composable
+private fun Pill(
```
ashiagr, line 240: There can be reusable components that can be extracted out like this `Pill` and the component for each option.  Can you describe the challenges you had in using the `OptionsDialog()` that you initially proposed to use?

This will help us decide the architecture for the rest of the "Playback Effects" dialogs.

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/ShareFragment.kt
```diff
@@ -4,124 +4,264 @@ import android.os.Bundle
...
-                    requireContext(),
-                    ShareType.EPISODE_FILE,
-                    SourceView.PLAYER,
-                    analyticsTracker
-                ).sendFile()
+            Divider(
+                color = Color.LightGray.copy(0.3f),
+                thickness = 1.dp,
+            )
+            Divider(thickness = 1.dp)
```
ashiagr, line 188: Duplicate divider? There's one above.

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/ShareFragment.kt
Mzazi25: I agree, the pill can be reused. Regarding the Options Dialog, it would work but one concern: 
If we migrate all fragments using `BaseDialogFragment()` to `optionsDialog()`, we would have to add new rows for each specific fragment. That means for each fragment that is being migrated, we would add new custom composables on the dialog. But if this would not be a problem, it would work.

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/ShareFragment.kt
```diff
@@ -4,124 +4,264 @@ import android.os.Bundle
...
+    private lateinit var composeView: ComposeView
     private val viewModel: PlayerViewModel by activityViewModels()
-    private var binding: FragmentShareBinding? = null
     private var disposable: Disposable? = null
-
     override fun onPause() {
         super.onPause()
 
         disposable?.dispose()
     }
```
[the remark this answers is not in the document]
Mzazi25: Sure thing

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/ShareFragment.kt
Mzazi25: okay

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/ShareFragment.kt
ashiagr: > That means for each fragment that is being migrated, we would add new custom composables on the dialog. 

Could we pass a custom option component (which is only shown on a single screen) to the `OptionsDialog()` so that we don't increase the class size too much and yet be able to have a common class across all these option fragments?

@mchowning, @geekygecko any thoughts?

ashiagr: @Mzazi25 👋 

I had a brief discussion with the team, with `OptionsDialog()` we'll be able to reuse common components and see some advantages. Can we give it a try?

Mzazi25: Sure thing. Let me work on that

mchowning:
👋 @Mzazi25 ! Are you still planning to work on this?

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/ShareFragment.kt
```diff
@@ -4,124 +4,264 @@ import android.os.Bundle
...
-                SharePodcastHelper(
-                    podcast,
-                    null,
-                    null,
-                    requireContext(),
-                    ShareType.PODCAST,
-                    SourceView.PLAYER,
-                    analyticsTracker
-                ).showShareDialogDirect()
+    @Suppress("DEPRECATION")
```
Mukuljangir372, line 79: `onActivityCreated` has been deprecated. Should we consider to use `OnViewCreated` function? wdyt?

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/ShareFragment.kt
```diff
@@ -4,124 +4,264 @@ import android.os.Bundle
...
-                    SourceView.PLAYER,
-                    analyticsTracker
-                ).showShareDialogDirect()
+        (dialog as? BottomSheetDialog)?.behavior?.state = BottomSheetBehavior.STATE_EXPANDED
+    }
+}
+
+@Composable
+private fun ShareScreen(
+    podcast: Podcast?,
```
Mukuljangir372, line 155: As we are passing `Podcast` and `BaseEpisode` object to composable function, Should we mark these classes with @Stable for compose-runtime?

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/ShareFragment.kt
```diff
@@ -4,124 +4,264 @@ import android.os.Bundle
...
+                thickness = 1.dp,
+            )
+            Divider(thickness = 1.dp)
+            if (podcast != null && episode is BaseEpisode) {
+                ShareButton(
+                    label = IR.string.podcast_share_episode,
+                    onClick = onShareEpisode
+                )
+                Divider(
+                    color = Color.LightGray.copy(0.3f),
```
Mukuljangir372, line 195: We should consider to use `MaterialTheme` outline color as divider colours rather than hardcoding them. Wdyt?

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/ShareFragment.kt
```diff
@@ -4,124 +4,264 @@ import android.os.Bundle
...
+    label: Int,
+    onClick: () -> Unit,
+    modifier: Modifier = Modifier
+) {
+    Box(
+        modifier = modifier
+            .fillMaxWidth()
+            .clickable(
+                interactionSource = remember { MutableInteractionSource() },
+                indication = rememberRipple(color = Color.LightGray)
```
Mukuljangir372, line 221: We should consider to use `MaterialTheme` outline color as divider colours rather than hardcoding them. Wdyt?

Mukuljangir372 (changes requested):

MiSikora:
Closing this one as `ShareFragment` no longer exists and `ShareDialogFragment` is migrated to compose.

PR #1190, opened 2023-07-21
Add share action to swipe menus

mchowning:
## Description

This adds a share option to the swipe menus as requested in the [Increase "Share" discoverability](https://github.com/Automattic/pocket-casts-android/discussions/1168) discussion. 

Because our swipe menus are custom made to handle up to two icons on the left and only a single icon on the right, adding a second icon to the right was a bit tricky. 

In addition, the way this was set up before was kind of complex because the click listener for the buttons came from a completely different place from where the icon and background color was set. This made it very difficult to avoid bugs due to these two sources of information getting out of sync. After spending a day or so playing whack-a-mole fixing a bug, and then fixing a new bug the previous fix caused, and then fixing a new bug the previous..., etc., I decided to do a broader refactor. Now a lot more of the swipe button logic is consolidated in the `SwipeButtonLayoutViewModel`.

I think this consolidates the logic around the share buttons nicely, but it does not consolidate the view handling itself—everywhere we want to use a swipe layout still requires a fair bit of manual XML work, but addressing that didn't seem worth it because (1) I already spent more time on this than I wanted and (2) the whole app will be written in compose soon anyway. 😉 

Screenshot 2023-07-21 at 2 29 46 PM

Related iOS PR: https://github.com/Automattic/pocket-casts-ios/pull/966

## Testing Instructions

Because of the refactoring, when testing this PR, we should test all the possible swipe actions, not just the share action. Apart from adding the share button to the Podcast, Filter, Downloads, Starred, and Listening History screens, all other swipe behavior should be unchanged.

Some things to watch out for include:
1. Make sure that when you release the swipe menu so the buttons remain visible, that the buttons have the right width (had a lot of bugs where it wasn't 1-button-wide when showing 1 button and 2-buttons-wide when showing two buttons).
2. Do multiple actions that change the buttons on a single row (i.e., adding and removing the same episode from the up next queue) to make sure the buttons up date appropriately for the episode's new state.

### 1. Podcast, Filter, Downloads, Starred, and Listening History:

* **For unqueued episodes:** LtR swipes reveal the add to up next options. The left button should be the default that user has selected under the General settings. Doing a full-swipe should apply the left button. The only way to apply the right button is the menu open a small amount, release the swipe so the buttons remain visible, and tap the right button.
* **For queued episodes:** LtR swipes reveal the remove from queue option.
* **For archived episodes:** RtL swipes reveal the share and archive buttons. The archive button is on the right and is the action that is taken with a full swipe.
* **For unarchived episodes:** RtL swipes are the same as for archived episodes except the archive action is changed to an unarchived action.

### 2. Files

The Files screen's swipe behavior should be unchanged by this PR. 
* LtR swipes will reveal the add to up next options (ordered based on the user's setting). 
* RtL swipes will reveal only the delete option.

### 3. Up Next Queue

* LtR swipes will always reveal the add-to-queue (or in this case move-in-queue) buttons. Unlike on the other screens, the order here never changes. The left button is move-to-top and the right button is move-to-bottom regardless of what the user's setting says should be the default action.
* RtL swipes reveal the remove from queue button

### 4. Analytics
* For each swipe action, we should send an appropriate`episode_swipe_action_performed` with `action` and `source`, and `episode_uuid` properties. 
* When sharing an episode from the swipe menu, the `podcast_shared` event is sent with a `source` of "swipe_action"
* I have added the `episode_uuid` as a property to the `episode_*_to_up_next` events since we were missing that, but it was identified on the tracking document

## Checklist
 
- [X] with different themes

### DIFF
--- a/modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/filters/FilterEpisodeListFragment.kt
+++ b/modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/filters/FilterEpisodeListFragment.kt
@@ -383,7 +383,18 @@ class FilterEpisodeListFragment : BaseFragment() {
         binding.btnChevron.setOnClickListener(clickListener)
         toolbar.setOnClickListener(clickListener)
 
-        val itemTouchHelper = EpisodeItemTouchHelper(this::episodeSwipedRightItem1, this::episodeSwipedRightItem2, viewModel::episodeSwiped)
+        val itemTouchHelper = EpisodeItemTouchHelper(
+            onLeftItem1 = this::episodeSwipeLeftItem1,
+            onLeftItem2 = this::episodeSwipeLeftItem2,
+            onRightItem1 = viewModel::episodeSwipeRightItem1,
+            onRightItem2 = { baseEpisode, _ ->
+                viewModel.episodeSwipeRightItem2(
+                    baseEpisode = baseEpisode,
+                    context = context ?: return@EpisodeItemTouchHelper,
+                    fragmentManager = parentFragmentManager,
+                )
+            },
+        )
         itemTouchHelper.attachToRecyclerView(recyclerView)
 
         val multiSelectToolbar = binding.multiSelectToolbar
@@ -563,7 +574,7 @@ class FilterEpisodeListFragment : BaseFragment() {
             .show(childFragmentManager, "confirm")
     }
 
-    private fun episodeSwipedRightItem1(episode: BaseEpisode, index: Int) {
+    private fun episodeSwipeLeftItem1(episode: BaseEpisode, index: Int) {
         when (settings.getUpNextSwipeAction()) {
             Settings.UpNextAction.PLAY_NEXT -> viewModel.episodeSwipeUpNext(episode)
             Settings.UpNextAction.PLAY_LAST -> viewModel.episodeSwipeUpLast(episode)
@@ -571,7 +582,7 @@ class FilterEpisodeListFragment : BaseFragment() {
         adapter.notifyItemChanged(index)
     }
 
-    private fun episodeSwipedRightItem2(episode: BaseEpisode, index: Int) {
+    private fun episodeSwipeLeftItem2(episode: BaseEpisode, index: Int) {
         when (settings.getUpNextSwipeAction()) {
             Settings.UpNextAction.PLAY_NEXT -> viewModel.episodeSwipeUpLast(episode)
             Settings.UpNextAction.PLAY_LAST -> viewModel.episodeSwipeUpNext(episode)
--- a/modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/filters/FilterEpisodeListFragment.kt
+++ b/modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/filters/FilterEpisodeListFragment.kt
@@ -49,6 +49,7 @@ import au.com.shiftyjelly.pocketcasts.views.fragments.BaseFragment
 import au.com.shiftyjelly.pocketcasts.views.fragments.BaseFragmentToolbar.ChromeCastButton.Shown
 import au.com.shiftyjelly.pocketcasts.views.helper.EpisodeItemTouchHelper
 import au.com.shiftyjelly.pocketcasts.views.helper.NavigationIcon.BackArrow
+import au.com.shiftyjelly.pocketcasts.views.helper.SwipeButtonLayoutFactory
 import au.com.shiftyjelly.pocketcasts.views.helper.ToolbarColors
 import au.com.shiftyjelly.pocketcasts.views.multiselect.MultiSelectEpisodesHelper
 import au.com.shiftyjelly.pocketcasts.views.multiselect.MultiSelectHelper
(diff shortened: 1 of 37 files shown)


modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/filters/FilterEpisodeListFragment.kt
```diff
@@ -92,7 +95,28 @@ class FilterEpisodeListFragment : BaseFragment() {
...
+            swipeButtonLayoutFactory = SwipeButtonLayoutFactory(
+                swipeButtonLayoutViewModel = swipeButtonLayoutViewModel,
+                onItemUpdated = this::lazyNotifyAdapterChanged,
+                defaultUpNextSwipeAction = { settings.getUpNextSwipeAction() },
+                context = requireContext(),
+                fragmentManager = parentFragmentManager,
+                swipeSource = EpisodeItemTouchHelper.SwipeSource.FILTERS,
+            )
+        )
+    }
```
mchowning, line 118: I moved the adapter initialization to be lazy because initializing it in `onAttach` was causing problems due to the fact that the `swibeButtonLayoutViewModel` had not been injected yet. Let me know if you think this might cause any problems.

modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/helper/SwipeButtonLayout.kt
```diff
@@ -0,0 +1,221 @@
...
+import au.com.shiftyjelly.pocketcasts.ui.extensions.getThemeColor
+import au.com.shiftyjelly.pocketcasts.images.R as IR
+import au.com.shiftyjelly.pocketcasts.ui.R as UR
+import au.com.shiftyjelly.pocketcasts.views.R as VR
+
+data class SwipeButtonLayout(
+    val leftPrimary: () -> SwipeButton,
+    val leftSecondary: () -> SwipeButton?,
+    val rightPrimary: () -> SwipeButton,
+    val rightSecondary: () -> SwipeButton?,
```
mchowning, line 18: Making these all lambdas to avoid the views loading with stale data.

CookieyedCodes:
Any thoughts on my comments to make both options on swipe left to right customisable?

mchowning:
> Any thoughts on my comments to make both options on swipe left to right customisable?

We've logged it as a possible future improvement @CookieyedCodes . The changes in this PR should make it easier to do that at some point in the future.

CookieyedCodes:
That it should 😉 ❤️

modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/helper/SwipeButtonLayout.kt
```diff
@@ -0,0 +1,216 @@
...
+            if (episode !is PodcastEpisode) {
+                throw IllegalStateException("Can only share podcast episodes: $episode")
+            }
+            viewModel.share(episode, fragmentManager, context, swipeSource)
+        }
+    }
+}
+
+/**
+ * - [onDeleteOrArchiveClick] will delete [UserEpisode]s and archive [PodcastEpisode]s
```
geekygecko, line 163: Is this right? `onDeleteOrArchiveClick` I can't see it anywhere else in the code.

modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/helper/SwipeButtonLayoutViewModel.kt
```diff
@@ -0,0 +1,259 @@
...
+    ) {
+
+        viewModelScope.launch(Dispatchers.Default) {
+
+            trackSwipeAction(
+                swipeSource = swipeSource,
+                swipeAction = EpisodeItemTouchHelper.SwipeAction.SHARE,
+            )
+
+            val podcast = podcastManager.findPodcastByUuid(episode.podcastUuid) ?: return@launch
```
geekygecko, line 51:
```suggestion
        viewModelScope.launch {

            trackSwipeAction(
                swipeSource = swipeSource,
                swipeAction = EpisodeItemTouchHelper.SwipeAction.SHARE,
            )

            val podcast = podcastManager.findPodcastByUuidSuspend(episode.podcastUuid) ?: return@launch
```

Should we use the coroutine version `findPodcastByUuidSuspend` and let Room decide the dispatcher it wants to use?

geekygecko (approved):
Excellent work changing this code, it's a complicated part of the app.
I'm unsure about another view model rather than including another class in the view model, but I didn't have to work with all this complicated code! 
Thanks for tackling this feature request.

modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/helper/SwipeButtonLayoutViewModel.kt
mchowning: Yes, this is much better. Thanks!

mchowning:
> I'm unsure about another view model rather than including another class in the view model,

The reason I went with the view model approach was to get the lifecycle handling for the coroutine scope for free. We could definitely do this without a view model though. Let me know if you think we should change it.

PR #1188, opened 2023-07-21
Add bookmarks from the player shelf

geekygecko:
## Description

You can now add bookmarks from the full screen player.

It uses another activity so I could add `android:windowSoftInputMode="adjustResize"` and have the "Save bookmark" button just above the keyboard. I'm not really happy with the amount of code to setup the activity but it might be worth it to be able to add a bookmark quickly. 

## Testing Instructions
1. Play an episode
2. Tap on the mini player to open the full screen player
3. Tap on the shelf three dots icon
4. Tap Add bookmark
5. ✅ Verify the "Add bookmark" page is shown
6. Tap "Save bookmark"
7. ✅ Verify a Snackbar is shown
8. Tap the "View" action in the Snackbar
9. ✅ Verify it takes you to the bookmarks tab and the new bookmark is present
10. Tap the "Now Playing" 
11. Tap on the shelf three dots icon
12. Tap the edit pencil
13. Use the drag handles to move the "Add bookmark" item up to the "Shortcut on player"
14. Tap the back button
15. ✅ Verify the "Add bookmark" icon is on the shelf
16. Tap the icon
17. ✅ Verify it opens the add bookmark page

## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/308331/1786c8dd-42d5-4642-b6aa-3c383cac2ed0

PR #1193, opened 2023-07-23
Seeking to a bookmark position

geekygecko:
| Merge after https://github.com/Automattic/pocket-casts-android/pull/1188 |
|---|

## Description

This contains a few changes:

1. On the player bookmarks tab, tapping on a bookmark play button seeks to that time.
2. The bookmark play button has been extracted to a common component and added to Showkase.
3. The show notes seek to toast has been added to the localisation strings.

## Testing Instructions
1. Play an episode
3. Tap add bookmark in the shelf
4. Change to the bookmarks tab
5. Tap a bookmark play button
6. ✅ Verify the playback jumps to the correct position

### DIFF
--- a/modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/buttons/TimePlayButton.kt
+++ b/modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/buttons/TimePlayButton.kt
@@ -32,7 +32,7 @@ import com.airbnb.android.showkase.annotation.ShowkaseComposable
 
 sealed class TimePlayButtonStyle {
     object Solid : TimePlayButtonStyle()
-    object Outline : TimePlayButtonStyle()
+    object Outlined : TimePlayButtonStyle()
 }
 
 @Composable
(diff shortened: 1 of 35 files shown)


modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/buttons/TimePlayButton.kt
```diff
@@ -0,0 +1,138 @@
...
+) {
+    val timeText by remember {
+        mutableStateOf(TimeHelper.formattedSeconds(timeSecs.toDouble()))
+    }
+    val description = stringResource(contentDescriptionId, timeText)
+    val border = when (buttonStyle) {
+        TimePlayButtonStyle.Outline -> BorderStroke(2.dp, borderColor)
+        TimePlayButtonStyle.Solid -> null
+    }
+    OutlinedButton(
```
geekygecko, line 57: I noticed the highlight was escaping the button shape so I have switched it to an OutlinedButton.
Screenshot 2023-07-22 at 10 06 51 am

modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/buttons/TimePlayButton.kt
```diff
@@ -0,0 +1,131 @@
...
+@Composable
+fun TimePlayButton(
+    timeSecs: Int,
+    @StringRes contentDescriptionId: Int,
+    onClick: () -> Unit,
+    modifier: Modifier = Modifier,
+    textColor: Color = MaterialTheme.theme.colors.primaryText01,
+    backgroundColor: Color = MaterialTheme.colors.surface,
+    borderColor: Color = MaterialTheme.theme.colors.primaryText01,
+    buttonStyle: TimePlayButtonStyle = TimePlayButtonStyle.Outlined
```
geekygecko, line 47: I didn't know if it would be better to have multiple components such as `OutlinedTimePlayButton` and `SolidTimePlayButton` but I went with a style parameter. Let me know if you think a different name would be better.

ashiagr: Works for me 👍

ashiagr (approved):

PR #1195, opened 2023-07-24
Wear/LoginWithGoogleViewmodel Unit Test

Mzazi25:
## Description

This PR adds unit tests for the wear/ Login with GoogleViewmodel. 
I also added a small testing library [turbine](https://github.com/cashapp/turbine) for Kotlin Flows. 

Fixes # 

## Checklist
 
- [ ] with different themes

### DIFF
--- a/base.gradle
+++ b/base.gradle
@@ -206,6 +206,7 @@ dependencies {
     implementation androidLibs.composeUiUtil
     implementation androidLibs.wearPlayServices
 
+    implementation libs.turbine
     implementation libs.kotlinCoroutines
     implementation libs.kotlinCoroutinesAndroid
     implementation libs.kotlinCoroutinesPlayServices
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -271,6 +271,8 @@ project.ext {
             lottieCompose: "com.airbnb.android:lottie-compose:5.2.0",
             // JUnit
             junit: "junit:junit:4.13.2",
+            //turbine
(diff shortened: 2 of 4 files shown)


base.gradle
```diff
@@ -206,6 +206,7 @@ dependencies {
     implementation androidLibs.composeUiUtil
     implementation androidLibs.wearPlayServices
 
+    implementation libs.turbine
```
mchowning, line 209: I think we can change this to `testImplementation` (and move it down with the other test dependencies). That way this doesn't get bundled into the app's apk.

wear/build.gradle
```diff
@@ -117,6 +117,7 @@ dependencies {
     implementation project(':modules:services:preferences')
     implementation project(':modules:services:repositories')
     implementation project(':modules:services:servers')
+    implementation project(':modules:services:sharedtest')
```
mchowning, line 120: I think this can also be `testImplementation` instead of `implementation` since it is only used for tests. If you wanted to also make this change in the other places where we depend on the `sharedtest` module as well, that would be great.

mchowning (changes requested):
Thanks so much for adding some more tests @Mzazi25 . I just saw two small changes I think we should make before merging this.

Mzazi25:
@mchowning I have update the PR with the changes requested

mchowning (approved):
Looks good! Thanks! 🙇

PR #1196, opened 2023-07-24
Remove duplicated MainCoroutine Rule

Mzazi25:
## Description

This PR adds a global MainCoroutine Rule, so we can reuse it instead of having multiple duplicates. 

Fixes # 

## Checklist
 
- [ ] with different themes


ashiagr:
👋 @Mzazi25 

I understand [why](https://stackoverflow.com/questions/49436559/android-multi-module-test-dependency) you've put the test rule inside the main source instead of the test source in a module that is at the bottom of the multi-module hierarchy.

I just have one concern the test rule doesn't seem to be at a place where it should belong and might not be easily discoverable. 

I wonder if having a separate shared test module like `services/sharedtest` makes more sense. I can see this pattern being followed in `Google`'s [ioshed](https://github.com/google/iosched/tree/main/test-shared/src/main/java/com/google/samples/apps/iosched/test/data) repository. I understand that it will have just one file at the moment but we will know where to look for and add shared test utils in the future. Wdyt?

Mzazi25:
> 👋 @Mzazi25 
> 
> I understand [why](https://stackoverflow.com/questions/49436559/android-multi-module-test-dependency) you've put the test rule inside the main source instead of the test source in a module that is at the bottom of the multi-module hierarchy.
> 
> I just have one concern the test rule doesn't seem to be at a place where it should belong and might not be easily discoverable. 
> 
> I wonder if having a separate shared test module like `services/sharedtest` makes more sense. I can see this pattern being followed in `Google`'s [ioshed](https://github.com/google/iosched/tree/main/test-shared/src/main/java/com/google/samples/apps/iosched/test/data) repository. I understand that it will have just one file at the moment but we will know where to look for and add shared test utils in the future. Wdyt?
> 

Its a good call. We could also add livedata test util helpers in the same module when we test the viewmodels that uses livedata. For now I can migrate the MainCoroutine rule to the suggested module.

Mzazi25:
@ashiagr kindly review the changes. I have created a new `sharedtest` module and refactored the rule into the module

ashiagr (approved):
Thank you for these updates!

PR #1197, opened 2023-07-24
Add logging for network selected by networking rules

mchowning:
## Description
We've had a few reports of issues downloading episodes on the watch. We're using the horologist network awareness library for downloads, and this PR logs which network is selected by that library. My thinking is that it could be helpful to see in the logs what network is being selected when the user reports that downloads are failing.

Fixes # 

## Testing Instructions
1. Download an episode in the watch app
2. Verify that the "Preferred network according to networking rules..." line appears in the logs

## Checklist

### DIFF
--- a/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/networking/PocketCastsNetworkingRules.kt
+++ b/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/networking/PocketCastsNetworkingRules.kt
@@ -2,6 +2,7 @@ package au.com.shiftyjelly.pocketcasts.wear.networking
 
 import androidx.annotation.VisibleForTesting
 import au.com.shiftyjelly.pocketcasts.BuildConfig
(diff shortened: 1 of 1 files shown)


Mzazi25:
@mchowning regarding this, could we also use [Chucker](https://github.com/ChuckerTeam/chucker) to help debug the network calls?

mchowning:
I hadn't seen Chucker before. Does it provide more than the network inspector in Android Studio?

I think Chucker is addressing a slightly different issue than the LogBuffer change I'm making here. For one thing, I don't think see anything indicating that Chucker would work on a watch, and this change is focused on issues users are reporting with the watch app. Second, it looks like Chucker could be helpful during development, but the LogBuffer isn't really for development—it is adding more information to the logs that users send to support when they have a problem.

geekygecko (approved):
That worked for me.
```
BgTask: Preferred network according to networking rules: NetworkStatus(id=175, status=Available, type=Wifi(name=wlan0, ssid=null), addresses=[/192.168.86.40])
```

Nice use of `also`.

Issue #424, opened 2022-10-20
Add to F-Droid


licaon-kter:
I see some stuff that needs to be separated under a gradle flavor (google services, cast, firebase, billing. oss-license: as these are either not FOSS or depend on not-FOSS libs). Sentry/Tracks is ok but it might warrant a `Tracking` antifeature if not opt-in.

A custom maven repo is used for which dep (Automattic-Tracks)? That [is FOSS](https://github.com/Automattic/Automattic-Tracks-Android). Is it available in other repos, eg. Maven Central? Or we need to build it.

/LE: updated deps info

eighthave:
As an F-Droid contributor, I'm here to offer my help getting Pocket Casts to be free software.

xrishox:
would love to see this project added to fdroid

IzzySoft:
Still not even a response from the team? Third F-Droid contributor here. There's already an [inclusion request](https://gitlab.com/fdroid/rfp/-/issues/2236) open with us for several month which we cannot proceed with, as we'd need your help. Count the thumbs how many folks wait to see your app there, doesn't that convince you it should be included? You were even offered a hand for that. So may I kindly ask you to join the efforts? Thanks in advance – @mchowning & all!

mchowning:
Thanks for following up @IzzySoft . This had fallen off my radar.

It would be great to see Pocket Casts on F-Droid, but unfortunately this is not a focus for the core team at this time. I'm happy to help move this forward and do what I can unblock anyone who wants to work on this though. 

What do you think would be the best first steps? Is there anything you need to unblock this work?

> I see some stuff that needs to be separated under a gradle flavor (google services, cast, firebase, billing. oss-license: as these are either not FOSS or depend on not-FOSS libs). 

How would we go about separating these? Something like calling a wrapper around these libraries that has a noop implementation in the f-droid build flavor, or something else?

> Sentry/Tracks is ok but it might warrant a Tracking antifeature if not opt-in.

Just applying the Tracking anti-feature flag sounds like the best option at this time to me.

IzzySoft:
[...]

PR #1206, opened 2023-07-25
set up free/play build flavors and move CastOptionsProvider to play

eighthave:
## Description

The sets up the standard pattern for building free releases of apps that also use proprietary libraries.  This draws on patterns used by Nextcloud, Element and other apps.  There are two build flavors:

* _free_: this should ultimately end up building a free open source APK.
* _gplay_: this builds the current version, with all the existing proprietary dependencies.

This includes one fix as an example of how build flavors work for this kind of thing.  The `CastOptionsProvider` comes from the proprietary Chromecast library.  This moves all that to the _gplay_ flavor.

first step towards #424

## Testing Instructions

```
./gradlew assemble
./gradlew assembleFreeDebug
./gradlew assembleGplayDebug
```

* Check whether Chromecast works in the _gplay_ APK.

## Checklist


base.gradle
```diff
@@ -149,6 +149,16 @@ android {
...
         }
     }
+
+    flavorDimensions "license"
+    productFlavors {
+        free {
+            dimension "license"
+        }
+        gplay {
+            dimension "license"
```
mchowning, line 159: Let's make the `gplay` dimension the default one.

```suggestion
            dimension "license"
            isDefault true
```

mchowning (changes requested):
Thanks for the contribution @eighthave!

### 1. Chromecasting is failing

Chromecasting isn't working for me on the `gplay` variant. Whenever I try to chromecast it gets stuck "Finding devices". Without the changes in this PR, I'm able to immediately find multiple chromecast devices. Is this working for you?

https://github.com/Automattic/pocket-casts-android/assets/4656348/b41f8edf-7b35-42d6-bc2e-9fadf5cecc13

### 2. CI is failing

CI is failing because it's trying to build without specifying the new dimension. I think updating the tasks [here](https://github.com/Automattic/pocket-casts-android/blob/b8287a4b97e4c1deac852e861b7ea17583412d3d/fastlane/Fastfile#L252) to be "['assembleGplayDebug', 'assembleGplayDebugAndroidTest']" will fix that.

### 3. [Nice to have] Automotive and Wear app modules and the free flavor

It doesn't have to be a part of this PR, but it seems like we should probably remove the free flavor from the automotive and wear app modules. I don't think it's possible to have a "free" automotive or wear os app (correct me if I'm wrong). What do you think?

eighthave:
> 1. Chromecasting is failing

Oops, sorry about that.  I used the wrong name for the flavor folder ("play" when it should have been "gplay".  I don't have an easy way to test with Chromecast.  So I used _diffoscope_ to do a before/after diff on the APKs.  I can see before the class and manifest entry were missing, now they are there.

> 2. CI is failing

Should be fixed.

> 3. [Nice to have] Automotive and Wear app modules and the free flavor

Sounds good to me. I worked with Gradle Build Flavors a fair amount, and I can't think of how to do this.  Do you have a suggestion?  Given the structure with _base.gradle_ being shared across modules, I think it would require a bigger change to the Gradle setup.

mchowning:
Thanks for the updates!

For removing the build variants, I think something along these lines in the app and wear modules works to exclude the free variants.

``` groovy
android {

    ...

    variantFilter { variant ->
        if (variant.flavors*.name.contains("free")) {
            // Automotive app should not have a free variant
            setIgnore(true)
        }
    }
}
```

Unrelated to this, CI is still failing (but it's a new failure at least). I haven't found the fix yet, but it looks like the `useGoogleServicesDebugFile` task is not getting run anymore, and this is causing the build to fail due to the missing `google-services.json` file. If I execute `./gradlew :app:preBuild --info` from the `main` branch, the `useGoogleServicesDebugFile` task runs, but it gets skipped with the changes on this PR (it gets skipped because "task onlyIf 'Task is enabled' is false"). 🤔

eighthave:
I'll try adding the code that removes the `free` tasks from _Wear_ and _TV_.

I also saw that build failure intermittently on my machine.  I can't actually see the build logs, it seems I need an Automattic account to see them on buildkite.com?  Other CI services just let me click and see them.  

I would solve it by removing that task, and just committing the `google-services.json` file in the `debug` flavor folder, e.g.:

    git mv app/google-services.json_debug-only app/src/debug/

I don't understand what exactly that task is doing beyond just putting the file somewhere where the build will pick it up.

eighthave:
I'll be AFK for vacation, I'll come back to this when I'm back.

eighthave:
I'd like to continue to work on this.  Any feedback on https://github.com/Automattic/pocket-casts-android/pull/1206#issuecomment-1666427743 ?

mchowning:
> I would solve it by removing that task, and just committing the google-services.json file in the debug flavor folder, e.g.:
>
> git mv app/google-services.json_debug-only app/src/debug/
>
> I don't understand what exactly that task is doing beyond just putting the file somewhere where the build will pick it up.

Seems like that should work—at least I can't think of any reason that wouldn't work now. It would be nice to simplify the build process. 👍 

I'm assuming we won't need to update [.configure](https://github.com/Automattic/pocket-casts-android/blob/main/.configure) since release builds would still fall back to the file at the root. We'll probably need to update [.gitleaksignore](https://github.com/Automattic/pocket-casts-android/blob/main/.gitleaksignore) though to ignore the new `google-services.json` files in the `debug` folders (and we can remove the currently ignored files if those get removed).

Since this change is directly dealing with how we handle secrets, let's make this change in a separate PR so it is easier to review and debug any issues that might crop up.

> I also saw that build failure intermittently on my machine. I can't actually see the build logs, it seems I need an Automattic account to see them on buildkite.com? 

I'll check into this.

eighthave:
> Since this change is directly dealing with how we handle secrets, let's make this change in a separate PR so it is easier to review and debug any issues that might crop up.

I posted https://github.com/Automattic/pocket-casts-android/pull/1434

eighthave:
I rebased this on #1434 and added https://github.com/Automattic/pocket-casts-android/pull/1206#issuecomment-1666200671 to remove the free flavors from Automotive and Wear

eighthave:
I rebased this now that #1434 is merged, so this should be ready to go.  Everything works locally for me.  The one thing I couldn't figure out was how to use the new version catalogs syntax in the `productFlavor` block.  The old syntax works there still.

eighthave:
Let me know when you're close to reviewing this and I can rebase and fix the merge conflicts.

eighthave:
I'm ready to pick this up again, just let me know when you can review it, and I'll rebase it on _master_.

mchowning:
I'm switching away from being a Pocket Casts core contributor, so someone else from @Automattic/pocket-casts-android will need to review this going forward.

julianfoad:
Could someone else from @Automattic/pocket-casts-android review this? It looks like this was well on its way to being a useful step towards F-Droid eligibility #424 ...

eighthave:
If I know that it'll be reviewed, I'll happily fix the merge conflicts.

MiSikora:
I'm closing this PR as outlined in #424.

PR #1213, opened 2023-07-28
Settings Sync - Refactor "General" settings to be UserSettings

mchowning:
## Description
This PR is the first step toward refactoring our settings to be `UserSettings` in an effort to make it easier to sync our settings in the future (and easy to have new settings be synced).

I'm only including the settings from the "General" screen in order to get early feedback on this before I invest more time in this direction. Also, this helps make it a bit easier to review the PR (only 32 changed files ! 😅 ).

The main point of this is to simplify what it takes to add a syncable setting so it is easier to add synced settings in the future. For example, with this change, instead of the SettingsImpl class needing to implement four methods/fields:

1. skipForwardInSecsObservable
2. getSkipForwardInSecs
3. setSkipForwardInSecs
4. getSkipForwardNeedsSync

We now just need a single val skipForwardInSecs: UserSetting field, and that also covers the coroutine flow as well (which would be a another field that needed to be implemented before this PR).

> **Note**
> I have this PR targeting a feature branch because I think it would be good to hold these changes back for now. That will allow me to make further changes to how `UserSetting`s work if necessary without having that directly impact our production users. Also, this reserves for us the option of not rolling this change out until we start syncing settings more broadly (I'm not sure that's a good idea, but I'm not sure that it isn't either). Let me know if you think we're better off merging to `main` because I don't feel strongly.

## Testing Instructions
For each setting in the "General" settings:
1. Fresh install the app
3. Log into an account
4. Change the setting to a new value
5. Force close the app
6. Verify that the setting has retained the new value
7. If the setting is one that is synced (skip-forward and skip-back times):
    1. Sync the app
    2. uninstall the app
    3. reinstall the app and log into the account
    4. Once the account has synced, verify that the updated setting value is shown
8. Verify that the app's behavior reflects the updated setting value (i.e., that skip forward actually skips forward based on the new value).

## Checklist

### DIFF
--- a/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/PodcastManagerTest.kt
+++ b/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/PodcastManagerTest.kt
@@ -1,11 +1,12 @@
 package au.com.shiftyjelly.pocketcasts.models.db
 
+import androidx.preference.PreferenceManager
 import androidx.room.Room
 import androidx.test.internal.runner.junit4.AndroidJUnit4ClassRunner
 import androidx.test.platform.app.InstrumentationRegistry
 import au.com.shiftyjelly.pocketcasts.models.db.dao.PodcastDao
 import au.com.shiftyjelly.pocketcasts.models.entity.Podcast
-import au.com.shiftyjelly.pocketcasts.preferences.Settings
+import au.com.shiftyjelly.pocketcasts.preferences.SettingsImpl
 import au.com.shiftyjelly.pocketcasts.repositories.playback.PlaybackManager
 import au.com.shiftyjelly.pocketcasts.repositories.podcast.EpisodeManager
 import au.com.shiftyjelly.pocketcasts.repositories.podcast.PlaylistManager
@@ -17,6 +18,7 @@ import au.com.shiftyjelly.pocketcasts.servers.cdn.StaticServerManager
 import au.com.shiftyjelly.pocketcasts.servers.podcast.PodcastCacheServerManager
 import au.com.shiftyjelly.pocketcasts.servers.refresh.RefreshServerManager
 import au.com.shiftyjelly.pocketcasts.utils.Optional
+import com.squareup.moshi.Moshi
 import io.reactivex.Single
 import org.junit.After
 import org.junit.Assert.assertTrue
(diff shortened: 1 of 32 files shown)


app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/PodcastManagerTest.kt
```diff
@@ -42,7 +44,15 @@ class PodcastManagerTest {
...
-        val settings = mock<Settings>()
+
+        val settings = SettingsImpl(
+            sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context),
+            privatePreferences = PreferenceManager.getDefaultSharedPreferences(context),
+            context = context,
+            firebaseRemoteConfig = mock(),
+            moshi = Moshi.Builder().build(),
+        )
+
```
mchowning, line 55: I don't feel great about this, but no matter what way I tried to mock out the settings, I got crashes in the tests when it tried to retrieve a `UserSetting::flow` object. Let me know if you have any ideas for getting this to work with mocks.

modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/UserSetting.kt
```diff
@@ -0,0 +1,162 @@
...
+                apply()
+            }
+        }
+
+    fun getSyncValue(): T? =
+        if (needsSync) get() else null
+
+    private val _flow by lazy { MutableStateFlow(get()) }
+    // lazy because the class needs to initialize before calling get()
+    val flow by lazy { _flow }
```
mchowning, line 25: Using `by lazy` here feels more like a hacky workaround than a proper solution, but without this the app crashes on startup because it seems to call `get()` when initializing `_flow` before the non-null constructor parameters are present. Let me know if you have any ideas for a better way to handle this.

modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/UserSetting.kt
```diff
@@ -0,0 +1,163 @@
...
+                apply()
+            }
+        }
+
+    fun getSyncValue(): T? =
+        if (needsSync) get() else null
+
+    private val _flow by lazy { MutableStateFlow(get()) }
+    // lazy because the class needs to initialize before calling get()
+    val flow: StateFlow<T> by lazy { _flow }
```
mchowning, line 26: Using `by lazy` here is more like a hacky workaround than a proper solution, but without this the app crashes on startup because it seems to call `get()` when initializing `_flow ` before the (non-null!) constructor parameters are present. I think this is because of the order in which superclasses and subclasses are initialized. I'm sure there's a better way to handle this, but I haven't figured it out yet.

Let me know if you have any ideas for a better way to handle this.

ashiagr (approved):
Your changes look good to me, @mchowning! Just to be careful with settings, I'll keep the PR open today till I do some more testing. 

Thank you for the commit breakdown, it made it easy to review the PR.

ashiagr:
All "General" settings were synced properly as described. Great work!

app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/PodcastManagerTest.kt
mchowning: I found a way to avoid this in https://github.com/Automattic/pocket-casts-android/pull/1234/commits/2767f55d32961a2516b758513b0fcb73e0bc120e

PR #1186, opened 2023-07-20
Bookmarks - Add sorting

ashiagr:
## Description

This adds sorting to the bookmarks list.

## Testing Instructions

1. Add a few bookmarks to the bookmarks table with different `create_at` and `time` values
2. Launch the app and login with Plus account
3. Open the Full Screen Player
4. Swipe to the bookmarks list
5. Tap the vertical <kbd>...</kbd> button
6. ✅ Verify you see an options menu with 2 options: Select bookmarks, Sort by
7. ✅ Verify the default sort by option is newest to oldest
8. Tap on Select bookmarks
9. ✅ Verify the view enters multi select mode
10. Tap back arrow, then tap the more (...) button again
11. Tap the Sort by option
12. ✅ Verify a new options menu opens with 3 options: "Newest to oldest", "Oldest to newest", and "Timestamp"
13. Tap on the "Oldest to newest"
14. ✅ Verify the list is now sorted with the oldest bookmarks first
15. Open the sort menu again, choose "Timestamp"
16. ✅ Verify the list is now sorted with the earliest timestamp at the top

## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/1405144/c57a1b1d-30d8-4bc1-bd66-17ba8f3ea1a8

## Checklist
- [ ] If this is a user-facing change, I have added an entry in CHANGELOG.md
- [x] I have considered whether it makes sense to add tests for my changes
- [x] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml`
- [ ] Any jetpack compose components I added or changed are covered by compose previews
 
#### I have tested any UI changes...

- [x] with different themes
- [x] with a landscape orientation
- [x] with the device set to have a large display and font size
- [ ] for accessibility with TalkBack

PR #1223, opened 2023-07-31
Bookmarks - Add sorting on podcast page

ashiagr:
## Description

This adds bookmark sorting on the podcast page.

## Testing Instructions
1. Go to the podcast page with a few bookmarks added to at least it's two different episodes
2. Go to the bookmarks tab
3. Tap on vertical "..." to see sort options
4. ✅ Verify that menu opens with 3 options: "Newest to oldest", "Oldest to newest", and "Episode"
5. Notice that the default option in "Newest to oldest"
6. Tap on the "Oldest to newest"
7. ✅ Verify the list is now sorted with the oldest bookmarks first
8. Open the sort menu again, choose "Episode"
9. ✅ Verify that the most recent episode is shown first, and then by timestamp within each episode.

#### Regression

Test that sort order on the player bookmarks screen works as before.

## Checklist
 
- [x] with different themes


CookieyedCodes:
I would also add a sort A-Z & Z-A option
(While you at it could you add sort A-Z & Z-A to filters, some would appreciate that for hosting their own audiobooks via RSS especially with the autoplay functionality 😉 🤔)

ashiagr:
> I would also add a sort A-Z & Z-A option

👍 I'll share this and [this one](https://github.com/Automattic/pocket-casts-android/pull/1186#issuecomment-1646584972) with the team @CookieyedCodes. If not in the initial release, we can include them in the next one.

CookieyedCodes:
@ashiagr no bothers 😉 ❤️, if your getting too many sort options it might be time to condense the sort options too 🤔

With a toggle on tap
 _sort option_          /\ to    _sort option_      \/

modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/SettingsImpl.kt
```diff
@@ -1454,16 +1456,32 @@ class SettingsImpl @Inject constructor(
...
 
-    override fun getBookmarksSortType() =
-        BookmarksSortType.values()[
+    override fun getBookmarksSortTypeForPlayer() =
+        BookmarksSortTypeForPlayer.values()[
+            getInt(
+                Settings.PREFERENCE_BOOKMARKS_SORT_TYPE_FOR_PLAYER,
+                BookmarksSortTypeForPlayer.DATE_ADDED_NEWEST_TO_OLDEST.ordinal
+            )
+        ]
```
mchowning, line 1478: I feel like persisting the enum's ordinal could make things difficult in the future, for example if we ever wanted to change the order of the items (since the order is currently tied to the order of the enums as well). It's also not super clear when debugging what a particular `Int` value means. What about storing it as a `String`, so it's not tied to the order? We could add something like a descriptive `persistenceKey: String` parameter to the enums and use that when we save it.

mchowning (approved):
Bookmarks is looking good. Nice work!

Just left a couple of comments on the code here, but no blockers, and the sorting on the podcast page worked great.

I did notice a couple of other possible issues, but they do not directly involve the sort options on the podcast page. They're just things I ran across testing this (probably because this is my first time testing bookmarks).

### 1. Tapping on the play icon on a bookmark doesn't start playback

When tapping on a bookmark while the player is paused, I get the toast message saying that it is playing the bookmark, but it does not actually start playback.

https://github.com/Automattic/pocket-casts-android/assets/4656348/474a712f-eb9c-44ae-bce0-d7fa0b50bcc0


</details>

### 2. UI froze for a second or so after adding a bookmark (sometimes, maybe)

After adding a bookmark from the player screen, I noticed that upon returning to the player the UI was frozen for about 1-2 seconds. This happened to me on 3 different times.  **I can't recreate this now though**, so maybe it was just something weird going on with my build (it was a debug build). Just mentioning it as something to possibly keep an eye out for.

### 3. Sorting bookmarks on player screen sometimes causes bookmarks to disappear

When sorting bookmarks on the player screen, I sometimes had some or all of the bookmarks disappear.

https://github.com/Automattic/pocket-casts-android/assets/4656348/7060a0f8-1a73-40e3-a4f9-737e7dd79b23

</details>

modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/SettingsImpl.kt
ashiagr: Good point, I updated it in 97593af

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/type/BookmarksSortType.kt
```diff
@@ -2,26 +2,48 @@ package au.com.shiftyjelly.pocketcasts.models.type
...
+private const val SORT_ORDER_DATE_ADDED_OLDEST_TO_NEWEST = "date_added_oldest_to_newest"
+private const val SORT_ORDER_EPISODE = "episode"
+private const val SORT_ORDER_TIMESTAMP = "timestamp"
+
+interface BookmarksSortType {
+    val labelId: Int
+    val analyticsValue: String
+}
+
+enum class BookmarksSortTypeForPlayer(
```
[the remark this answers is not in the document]
ashiagr: I feel like having separate enums can help when we want to have separate sort options on the two screens. Let me think more about it though.

ashiagr:
Thanks for the review, @mchowning!

> 1. Tapping on the play icon on a bookmark doesn't start playback
> 2. UI froze for a second or so after adding a bookmark (sometimes... maybe

I've noted them down on our project board.

> 3. Sorting bookmarks on player screen sometimes causes bookmarks to disappear

Good catch! The reason why it happened is that the bookmarks fragment remains in the view pager's fragments list. Since it isn't destroyed, whenever we update the sort order, bookmarks may be fetched for any previously selected episode. 

I now cancel all children of the `Job` in the `viewModelScope` context when `loadBookmarks` is called (everytime when episode changes) in 8f275ca. This is working well in my testing.

https://github.com/Automattic/pocket-casts-android/assets/1405144/cc33f01c-ab48-4f45-8d50-30317d8352f1

PR #1224, opened 2023-07-31
Bookmarks - Add parent container on episode page

ashiagr:
## Description

This is a preparatory PR to add the bookmarks fragment to the episode screen later. 

It
- Adds a parent container for the episode fragment with a view pager to contain child fragments
- Extracts toolbar from episode fragment to the parent container
- Adds tab bar to the toolbar

** I did not use a Compose Pager as the `EpisodeFragament` is not converted to a Compose yet and Compose Pager doesn't seem to support fragments properly. `ViewPager` has a `FragmentStateAdapter` that manages fragment states well.

## Testing Instructions
1. Install debug build
2. Open the episode screen
3. ✅  Notice there's a `Details` tab at the top
4. Disable the bookmarks feature in `Settings` -> `Beta features`
5. Re-open the episode screen
6. ✅  Notice that the `Details` tab is not present

#### Regression

Test that close, star, and share toolbar buttons work as before.

## Screenshots or Screencast 

Bookmarks Enabled | Bookmarks Disabled
----|------
Screenshot_20230731_180146 | Screenshot_20230731_180244

## Checklist
 
- [x] with different themes


mchowning (approved):
This looks good! 👍 

I only noticed that the top bar used to scroll off the screen, but now it does not. I'm not sure if that change was intended.

| Before | Now |
| --- | --- |
| <video src="https://github.com/Automattic/pocket-casts-android/assets/4656348/4f44201b-461d-42f6-88a0-943e25fc218c" /> | <video src="https://github.com/Automattic/pocket-casts-android/assets/4656348/66136607-3bff-4637-bb22-6325f1c33005"/> |

CookieyedCodes:
I kind of prefer star & share to be permanently fixed at the top of the screen truthfully 😅 but I get if it looks less seemless :)

ashiagr:
> I only noticed that the top bar used to scroll off the screen, but now it does not. I'm not sure if that change was intended.

I like how the tabs and toolbar options are easily accessible. I'll keep it this way until I hear any feedback against it.

PR #1227, opened 2023-07-31
Settings sync - Refactor Notification settings to be `UserSetting`s

mchowning:
## Description

This builds on https://github.com/Automattic/pocket-casts-android/pull/1213 to convert more settings over to use the `UserSetting` object in preparation for syncing our settings.

Some of the work on this PR was converting settings that used to be persisted directly from the XML layout to now being persisted via the relevant `UserSetting` object. This is needed to insure that the `UserSettings` object has the latest setting value at all times.

## Testing Instructions
Some of the settings (vibration ,ringtone) here are only used if the app is running on a Android versions older than O, so you'll need to use a build with an old version of Android to see them.

Also, none of these settings are synced (yet), so you don't need to worry about testing that.

For each setting in the "Notifications" section:
1. Fresh install the app
4. Change the setting to a new value
5. Force close the app
6. Verify that the setting has retained the new value
8. Verify that the app's behavior reflects the updated setting value (i.e., that skip forward actually skips forward based on the new value).

## Checklist


ashiagr (approved):
Works well in my testing. Good job 👍

PR #1229, opened 2023-08-01
What's new tracks

mchowning:
## Description
This adds tracks for the Whats new dialog

## Testing Instructions

For testing this, I would comment out [these lines](https://github.com/Automattic/pocket-casts-android/blob/d4b3363fbdb0793d9c5c9687768be00a19aa047b/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt#L740-L742) so the what's new dialog is always presented.

1. Run the app
2. When the What's New popup appears you should see a Tracked: whatsnew_displayed ["version": 9115, ...] in the logs
3. If you tap "Maybe later" or tap outside the dialog, you should see Tracked: whatsnew_dismissed ["version": 9115, ...] in the logs. If you tap "Enable it now" you should see Tracked: whatsnew_confirm_button_tapped ["version": 9115, ...] in the logs. 

## Checklist


ashiagr (approved):
Looks good 👍

Issue #1228, opened 2023-08-01
Theme selection UI can incorrectly indicate currently selected theme


mchowning:
### Description

When using a free account, selecting a Plus "theme" can result in subsequently selected themes not reflecting the current state of the UI—the UI will have the just-selected theme active, but the UI will indicate that a previously selected theme is active.

This is related to https://github.com/Automattic/pocket-casts-android/issues/68

### Step-by-step reproduction instructions

1. With a Free account,
2. Go to "Profile" → ⚙️  → "Appearance"
3. Select a "Free" theme
4. ✅ Observe that the UI updates to that theme and indicates that theme is selected
5. Select a "Plus" theme
6. Dismiss the Plus promotion screen
7. 💥 Observe that the UI indicates that the Plus theme is selected, but the UI displays the previously selected theme
8. Select a "Free" theme
9. 💥 💥  Observe that the UI updates to the selected theme, but indicates that the previously selected Free theme is currently selected.

### Screenshots or screen recording

https://github.com/Automattic/pocket-casts-android/assets/4656348/f1d1c875-e47a-4289-918f-5fdb3b7434de

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

Samsung A13, Android 13, Pocket Casts 7.44 pre-release build

neeldoshii:
Can I get this issue assigned to me?

mchowning:
All yours @neeldoshii ! 😄

P-Burk:
I'm not trying to step on anyone's toes, but I've got a PR up for a fix to this issue. I don't know how to bring this up without sounding petty, but I just wouldn't want for someone to put time/effort into fixing something only to find the issue has already been closed when they submit their own PR.

neeldoshii:
Sure @P-Burk, sure no problem it slipped of my mind. I didn't checked a PR has been merged.

mchowning:
Thanks for chiming in @P-Burk ! Sorry I overlooked that.

P-Burk:
No worries. Going forward, I can ask to get assigned to issues I'm working on if that's the standard practice/etiquette for this sort of thing (I'm still new to OSS contributions). I just usually don't do it since my fixes start out with me just being like "let's see if I can even figure out where this issue stems from" and I don't want to "claim" an issue while I'm literally just searching through the code base on where to even start trying to fix said issue.

mchowning:
[...]

PR #1230, opened 2023-08-01
Settings sync - Refactor appearance settings to be `UserSetting`s

mchowning:
## Description
This builds on #1227 by converting the Appearance settings to be `UserSetting`s.

## Testing Instructions

Note that there is a pre-existing bug [when switching between Plus and non-Plus themes on a Free account](https://github.com/Automattic/pocket-casts-android/issues/1228).

For each setting in the "Appearance" section:

1. Fresh install the app
4. Force close the app
6. Verify that the app's behavior reflects the updated setting value.

## Checklist


ashiagr (approved):
LGTM 👍

PR #1231, opened 2023-08-02
Bookmarks - Add to episode details page

ashiagr:
## Description

This adds bookmarks on the episode details page.

** Note that some of the components like header row,  no bookmarks view, plus upsell view and loading view have been extracted outside `BookmarksFragment` to reduce the class size and make it more readable. This + adding a few components to `showkase` (as requested in https://github.com/Automattic/pocket-casts-android/pull/1210#discussion_r1275927045) has increased PR size.

I also wanted to move `BookmarksFragment` to the bookmark package and rename file `BookmarkItem` to  `BookmarkRow` but for some reason doing that created new files instead of showing them as changes on GitHub. I will do that before merging this PR.

## Testing Instructions
1. Install debug build
2. Open a podcast episode with a few bookmarks added to it
3. ✅ Notice that bookmarks tab is shown 
4. Tap on the bookmarks tab
5. ✅ Notice that the bookmarks page is shown
6. Long press a bookmark row
7. ✅ Notice that the multi-select toolbar appears
8. Dismiss the dialog
9. ✅ Notice that the multi-select toolbar is not shown

#### Other tests

1. The screen reuses `MultiSelectEpisodeHelper`. Smoke test that the multi-select operations work correctly on the screen. (** edit action is not yet functional)

2. Disable the bookmarks feature and test that the bookmarks page cannot be accessed on the episode details screen (no tabs and no scroll to the page).

#### Notes

1. Bookmark sorting works but it reuses the preferences for the playing episode screen. Since they both apply to a single selected episode, maybe I should continue using it. I'll address it in a later PR.
2. Header row, bookmark row, no bookmarks view, and plus upsell view have been reused from the Player screen with color changes. Plus upsell view logo needs to be updated for lighter themes. I'll address it when I integrate Plus upsell screens.
3. Design - Known issues:
    - Episode Details screen layout changes (date and duration) not yet integrated 
    - Header row option needs to go next to the tabs 
    - Tabs design does not match Figma

## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/1405144/7ae3962a-5308-497e-82f3-98fc61f7634f

## Checklist
 
- [x] with different themes


modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/buttons/TimePlayButton.kt
```diff
@@ -29,36 +29,70 @@ import au.com.shiftyjelly.pocketcasts.localization.R
...
 fun TimePlayButton(
     timeSecs: Int,
     @StringRes contentDescriptionId: Int,
     onClick: () -> Unit,
     modifier: Modifier = Modifier,
-    textColor: Color = MaterialTheme.theme.colors.primaryText01,
-    backgroundColor: Color = MaterialTheme.colors.surface,
-    borderColor: Color = MaterialTheme.theme.colors.primaryText01,
-    buttonStyle: TimePlayButtonStyle = TimePlayButtonStyle.Outlined
+    colors: TimePlayButtonColors = TimePlayButtonColors.Default,
```
ashiagr, line 80: I combined all colors into `TimePlayButtonColors` just like the compose `ButtonColors` and included predefined colors: `Default` and `Player` just like compose `DefaultButtonColors`.

```
private class DefaultButtonColors(
    private val backgroundColor: Color,
    private val contentColor: Color,
    private val disabledBackgroundColor: Color,
    private val disabledContentColor: Color
) : ButtonColors
```

I did that to other bookmark views and rows for theming.

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/viewmodel/BookmarksViewModel.kt
```diff
@@ -72,19 +82,20 @@ class BookmarksViewModel
...
                                     bookmarks = bookmarks,
                                     isMultiSelecting = isMultiSelecting,
                                     isSelected = { selectedList.contains(it) },
                                     onRowClick = ::onRowClick,
+                                    sourceView = sourceView,
                                 )
                             }
                         }.stateIn(viewModelScope)
                     } ?: run { // This shouldn't happen in the ideal world
                         Timber.e("Episode not found.")
```
mchowning, line 97: Would it be worth making this a `LogBuffer.e` call so it shows up in support logs if it happens? I don't feel strongly, just wondering.

mchowning (approved):
Looking good! 👍

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/viewmodel/BookmarksViewModel.kt
ashiagr: Changed in [e16a4cd](https://github.com/Automattic/pocket-casts-android/pull/1231/commits/e16a4cde0e261a27ddb38c69ed5fdae59d4f2773)

PR #1232, opened 2023-08-02
Fix Notification permissions warning

### LINKED ISSUE #1126: Address warning about missing permission check when PlaybackManager calls notify
### Description

Android Studio is warning that we're posting a notification [here](https://github.com/Automattic/pocket-casts-android/blob/1bc7767bb1218b3cf50ce5cac091ff3fe756370c/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt#L1597) without checking that we have the (user-revokable) permission to do so.

N/A

image

_No response_

vkundapur:
## Description
Fixed Notification permissions warning by adding the Permission to AndroidManifest.xml and checking if the permissions are granted.

Fixes #1126

## Checklist


ashiagr (approved):
This is working well. Thanks for the contribution. 🙏

Issue #741, opened 2023-01-30
Storage choice is persisted as localized string


mchowning:
### Description

When selecting a storage choice in the "Storage & data use" section of the app's settings, we are saving (and retrieving) the localized label for the storage choice. This means that if you set a storage choice with the OS under one language, and then switch the OS to a different language, the rest of the app updates to the new language, but your storage choice is still displayed in the previous language.

This is occurring because we persist (and retrieve) the _localized_ label for the storage choice in Shared Preferences under the `Settings.PREFERENCE_STORAGE_CHOICE` key.

### Step-by-step reproduction instructions

1. Go to the Profile tab → ⚙️ → "Storage & data use"
2. Select an option under "Store podcasts on"
3. Switch the language the OS is using
4. 💥 Observe that the selected option under "Store podcasts on" is still in the previous language.

### Screenshots or screen recording

[device-2023-01-30-173349.webm](https://user-images.githubusercontent.com/4656348/215611405-22719488-9fdc-4b2a-a7df-ffda4e5794e1.webm)

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

Pixel 4 emulator running Android api 24, Pocket Casts 7.31 pre-release build.

mchowning:
This same issue was also occuring with the "Archive played episodes" and "Archive inactive episodes" settings, but the effect was more significant: users would actually lose their custom settings when switching languages. I will have a fix up for this shortly.

<details><summary>Video</summary>

https://github.com/Automattic/pocket-casts-android/assets/4656348/08ab5436-c809-46e6-b96d-a12ea1d9c824

</details>

PR #1234, opened 2023-08-03
Settings Sync - Refactor archive settings to be UserSettings

mchowning:
## Description
This updates the archive settings to be user settings.

In doing this, I found that we were persisting some of the archive settings as localized strings. Not surprisingly, this can cause [some issues](https://github.com/Automattic/pocket-casts-android/issues/741#issuecomment-1663067646) when a user changes the language on their phone. This PR fixes this by migrating the settings to a new key in shared preferences that is no longer tied to the app's language.

I also updated our support logs so that the auto archive settings are no longer displayed with a localized string.

### 1. General refactor testing
For each setting in the "Auto Archive" section:

1. Fresh install the app
4. Force close the app
6. Verify that the app's behavior reflects the updated setting value.

### 2. Language change bug fix
1. Fresh install the app
2. Go the Auto Archive settings
3. Change the "Archive played episodes" and "Archive inactive episodes" settings to a non-default value
4. Change the language of your phone to French
5. Return to the Pocket Casts app
6. Wait for the Pocket Casts app to update to the new language
7. Confirm that the non-default settings are still visible on the Auto Archive settings screen (without this PR fix, the previously set settings [would disappear](https://github.com/Automattic/pocket-casts-android/issues/741#issuecomment-1663067646)).

### 3. Migration of archive settings
1. Fresh install a build from `main`
2. Change the "Archive played episodes" and "Archive inactive episodes" settings to a non-default value
4. Verify that your updates to those settings are retained

### 4. Support logs no longer use localized strings for archive settings
1. Change your phone's language to French
5. View the app logs (Profile → "Help & feedback" → logs icon in the toolbar)
6. Verify that the values for "Auto archive played episodes after" and "Auto archive inactive episodes after" are not in French.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+Unknown
+-----
+*   Bug Fixes:
+    *   Fixed auto archive settings getting lost when switching languages
+        ([#1234](https://github.com/Automattic/pocket-casts-android/pull/1234))
+
 7.45
 -----
 *   Bug Fixes:
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/support/Support.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/support/Support.kt
@@ -313,11 +313,10 @@ class Support @Inject constructor(
                 Timber.e(e)
             }
 
-            val afterPlaying = context.resources.getStringArray(R.array.settings_auto_archive_played_values)
             val inactive = context.resources.getStringArray(R.array.settings_auto_archive_inactive_values)
             output.append(eol)
             output.append("Auto archive settings").append(eol)
-            output.append("Auto archive played episodes after: " + afterPlaying[settings.getAutoArchiveAfterPlaying().toIndex()]).append(eol)
+            output.append("Auto archive played episodes after: " + settings.autoArchiveAfterPlaying.flow.value.analyticsValue).append(eol)
             output.append("Auto archive inactive episodes after: " + inactive[settings.getAutoArchiveInactive().toIndex()]).append(eol)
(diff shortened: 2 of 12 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+Unknown
```
mchowning, line 1: I know this won't be in 7.45, so putting this here in order to create a merge conflict when this gets merged to `main`. At that point, I'll add this release note under the correct release heading.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/support/Support.kt
```diff
@@ -313,13 +313,11 @@ class Support @Inject constructor(
                 Timber.e(e)
             }
 
-            val afterPlaying = context.resources.getStringArray(R.array.settings_auto_archive_played_values)
-            val inactive = context.resources.getStringArray(R.array.settings_auto_archive_inactive_values)
             output.append(eol)
             output.append("Auto archive settings").append(eol)
-            output.append("Auto archive played episodes after: " + afterPlaying[settings.getAutoArchiveAfterPlaying().toIndex()]).append(eol)
-            output.append("Auto archive inactive episodes after: " + inactive[settings.getAutoArchiveInactive().toIndex()]).append(eol)
```
mchowning, line 321: I changed this to no longer use localized strings here since localized strings just seem like they would make things more difficult for support. I'm using the analytics value because that seemed good enough, but if you think we'd be better off creating some even-more-readable support-specific strings, let me know.

ashiagr:
All tests pass. 

> This PR fixes this by migrating the settings to a new key in shared preferences that is no longer tied to the app's language.

Thank you for fixing it 🙇‍♀️ I've verified that it works correctly now.

ashiagr (approved):

PR #1235, opened 2023-08-03
Fix/issue 1192 toolbar disappear on add file screen

### LINKED ISSUE #1192: Top navigation bar and podcast art disappears on orientation change
### Description

When you are on the final screen to add a personal file, if you change the screen orientation, the top navigation bar and the podcast artwork disappear. You are unable to click the back button and/or save button, leaving the user stuck on the screen unless using the phone's navigation buttons/gestures.

1. Go to the user account page.
2. Click on the Files button.
3. Click the Add button (Floating Action Button).
4. Select a file to upload.
5. You will automatically be taken to the Add File screen.
6. Change the orientation of the screen/phone.

https://github.com/Automattic/pocket-casts-android/assets/68164093/b7767dd7-6f25-43f3-9f08-e5e71881265f

Google Pixel 6 on API 33. Pocket Casts version 7.43.

P-Burk:
## Description

This PR is to fix a bug on the *Add File* Activity/Screen where when a user has a file selected for upload and changes the screen orientation the Toolbar does not load and the file art disappears. When this occurs, the save and back navigation buttons within the Toolbar also do not work as they are not loaded.

Fixes #1192

## Testing Instructions

1. Go to the user account page.
2. Click on the Files button.
3. Click the Add button (Floating Action Button).
4. Select a file to upload.
5. You will automatically be taken to the Add File screen.
6. Change the orientation of the screen/phone and the Toolbar + file art should reappear when the Activity is recreated.

#### Before:
Check the original issue for a Before screen cap of the issue.
#### After:
https://github.com/Automattic/pocket-casts-android/assets/68164093/ea22d2e1-d867-4eae-baad-ffd8ac3e77b0

## Checklist
	- No strings added.
	- No Jetpack compose components added or changed.
 
- [x] with different themes
	- This was a bug fix to bring back a missing UI element

#### Notes:
`savedInstanceState.getParcelable(key: String)` is deprecated. I did have a solution to fix this, which involves using `savedInstanceState.getParcelable(key: String, clazz: Class<T>)`, but it requires Android version 33 and the minimum SDK version is 23. So for the time being, I suppressed the deprecation warning because I saw the same suppression used elsewhere in the `AddFileActivity.kt` file (where my changes take place).
image

### DIFF
--- a/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/cloud/AddFileActivity.kt
+++ b/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/cloud/AddFileActivity.kt
@@ -291,6 +291,12 @@ class AddFileActivity :
                 updateColorItems()
             }
         }
+
+        @Suppress("DEPRECATION")
+        if (savedInstanceState != null) {
+            dataUri = savedInstanceState.getParcelable(STATE_DATAURI)
+            setupForNewFile(dataUri)
+        }
     }
 
     override fun onSaveInstanceState(outState: Bundle) {
--- a/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/cloud/AddFileActivity.kt
+++ b/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/cloud/AddFileActivity.kt
@@ -83,7 +83,7 @@ private const val ACTION_PICK_FILE = 2
 private const val EXTRA_EXISTING_EPISODE_UUID = "fileUUID"
 private const val EXTRA_FILE_CHOOSER = "filechooser"
 private const val STATE_LAUNCHED_FILE_CHOOSER = "LAUNCHED_FILE_CHOOSER"
-private const val STATE_DATAURI = "null"
+private const val STATE_DATAURI = "DATAURI"
 
 @AndroidEntryPoint
 class AddFileActivity :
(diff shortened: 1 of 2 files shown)


geekygecko (commented):
Thanks for fixing this issue, it worked well for me. Is there a reason the constant value is "null" and not something like "DATAURI"?

P-Burk:
> Thanks for fixing this issue, it worked well for me. Is there a reason the constant value is "null" and not something like "DATAURI"?

To be 100% honest, no reason other than I'm new to this (OSS contributions) and Kotlin. I'm not sure what the "default" values should be for constants.

modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/cloud/AddFileActivity.kt
```diff
@@ -83,6 +83,7 @@ private const val ACTION_PICK_FILE = 2
 private const val EXTRA_EXISTING_EPISODE_UUID = "fileUUID"
 private const val EXTRA_FILE_CHOOSER = "filechooser"
 private const val STATE_LAUNCHED_FILE_CHOOSER = "LAUNCHED_FILE_CHOOSER"
+private const val STATE_DATAURI = "null"
```
geekygecko, line 86:
```suggestion
private const val STATE_DATAURI = "DATAURI"
```
Can we change it to this? It just needs to be unique from the other state constants and null looks like we forgot to add a value.

geekygecko (approved):

PR #1237, opened 2023-08-04
Hide ratings if the podcast doesn't have any.

geekygecko:
## Description

When a podcast doesn't have any ratings it looks like they have zero stars. This change hides the star if there are no ratings.

## Testing Instructions
1. Tap the Discover tab
2. Tap search
3. Paste in the staging share URL https://pcast.pocketcasts.net/51phk3pk
4. ✅ Verify no stars are visible
5. Go back to search and find The Daily
6. ✅ Verify the stars are visible

## Screenshots

| Before | After |
| --- | --- |
| Screenshot_20230804_140129 | Screenshot_20230804_135933 |


ashiagr (approved):
Looks good 👍 
Thanks.

PR #1240, opened 2023-08-04
Make auto add to up next settings a `UserSetting`

mchowning:
## Description
This updates the auto add to up next limit and behavior to each be a `UserSetting`

## Testing Instructions

For the Auto Add to Up Next setting, test the "Aut add limit" and "If limit reached" setting by:

1. Fresh install the app
4. Force close the app
6. Verify that the app's behavior reflects the updated setting value. Note that you can force new episodes to be received from the "Developer Settings" section by tapping "Delete first episodes" and then tapping "force refresh". When testing whether the up next queue limit be aware that "delete first episodes" will remove items from your queue (i.e., if you set your queue to the max, and then tap delete first episodes, your queue may no longer be at the max).

## Checklist


ashiagr (approved):
All tests pass. 

Thanks for the extra test details that you include to make testing and reviewing easier.

PR #1241, opened 2023-08-04
Make auto download a `UserSetting`

mchowning:
## Description
This updates the auto download settings to each be a `UserSetting`

## Testing Instructions
1. Fresh install the app
1. Verify that the setting has the same default value that it had before
1. Change the setting to a new value
1. Force close the app
1. Verify that the setting has retained the new value
1. Verify that the app's behavior reflects the updated setting value. Note that you can force new episodes to be received from the "Developer Settings" section by tapping "Delete first episodes" and then tapping "force refresh".

## Checklist


CookieyedCodes:
I personally wouldn't have auto download  upnext syncing across devices without an first number of episodes limit in place truthfully
(ie like the the apple watch it only downloads the first NUM which ignores auto download/auto add to upnext/filters auto download if that's possible)

mchowning:
> I personally wouldn't have auto download upnext syncing across devices...

Right, we're not planning to sync auto download, this PR is just converting that setting to the common structure.

CookieyedCodes:
> > I personally wouldn't have auto download upnext syncing across devices...
> 
> Right, we're not planning to sync auto download, this PR is just converting that setting to the common structure.

Cool cool, happy setting sync is finally happening tho ❤️❤️, hopefully playback FX sync is being considered 😉

PR #1242, opened 2023-08-06
Bookmark initial sync

geekygecko:
## Description

This change introduces Protocol Buffers to the user sync to improve the performance.  
The script `scripts/update_protobuf.sh` generates the request and response object in a new module `modules/services/protobuf/`. Sorry this will make it hard to review this PR. 

## Testing Instructions

1. 

## Checklist
 
- [ ] with different themes


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/PodcastSyncProcess.kt
```diff
@@ -132,6 +132,7 @@ class PodcastSyncProcess(
                 cacheStats()
                     .andThen(downloadAndImportHomeFolder())
                     .andThen(downloadAndImportFilters())
+                    .andThen(rxCompletable { downloadAndImportBookmarks() })
```
geekygecko, line 135: I would like to push towards sync using Coroutines rather than Rx, this is why I have used them where I can.

ashiagr (approved):
Works well as described. Nice to see Protocol Buffers in action 🥇 

> There are many generated files but I have deleted all those unrelated to this bookmark initial sync to make it easier to review. 

Thank you for this ^

PR #1244, opened 2023-08-07
Fixes: #1217 Notification setting UI update bug fixed

### LINKED ISSUE #1217: Notification setting does not update properly when toggled on
### Description

When toggling on the global "Notify me" setting it initially says that no podcasts are chosen. If you leave the screen and return though, it now says all podcasts are selected. It looks like the UI for the selected podcasts isn't getting updated appropriately when you first toggle "Notify me" on.

1. Fresh install
2. Subscribe to at least 2 podcasts
3. Go to "Profile" tab → ⚙️ → Notifications
4. Tap to toggle "Notify me" on
5. Observe that under "Choose podcasts" it says "No podcasts selected"
6. Tap back
7. Open "Notifications" again
8. Observe that under "Choose podcasts" it says "All podcasts"

Note that there are some times when following different flows that toggling on "Notify me" correctly shows "All podcasts" but I think this is still showing stale state, it just happens to be accurate stale state.

https://github.com/Automattic/pocket-casts-android/assets/4656348/ec14886c-0c49-4268-b71f-a469b0715028

Samsung A13 running Android 13, Pocket Casts 7.44-rc-1.

Yashraj254:
## Description

 I have changed the use of findSubscribed() method which was causing this unexpected behaviour (because of not retrieving the latest updated data), to findSubscribedFlow() method to get the latest updated data from database which fixes this issue.

Fixes #1217 

## Testing Instructions

Follow the procedure as described in issue #1217 

## Checklist

### DIFF
--- a/modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/NotificationsSettingsFragment.kt
+++ b/modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/NotificationsSettingsFragment.kt
@@ -33,8 +33,6 @@ import com.afollestad.materialdialogs.list.MultiChoiceListener
 import com.afollestad.materialdialogs.list.listItemsMultiChoice
 import com.afollestad.materialdialogs.list.updateListItemsMultiChoice
 import dagger.hilt.android.AndroidEntryPoint
-import io.reactivex.android.schedulers.AndroidSchedulers
-import io.reactivex.schedulers.Schedulers
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.DelicateCoroutinesApi
 import kotlinx.coroutines.Dispatchers
@@ -170,9 +168,6 @@ class NotificationsSettingsFragment :
                         mapOf("enabled" to checked)
                     )
 
-                    podcastManager.updateAllShowNotificationsRx(checked)
-                        .observeOn(AndroidSchedulers.mainThread())
-                        .subscribeOn(Schedulers.io()).onErrorComplete().subscribe()
                     if (checked) {
                         settings.setNotificationLastSeenToNow()
                     }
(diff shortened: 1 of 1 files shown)


Yashraj254:
@geekygecko hey, can you review my PR..?

geekygecko:
> hey, can you review my PR..?

Thanks for your contribution! I might ask @mchowning if he wouldn't mind looking at this as he has recently worked on this settings page.

mchowning (changes requested):
Thanks for the contribution @Yashraj254 ! Switching this over to collecting the flow is a great improvement! 😄  I think we just need to address a couple of issues, and this will be good to merge.

#### 1. Collecting the flow multiple times

We were calling this `changePodcastsSummary` method from multiple places before because we were just retrieving the latest data each time it was called before. Now that we're collecting the flow we don't need to keep doing that. In fact, we need to stop doing that because it is starting multiple collectors for the flow (for example, this adds an additional collector each time the user enables/disables the notify me option). As a result, we end up processing each item in the flow multiple times.

How about updating the usages of this method so it is only called one time when the fragment is created? Maybe at the end of `onCreatePreferences` would make sense?

#### 2. GlobalScope problems

We shouldn't have been using GlobalScope, but its usage before wasn't too problematic until now because we were just quickly processing the latest data. Now that we're listening for flow changes, launching this with GlobalScope means that it will continue collecting these changes even after the user changes screens (i.e., it will get fired if the user subscribes to a new podcast after backing out of the notification settings screen).

How about changing `GlobalScope.launch(Dispatchers.Default)` on line 380 to `lifecycleScope.launch(Dispatchers.Default)`?

Yashraj254:
@mchowning thanks for pointing out those issues, here i have made the necessary changes..

mchowning (approved):
Thanks for the updates! This looks good to merge.

I did push a small commit (https://github.com/Automattic/pocket-casts-android/pull/1244/commits/9d69a00b86549fcc6b838787aceff8ae6e0f7cc7) to remove an annotation we don't need anymore.

Thanks again for contributing to Pocket Casts!

Issue #1248, opened 2023-08-07
Automatic delete settings for files do not work when an episode is marked as played


mchowning:
### Description

When the settings for automatically deleting a local file after playing or automatically deleting the cloud file after playing ("Profile" tab → "Files" → three-dot-overflow-menu → "Files Settings") are turned on. The files will be deleted if they are played to the end, but they will not be deleted if the user taps "Mark as Played" while the file is playing.

### Step-by-step reproduction instructions

1. Turn on the settings for deleting local and cloud files in "Profile" tab → "Files" → three-dot-overflow-menu → "Files Settings"
2. Load a file and upload it to the cloud (requires a Plus account)
3. Start playing a file
4. Mark the file as played without playing through the end of the file
5. Observe that the file is not deleted either locally or from the cloud
6. Play the file again, but use the skip forward button to skip through the end of the file
7. Observe that tthe file is deleted locally and removed from the cloud

### Screenshots or screen recording

https://github.com/Automattic/pocket-casts-android/assets/4656348/fe3cef0a-1e17-4168-a1f1-a69e921f4d56

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

Samsung A13, Android 13, 7.45-pre-release build

PR #1249, opened 2023-08-07
Settings Sync - Make file settings into UserSettings

mchowning:
## Description
This converts the file settings to be UserSettings.

### 1. Check defaults
1. Fresh install of build from this PR
2. Log into a Plus account
3. Go to "Profile" tab → "Files" → three-dot-overflow-menu → "Files Settings"
4. Verify that the only setting that defaults to being on is "Only on WiFi". All other settings should be turned off.

### 2. Settings are migrated
1. Fresh install of app built from `main`
2. Log into a Plus account
3. Go to "Profile" tab → "Files" → three-dot-overflow-menu → "Files Settings"
4. Update all the settings to not be their default (so you want them all turned on except "Only on WiFi")
5. Install a build of the app from this PR
6. Verify that all the settings have their updated values

### 3. Settings work as they did before
Verify that toggling each setting still works.
- The automatic deletion settings only appear to work when skipping to the end of an episode, they [do not work when marking an episode as played](https://github.com/Automattic/pocket-casts-android/issues/1248).
- The "Only on WiFi" setting is really only checking for an unmetered network connection, so this can be tested by switching your wifi network to be metered/unmetered.

## Checklist

### DIFF
--- a/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/cloud/CloudSettingsFragment.kt
+++ b/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/cloud/CloudSettingsFragment.kt
@@ -103,7 +103,7 @@ class CloudSettingsFragment : BaseFragment() {
             }
         }
         with(binding.swtAutoUploadToCloud) {
-            isChecked = settings.getCloudAutoUpload()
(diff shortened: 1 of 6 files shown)


ashiagr (approved):
LGTM. All tests pass.

PR #1250, opened 2023-08-08
Settings Sync - Make refresh and stream warning into UserSettings

mchowning:
## Description
This makes the stream warning and background refresh settings into UserSettings.

I also moved the MockUserSetting class out of the separate test classes and instead into the UserSetting class itself. Ideally, we would only include this as a test artifact, and I could do that, but I don't see any way to accomplish that without quite a bit of boilerplate, and I just didn't think it was worth it to avoid bundling this small mock class in the app. Let me know if you disagree or have a better idea how to handle this mock class.

## Testing Instructions

1. Fresh install of the app
2. Go to "Profile" → ⚙️ → "Storage & data use"
3. Verify that "Background refresh" is defaulted to on (on the watch this defaults to off) and "Warn before using data" is defaulted to off
4. Toggle the warning before using data to be on
5. Switch your network to be metered
6. Begin streaming an episode
7. Observe the stream warning is presented
8. Switch your network to be unmetered
9. Begin streaming an episode and observe the stream warning is NOT presented
10. Monitor the app logs to observe background refresh is occurring. I triggered this to occur more often [by updating the refresh to occur every 20 seconds](https://gist.github.com/mchowning/eb2e5f8af81e0efff6e869215be5d323). Also filtering the logs for "BgTask" helps.
11. Toggle the background refresh to be off
12. Monitor the app logs and observe background refresh is NOT occurring

## Checklist

### DIFF
--- a/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/AutoArchiveTest.kt
+++ b/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/AutoArchiveTest.kt
@@ -1,7 +1,6 @@
 package au.com.shiftyjelly.pocketcasts.models.db
 
 import android.content.Context
-import android.content.SharedPreferences
 import androidx.room.Room
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import androidx.test.platform.app.InstrumentationRegistry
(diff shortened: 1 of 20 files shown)


ashiagr (approved):
Works well in my testing. Thanks for providing clear testing instructions.

> I also moved the MockUserSetting class out of the separate test classes and instead into the UserSetting class itself. Ideally, we would only include this as a test artifact, and I could do that, but I don't see any way to accomplish that without quite a bit of boilerplate, and I just didn't think it was worth it to avoid bundling this small mock class in the app. Let me know if you disagree or have a better idea how to handle this mock class.

Makes sense, looks neat 👍

PR #1239, opened 2023-08-04
Bookmarks - Display for user episodes

ashiagr:
## Description

This displays bookmarks for user episodes (files).

There's no Figma design, but here's the iOS PR: https://github.com/Automattic/pocket-casts-ios/pull/1018

## Testing Instructions
- Open a user episode (files)
- Play it and add a few bookmarks
- Return to the Files screen
- ✅ Notice that the file has a bookmark icon
- Tap on the file row
- ✅ Notice that the file details bottom sheet has a bookmark menu
- Tap on the bookmarks menu
- ✅ Notice that the bookmarks dialog is shown with bookmarks for the selected file

## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/1405144/8084a6a6-7017-4593-b83f-654b589037ed

#### Known Issues:
- Theming doesn't work well
- On deleting all bookmarks for a file, the bookmark icon is not removed for it until the view is reloaded
- Reuses sort order previously set for episodes 

## Checklist
- [ ] If this is a user-facing change, I have added an entry in CHANGELOG.md
- [x] Ensure the linter passes (`./gradlew spotlessApply` to automatically apply formatting/linting)
- [ ] I have considered whether it makes sense to add tests for my changes
- [x] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml`
- [ ] Any jetpack compose components I added or changed are covered by compose previews
 
#### I have tested any UI changes...

- [ ] with different themes
- [ ] with a landscape orientation
- [ ] with the device set to have a large display and font size
- [ ] for accessibility with TalkBack

PR #1254, opened 2023-08-08
Bookmark - Play episode if not playing

ashiagr:
## Description

This 
1. Fixes an issue with adding a bookmark in place of an already deleted and not synced bookmark
2. Plays selected bookmark episode if not already playing

#### Test.1 Already deleted but not synced bookmark

1. Play an episode and pause it so that playback position doesn't change
2. Open the full-screen player and add a bookmark
3. Go to the `Bookmarks` tab
4. Long press the newly created bookmark and delete
5. Go back to the `Player` tab
6. Re-add a bookmark at the same position
7. ✅  Notice that the "Add bookmark" screen is shown and not the "Change title" screen as the bookmark is in "to be synced - deleted" state.

https://github.com/Automattic/pocket-casts-android/assets/1405144/9393c246-5dc5-493a-989e-6077ad853847

#### Test.2 Play the selected bookmark episode if not already playing

1. Play an episode and pause it 
2. Open the full-screen player and add a bookmark
3. Go to the `Bookmarks` tab
4. Tap on the Play button for the newly created bookmark
5. Notice that the bookmark episode starts playing from the selected bookmark position

`Episode Details` and `Files` screens
** Make sure that the behavior is the same for the `Episode Details` and `Files` (https://github.com/Automattic/pocket-casts-android/pull/1239) screens

`Podcast Details` screen
** For a podcast with multiple episodes, make sure that the correct episode is selected and played at the bookmark position

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastViewModel.kt
+++ b/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastViewModel.kt
@@ -380,6 +380,19 @@ class PodcastViewModel
         settings.setBookmarksSortType(order)
     }
 
+    fun play(bookmark: Bookmark) {
+        val time = bookmark.timeSecs
+        val bookmarkEpisode = (uiState.value as? UiState.Loaded)?.episodes?.firstOrNull { it.uuid == bookmark.episodeUuid }
+        bookmarkEpisode?.let {
+            val shouldPlayEpisode = !playbackManager.isPlaying() ||
+                playbackManager.getCurrentEpisode()?.uuid != bookmarkEpisode.uuid
+            if (shouldPlayEpisode) {
+                playbackManager.playNow(it, sourceView = SourceView.PODCAST_LIST)
+            }
+        }
+        playbackManager.seekToTimeMs(time * 1000)
+    }
+
     fun multiSelectSelectNone() {
         val uiState = uiState.value as? UiState.Loaded ?: return
         when (uiState.showTab) {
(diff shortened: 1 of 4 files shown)


modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastViewModel.kt
```diff
@@ -380,6 +380,19 @@ class PodcastViewModel
...
+        val time = bookmark.timeSecs
+        val bookmarkEpisode = (uiState.value as? UiState.Loaded)?.episodes?.firstOrNull { it.uuid == bookmark.episodeUuid }
+        bookmarkEpisode?.let {
+            val shouldPlayEpisode = !playbackManager.isPlaying() ||
+                playbackManager.getCurrentEpisode()?.uuid != bookmarkEpisode.uuid
+            if (shouldPlayEpisode) {
+                playbackManager.playNow(it, sourceView = SourceView.PODCAST_LIST)
+            }
+        }
+        playbackManager.seekToTimeMs(time * 1000)
```
mchowning, line 393: Nitpick: I would inline this `time` variable since it's not used anywhere else and `seekTimeToMs(bookmark.timeSecs * 1000)` seems nice since it makes the seconds to ms conversion more explicitly. Certainly not a big deal though.

ashiagr: Nice pick. 🙂

Updated in https://github.com/Automattic/pocket-casts-android/pull/1254/commits/e3031f750d3309994ba2e18ea60a4398778c7ef2

PR #1258, opened 2023-08-09
Settings Sync - Make global playback effects into UserSettings

mchowning:
## Description
This refactors the global playback efffets settings to use the UserSettings structure. I really liked how it was easy to just create a UserSetting for the `PlaybackEffects` that delegated to UserSettings for the speed, trim silence, and volume boost settings.

### 1. Settings from previous versions get picked up
1. Install a build from the `feature/sync-settings-refactor` branch
2. Start playing an episode and update all the playback effects (speed, trim silence, volume boost)
4. Verfiy that all the updated settings are carried over

### 2. Settings have proper defaults
1. Do a fresh install from this branch
2. Check that the speed value is defaulted to 1.0
3. Check that the trim silence setting is defaulted to off
4. Check that the volume boost setting is defaulted to off

## Checklist


ashiagr (approved):
Awesome work!

PR #1259, opened 2023-08-09
Settings Sync - Make newsletter settings a UserSetting

mchowning:
## Description
This makes the newsletter preference into a `UserSetting`. I also did a slight tweak to how we indicate whether a setting needs to be synced, but the actual behavior is unchanged.

## Testing Instructions
1. Veriify the newsletter setting value for a PC account
2. Log into that account on a clean install of the app
3. Verify that the account screen reflects the proper newsletter setting value
4. Change the newsletter setting value
5. Sync the app
6. Uninstall and reinstall the app
7. Verify that the new newsletter setting is picked up

## Checklist

### DIFF
--- a/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/AccountDetailsViewModel.kt
+++ b/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/AccountDetailsViewModel.kt
@@ -3,8 +3,10 @@ package au.com.shiftyjelly.pocketcasts.profile
 import androidx.lifecycle.LiveData
 import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.ViewModel
+import androidx.lifecycle.asLiveData
 import androidx.lifecycle.map
 import androidx.lifecycle.toLiveData
+import androidx.lifecycle.viewModelScope
 import au.com.shiftyjelly.pocketcasts.account.viewmodel.NewsletterSource
 import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsEvent
 import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsTrackerWrapper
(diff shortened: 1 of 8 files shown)


modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/AccountDetailsViewModel.kt
```diff
@@ -72,10 +73,9 @@ class AccountDetailsViewModel
     val accountStartDate: LiveData<Date> = MutableLiveData<Date>().apply { value = Date(statsManager.statsStartTime) }
 
     val marketingOptInState: LiveData<Boolean> =
-        settings.marketingOptObservable
-            .distinctUntilChanged()
```
mchowning, line 76: Because this is a state flow now, we don't need `distinctUntilChanged` anymore.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/user/UserManager.kt
```diff
@@ -114,8 +114,8 @@ class UserManagerImpl @Inject constructor(
                     userEpisodeManager.removeCloudStatusFromFiles(playbackManager)
                 }
 
-                settings.setMarketingOptIn(false)
-                settings.setMarketingOptInNeedsSync(false)
+                settings.marketingOptIn.set(false)
+                settings.marketingOptIn.needsSync = false
```
mchowning, line 118: We can't just pass a false value for needsSync when we set the value here (i.e., `marketingOptIn.set(false, needsSync = false)` because that parameter only has effect when it is `true`. We have to explicitly set the needsSync value on its own. This certainly isn't ideal, and if this is something we end up wanting to do more often, I should probably rethink how this works, but my hope is that this isn't something we'll be doing apart from this one instance, so I left it as-is since I think the current handling makes more sense for the more common scenario where we don't want a call to set to unintentionally overwrite a previous request to sync a value.

modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/AccountDetailsViewModel.kt
ashiagr: Nice! Thanks for pointing it out. 🙇‍♀️

ashiagr (approved):
All tests pass. Code changes look good too.

PR #1264, opened 2023-08-09
Improve handling of notification enabled setting

mchowning:
## Description
This PR addresses an issue where we were not using the notifications enabled setting to determine the state of the notifications enabled toggle. Instead we were checking whether any podcasts had notifications enabled. This PR updates this and makes the behavior consistent with iOS.

This issue exists on `main` and is not a result of the recent Settings refactorings.

This change feels a bit tricky, so let me know if you can think of any undesireable behavior that these changes might introduce.

I am targeting the settings sync refactor branch with this change because it is easier to just include this with the other settings changes, and I anticipate that I'll be merging this branch to `main` pretty soon. 

## Testing Instructions

1. Go to the notification settings
2. Change the notification setting
3. Exit the screen and return to the screen
4. Note that the new setting value persists
5. Update the setting to be turned on, but with no podcasts selected
6. Exit the screen and return to the screen
7. Note that notifications are still enabled (previously, notifications would have been turned off because we were just checking if any podcasts had notifications enabled).
8. Turn off notifications
9. Tap on the podcasts tab (do not back out of the notifications screen), and open a subscribed podcast
10. Tap the notification bell to turn on notifications
11. Tap on the profile tab to return to the notifications screen
12. Note that notifications are now on (previously, notifications would have been turned on, but the UI would have shown that notifications were off because the UI was out of date).

## Checklist

### DIFF
--- a/modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/NotificationsSettingsFragment.kt
+++ b/modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/NotificationsSettingsFragment.kt
@@ -114,8 +114,6 @@ class NotificationsSettingsFragment :
             }
         }
 
-        updateNotificationsEnabled()
-
         hidePlaybackNotificationsPreference?.setOnPreferenceChangeListener { _, newValue ->
             val newBool = (newValue as? Boolean) ?: throw IllegalStateException("Invalid value for hide notification on pause preference: $newValue")
             settings.hideNotificationOnPause.set(newBool)
@@ -155,48 +153,6 @@ class NotificationsSettingsFragment :
             )
 
             true
-        }
-    }
-
-    private fun updateNotificationsEnabled() {
-        launch(Dispatchers.Default) {
-            val enabled = settings.notifyRefreshPodcast.flow.value
-
-            launch(Dispatchers.Main) {
-                enabledPreference?.isChecked = enabled
-                enabledPreferences(enabled)
-
-                enabledPreference?.setOnPreferenceChangeListener { _, newValue ->
-                    val checked = newValue as Boolean
-                    settings.notifyRefreshPodcast.set(checked)
-
-                    analyticsTracker.track(
-                        AnalyticsEvent.SETTINGS_NOTIFICATIONS_NEW_EPISODES_TOGGLED,
-                        mapOf("enabled" to checked)
-                    )
-
-                    podcastManager.updateAllShowNotificationsRx(checked)
-                        .observeOn(AndroidSchedulers.mainThread())
-                        .subscribeOn(Schedulers.io()).onErrorComplete().subscribe()
-                    if (checked) {
-                        settings.setNotificationLastSeenToNow()
-                    }
-                    changePodcastsSummary()
-                    enabledPreferences(checked)
-
-                    true
-                }
-
-                notificationPodcasts?.setOnPreferenceClickListener {
-                    openSelectPodcasts()
-                    true
-                }
-
-                changePodcastsSummary()
-                changeVibrateSummary()
-                changeNotificationSoundSummary()
-                setupActions()
-            }
         }
     }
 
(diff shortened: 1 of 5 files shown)


modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/NotificationsSettingsFragment.kt
```diff
@@ -445,6 +402,48 @@ class NotificationsSettingsFragment :
...
+        launch(Dispatchers.Default) {
+            val enabled = settings.notifyRefreshPodcast.flow.value
+
+            launch(Dispatchers.Main) {
+                enabledPreference?.isChecked = enabled
+                enabledPreferences(enabled)
+
+                enabledPreference?.setOnPreferenceChangeListener { _, newValue ->
+                    val checked = newValue as Boolean
+                    settings.notifyRefreshPodcast.set(checked)
```
mchowning, line 415: Apart from moving this entire method to be run in `onResume` instead of only when the fragment was created, the only change to the logic was to add this `settings.notifyRefreshPodcast.set(checked)` line so that we now persist the enabled setting.

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/NotificationsSettingsFragment.kt
```diff
@@ -445,6 +402,48 @@ class NotificationsSettingsFragment :
         }
     }
 
+    private fun setupEnabledNotifications() {
+        launch(Dispatchers.Default) {
+            val enabled = settings.notifyRefreshPodcast.flow.value
```
geekygecko, line 407: Do you think it's worth having a `value` method that just calls `flow.value` as we seem to call it lots? Feel free to ignore this.

geekygecko (approved):
There is something slightly strange happening where you turn on "Notify me", it says "No podcasts selected", you exit the screen, come back, and now it says "All podcasts".

https://github.com/Automattic/pocket-casts-android/assets/308331/f9f30e6d-5d81-4869-a3ea-1a13c83c0548

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/NotificationsSettingsFragment.kt
mchowning: My initial thought was to intentionally not add something like that because I wanted to provide a push toward reacting to the flow instead of just fetching the latest value. There are enough places where we just get the latest value though that you're probably right that it makes sense to just make that more convenient. I'll make this change in a separate PR. Thanks for bringing this up.

mchowning:
> There is something slightly strange happening where you turn on "Notify me", it says "No podcasts selected", you exit the screen, come back, and now it says "All podcasts".

Looks like it was a race condition, but I'm pretty sure you just brought this up to bait me into removing some RxJava 🙂 . Should be fixed by [a542615](https://github.com/Automattic/pocket-casts-android/pull/1264/commits/a542615e185e926b4e9f72bb8a0195a9c496d86a). Since this isn't a trivial change, I'll hold off on merging until you give this another look.

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/NotificationsSettingsFragment.kt
```diff
@@ -445,6 +399,49 @@ class NotificationsSettingsFragment :
...
+                enabledPreference?.setOnPreferenceChangeListener { _, newValue ->
+                    val checked = newValue as Boolean
+                    settings.notifyRefreshPodcast.set(checked)
+
+                    analyticsTracker.track(
+                        AnalyticsEvent.SETTINGS_NOTIFICATIONS_NEW_EPISODES_TOGGLED,
+                        mapOf("enabled" to checked)
+                    )
+
+                    lifecycleScope.launch {
```
mchowning, line 419: Interestingly, if I used the Fragment's CouroutineScope directly here by calling `launch` instead of `lifecycleScope.launch` the coroutine wasn't getting run. I'm not sure why that is. 😕 If you have any insight, let me know.

geekygecko:
Thanks for those changes. It worked well for me.

PR #1265, opened 2023-08-10
Remove no longer used settings from Settings class

mchowning:
## Description
This PR is removing mthods from the Settings class where either the setting is never getting read/used or the setting is never getting set (meaning that only the default value ever gets used, which means we don't need the setting). The goal of this is to clean out unused code and to reduce the quite large public interface of our Settings class.

There are a lot of changes here. I considered breaking this out into multiple PRs, but putting each setting in its own PR would lead to a LOT of PRs. Instead, I made each setting that got removed its own commit. In reviewing this, I would go through each commit instead of trying to look at all the changes at once.

One effect this will have is that for settings that are used but never set, it is possible that a user could have set the setting when they had a version of the app that still allowed that, and they've just been upgrading the app ever since, so the setting is still there in their shared preferences and getting read. I still think it's better to go ahead and remove the setting because the user can't change the setting anymore, so although they might like the setting, they also might not like it, and there's nothing they can do about it (and presumably we removed the setting for a good reason). 

Note that I am making these changes on top of the other settings refactorings that I've been doing recently, and I'm merging this first to the feature branch.

This does feel like a bit of a risky PR to me. I think it's worth it to get some overdue cleanup in, but if you disagree let me know.

## Testing Instructions

If CI is happy, we're probably fine on these. One thing that CI could miss would be if we have one of the Preference views that is keyed to a hardcoded string matching a setting that was removed. I was pretty careful about doing string searches to avoid this, but I wanted to mention it as a possible risk.

If you can think of any other ways that this could introduce issues that CI might miss, let me know because I'd like to test those as well.

Since CI might not catch everything, it would be good to do a bit of smoketesting of the app with these changes.

## Checklist

### DIFF
--- a/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/AutoArchiveTest.kt
+++ b/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/AutoArchiveTest.kt
@@ -62,13 +62,11 @@ class AutoArchiveTest {
         played: AutoArchiveAfterPlayingSetting,
         inactive: AutoArchiveInactiveSetting,
         includeStarred: Boolean = false,
-        excludedPodcasts: List<String> = emptyList()
     ): EpisodeManager {
         val settings = mock<Settings> {
             on { autoArchiveInactive } doReturn UserSetting.Mock(inactive, mock())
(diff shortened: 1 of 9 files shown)


geekygecko (approved):
Great work cleaning up the Settings class.

Issue #1220, opened 2023-07-31
OptionDialog Crashes on Configuration Changes


Mzazi25:
### Description

The ` optionDialog ` doesn't display as intended when there's a configuration change. 
If you change the screen orientation, the dialog crashes.

### Step-by-step reproduction instructions

1. Go to Discover 
2. Click any podcast 
3. Go the the episodes 
4. Long press one episode 
5. long press the selected episode to trigger the `optionDialog`
6. Change screen orientation 
7. Observe the changes, 
8. Go back to the previous orientation
9. Observe the changes again

### Screenshots or screen recording

https://github.com/Automattic/pocket-casts-android/assets/95022986/7260fdb5-4acd-47c6-9e93-ba175d09b0f3

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

Samsung Galaxy S22+, Android 13. 
Tested on debug builds 7.44-rc-2 and release 7.43

PR #1266, opened 2023-08-10
Option dialog configuration fix

CalebKL:
## Description

This PR fixes optionDialog configuration bug by saving and retrieving the data as a Parcelable object. 
Fixes # 
https://github.com/Automattic/pocket-casts-android/issues/1220

## Testing Instructions
1. Go to Profile
2. Go to settings(Top Right)
3. Tap Auto add to up next
4. Click Auto add limit to trigger the option Dialog
5. Change Screen Orientation
6. Observe the dialog doesn't crash

## Screenshots or Screencast 
https://github.com/Automattic/pocket-casts-android/assets/95022986/c6396a2f-88bd-4a5c-a5e1-dc5fc3c464e2

## Checklist
 
- [ ] with different themes


modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/dialog/OptionsDialog.kt
```diff
@@ -37,6 +39,32 @@ class OptionsDialog : BottomSheetDialogFragment() {
...
+        private const val KEY_TITLE = "title"
+        private const val KEY_OPTIONS = "options"
+        private const val KEY_ICON_COLOR = "iconColor"
+    }
+
+    @RequiresApi(Build.VERSION_CODES.TIRAMISU)
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        savedInstanceState?.let {
+            options.addAll(it.getParcelableArrayList(KEY_OPTIONS, OptionsDialogOption::class.java) ?: mutableListOf())
```
mchowning, line 53: This works fine on phones running Android 13, but any phones running older versions of Android will crash when they try to call the `getParcelableArrayList` method that only exists on Android 13. We need to do a version check and call the appropriate method for both Android 13 and for earlier versions. I would suggest creating an extension method similar to what we did with [getSerializableCompat](https://github.com/Automattic/pocket-casts-android/blob/aa9ae49b05e94fae69cdf7c1aa71241a8e0120f9/modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/extensions/Bundle.kt#L1-L0).

mchowning (changes requested):
Thanks for noticing this issue and putting together this fix!

This generally looks good, I just think there are two issues that we need to address before we can merge it.

First, after rotation the device, it doesn't seem like tapping on the buttons is working for me.

https://github.com/Automattic/pocket-casts-android/assets/4656348/07ccfa4b-1abd-4d80-b86d-de8cde00084e


</details>

Second, we need to make sure this code runs on older versions of Android (this is what I brought up in my other comment).

MiSikora:
@CalebKL Do you intend to work on this or should I close the PR?

MiSikora:
@CalebKL I'm going to close this issue. If you want to follow up on it with the changes please make a new PR.

PR #1267, opened 2023-08-10
Bookmarks - Add bookmark from headset hook

ashiagr:
## Description

This adds bookmarks using the [headset hook](https://developer.android.com/reference/android/view/KeyEvent#KEYCODE_HEADSETHOOK) button.

The headset hook button is already used to skip forward on double tap. The functionality is extended to add a bookmark based on how it is configured in "Headphone controls" settings.

## Testing Instructions
1. Install debug build
2. Launch the app 
3. Login to a paid account
4. Go `Profile` tab -> `Settings` -> `Headphone controls `
5. Tap on "Double tap headset hook button"
6. Select "Add bookmark" option
7. Go to "Discover" tab and play any episode
8. Connect headphone to the device 
9. With the app open, double tap headset hook button
    [Or you can run command `adb shell input keyevent 79` quickly twice to emulate the action]
10. ✅ Notice that "Add bookmark "screen is displayed
11. Dismiss the screen
12. Background the app
13. Again trigger the double tap headset hook button action
14. ✅ Notice that a notification is shown that the bookmark is added and sound is played
15. Tap on the notification 
16. App opens and navigates to the Player - Bookmarks screen 
17. Dismiss the screen and background the app again
18. Again trigger the double tap headset hook button action
19. ✅ Notice that a notification is shown that the bookmark is added and sound is played
20. Tap on the "Change title" notification action
21. ✅ Notice that "Change title" bookmark screen is shown
22. Update the title
23. ✅ Notice that snackbar appears with an action to view bookmark
24. Tap on view bookmark
25. ✅  Notice that  the app navigates to the Player - Bookmarks screen 

## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/1405144/95bea484-055c-4a01-849e-b1453f76d0f8

Notification on Android API Level 23 (Action icon not shown on [versions above)](https://android-developers.googleblog.com/2016/06/notifications-in-android-n.html)

notification

## Checklist
 
- [ ] with different themes


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
```diff
@@ -144,6 +148,13 @@ open class PlaybackManager @Inject constructor(
     private var audioNoisyManager =
         AudioNoisyManager(application)
 
+    private val tonePlayer: MediaPlayer by lazy {
+        MediaPlayer().apply {
+            setDataSource(application, Uri.parse("android.resource://${application.packageName}/${R.raw.tone}"))
```
ashiagr, line 153: This is a temporary sound file to test add bookmark confirmation sound.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/bookmark/BookmarkHelper.kt
```diff
@@ -0,0 +1,117 @@
...
+    fun handleAddBookmarkAction(
+        context: Context,
+    ) {
+        if (context.isAppForeground()) {
+            val bookmarkIntent =
+                context.packageManager.getLaunchIntentForPackage(context.packageName)
+                    ?.apply { action = INTENT_OPEN_APP_ADD_BOOKMARK }
+            bookmarkIntent?.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+            context.startActivity(bookmarkIntent)
+        } else {
```
ashiagr, line 42: I'm not sure about this. Maybe we don't need this if the bookmark is added using headset controls 🤔

CookieyedCodes:
You know how you can long press on skip forward and mark played & next episode option appears



With bookmarks I feel like a popup should be added to the skip backwards with the following options

Set bookmark at current time
.Go back 10 seconds and set bookmark
.Go back 15 seconds and set bookmark
.Go back 60 seconds and set bookmark
.Go back NUM seconds and set bookmark (customisable in settings or go off the existing skip back 🤔)
.Restart the current podcast to beginning 
(I don't know why I think doing a long press for this would be faster then using the scrub bar but it would be a nice other functionality for the long press at least 😅)

Would that be possible?

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/bookmark/BookmarkHelper.kt
```diff
@@ -0,0 +1,117 @@
...
+        Settings.NotificationChannel.NOTIFICATION_CHANNEL_ID_BOOKMARK.id
+    )
+        .setVisibility(NotificationCompat.VISIBILITY_PUBLIC)
+        .setContentTitle(context.getString(LR.string.bookmark_notification_title_added))
+        .setContentText(context.getString(LR.string.bookmark_notification_content_tap_to_view))
+        .setSmallIcon(IR.drawable.ic_launcher_foreground)
+        .setAutoCancel(true)
+        .setOnlyAlertOnce(true)
+        .setContentIntent(buildPendingIntent(context, INTENT_OPEN_APP_VIEW_BOOKMARKS))
+        .addAction(changeTitleAction)
```
geekygecko, line 92: Not for this version but I wonder if we should also have an action for "Delete". It might be a common action they want to take if they accidentally add a bookmark or want to move it. 

Screenshot_20230811_112853

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/notification/NotificationHelperImpl.kt
```diff
@@ -78,6 +78,14 @@ class NotificationHelperImpl @Inject constructor(@ApplicationContext private val
         }
         channelList.add(signInErrorChannel)
 
+        val bookmarkChannel = NotificationChannel(Settings.NotificationChannel.NOTIFICATION_CHANNEL_ID_BOOKMARK.id, "Bookmark", NotificationManager.IMPORTANCE_HIGH).apply {
+            description = "Shows when bookmark is added using broadcast"
```
geekygecko, line 82: I noticed none of the other strings are localised in this file but would it be a good idea?

geekygecko (approved):

geekygecko:
> You know how you can long press on skip forward and mark played & next episode option appears

@CookieyedCodes I forgot about that hidden action! I like the idea of adding more to that menu but I'm unsure about all the bookmark options. I think I would use the skip buttons to get the correct position first but I need to try it some more. 🤔

Thanks for all your comments. I'm amazed at the amount of feedback you give us.

CookieyedCodes:
> @CookieyedCodes I forgot about that hidden action! I like the idea of adding more to that menu but I'm unsure about all the bookmark options. I think I would use the skip buttons to get the correct position first but I need to try it some more. 🤔
> @geekygecko A lot of people dont even know truthfully 😅, I use it like once a week I would say truthfully, quite handy for rebroadcast episodes which I don't realise are rebroadcast at the time(hence why I do very much want auto archive by word), the main reason I have suggested so many options is to give headphone users more options too mainly because of the ease of use factor I am one take screenshots as bookmarks currently (as I will clip them up later) and I allways kind of hate that I have to go back/scrub the episodes to nearer the correct point hence the suggestion (I would probably get rid of the 10 second option but that is the default in the app 🤔🤔)
How about this?
.user rewind default +bookmark(in my case15)
.rewind 30 seconds + bookmark
.rewind 60 seconds + bookmark
.rewind NUM seconds (in headphone settings spefically) + bookmark
.restart to the beginning 
)
> Thanks for all your comments. I'm amazed at the amount of feedback you give us.
> It is no bother half the reason is because I am trained in UX for one (tho I suck at actual design) the other reason is I don't want more niggles in my favourite appXD

PR #1268, opened 2023-08-10
Replace UserSetting.flow.value with UserSetting.flow

mchowning:
## Description
This gives the `UserSetting` object a `value` field to access `flow.value`. The other changes are just replacing calls to `.flow.value` with `.value` as suggested in [this comment](https://github.com/Automattic/pocket-casts-android/pull/1264#discussion_r1289708778).

This PR is targeting the settings-refactor feature branch.

## Testing Instructions
If CI is happy, then we should be good.

## Checklist


geekygecko (approved):

PR #1204, opened 2023-07-25
Continue to upgrade screen from google sign in when appropriate

mchowning:
## Description
This fixes the onboarding flow when someone clicks an upgrade prompt and then logs into a free account using Google. See #1184 for more info.

Fixes #1184 

## Testing Instructions

If you test this on a debug build, note that tapping on the "Upgrade to Plus" button can freeze the app if you're already logged in (https://github.com/Automattic/pocket-casts-android/issues/1158). Also note that the upgrade screen won't have any subscription options unless you make the changes needed to pull in subscriptions (which are not needed to test this PR).

### 1. Fixes upgrade flow
1. Fresh install of the app
2. Go to `Profile` tab → ⚙️ → Pocket Casts Plus
3. Tap on "Upgrade to Plus"
4. Tap "Continue with Google"
5. Log into an already existing free Pocket Casts account
6. ✅ Observe that you are taken to the onboarding flow's upgrade screen.

#### 2a. Upgrading and creating a new Pocket Casts account with Google sign in
1. Fresh install of the app
2. Go to `Profile` tab → ⚙️ → Pocket Casts Plus
3. Tap on "Upgrade to Plus"
4. Tap "Continue with Google"
5. Create a new Pocket Casts account (i.e., select a Google login that is not associated with a PC account already)
6. ✅ Observe that you are taken to the recommendations screen in the onboarding flow.

#### 2b. Signing into a free account using Google outside the upgrade flow
1. Fresh install of the app
2. Tap "Continue with Google"
3. Create a new Pocket Casts account (i.e., select a Google login that is not associated with a PC account already)
4. ✅ Observe that you are taken to the recommendations screen in the onboarding flow.

#### 2c. Creating a new Google account in the regular onboarding flow
1. Fresh install of the app
2. Tap "Continue with Google"
3. Log into an already existing free Pocket Casts account
4. ✅ Observe that you are taken out of the onboarding flow and into the app

## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/4656348/ba087724-f960-4bfd-89ea-84f386f3a29f

## Checklist
- [x] If this is a user-facing change, I have added an entry in CHANGELOG.md
- [X] Ensure the linter passes (`./gradlew spotlessApply` to automatically apply formatting/linting)
- [X] I have considered whether it makes sense to add tests for my changes
- [X] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml`
- [X] Any jetpack compose components I added or changed are covered by compose previews

PR #1275, opened 2023-08-11
Ensure upgrade flow continues after signing into account with google sign in

mchowning:
## Description

This addressses an issue where the upgrade flow for logged-out users would exit after the user signed in with a google account. Basically, it's a follow-up to https://github.com/Automattic/pocket-casts-android/pull/1204 that further improves the flow.

As part of this I added test scenarios for the onboarding flow to our testing checklist. I also created a very basic first step toward testing the onboarding flow's navigation. All it does now is test the start destination though because this ended up taking a bit to get working and I ran out of time. I suspect we may be better off moving as much of the navigation logic as we can into a view model and testing that in isolation. If you think it's not worth adding the integration test here since it's not doing much yet, just let me know and I can remove that commit.

Fixes # 

## Testing Instructions
1. When not signed into an account
2. Tap on an upgrade prompt like the folder icon on the Podcasts tab
3. Sign in with a google account
4. After sign in you should proceed through the upgrade flow (before this change you would be returned to the screen from which you tapped on the upgrade prompt).

Also test a few of the onboarding flows I've added to the testing-checklist to make sure there aren't any regressions.

## Checklist

### DIFF
--- a/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingFlowComposable.kt
+++ b/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingFlowComposable.kt
@@ -1,5 +1,6 @@
 package au.com.shiftyjelly.pocketcasts.account.onboarding
 
+import androidx.annotation.VisibleForTesting
 import androidx.compose.runtime.Composable
 import androidx.navigation.NavHostController
 import androidx.navigation.NavType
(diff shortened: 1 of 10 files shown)


modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/OnboardingFlowComposable.kt
```diff
@@ -125,21 +129,7 @@ private fun Content(
                     if (state.isNewAccount) {
                         onAccountCreated()
                     } else {
-                        when (flow) {
-                            OnboardingFlow.InitialOnboarding,
-                            OnboardingFlow.LoggedOut,
-                            is OnboardingFlow.PlusAccountUpgrade,
-                            is OnboardingFlow.PlusUpsell -> exitOnboarding()
```
mchowning, line 132: Having these two flows _not_ exit onboarding after signing in with Google is the fix this PR is making.

geekygecko (approved):
This worked well for me.

Issue #1273, opened 2023-08-11
On deleting all bookmarks for a file, the bookmark icon is not removed for it until the view is reloaded


ashiagr:
Steps to reproduce

- Login using paid account
- Go to `Profile` tab ->  `Files` screen
- Add a file
- Play the file and a few bookmarks using the `Player` screen bottom toolbar "Add bookmark" option
- Return to `Files` screen
- Notice that the file has a bookmark icon
- Tap on the file 
- On the bottom sheet that appears, tap on `Bookmarks` menu
- Delete all bookmarks
- Return to `Files` screen
- Notice that the bookmark icon on the file is still present and not removed until the view is reloaded

PR #1276, opened 2023-08-13
Bookmark icon on Delete fix

Mzazi25:
## Description

This PR fixes on deleting bookmark icon without need to reload the view. 

Fixes #  #1273 

## Testing Instructions

1. Login using paid account
2. Go to Profile tab -> Files screen
3. Add a file
4. Play the file and add a few bookmarks using the Player screen bottom toolbar "Add bookmark" option
4. Return to Files screen
5. Notice that the file row has a bookmark icon
6. Tap on the file
7. On the bottom sheet that appears, tap on Bookmarks menu
8. Delete all bookmarks
9. Check the bookmark icon is deleted.

## Checklist
 
- [ ] with different themes


ashiagr:
Thanks for this PR, @Mzazi25 🙏 

Can you give it a try with a playback paused state? It seems that when the episode is playing, the UI is refreshed at regular intervals and the icon is updated based on the latest bookmarks state. But when the playback is paused, we need to pull to refresh the screen to see the updated state.

https://github.com/Automattic/pocket-casts-android/assets/1405144/111b41a7-761a-4f79-a2e1-a9753b39371b


Notice that in the `FileStatusIconsView`, we do not observe bookmark changes [here](https://github.com/Automattic/pocket-casts-android/blob/67420270cb8a4ea96797507cac3cccadf2118295/modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/component/FileStatusIconsView.kt#L109):

```
Observables.combineLatest(combinedProgress, playbackStateForThisEpisode, isInUpNextObservable)
```

modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/component/FileStatusIconsView.kt
```diff
@@ -77,9 +77,12 @@ class FileStatusIconsView @JvmOverloads constructor(
         downloadProgressUpdates: Observable<DownloadProgressUpdate>,
         playbackStateUpdates: Observable<PlaybackState>,
         upNextChangesObservable: Observable<UpNextQueue.State>,
-        hideErrorDetails: Boolean = false
-    ) {
+        hideErrorDetails: Boolean = false,
 
+    ) {
+        val bookmarkVisibilityObservable = BehaviorRelay.create<Boolean?>().apply {
+            accept(episode.hasBookmark)
```
ashiagr, line 84: Observing changes to the episode's bookmark status will have any effect as the episode is not changing here.

modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/component/FileStatusIconsView.kt
```diff
@@ -101,24 +104,39 @@ class FileStatusIconsView @JvmOverloads constructor(
...
             .toFlowable(BackpressureStrategy.LATEST)
             .subscribeOn(Schedulers.io())
             .observeOn(AndroidSchedulers.mainThread())
             .doOnTerminate { UploadProgressManager.stopObservingUpload(episode.uuid, uploadConsumer) }
-            .doOnNext { (streamingProgress, playbackState, isInUpNext) ->
-                episode.playing = playbackState.isPlaying && playbackState.episodeUuid == episode.uuid
-                imgUpNext.visibility = if (isInUpNext) View.VISIBLE else View.GONE
+            .doOnNext { combinedData ->
+                episode.playing = combinedData.playbackState.isPlaying && combinedData.playbackState.episodeUuid == episode.uuid
+                episode.hasBookmark = combinedData.playbackState.episodeUuid == episode.uuid
```
ashiagr, line 134: As you mentioned, this is not required. When you'll receive the updated bookmarks list, it'll have an updated `hasBookmark` status.

ashiagr:
👋 @Mzazi25 

As discussed in DM, the issue is resolved for user episodes but a similar problem exists for podcast episodes on the podcast details screen. I'm merging this PR and we will address it in a separate PR.

ashiagr (approved):

PR #1279, opened 2023-08-14
Add scroll to top/bottom to logs screen

mchowning:
## Description
This adds buttons to make it easy to get to the top or bottom of the logs screen (it can get very long).

## Testing Instructions
1. Go to logs screen (Profile tab → ⚙️ → "Help & feedback" → logs button on toolbar)
2. Verify that the toolbar has scroll to top and scroll to bottom buttons and they work as expected

## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/4656348/a1eb092f-8848-420f-9556-ff19d241cb74

## Checklist


geekygecko (approved):
Great idea. This worked well for me. 

It would be good to fix the toolbar icon tint colours as they are different to the previous screen but that isn't related to this PR.

PR #1280, opened 2023-08-15
Stop episodes being removed from the Up Next when clearing downloads.

geekygecko:
## Description

This change fixes the issue that if any episodes in the Up Next are downloaded when the downloads are cleared in the settings area they are removed from the Up Next. 

## Testing Instructions
1. Tap on the 'Profile' tab
2. Tap on the settings cog
3. Tap on 'Auto download'
4. Turn on 'Episodes added to Up Next'
5. Add multiple episodes to the Up Next
6. Open the Up Next
7. ✅ Verify the downloaded icon is on the episode rows
8. Go back to the settings section
9. Tap on 'Storage & data use'
10. Tap on 'Downloaded files'
11. Tick all the checkboxes and tap 'Clean up'
12. ✅ Verify the Up Next episodes haven't been removed

## Screencast 

https://github.com/Automattic/pocket-casts-android/assets/308331/6396fe63-b5b8-4f2b-a332-de6458a49c00

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
     *   Added a support section to the Automotive app
         ([#1277](https://github.com/Automattic/pocket-casts-android/pull/1277))
 *   Bug Fixes:
+    *    Fixed episodes being removed from the Up Next when clearing downloads
+         ([#1280](https://github.com/Automattic/pocket-casts-android/pull/1280))
     *    Improved upgrade flow when signing in with Google account
          ([#1275](https://github.com/Automattic/pocket-casts-android/pull/1275))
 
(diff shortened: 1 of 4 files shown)


mchowning (approved):
This looks good and works well.

I do kind of wonder if it might make sense to leave episodes in the Up Next queue in other circumstances where the download is deleted too (from the episode details screen and when multiselecting episodes), but that is a more significant behavior change, so we shouldn't do that lightly.

CookieyedCodes:
I'll agree with matt to me it dose make sense that clean up would do it but the others delete methods shouldn't, I would mention on the clean up warning popup that in bold that this will also remove files from upnext now too tho 😅

geekygecko:
> I'll agree with matt to me it dose make sense that clean up would do it but the others delete methods shouldn't

I agree with Matt, but I understood it slightly differently, that anywhere you can delete an episode shouldn't remove it from Up Next. I didn't make that change as it would need a lot more testing in areas such as unsubscribing from a podcast.

CookieyedCodes:
Ah I got you the video I thought was showing the new functionality not the bug, I suppose I wouldn't mind a check box truthfully in the cleanup menu that would wipe stuff from upnext, just because some people it might be their prefences 🤔

PR #1284, opened 2023-08-17
Bookmarks - Update headphone controls options display

ashiagr:
## Description

This
- Updates headphone controls options layout from `SettingRadioDialogRow` to `OptionsDialog`
- Displays "Add bookmark" option only if the user is eligible for `Bookmarks` feature
- Returns default values for headphone controls if the previous value is "Add bookmark" and the user is not eligible for the `Bookmarks` feature

Slack discussion p1692252268904039/1692162285.498199-slack-C055BDMU095

P.S: Since Android didn't have a skip chapters option for headphone controls, we only focussed on adding skip actions and "Add bookmark" in this PR. Other options will be added as part of a separate PR.

## Testing Instructions

Prerequisites:

- Release build
- Patron + Bookmarks features enabled in `Feature` class
- License test account

1. Upgrade to `Patron` plan
2. Go to `Profile` -> `Settings`
3. Notice that "Headphone controls" menu is present
4. Tap on "Headphone settings" menu 
5. Tap on "Previous" or "Next" action
6. Notice that the options layout is changed to use`OptionsDialog`
7. Notice that options have "Add bookmark" option
8. Select "Add bookmark" options for both previous and next actions
9. Downgrade to `Plus` or `Free` plan
10. Go to `Profile` -> `Settings` 
11. Notice that "Headphone controls" menu is present
12. Tap on "Headphone controls" menu 
13. Notice that "Previous" and "Next" actions do not have "Add bookmark" option and have been set to default values. 

## Screenshots or Screencast 

Before | Now
----| ----
Screenshot 2023-08-17 at 12 55 07 PM |  

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/HeadphoneControlsSettingsFragment.kt
+++ b/modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/HeadphoneControlsSettingsFragment.kt
@@ -4,6 +4,7 @@ import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.ViewGroup
 import androidx.annotation.StringRes
+import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.rememberScrollState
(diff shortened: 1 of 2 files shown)


ashiagr:
Hey @CookieyedCodes 

This PR only focuses on adding "Add bookmark" at the current position through previous/ next headphone controls. I remember that discussion in https://github.com/Automattic/pocket-casts-android/pull/1267 but we need to try them out and get a formal approval. We appreciate your patience till then. 🙏

mchowning (approved):
This looks good. 

The only issue I noticed was that if I signed out of my account and then immediately checked the headphone controls settings, the UI still showed my previously-selected "Add bookmark" actions. If I tapped on either action to select a new one, there was no "Add bookmark" action. Swiping the app away, reopening it, and then returning to the headphone settings then properly showed the default actions. 

This feels like an edge case that is probably just due to the settings screen not reloading the setting to see that the value had changed.

_If_ you want to change this to use a `Flow`, I think it would be good to try to covert these headphone action settings into UserSettings since that will help make these easy to sync in the future, and I'm happy to help with that conversion since I may have just a bit of practice converting settings into UserSettings, just let me know. Making this setting into a flow is certainly not the only way to address this though, so don't feel like you have to go that direction.

ashiagr:
Hey @mchowning, I'm on board to convert it to a user setting. Since user setting changes are in a separate branch, and considering it is an edge case, I'm merging this PR and we can take up the conversion in a separate PR.

mchowning:
> Since user setting changes are in a separate branch, and considering it is an edge case, I'm merging this PR and we can take up the conversion in a separate PR.

I've opened a PR (https://github.com/Automattic/pocket-casts-android/pull/1291) where we can continue this conversation.

PR #1290, opened 2023-08-18
Merging UserSettings to main

mchowning:
## Description
This is merging to `main` the refactoring of the to-be-synced settings to be UserSettings.

## Testing Instructions
All of these changes have already been tested in separate PRs, so they shouldn't need to be retested individually.

The one exception to this is the settings on the Notification screen because I had to resolve some non-trivial merge conflicts when merging in main due to https://github.com/Automattic/pocket-casts-android/pull/1264 and https://github.com/Automattic/pocket-casts-android/pull/1244. It's worth giving those another quick test to just make sure there aren't any issues.

> **Warning**
> 7.46 has been cut, so this is no longer relevant. ~This can be reviewed now, but let's not merge it until after the 7.46 release is cut since the number of changes in here creates a bit of a risk that there could be a regression, so it would be good to have as much time as possible to catch any issues.~

## Checklist


ashiagr (approved):
I smoke-tested a few settings, including notifications. They all work as expected. 

Great job getting this done!

mchowning:
I pushed an update in https://github.com/Automattic/pocket-casts-android/pull/1290/commits/43c3edab84b3807486b38837dead35765430ab2d to address the proguard issue in release builds. Let me know what you think.

ashiagr:
LGTM 👍

CookieyedCodes:
Now I know this project is makeing progress for a future release, and I know notification, add to upnext (including which upnext/uplast selection) & auto downloads & per podcast playback FX are unlikely to sync but if their is a chance you can make this more seemless alot of people wouldn't mind and with individual experimental advanced option to turn on sync for these (obviously with clear warnings if required 

the switching device pain is one of the few complaints I regularly see about podcatchers (now should this be a patron level feature I'll leave you decide but it might erk people & you'd get good brownie points with alot of users if the options were at least their🤔) let me know tho some of the technical reasons why it would be difficult if it is tho id be very interested just so I can tell others as I see it popup 🤔 😅

mchowning:
>  I know notification, add to upnext (including which upnext/uplast selection) & auto downloads & per podcast playback FX are unlikely to sync

We're actually planning to sync all of these except for auto downloads (because we feel people might have different auto download preferences on different devices, not because of technical limitations).

CookieyedCodes:
Great stuff, I'll probably be the person who checks with the high numbers sync as I have 413 notifications, 335 set to auto download & 326 set to add to upnext 😅, so please be careful, I'll definitely let you know tho, I'll also note add to upnext might cause complications if archive/stared sync isn't timed right

I think I would agree tho I still think maby an advanced option would be appropriate to try cut down on list comparisons, I also think if you figure out how local playback effects sync is done it should apply to the webapp too as episodes switch which would add some added complexity 🤔

Looking forward to testing tho I will be cautious ❤️❤️

PR #1291, opened 2023-08-18
Update headphone control to be UserSettings

mchowning:
## Description
This updates the headphone controls to be UserSettings and also makes the headphone control settings react to changes in the user subscription to insure it is always up-to-date (see discussion in https://github.com/Automattic/pocket-casts-android/pull/1284#pullrequestreview-1583549610).

This PR is targeting the `feature/sync-settings-refactor` branch (which currently has a PR open to merge to `main`)

## Testing Instructions
1. Make sure the headphone controls update and work as expected.
2. When logged into a Patron account, set the headphone controls to "Add bookmark"
3. Log out of your account
4. Note that the headphone controls are no longer "Add bookmark"
5. Log back into your Patron account
6. Note that the headphone controls have returns to be "Add bookmark" if they were that before (and assuming you didn't update the headphone control setting when you were logged out). Let me know if you think we should overwrite the persisted headphone control when a user logs out automatically. I thought it was better to leave it so someone that logs out and logs right back in doesn't lose their setting, but I don't think there's a clear correct answer here.

## Checklist

### DIFF
--- a/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/CancelConfirmationViewModel.kt
+++ b/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/CancelConfirmationViewModel.kt
@@ -17,7 +17,7 @@ class CancelConfirmationViewModel
 ) : ViewModel() {
     var expirationDate: String? = null
     private val paidSubscription: SubscriptionStatus.Paid?
-        get() = settings.getCachedSubscription() as? SubscriptionStatus.Paid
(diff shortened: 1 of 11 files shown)


ashiagr (approved):
This is working well. 👍 Thanks for the updates.

> Note that the headphone controls have returns to be "Add bookmark" if they were that before (and assuming you didn't update the headphone control setting when you were logged out). Let me know if you think we should overwrite the persisted headphone control when a user logs out automatically. I thought it was better to leave it so someone that logs out and logs right back in doesn't lose their setting, but I don't think there's a clear correct answer here.

I too think it doesn't have a clear answer. I'd leave the current implementation as it is.

Issue #1295, opened 2023-08-21
Initial Onboarding Screen Missing Episode Artwork


mchowning:
### Description

The onboarding screen appears to sometimes not have the episode artwork showing. Earlier I thought this was only occurring on debug builds, but I am now seeing it on the 7.46-rc-1 build.

### Step-by-step reproduction instructions

1. Fresh install the app
2. Observe that the initial onboarding screen that opens does not have any epsiode artwork.

### Screenshots or screen recording

| 7.45 | 7.46-rc-1 |
| --- | --- |
| ![Screenshot 2023-08-21 at 11 17 33 AM](https://github.com/Automattic/pocket-casts-android/assets/4656348/95911d26-8949-4170-b42c-7ed4dfcfa54c) | ![Screenshot 2023-08-21 at 11 23 34 AM](https://github.com/Automattic/pocket-casts-android/assets/4656348/14fb3eb2-9c55-45e4-a512-6ac759eb25fc) |

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

Samsung A13, Android 13, Pocket Casts 7.46-rc-1

PR #1299, opened 2023-08-22
Fix the artwork not loading on the onboarding page

geekygecko:
## Description

The issue is that the `LocalView.current` returns a view with zero width and height. To fix this I have replaced this logic with a `BoxWithConstraints`. 

Before I found the issue I thought it was related to the podcasts in the ViewModel being a list and not a state variable. Whilst these changes didn't fix it I believe they are still worth making. Let me know what you think, I will remove them if you don't agree.

Fixes https://github.com/Automattic/pocket-casts-android/issues/1295

## Testing Instructions

1. Do a refresh install of the app
2. ✅ Verify you are shown the podcast artwork when the app launches
3. Don't sign in and close the dialog
4. Subscribe to a few podcasts
5. Tap on the 'Profile' tab
6. Tap "Set Up Account'
7. ✅ Verify the subscribed podcasts are included in the artwork

## Screencast 

https://github.com/Automattic/pocket-casts-android/assets/308331/deaddbd0-3a31-47de-ac3a-da4af77be6cb

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 *   Updates:
     *   Enable copying logs from in-app logs viewer
         ([#1298](https://github.com/Automattic/pocket-casts-android/pull/1298))
+*   Bug Fixes:
+    *    Fixed the artwork not appearing on the onboarding page
+        ([#1299](https://github.com/Automattic/pocket-casts-android/pull/1299))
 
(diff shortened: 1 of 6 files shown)


ashiagr (approved):
I can see the artwork again. Code changes look good as well. Thanks for fixing it!

PR #1300, opened 2023-08-22
Avoid crash when accessing and dismissing episode details from up next queue

mchowning:
## Description
This avoids a crash that can occur when dismissing the episode details view.

This issue was introduced in https://github.com/Automattic/pocket-casts-android/pull/1282.

## Testing Instructions
1. Make sure you have the "General" setting for playing episodes when you tap them in your up next queue turned off (we need it to open the episode details screen).
2. Open the up next queue without opening a podcast screen (if you go to a podcast screen first the crash doesn't happen)
3. Tap on an episode to open the episode details view
4. Dismiss the episode details view
5. Observe that the app doesn't crash

## Checklist
- [X] If this is a user-facing change, I have added an entry in CHANGELOG.md (no changelog because this is a regression since the last release)


mchowning:
To add some more context. This issue arose because I mistakenly thought the only way to access the episode details screen was from the podcast screen. So I made this field `lateinit` since it needs to be initialized anytime the podcast screen loads. Since the episode details screen can be accessed without going through the podcast screen, we no longer need to ensure that the field is initialized and it is safe for it to be nullable.

ashiagr (approved):
I was able to reproduce the crash without the changes and I confirm that the crash is fixed now. 👍

PR #1302, opened 2023-08-23
Fix Protocol Buffers in the release build

geekygecko:
## Description

This change fixes two issues.

1. The new Protocol Buffers don't work after Proguard has run.

2. After your first login if it downloaded any bookmarks they are uploaded to the server again.

### Test Proguard
1. In `Feature.kt` set the default value for bookmarks to true.
2. Build a release version of the app `./gradlew app:assembleRelease`
4. Install the build
5. Login to a Pocket Casts Plus account
6. Add a bookmark
7. On the 'Profile' tab manually refresh
8. ✅ Verify the refresh is successful

### Test first login
1. Add some bookmarks to a Pocket Casts account
2. Sign out and clear the app storage (or reinstall)
3. Open the App Inspection window
4. Open the Network Inspector
5. Tap Refresh Now
6. ✅ Verify the https://api.pocketcasts.net/sync/update call doesn't send any changes to the server. `"changes":[]`

### DIFF
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -185,3 +185,6 @@
 # R8 full mode strips generic signatures from return types if not kept.
 -if interface * { @retrofit2.http.* public *** *(...); }
 -keep,allowoptimization,allowshrinking,allowobfuscation class <3>
+
+# Protocol Buffers - keep the field names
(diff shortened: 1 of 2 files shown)


ashiagr (approved):
This is looking good. Thanks for these updates!

PR #1303, opened 2023-08-23
Remove Sentry NDK

geekygecko:
## Description

As we don't use the NDK we don't need to include `sentry-android-ndk`. Removing these libraries reduces the APK from 28.4MB to 23.8MB. 

We have also been contacted by an Automotive partner as these libraries include old C functions which it recommends not using. I thought we could have tried just the library upgrade but if we don't need these anyway maybe it's best to remove them.

Here is the [Sentry documentation](https://docs.sentry.io/platforms/android/configuration/using-ndk/#using-the-sdk-without-the-ndk).

## Testing Instructions
1. Change `modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/SettingsFragmentPage.kt` so the developer menu is visible in a release build.
2. Install a release build
3. Go to 'Profile' tab -> settings cog -> Developer
4. Tap 'Report a crash'
5. ✅ Verify in Sentry the exception has the same information as previous reports

### DIFF
--- a/base.gradle
+++ b/base.gradle
@@ -241,7 +241,7 @@ dependencies {
     implementation libs.accompanistFlowLayout
     implementation libs.accompanistSystemUiController
     implementation platform(libs.sentryBom)
-    implementation libs.sentry
+    implementation libs.sentryAndroidCore
     implementation libs.sentryFragment
     implementation libs.sentryTimber
     implementation libs.media3Cast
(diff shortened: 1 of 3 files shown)


ashiagr (approved):
Works as described. 👍  Thanks for adding the `SendCrashSetting` developer option!

PR #1305, opened 2023-08-24
In-app review

ashiagr:
## Description

This matches iOS PR: https://github.com/Automattic/pocket-casts-ios/pull/240 and requests in-app review if:
- The user checks Stats and stays there for more than a second after the information appears
- The user has been using the app for more than a week
- The user has a total of playedUpTo in the last 7 days bigger than 2.5 hours
- The request hasn't been made to this user before

Logic to display the review dialog only once and saving review request date matches iOS logic for consistency and future extensibility.

## Testing Instructions

Prerequisites: 

[Optional]
1. Apply [this patch](https://github.com/Automattic/pocket-casts-android/files/12428599/in-app-review.patch) (that reduces playedUpto duration check from 2.5 hours to mins)
2. Build apk and upload to [internal app sharing](https://play.google.com/console/internal-app-sharing/)
3. Open the uploaded apk url on device and fresh install the app

[Or]

Install already uploaded apk with this patch from [this url](https://play.google.com/apps/test/RQyLMLSAri8/ahAO29uNQSXaVkDTTiNZnDfwX5J9-NsL863dS6oYkI7esWzYkxFQvZWiu7o6bOl7Ifu7vAARS7KgzJeqIcuDGG6M61
)

> Important: When using an app installed with internal app sharing, reviews can't be submitted. To emphasize this difference, the button is disabled in the UI.

### Account with more than a week
1. Open the app and login to an account that you created more than a week ago and with 
2. Go to Stats
3. Wait a few seconds
5. ✅ Make sure the request review appears
6. ✅ Make sure that `Tracked: app_store_review_requested ["source": "stats"]` is logged
7. Tap any star or dismiss it
8. Close the app and reopen it again
9. Go to Stats
10. Wait a few seconds
11. ✅ You should not see the review dialog

### Account with less than a week

Put a breakpoint on lines `112` and `117` of `StatsViewModel`

1. Signout and clear app storage
2. Create an account
3. Listen to a podcast for more than the time you specified on `StatsViewModel` line `110` 
4. Go to Profile > Refresh Now
5. Open Stats
6. ✅ Check that the breakpoint `112` is hit
7. Continue execution
8. ✅ Check that breakpoint `117` is not hit because your account is less than a week old

## Checklist
 
- [ ] with different themes

### DIFF
--- a/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/views/review/InAppReviewHelperTest.kt
+++ b/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/views/review/InAppReviewHelperTest.kt
@@ -15,8 +15,8 @@ import org.mockito.Mock
 import org.mockito.MockitoAnnotations
 import org.mockito.kotlin.mock
 import org.mockito.kotlin.spy
+import org.mockito.kotlin.times
 import org.mockito.kotlin.verify
-import org.mockito.kotlin.verifyNoMoreInteractions
 import org.mockito.kotlin.whenever
 import java.util.Date
 
@@ -50,8 +50,7 @@ class InAppReviewHelperTest {
         launchReviewDialog()
         launchReviewDialog()
 
-        verify(reviewManager).requestReviewFlow()
-        verifyNoMoreInteractions(reviewManager)
+        verify(reviewManager, times(1)).requestReviewFlow()
     }
 
     private fun launchReviewDialog() = runTest() {
--- a/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/views/review/InAppReviewHelperTest.kt
+++ b/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/views/review/InAppReviewHelperTest.kt
@@ -1,0 +1,72 @@
+package au.com.shiftyjelly.pocketcasts.views.review
+
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.test.platform.app.InstrumentationRegistry
+import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsTrackerWrapper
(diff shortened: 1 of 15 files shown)


mchowning (approved):
Nice work on this @ashiagr ! This works well. There were just a few small issues I noticed.

## 1. The server returns the UNIX epoch as the `statsStartedAt` Date until the user syncs their listening history

Until a user syncs their listening history, the server returns the unix-epoch as the `statsStartedAt` Date (internal discussion: p1693498414372639-slack-C029BMW150R). As a result, we're finding that they've used the app for a _lot_ more than 7 days. This usually wouldn't matter, but it does mean that we basically ignore the check for how long they've used the app until they sync their stats.

I pushed a commit addressing this by considering the value to be 0 days if we get the UNIX epoch for the `statsStartedAt` time (https://github.com/Automattic/pocket-casts-android/pull/1305/commits/7cbe90531fe017ca65f9356565560f6996df0177). I also added a test to account for that edge case. I didn't observe any undesireable side effects, but let me know if you see any issues.

## 2. Listening stats are only counted for subscribed podcasts

I noticed that we're not considering any listening stats for podcasts that haven't been subscribed to or for user-uploaded files. This feels like an edge case, so I don't think this is a blocker to getting this out the door, but I wanted to mention it. Assuming this isn't an intentional decision, maybe we should log an issue for this just to keep track of it? (I don't feel strongly)

## 3. Changelog entry?

Seems like we want a changelog entry for this? I would have gone ahead and added it, but I wasn't sure whether this would make it into `7.47`, so I held off.

ashiagr:
Thanks for catching those @mchowning 

For 2, I included listening stats for unsubscribed podcasts also in https://github.com/Automattic/pocket-casts-android/pull/1305/commits/0f4219b73ca671c67ab90aef823384f188edf148.

User episodes are excluded for now to match iOS behavior.

Issue #1288, opened 2023-08-17
Changing title on a bookmark shows "Bookmark added" toast


mchowning:
### Description

When changing the title on an existing bookmark from the full screen player. The same "Bookmark added" toast that is used when adding a bookmark is shown. It would be better to update the language here to reflect that just the title was updated.

### Step-by-step reproduction instructions

1. Go to the full screen player
2. Pause playback
3. Add a new bookmark
4. Note that the "Bookmark added" toast is shown
5. "Add" another bookmark at the same location (which means you're changing the bookmarks title, as reflected by the confirmation button when you input the title)
6. Observe the "Bookmark added" toast is shown.

### Screenshots or screen recording

https://github.com/Automattic/pocket-casts-android/assets/4656348/c7fc6912-ffa4-4c01-99e7-9fdf3f9036af

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

Samsung A13 running Andorid 13, PC 7.46 pre-release-build

PR #1309, opened 2023-08-25
Change Snackbar message on existing bookmark

Mzazi25:
## Description

This PR updates the snackbar message to "bookmark exists" when a bookmark is already added. 

Fixes #  [1288](https://github.com/Automattic/pocket-casts-android/issues/1288)

## Testing Instructions
1. Go to full screen player 
2. Pause playback
3. Add a bookmark at any minute
4. Add bookmark again( Check the Toast message)

## Checklist
 
- [ ] with different themes

### DIFF
--- a/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
+++ b/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
@@ -1432,7 +1432,7 @@ class MainActivity :
         if (result == null) return
 
         val snackbarMessage = if (result.isExistingBookmark) {
-            getString(LR.string.bookmark_updaed, result.title)
+            getString(LR.string.bookmark_updated, result.title)
         } else {
             getString(LR.string.bookmark_added, result.title)
         }
--- a/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
+++ b/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
@@ -1432,7 +1432,7 @@ class MainActivity :
         if (result == null) return
 
         val snackbarMessage = if (result.isExistingBookmark) {
-            getString(LR.string.bookmark_exists, result.title)
+            getString(LR.string.bookmark_updaed, result.title)
         } else {
             getString(LR.string.bookmark_added, result.title)
         }
--- a/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
+++ b/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
@@ -1431,11 +1431,17 @@ class MainActivity :
         val view = snackBarView()
         if (result == null) return
 
+        val snackbarMessage = if (result.isExistingBookmark) {
+            getString(LR.string.bookmark_exists, result.title)
+        } else {
+            getString(LR.string.bookmark_added, result.title)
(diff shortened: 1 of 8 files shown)


modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/PlayerHeaderFragment.kt
```diff
@@ -579,14 +579,30 @@ class PlayerHeaderFragment : BaseFragment(), PlayerClickListener {
...
+        } else {
+            getString(LR.string.bookmark_added, result.title)
+        }
         val viewBookmarksAction = View.OnClickListener {
             (parentFragment as? PlayerContainerFragment)?.openBookmarks()
         }
-        Snackbar.make(view, getString(LR.string.bookmark_added, result.title), Snackbar.LENGTH_LONG)
+
+        val existingBookmarkUuid = result.bookmarkUuid
+        settings.setExistingBookmarkUuid(existingBookmarkUuid)
```
ashiagr, line 594: There's a problem with this approach. If we add a bookmark at the first bookmark position after creating a second bookmark at another position, my first bookmark uuid will not be counted as an existing one since the second one has replaced it.

See if this video helps:

https://github.com/Automattic/pocket-casts-android/assets/1405144/447d6c2c-8d19-4b37-817e-1e301dfbe64e

We can see if passing `isExistingBookmark` from the `BookmarkViewModel`'s `saveBookmark(...)` down to  `BookmarkActivityContract.BookmarkResult` just like other params `bookmarkUuid`, `title`, `tintColor` helps.

Note that there's also a `showViewBookmarksSnackbar` on `MainActivity`.

ashiagr (changes requested):

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/PlayerHeaderFragment.kt
Mzazi25: > There's a problem with this approach. If we add a bookmark at the first bookmark position after creating a second bookmark at another position, my first bookmark uuid will not be counted as an existing one since the second one has replaced it.
> 
> See if this video helps:
> 
>  Screen_recording_20230825_161002.mp4 
> We can see if passing `isExistingBookmark` from the `BookmarkViewModel`'s `saveBookmark(...)` down to `BookmarkActivityContract.BookmarkResult` just like other params `bookmarkUuid`, `title`, `tintColor` helps.
> 
> Note that there's also a `showViewBookmarksSnackbar` on `MainActivity`.

mmh interesting. I didn't notice that. Let me try passing the `isExistingBookmark` flag from the viewmodel.

Mzazi25:
@ashiagr I have changed the implementation as suggested. I think its okay now, I checked the logs and its working well now. 
Lmk if it looks good to you?

app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
```diff
@@ -1432,7 +1432,7 @@ class MainActivity :
         if (result == null) return
 
         val snackbarMessage = if (result.isExistingBookmark) {
-            getString(LR.string.bookmark_exists, result.title)
```
Mzazi25: one minor typo on the last commit

Mzazi25 (commented):
@ashiagr one minor typo here

Issue #1307, opened 2023-08-24
Bookmarks: Final Touches


emilylaguna:
### Player
- [x] Rename the Show Notes tab to 'Details' (https://github.com/Automattic/pocket-casts-android/pull/1311)
- [x] Show the Add Bookmark option even if bookmarks are not unlocked (https://github.com/Automattic/pocket-casts-android/pull/1332/commits/f2a946e1307516950cb08b4fc4e8d77b8bb3d35c)
    - Tapping it should show the upsell

### Upgrade View
- [x] Show the Pocket Casts logo image instead of text (https://github.com/Automattic/pocket-casts-android/pull/1311)
- [x] Change the upgrade button title to be: (https://github.com/Automattic/pocket-casts-android/pull/1331)
   - If the tier has a free trial: `Start Free Trial`
   - If not: `Upgrade to X`
      - Where X is name of the tier (Plus, or Patron)

### Empty State
- [x] The text color of the title appears too light (https://github.com/Automattic/pocket-casts-android/pull/1311)

### Bookmarks List
- Bookmarks may not display if the episode is archived
   - I'm not able to reproduce this one again but noting it here since it came up - (Unable to reproduce)
- [x] The oldest/newest sort options are reversed
    - When sorting Oldest to Newest displays the newest bookmark first, and newest to oldest shows the oldest bookmark first. (https://github.com/Automattic/pocket-casts-android/pull/1321)
- [x] Store the sort option on a "per list" basis where each bookmark lists option is saved separately (https://github.com/Automattic/pocket-casts-android/pull/1321)

### Player Bookmarks 
- [x] Select All Above/Below is not selecting any bookmarks (https://github.com/Automattic/pocket-casts-android/pull/1321/commits/e48dacf2ad3fcc660ac800472771597984012918)
- [x] Select All only selects the last item (https://github.com/Automattic/pocket-casts-android/pull/1321/commits/e48dacf2ad3fcc660ac800472771597984012918)

### Podcast Bookmarks
- [x] Display the episode title in the bookmark row (https://github.com/Automattic/pocket-casts-android/pull/1311)
- [x] When sorting by Episode, the bookmarks are sorted by oldest episode first, but should be sorted by newest. (https://github.com/Automattic/pocket-casts-android/pull/1311)
- [x] When searching (https://github.com/Automattic/pocket-casts-android/pull/1311)
    - [x] If there are no search results while typing the search field is replaced with the 'No Bookmarks yet' view
[...]

PR #1311, opened 2023-08-25
Bookmarks misc - part 1

ashiagr:
## Description

This includes miscellaneous fixes for bookmarks (https://github.com/Automattic/pocket-casts-android/issues/1307):

### Player
- Rename the Show Notes tab to 'Details'

### Upgrade View
- Show the Pocket Casts logo image instead of text

   (Body text not yet available, so ignore it for now)

Dark Logo | Light Logo
---|----
|

### Empty State
- The text color of the title appears too light

Dark Theme | Light Theme
---|----
 |

### Podcast Bookmarks
- Display the episode title in the bookmark row

- When sorting by Episode, the bookmarks are sorted by oldest episode first, but should be sorted by newest.

https://github.com/Automattic/pocket-casts-android/assets/1405144/9c15189e-138b-49af-9b1a-ba2865cfaa06

- When searching
    - If there are no results we should display the empty state

       This reuses the no podcast episodes found view and replaces "episodes" with "bookmarks". 

### Bookmarks List

-  The oldest/newest sort options are reversed
    - When sorting Oldest to Newest displays the newest episode first, and newest to oldest shows the oldest episode first.
    
        I'm not sure if "Oldest to Newest" and  "Newest to Oldest" on podcast details screen bookmarks should show bookmarks by episodes or by bookmark creation date only. It seems like only sorting by "Episode" should consider episode published dates. I'll confirm this and update it in the next PR.

## Checklist
 
- [ ] with different themes


CookieyedCodes:
@ashiagr the episode sorting option to me should go off of the episode lists sort order so technically it depends,  (I would also recommend a global sort option in settings, tho in my opinion also have an episode reversed option too because personally I would like bookmarks to be episode inversed as I would like a list of bookmarks based on release order but my lists are not necessarily allways sorted oldest to newest since I had to reinstall 😅)

where's oldest to newest/newest to oldest is made when

mchowning:
Unrelated to these changes, but just a heads up that one slightly odd thing I noticed was that switching themes seems to reset the selected tab away from bookmarks. Not a big deal at all, and probably not worth us taking the time to fix at this point, but I wanted to mention it.

https://github.com/Automattic/pocket-casts-android/assets/4656348/e3c165ec-a870-46e0-ab36-49a316a7c4b7

</details>

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/bookmark/components/UpsellView.kt
```diff
@@ -26,17 +30,44 @@ import au.com.shiftyjelly.pocketcasts.ui.R as UR
...
+                modifier = Modifier
+                    .clearAndSetSemantics { contentDescription = description }
             ) {
-                TextH20(text = stringResource(LR.string.pocket_casts))
+                Image(
+                    painter = painterResource(IR.drawable.logo_pocket_casts),
+                    contentDescription = null,
+                    colorFilter = ColorFilter.tint(logoColor),
+                )
                 Spacer(modifier = Modifier.width(8.dp))
```
mchowning, line 71: Just wanted to confirm that we don't want the Pocket Casts icon here. If we do want that included, we could reuse the `HorizontalLogo` composable.

Even if we don't want to reuse that, it might be worth making the `Image` part of this a reusable component in the [same folder as the other logo components we have](https://github.com/Automattic/pocket-casts-android/tree/main/modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/images). That's certainly not a must-have though.

Eventually, we should probably add a "Logo" section to Showkase.

mchowning (approved):
Looks good. I just left one small question.

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/bookmark/components/UpsellView.kt
ashiagr: We don't want the icon with the logo from what I see in Figma. 

> it might be worth making the Image part of this a reusable component in the [same folder as the other logo components we have](https://github.com/Automattic/pocket-casts-android/tree/main/modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/images). That's certainly not a must-have though.

Yep, I plan to add it to the showcase too.

Issue #157, opened 2022-07-21
Playback skipping around randomly


mchowning:
Original report from 5359958-zen:

> My app is sometimes skipping around different parts of the episode I am listening to randomly. This has been happening for a couple of days. My phone is in my pocket listening on a Bluetooth headset and the episode will just all the suddenly jump to a different time stamp. Phone is still locked with screen off in pocket, I have ruled out accidental touches.

### Update
The most common cause of this is what was suggested in https://github.com/Automattic/pocket-casts-android/issues/157#issuecomment-1251847030, basically that dynamic ads within an episode are changing as an episode is being streamed and that's causing the jumps. If this is true then that means:
1. **Downloading the episode avoids the issue** (let us know if you have the issue with a downloaded episode) and
2. The issue is more likely to occur if you take a break in listening to the episode (if you take a 2 week break in the middle of listening to an episode, there is a much higher chance that a dynamic ad will change compared to if you listen to an episode straight through).
3. The issue is more likely to occur the longer an episode is and the further into the episode you are

martinmcwhorter:
I am having the same issue. The ads theory makes sense to me. This seems to happen more since local ads based on geoip have started being spliced into podcasts. 

Have other podcast apps had issues since the length of a podcast became dynamic?

mdrockwell:
5782439-zd-woothemes:

> Whenever I go back to a podcast, it forgets where I was and puts me back like 15 or 20 minutes. I have a galaxy s22 with all the latest updates.

They've already tried disabling intelligent playback resumption and "It happens all the time. All podcasts all the time."

(Debug logs in the ticket.)

App version : 7.27.1 (9055)
Phone: samsung - b0q - SM-S908U - b0q
Kernel: 5.10.81-android12-9-25490797-abS908USQU2BVKD - 21DK7B12
Android version: 13 SDK 33

thabotswana:
Another report in 5830791-zen

>When I open pocket casts, whatever this app is still in the background or has been killed from background, the current episode will keep rewind 8 times, just like im touch the rewind button. But that's not because of touch by mistake because there is no touch behavior when I open the “show touch on screen” on system settings.
>
>This weird thing happens every times when I open Pocket Casts and only happens on my Android Device.

App version : 7.28 (9059)
Phone: Xiaomi - MI 10 - Mi 10 - umi
Android version: 11 SDK 30

joashrajin:
Feedback from 5825326-zd-woothemes

[...]

PR #1312, opened 2023-08-25
Log and notify users when episode duration changes unexpectedly

mchowning:
## Description
We have many complaints about playback skipping to different positions in an episode unexpectedly (https://github.com/Automattic/pocket-casts-android/issues/157). I think that at least the majority of the times this is happening is because the podcast provider is changing the dynamic ads in the episode. If those ads are of a different length, then our playback position value will end up representing a different part of the episode.

I do not see any straightforward ways to address this, so this PR is attempting to mitigate the issue a bit by checking if the duration of a podcast episode has changed by more than 30 seconds and
1. Presenting a toast message to the user advising them that the duration of the podcast has changed. I figured that this at least gives the listener a clue that something has changed and, if they notice a jump in the playback position, the reason why.
2. Adding logs so that support can see this if the user reaches out to support 
3. Adding a tracks event so that we can get more specific data about when/how-often this issue is occurring. Let me know if you can think of any other event properties it would be good to include.

Adding the logs and tracks event feel like pretty clear wins to me, but I'm slightly less confident about presenting a notification to the user since I'm not sure how many users will see it and I wouldn't want a bug to cause it to get presented too often (the tracks will let us check how often its getting shown at least). 

I toyed with only showing the toast if the duration change was more than a minute instead of the 30 seconds that I arrived at in order to show it less often.  Let me know if you think requiring a minute (or longer) duration would be better than 30 seconds.

## Testing Instructions

In addition to the following test, it would be good to just generally smoke-test different playback scenarios to make sure the toast isn't getting shown when it shouldn't be shown.

1. Create a podcast where you can insert/remove dynamic ads.
2. Load an episode of the podcast you created and start playing it.
4. Make sure the episode is streaming and is not being downloaded.
5. Skip to somewhere at least 5 minutes into the podcast.
6. Pause the podcast
7. Swipe the Pocket Casts app away
8. Update the dynamic ads in the episode of the podcast that you were listening to so that the duration of episode changes by at least 30 seconds.
9. Clear the cache of Pocket Casts via the Android App settings options
10. Reopen the Pocket Casts app
11. Resume playing the podcast where you left off
12. Note that
    1. the duration of the podcast has changed,
    2. the audio jumps to a different place,
    3. there is a toast message presented notifying you that the duration of the episode has changed, and
    4. There is a `playback_episode_duration_changed` tracks event sent off with the appropriate values set for "duration_change", "duration", "notified_user", "is_user_file", "is_downloaded", "episode_uuid", and "podcast_uuid"

## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/4656348/354b334e-c176-4d52-87ee-20d61f0928c0

## Checklist

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
@@ -95,6 +95,7 @@ import java.util.concurrent.TimeUnit
 import javax.inject.Inject
 import javax.inject.Singleton
 import kotlin.coroutines.CoroutineContext
+import kotlin.math.abs
 import kotlin.math.min
 import au.com.shiftyjelly.pocketcasts.images.R as IR
 import au.com.shiftyjelly.pocketcasts.localization.R as LR
@@ -1288,6 +1289,15 @@ open class PlaybackManager @Inject constructor(
         val durationMs = player?.durationMs() ?: 0
         if (durationMs <= 0) {
             return
+        }
+
+        val durationDiffSeconds = (durationMs - episode.durationMs) / 1000
+        if (abs(durationDiffSeconds) > 30) {
+            LogBuffer.e(LogBuffer.TAG_PLAYBACK, "The total episode duration has changed significantly ($durationDiffSeconds seconds)")
+            launch(Dispatchers.Main) {
+                val message = application.getString(LR.string.episode_duration_change, durationDiffSeconds)
+                Toast.makeText(application, message, Toast.LENGTH_LONG).show()
+            }
         }
 
         withContext(Dispatchers.Main) {
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
@@ -1292,12 +1292,27 @@ open class PlaybackManager @Inject constructor(
         }
 
         val durationDiffSeconds = (durationMs - episode.durationMs) / 1000
-        if (abs(durationDiffSeconds) > 30) {
-            LogBuffer.e(LogBuffer.TAG_PLAYBACK, "The total episode duration has changed significantly ($durationDiffSeconds seconds)")
-            launch(Dispatchers.Main) {
-                val message = application.getString(LR.string.episode_duration_change, durationDiffSeconds)
-                Toast.makeText(application, message, Toast.LENGTH_LONG).show()
+        if (abs(durationDiffSeconds) > 0) {
+            val notifyUser = abs(durationDiffSeconds) > 30
+            if (notifyUser) {
+                LogBuffer.e(LogBuffer.TAG_PLAYBACK, "The total episode duration has changed significantly ($durationDiffSeconds seconds)")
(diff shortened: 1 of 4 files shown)


CookieyedCodes:
Hi @mchowning ya this issue is mostly related to dynamic ad insertion and has really only became a major issue in the past year(maby because of more ads because of the global ad market slump & maby more dynamic ads too). I regularly have to explain this to the people on reddit 😅

The issue only appears for users who stream episodes, episodes that are downloaded don't have this issue because dynamic ad insertion doesn't get incerted on the fly, I don't think a 30-45 second toast is a bad idea tho might be missed if people immediately pause their phone or unpause from headphones which are where most of these complaints are coming from, I do think their could be more elegant solutions though 🤔

1.Such as maby blocking dynamic ads being reincerted once an episode is in now playing (because I do think its cheeky to haveing multiple dynamic ad calls while an episode is playing, tho I get that might be difficult because your just calling on the stream-if people are hopping around I would still recommend the toast)

OR 
2.Required background downloading of any episode in now playing (which I thought it did this -or it did- but maby not anymore I suppose as long as it doesn't effect autoplay functionality & respects mobile data rules 🤔)
(I also think this would be a great justifier for an auto download top upnext NUM like the apple watch tho 😅-& if it could be tied into autoplay/only when people are on wifi even better ❤️❤️-tho I do get it might be difficult to count with auto download of pods/filters 🤔)

Just my 2 cents, let me know your thoughts, I do think this should be resolved tho😅

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
```diff
@@ -1300,13 +1325,7 @@ open class PlaybackManager @Inject constructor(
...
 
         val playerDurationSecs = durationMs.toDouble() / 1000.0
-
-        val currentDurationMs = episode.durationMs
-        if (currentDurationMs < 10000) {
-            episodeManager.updateDuration(episode, playerDurationSecs, true)
-        } else {
-            episodeManager.updateDuration(episode, playerDurationSecs, true)
-        }
+        episodeManager.updateDuration(episode, playerDurationSecs, true)
```
mchowning, line 1328: This was just a bit of cleanup I ran across and included in this PR since it was nearby. This is not intended to be a change in the actual behavior but instead is just simplifying the logic since we were doing the same thing in both branches of the if-statement.

ashiagr (approved):
Showing a non-obtrusive toast for a diff more than 30 secs makes sense to me as the issue is most noticeable within this limit of 30-60 secs.

Let's also keep a watch on the tracking event to make sure that the toast is not shown too often.

CookieyedCodes:
Like if you know how much time the episode has been disrupted by their might be a way to calculate the change and rewind it so it feels more like intelligent playback resumption but that obviously is difficult to do too hence my other proposals but maby something to think about 😉

CookieyedCodes:
You could probably tell it not to do it for video episodes or files over a certain size then 🤔, I haven't heard of dynamic ad insertion tech being put on video podcasts yet anyways (I have heard of some tv companies doing it tho I find it more manipulative then audio dynamic ad tech) tho the market for video podcasts is tiny in my view tho I'm one of the few that actually uses video podcasts 😅, to prevent miss downloads you could maby have it on a delay 🤔 (tho I wouldn't mind some resolution restrictions settings on video podcasts regardless to cut down on storage  size if that's possible 😅)

Fair enough with the second point if only it was easy to calculate positive & negative numbers 😅, I do think people definitely notice tho

Issue #1314, opened 2023-08-28
Multiselect causing ANRs


mchowning:
### Description

With the 7.45 release, we've seen a significant increase in ANRs. They mostly seem related to multi-selection.

For example:
* https://a8c.sentry.io/issues/4127312862/events/38682e5581664ac0acc624aecd4c1f96/
* https://a8c.sentry.io/issues/4379367564/events/68534d8466fd4dada916a25d0daec7d7/
* https://a8c.sentry.io/issues/4379479942/events/49d29151c5bc45fea09bd02f990b5778/
* https://a8c.sentry.io/issues/4379054706/events/9813858b8d104c8c969da31d1b511a64/
* https://a8c.sentry.io/issues/4411476095/events/3ee1b7da704745768c7753a6a68b6b30/
* https://a8c.sentry.io/issues/4412473142/events/0a8a909811cf453ba28e117850a9a4af/

### Step-by-step reproduction instructions

1. Make sure you have a bunch of episodes in your Up Next queue
2. Go to a filter with a lot of episodes
3. Enter multiselection
4. Select all
5. Tap the miniplayer and then swipe up to get your Up Next queue (don't close the filters screen)
6. Enter multiselect mode
7. Select all
8. Select None
9. Observe the app locks up for a while

### Device, Operating system, and Pocket Casts app version

Samsung A13, Android 13, Pocket Casts 7.45

mchowning:
Closing this because the ANRs were addressed by https://github.com/Automattic/pocket-casts-android/issues/1378 and the global state has been removed by https://github.com/Automattic/pocket-casts-android/pull/1579 and https://github.com/Automattic/pocket-casts-android/pull/1580.

PR #1315, opened 2023-08-28
Avoid multiselect ANRs

mchowning:
## Description
With the 7.45 release we've seen a significant increase in ANRs: https://github.com/Automattic/pocket-casts-android/issues/1314. This PR avoids the ANRs by addressing part of the underlying issue. 

Note that this PR is targeting a new `7.45.1` hotfix branch because I think this should be a hotfix since it is causing thousands of ANRs a day. For that reason, I am keeping this PR small and lean (basically the simplest changes that address the ANRs), and I will open a separate PR that builds on this for a betafix release (https://github.com/Automattic/pocket-casts-android/pull/1316).

This fix is basically just to make sure that we avoid iterating through the multiselect items when we select or deselect all of the multiselect items.

## Testing Instructions
1. Make sure you have a bunch of episodes in your Up Next queue
2. Go to a filter with a lot of episodes
3. Enter multiselection
4. Select all
5. Tap the miniplayer and then swipe up to get your Up Next queue (don't close the filters screen)
6. Enter multiselect mode
7. Select all
8. Select None
9. Observe the app locks up for a while

| Before | After |
| --- | --- |
| This video just shows a freeze, not a full ANR, but I've seen these steps result in ANRs on my device multiple times
<video src="https://github.com/Automattic/pocket-casts-android/assets/4656348/4381d489-6e53-47d1-86ff-9b2c692a2acc"/> | <video src="https://github.com/Automattic/pocket-casts-android/assets/4656348/6e1d1ec4-ef85-4017-82fc-76472469fd69"/> |

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 7.46
 -----
 
+7.45.1
+-----
+*   Bug Fixes:
+    *   Fixed issue where that could cause app freeze when using multiselect
+        ([#1315](https://github.com/Automattic/pocket-casts-android/pull/1315))
 
 7.45
 -----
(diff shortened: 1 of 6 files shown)


ashiagr (approved):
This is much improved. Thanks for taking a look and fixing it. 🙏

Issue #98, opened 2022-07-12
Multi-Select not always clearing


mchowning:
Sometimes when switching between different episode lists, the multi-select state persists. For example, when going from the Up Next list to the Downloads list.

### Steps to reproduce
1. Open Up Next, turn on multiselect, and select a few downloaded episodes
2. Swipe down until you return to the home screen
3. Tap on the Profile Tab
4. Tap on Downloads
5. Observe that the same items are already selected

https://user-images.githubusercontent.com/4656348/178579266-81bee586-1da4-4f64-92af-aa6d0fa0b87b.mp4

mchowning:
I've put together some test scenarios with the behavior I think we want. The 🔴 indicate steps that are currently failing in my tests.

## 1. Podcast screen: switching between episodes and bookmarks

1. Go to the podcast screen for a podcast with some bookmarks
2. Start multiselecting an episode
3. Observe that the multiselect toolbar is showing
4. Tap on the bookmarks tab
5. Observe that the multiselection toolbar went away
6. Tap on the episodes tab
7. Observe that multiselection is not active even though it was active when you last left episodes tab
8. Tap on the bookmarks tab
9. Start multiselecting a bookmark
10. Observe that the multiselect toolbar is showing
11. Tap on the episodes tab
12. Observe that the multiselection toolbar went away
13. Tap on the bookmarks tab
14. Observe that multiselection is not active even though it was active when you last left the bookmarks tab

## 2.  Podcast Bookmarks → Full Screen Player Bookmarks

1. Begin playing an episode from that podcast that has some bookmarks
2. Move the full-screen player to the bookmarks tab
3. Go to the podcast screen for that episode
4. Tap on the bookmarks tab
5. Enter multi-select mode
6. Open the full screen (which should be on the bookmarks tab already)
7. 🔴 Observe that multiselect mode has been cleared and the full screen player is not in multiselect mode

## 3. Full Screen Player Bookmarks → Podcast Bookmarks

1. Begin playing an episode from that podcast that has some bookmarks
2. Go to the podcast screen for that episode and tap on the bookmarks tab
3. Open the full-screen player and go to the bookmarks tab
4. Enter multiselect mode
5. Swipe down to dismiss the full screen player
6. Observe that the podcast's bookmarks tab is not in multiselect mode


## 4. Podcast Screen → Up Next

1. Go to the podcast screen for a podcast
2. Add some episodes from that podcast to the up next queue
[...]

PR #1316, opened 2023-08-28
Avoid sharing multiselect state

mchowning:
## Description
This PR builds on the proposed `7.45.1` hotfix (https://github.com/Automattic/pocket-casts-android/pull/1315), to address the multiselect ANRs we're seeing in the app (https://github.com/Automattic/pocket-casts-android/issues/1314).

~In addition to fixing those ANRs, this PR also avoids us sharing state between screens such that multiselecting on one screen effects the multiselect state on other screens. The change that accomplishes this is to not have the multi-select helpers be singletons. Let me know if you think there is a reason we do need these to be singletons because this does feel like a slightly risky change. Or, if you think we should just merge this to `main` instead of having it be a betafix, let me know. I went with the betafix since it addresses some multiselect bugs that seemed pretty rough, but I'm open to just merging the singleton change to `main`.~

UPDATE: This PR is now targeting `main` (instead of `release/7.46`) and is limited to avoiding us sharing state between screens such that multiselecting on one screen effects the multiselect state on other screens. The change that accomplishes this is to not have the multi-select helpers be singletons. 

Fixes https://github.com/Automattic/pocket-casts-android/issues/1314
Fixes https://github.com/Automattic/pocket-casts-android/issues/98

### 1. Avoids ANR

Verify that the ANR fixed by https://github.com/Automattic/pocket-casts-android/pull/1315 is still fixed

1. Make sure you have a bunch of episodes in your Up Next queue
2. Go to a filter with a lot of episodes
3. Enter multiselection
4. Select all
5. Tap the miniplayer and then swipe up to get your Up Next queue (don't close the filters screen)
6. Enter multiselect mode
7. Select all
8. Select None
9. Observe the app locks up for a while

<details><summary>Before/After videos</summary>

| Before | After |
| --- | --- |
| This video just shows a freeze, not a full ANR, but I've seen these steps result in ANRs on my device multiple times. Also, this before video is showing the behavior without the hotfix from #1315.
<video src="https://github.com/Automattic/pocket-casts-android/assets/4656348/4381d489-6e53-47d1-86ff-9b2c692a2acc"/> | <video src="https://github.com/Automattic/pocket-casts-android/assets/4656348/6e1d1ec4-ef85-4017-82fc-76472469fd69"/> |

</details>

### 2. Avoids sharing multiselect state

1. Add an episode from a list (podcast, filter, starred, etc.) to the top of your up next queue
2. Enter multiselect and select that episode
3. Tap on the miniplayer
4. Swipe up to your Up Next queue
5. Note that multiselect is not active (before this fix, multiselect would have been active with the same episode you selected on the previous screen also selected in your Up Next queue).

<details><summary>Before/After videos</summary>

| Before | After |
| --- | --- |
| <video src="https://github.com/Automattic/pocket-casts-android/assets/4656348/97a33647-d721-4e79-9f58-b2ee6e206764"/> | <video src="https://github.com/Automattic/pocket-casts-android/assets/4656348/2505bf80-8001-4fbb-a9fd-3ed4429945c7"/> |

</details>

### 3. Select all/none works even when multiselecting on different screens

1. Start multiselecting on a list of episodes (like a Filter)
2. Switch to a list of episodes that has some overlap with the list on the first screen, but is somewhat different (i.e., go to a filter, add some of the episodes to your up next queue using multiselect, then go to the up next queue).
3. Note that select all and select none behave as expected (before this fix, it seems like the select all and select none buttons would often update the multiselect options on the previous screen instead of the current screen).

<details><summary>Before/After videos</summary>

| Before | After |
| --- | --- |
| <video src="https://github.com/Automattic/pocket-casts-android/assets/4656348/d8873e2b-2b62-498d-8542-b29762780fc1"/> | <video src="https://github.com/Automattic/pocket-casts-android/assets/4656348/12fe77ba-f535-4c03-90a7-f3f57ed77527"/> |

</details>

## Checklist


CookieyedCodes:
So I'm kind of weird, I've gotten really used to these multiselection bugs, I do agree most of them should be fixed but try to keep one specific one as I find it extremely helpful, that one spefically is the upnext multiselect with downloads in the background as this allows me to quickly distinguish what podcasts are not in upnext (now an upnext filter pill -ie with an "not in upnext"- or a quick toggle at the top of downloads to show/hide downloads in upnext would achieve the same effect-it can appear in the non multiselected topbar in my view & the icon would be the upnext icon with an open/closed eye in the corner)

Now if this dose improve performance that much & you do decide to get rid of the visual bugs may I also suggest an invert selection action because that would also be helpful  🤔 😉

Truthfully I've only really noticed lag 7.45 before then their was no lag on android 12 for me 🤔 (besides a tiny bit with batch deselect but I kind of expect that 😅-which probably should exit multiselect mode regardless) I do wonder what happened between 7.44 & 7.45 tho I assume it's related to that crash in RC 1 🤔-but what caused that I don't think code was touched 😅 (whichever choice don't forget the inverted multiselect actions those are what makes pocketcasts powerful)

ashiagr:
👋 @mchowning 

I just cherry-picked https://github.com/Automattic/pocket-casts-android/pull/1316/commits/7f867ad81f2aec863a09ab7ac83a758d8070eca4 and https://github.com/Automattic/pocket-casts-android/pull/1316/commits/d3a97488e7097a65af02716a2e3cbe38627ac0b3 into https://github.com/Automattic/pocket-casts-android/pull/1317. I think https://github.com/Automattic/pocket-casts-android/commit/ab47dc1e0fdc35cc6239605eab054a8c7cacc526 requires more testing. Let's target `main` for that change.

mchowning:
> I think https://github.com/Automattic/pocket-casts-android/commit/ab47dc1e0fdc35cc6239605eab054a8c7cacc526 requires more testing. Let's target main for that change.

Yeah, that's probably a good call. I've updated this PR to target `main` with just that change.

mchowning:
Marking this as a draft for now until I get a chance to look at how this works with bookmarks.

ashiagr:
I pushed further changes to ensure that separate multi-select helper instances are used on screens with multi-select support.

I also cherry-picked two fixes that I thought were important  for proper testing: 

https://github.com/Automattic/pocket-casts-android/pull/1316/commits/d076950515280abe027202f94cc54bd4066955de
https://github.com/Automattic/pocket-casts-android/pull/1316/commits/b7f04f4ea01be3801506b96cf02e271029f1f337

## Testing Instructions: 
#### Test 1 Multi-select regression

1. Open screen with multi-select support and having sufficient multi-selectable (episodes or bookmarks)
2. Enter into multi-select mode
3. Perform multi-selection actions on the screen
4. ✅ Notice that multi-selection actions work without issues

#### Test 2 Back navigation for bottom sheet dialogs

1. Open bottom sheet dialogs with multi-select support
2. Enter into multi-select mode
3. Perform back action using a back gesture or device back button
4. ✅ Notice that multi-select toolbar is hidden
5. Perform back action again
6. ✅ Notice that the  bottom sheet dialog dismisses

Here's the list of screens I tested:

Fragment with MultiSelect support | For Episodes | For Bookmarks | Screen | Is Bottom Sheet Dialog
-- | -- | -- | -- | --
  |   |   |   |  
UpNextFragment | ✅ |   | Up Next | Yes
FilterEpisodeListFragment | ✅ |   | Episodes in Filters tab |  
ProfileEpisodeListFragment | ✅ |   | Listening History, Starred, Downloads |  
PodcastFragment | ✅ | ✅ | Podcast Details |  
CloudFilesFragment | ✅ |   | Files |  
BookmarksContainerFragment |   | ✅ | Bookmarks from Files bottom sheet | Yes
EpisodeContainerFragment |   | ✅ | Episode Details | Yes
PlayerContainerFragment |   | ✅ | Full Screen Player |

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/podcast/PodcastFragment.kt
```diff
@@ -656,7 +658,12 @@ class PodcastFragment : BaseFragment(), Toolbar.OnMenuItemClickListener, Corouti
     }
 
     private fun FragmentPodcastBinding.setupMultiSelect() {
+        viewModel.multiSelectEpisodesHelper = multiSelectEpisodesHelper
+        viewModel.multiSelectBookmarksHelper = multiSelectBookmarksHelper
+        sharedBookmarksViewModel.multiSelectHelper = multiSelectBookmarksHelper
```
mchowning, line 663: When resolving merge conflicts, I had to determine where to put these assignments. I'm pretty confident this is the right place, but it's worth taking a close look here to double-check.

ashiagr: This looks good to me 👍

mchowning:
This PR is out of date and the changes in it actually cause some regresssions on multi-select issues that we've fixed, so I'll close it, and work some of the improvements from this PR into a new PR.

PR #1324, opened 2023-08-31
Fix refresh interval

ashiagr:
## Description

This reverts refresh interval to original values (temporarily changed in #1290)


mchowning (approved):
Thanks for catching this! 👍

PR #1325, opened 2023-08-31
Temporary fix for notification permission checker

ashiagr:
## Description

This includes a temporary fix for https://a8c.sentry.io/share/issue/a00791f201394715be1cdffda7814fda/ 

> **Warning**
> Targets release/7.46

## Testing Instructions

I have not been able to reproduce the issue. I have converted `notificationPermissionChecker` to be nullable, and attempted `notify(...)` even when `notificationPermissionChecker`  is null to bring back the original implementation. I have also added logs to `Sentry` and `LogBuffer` if it still fails.

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
@@ -1726,7 +1726,9 @@ open class PlaybackManager @Inject constructor(
                     manager.notify(notificationTag, NotificationBroadcastReceiver.NOTIFICATION_ID, notification)
                 }
             } ?: run {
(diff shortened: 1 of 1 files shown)


mchowning (approved):
This looks good. I made a small tweak so I'll approve but hold off on merging until you can take a look.

The change I made was just to make it clearer that the `?:` is responding to the `notificationPermissionChecker` being `null` and not to the `checkNotificationPermission` function returning null (what you had was totally fine, I just found this easier to read). Then I added a log message and a Sentry breadcrumb if the `notificationPermissionChecker` is `null`. Since that means we're in some unexpected state I thought the extra logging would be nice. You can see what the breadcrumb looks like [here](https://a8c.sentry.io/issues/4446654880/?project=6711064&query=is%3Aunresolved&referrer=issue-stream&stream_index=2).

image

_* ignore the `RuntimeException: breadcrumb test exception` line, I just hardcoded that exception to see what the breadcrumb (shown on the next line) would look like in Sentry.

ashiagr:
Thanks for the extra logging, it looks fine to me. 👍

PR #1327, opened 2023-08-31
Use transaction when archiving multiple episodes

mchowning:
## Description
This avoids ANRs when archiving multiple episodes by wrapping all the db calls into a single transaction.

> **Note**
> Note that this change is targeting the beta branch. Let me know if you think that's not the right call this late in the beta period.

There are a lot of other places where we should consider creating transactions like this, but this is the most significant performance issue I'm seeing so far (by a wide margin), so I kept this PR small (and hopefully safe).

## Testing Instructions
1. Fresh install of the app
2. Open the Tim Ferriss Show podcast page
3. Tap the three dots
4. Tap Archive All and confirm
5. Observe that the UI locks for <1 second (without this change it would lock the UI on my device for ~30 seconds)

## Checklist

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/EpisodeManagerImpl.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/EpisodeManagerImpl.kt
@@ -28,6 +28,7 @@ import au.com.shiftyjelly.pocketcasts.models.type.EpisodeStatusEnum
 import au.com.shiftyjelly.pocketcasts.models.type.EpisodesSortType
 import au.com.shiftyjelly.pocketcasts.models.type.UserEpisodeServerStatus
 import au.com.shiftyjelly.pocketcasts.preferences.Settings
+import au.com.shiftyjelly.pocketcasts.repositories.di.IoDispatcher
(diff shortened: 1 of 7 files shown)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/EpisodeManagerImpl.kt
```diff
@@ -954,12 +955,19 @@ class EpisodeManagerImpl @Inject constructor(
...
-            playbackManager?.let { playbackManager ->
-                chunked.forEach {
-                    cleanUpEpisode(it, playbackManager)
+    override fun archiveAllInList(
+        episodes: List<PodcastEpisode>,
+        playbackManager: PlaybackManager?
+    ) {
+        launch(Dispatchers.IO) {
+            appDatabase.withTransaction {
+                episodes.filter { !it.isArchived }.chunked(500).forEach { chunked ->
```
mchowning, line 964: With this all wrapped in a transaction, I don't think there's any need to keep chunking this list. In my testing, there was no noticeable difference with and without the chunks. I left them in for now to minimize the amount of change in this PR since it is targeting the beta branch right before release.

ashiagr (approved):
This is working really well! 🏅

PR #1339, opened 2023-09-04
Fix docsbot not loading

ashiagr:
## Description
This fixes docsbot not loading issue
p1693818175061479-slack-C02A333D8LQ

## Testing Instructions
1. Launch the app 
2. Go to `Profile` -> `Help & feedback`
3. Tap on Help button (in red) at the bottom right corner
4. Notice that the docsbot loads

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/features/settings/src/main/AndroidManifest.xml
+++ b/modules/features/settings/src/main/AndroidManifest.xml
@@ -1,5 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
-
-
+    <application>
+        <activity
+            android:name=".HelpActivity"
+            android:launchMode="singleTask"
+            android:windowSoftInputMode="adjustResize"
+            android:configChanges="orientation|layoutDirection|screenSize" />
+    </application>
 </manifest>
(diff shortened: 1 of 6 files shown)


geekygecko (approved):
The bot now loads but it isn't possible to see what you type. I wonder if we should open the support page URL in a browser or open a separate activity that stops the keyboard from being over the content.

Screenshot_20230905_090208

ashiagr:
This is fixed in https://github.com/Automattic/pocket-casts-android/pull/1339/commits/904458326329ac7af60436aa817615ffa2c99c47

I opened a separate activity with the ability to resize the screen if the keyboard is present and updated the menu fragments to open an close from this activity. SInce `LogsFragments` is also used on `Automotive` app, it is closed based on the activity type.

https://github.com/Automattic/pocket-casts-android/assets/1405144/6f4b2f78-b4ab-42a7-a262-edc5b5b5dd05


https://github.com/Automattic/pocket-casts-android/assets/1405144/710c8860-3b60-417f-a76e-52f4b82829b1

geekygecko:
Great work fixing entering the text!

geekygecko (approved):

PR #1342, opened 2023-09-05
Update cloud storage limit title

ashiagr:
## Description

This updates cloud storage string title from 10GB to 20GB.

> This targets v7.47 beta branch.

## Testing Instructions
It should be sufficient to just check the strings file. 

## Screenshots or Screencast 

Old| New
---|---
|

## Checklist
 
- [ ] with different themes


mchowning (approved):
Looks good. I'll hold off on cutting a new beta since this is a relatively trivial change. Let me know if you think we should go ahead and cut a new beta though.

PR #1345, opened 2023-09-05
Prefer suspended method for finding podcast by uuid

mchowning:
## Description
This PR makes the suspended function for finding podcasts by id the preferred method by deprecating the other methods and removing the usages of the other methods where it is easy to do so. I did not remove all usages of the other methods because I wanted to keep this PR reviewable and reduce the risk of regressions. My thinking is that we'll be better off trying to remove the big RxJava chains when we are actually making changes to them because we'll be more familiar with what they're doing then and there will already be a risk of regression at that point since we'll be making changes.

I'm planning to make some more changes similar to this, so let me know if you think any other approach to this makes more sense.

## Testing Instructions
Smoke test interacting with a few podcasts in different ways (playing episodes, queueing episodes, filters, subscribing to podcasts, etc.).

## Checklist

### DIFF
--- a/modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/viewmodel/PodcastListViewModel.kt
+++ b/modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/viewmodel/PodcastListViewModel.kt
@@ -23,6 +23,7 @@ import io.reactivex.rxkotlin.addTo
 import io.reactivex.rxkotlin.combineLatest
 import io.reactivex.rxkotlin.subscribeBy
 import io.reactivex.schedulers.Schedulers
+import kotlinx.coroutines.rx2.rxMaybe
 import timber.log.Timber
 import javax.inject.Inject
 
@@ -135,8 +136,9 @@ class PodcastListViewModel @Inject constructor(
     fun findOrDownloadEpisode(discoverEpisode: DiscoverEpisode, success: (episode: PodcastEpisode) -> Unit) {
         podcastManager.findOrDownloadPodcastRx(discoverEpisode.podcast_uuid)
             .flatMapMaybe {
-                @Suppress("DEPRECATION")
-                episodeManager.findByUuidRx(discoverEpisode.uuid)
+                rxMaybe {
+                    episodeManager.findByUuid(discoverEpisode.uuid)
+                }
             }
             .subscribeOn(Schedulers.io())
             .observeOn(AndroidSchedulers.mainThread())
--- a/modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/viewmodel/PodcastListViewModel.kt
+++ b/modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/viewmodel/PodcastListViewModel.kt
@@ -134,7 +134,10 @@ class PodcastListViewModel @Inject constructor(
 
     fun findOrDownloadEpisode(discoverEpisode: DiscoverEpisode, success: (episode: PodcastEpisode) -> Unit) {
         podcastManager.findOrDownloadPodcastRx(discoverEpisode.podcast_uuid)
-            .flatMapMaybe { episodeManager.findByUuidRx(discoverEpisode.uuid) }
+            .flatMapMaybe {
+                @Suppress("DEPRECATION")
+                episodeManager.findByUuidRx(discoverEpisode.uuid)
+            }
             .subscribeOn(Schedulers.io())
             .observeOn(AndroidSchedulers.mainThread())
             .subscribeBy(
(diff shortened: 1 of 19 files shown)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
```diff
@@ -1462,20 +1462,22 @@ open class PlaybackManager @Inject constructor(
...
+                        }
                     }
-                }
-                .awaitSingleOrNull()
-        } else {
-            null
+                    .awaitSingleOrNull()
+            }
+
+            else -> null
```
mchowning, line 1480: Because `loadCurrentEpisode` was already a suspended function, the only change here was to accept the linter's suggestion of converting this `if`-`else if`-`else` into a `when` statement.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/file/FileStorage.java
```diff
@@ -322,7 +322,8 @@ public void moveStorage(File oldDirectory, File newDirectory, PodcastManager pod
 							if (fileName.length() < 36) {
 								continue;
 							}
-							PodcastEpisode episode = episodeManager.findByUuid(fileName);
+							//noinspection deprecation
+							PodcastEpisode episode = episodeManager.findByUuidSync(fileName);
```
mchowning, line 326: Leaving the synchronous calls in this class since using coroutines in java at our API level is not trivial. It's certainly doable though, so let me know if you'd rather us get rid of these.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/EpisodeManagerImpl.kt
```diff
@@ -89,31 +89,31 @@ class EpisodeManagerImpl @Inject constructor(
...
         }
 
         return userEpisodeManager.findEpisodeByUuid(uuid)
     }
 
-    override fun findByUuid(uuid: String): PodcastEpisode? {
-        return episodeDao.findByUuid(uuid)
-    }
+    override suspend fun findByUuid(uuid: String): PodcastEpisode? =
+        episodeDao.findByUuid(uuid)
```
mchowning, line 101: I changed it so that `findByUuid` is now the suspended function because I felt like it makes sense to make the preferred, default function the one with the shortest name. wdyt?

modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/viewmodel/PodcastListViewModel.kt
```diff
@@ -134,7 +134,10 @@ class PodcastListViewModel @Inject constructor(
 
     fun findOrDownloadEpisode(discoverEpisode: DiscoverEpisode, success: (episode: PodcastEpisode) -> Unit) {
         podcastManager.findOrDownloadPodcastRx(discoverEpisode.podcast_uuid)
-            .flatMapMaybe { episodeManager.findByUuidRx(discoverEpisode.uuid) }
+            .flatMapMaybe {
+                @Suppress("DEPRECATION")
+                episodeManager.findByUuidRx(discoverEpisode.uuid)
+            }
```
geekygecko, line 140: I like the way you have added `@Deprecated` to the method. 
Is it a step too far to change the code to something like the following:
```suggestion
            .flatMapMaybe {
                rxMaybe {
                    episodeManager.findByUuid(discoverEpisode.uuid)
                }
            }
```

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/download/UpdateEpisodeDetailsTask.kt
```diff
@@ -45,7 +46,9 @@ class UpdateEpisodeDetailsTask @AssistedInject constructor(
             Timber.i("Downloading Meta Data for ${episodeUuids.size} episodes")
 
             for (episodeUuid in episodeUuids) {
-                val episode = episodeManager.findByUuid(episodeUuid) ?: continue
+                val episode = runBlocking {
+                    episodeManager.findByUuid(episodeUuid)
+                } ?: continue
```
geekygecko, line 51:
```suggestion
                val episode = episodeManager.findByUuid(episodeUuid) ?: continue
```
The `doWork` method is a suspend function so this might not need a runBlocking.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/EpisodeManagerImpl.kt
geekygecko: I like this change too.

geekygecko (approved):
Thanks for working on converting to Coroutines.

modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/viewmodel/PodcastListViewModel.kt
mchowning: `rxMaybe` works great there. Thanks!

PR #1348, opened 2023-09-05
Make star method a suspended function

mchowning:
## Description
This updates the star methods to be suspended functions. This avoids us needing to launch new coroutines within the `EpisodeManager`.

As part of this, I moved the tracks event logic into the `EpisodeManagerImpl` to avoid duplicating in all the view models where we update an episode's starred status.

## Testing Instructions
Test starring/unstarring episodes
1. From the media notification
2. From the episode details screen
3. From multiselect on the podcast screen
4. From the now playing screen
5. From the watch app

## Checklist

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/EpisodeManagerImpl.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/EpisodeManagerImpl.kt
@@ -447,7 +447,6 @@ class EpisodeManagerImpl @Inject constructor(
     }
 
     override suspend fun starEpisode(episode: PodcastEpisode, starred: Boolean, sourceView: SourceView) {
-        episode.isStarred = starred
         episodeDao.updateStarred(starred, System.currentTimeMillis(), episode.uuid)
         val event =
             if (starred) AnalyticsEvent.EPISODE_UNSTARRED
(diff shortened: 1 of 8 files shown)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/EpisodeManagerImpl.kt
```diff
@@ -445,9 +446,12 @@ class EpisodeManagerImpl @Inject constructor(
         }
     }
 
-    override fun starEpisode(episode: PodcastEpisode, starred: Boolean) {
-        episode.isStarred = starred
```
mchowning, line 449: I removed this call to `episode.isStarred` because this is already being done in the `MediaSessionManager`. The only other place that calls this method doesn't need to do this because they are just using an episode instance that is being fetched from the db, and not being used after that.

The `MediaSessionManager`, however, is using the episode instance from `PlaybackManager::getCurrentEpisode` which is retrieved from our `UpNextQueue` object. It appears that the `UpNextQueue` does not listen for changes to episodes in the queue, which I'm guessing is the reason we're manually updating it in the `MediaSessionManager` when we change the starred value.

Ideally, we would be listening for changes in the `UpNextQueue`, but until then I left the manual updates to the episode instance in `MediaSessionManager` to keep that object up-to-date. I didn't observe any adverse effects from removing that update in my testing, but it seems better to just keep the current behavior, which is why I only stopped setting the starred value a second time inside the `EpsiodeManagerImpl`. I went ahead and stopped setting it twice since ideally we shouldn't ever be setting it, so avoiding setting it a second time gets us one (very tiny) step closer to that.

geekygecko (approved):

PR #1349, opened 2023-09-06
Convert the chapters fragment to Compose

geekygecko:
## Description

This change converts the chapters fragment to Compose.

## Testing Instructions
1. Open the podcast Upgrade https://pca.st/upgrade
2. Play an episode
3. Tap on the mini player to open the full screen player
4. Tap on the Chapters tab
5. ✅ Verify the chapters are shown

## Screenshots or Screencast 

| Before Android | iOS | After Android |
| --- | --- | --- |
| Screenshot_20230906-135005 | IMG_5576AA3ADA93-1 | Screenshot_20230906_134911 |


modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/chapters/ChaptersViewModel.kt
```diff
@@ -0,0 +1,109 @@
...
+
+    data class UiState(
+        val chapters: List<ChapterState> = emptyList(),
+        val backgroundColor: Color
+    )
+
+    data class ChapterState(
+        val chapter: Chapter,
+        val isPlayed: Boolean,
+        val isPlaying: Boolean,
```
geekygecko, line 44: I did try using a playing status for this but it seemed to add more code so I ended up removing it. I'm not sure if that was the right decision. 🤔 
```
    sealed class ChapterPlayingStatus {
        object Played: ChapterPlayingStatus()
        data class Playing(val progress: Float): ChapterPlayingStatus()
        object NotPlayed: ChapterPlayingStatus()
    }
```

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/viewmodel/PlayerViewModel.kt
```diff
@@ -618,14 +615,24 @@ class PlayerViewModel @Inject constructor(
...
+            field = value
+        }
+
+    private var cacheScrollToChapter: Chapter? = null
+
+    fun scrollToChapter(chapter: Chapter) {
+        if (scrollToChapterListener == null) {
+            cacheScrollToChapter = chapter
+        } else {
+            scrollToChapterListener?.invoke(chapter)
```
geekygecko, line 635: I found it quite tricky to open the chapter fragment and scroll to the correct chapter. Let me know if there is a cleaner way to do this.

mchowning: This is tricky! The first few things I tried were complete failures, but I eventually came up [with this](https://github.com/Automattic/pocket-casts-android/commit/4b43f28171aff8a153982050217b985367ff5b22) as an alternative way to accomplish this. I think it's maybe a bit cleaner, but I'm not sure if there might be some edge cases it doesn't cover.

mchowning (approved):
This looks good and works well. Left a couple of code comments, but no blockers. Great work!

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/chapters/ChaptersViewModel.kt
geekygecko: I like how you changed the ChapterState instead I added a new ChapterPlayingStatus sealed class. I'm going to steal your diff. Thanks!

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/viewmodel/PlayerViewModel.kt
geekygecko: Wow, thanks for improving the code. That worked well when I tested this.

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/chapters/ChapterRow.kt
```diff
@@ -0,0 +1,170 @@
...
+                text = chapter.title,
+                color = textColor,
+                maxLines = 2,
+                modifier = Modifier
+                    .weight(1f)
+                    .padding(vertical = 16.dp)
+            )
+            Spacer(Modifier.width(4.dp))
+            if (chapter.url != null) {
+                LinkButton(
```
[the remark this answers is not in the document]
geekygecko: Thanks for the suggestion, I have improved this.

| Before | After |
| --- | --- |
| Screenshot_20230908_203847 | Screenshot_20230908_205110 |

geekygecko:
@CookieyedCodes thanks again for your ideas.

> Ability to skip/turn off spefic chapters so they get skip over certain chapters

I'm not sure about this feature as it would be handy but I wouldn't want people to be able to skip the podcasters adverts too easily as that's what normally pays for the great content.

> Reading RSS timetags to create chapters(takes chapter title, until next timetag -last one defaults to "wrap up?"-).

This is an awesome idea. Do you have some example podcasts you would use this feature with that I could try?
I'm currently trying to add Podlove and Podcast Index chapters from the RSS feed XML. Hopefully they will be available soon.

> Headphone settings for skip over chapters as your working on bookmarks, like iOS

It will be good to add this, we can't have iOS being better than Android. 😄

CookieyedCodes:
No bothers at all, again I can't help it tho, you guys are so close to being niggle free 😅 (still the best app tho),

Re 1. I actually have never seen any podcaster ever use chapters to skip over ads really only over intros and jumping to spefic segments. Also from my understanding of dynamic ad tech they only count the downloads/the number of times a stream has started for the ad rates users skipping over ads doesn't effect podcasters currently because of this(also hence why I say plus feature too a certain extent, just don't market it as ad free😅)
(& if a podcast dose do this and it dose effect them the podcaster will shake stuff up, I've noticed less podcasters doing front and end ads since you began doing skip last, tho thats probably because of them potentially being less effective anyways😅)

Re 2.
https://pca.st/podcast/51857050-3e51-0135-9028-63f4b61a9224 the last 12 episodes began useing RSS timetags I think
Also YouTube RSS feeds from listenbox tend to also convert YouTube description timemarks to RSS time tags if I remember rightly 😉( so it would be helpful their too ❤️❤️, again it works from the show notes page but I'd prefer a dedicated UI), I can get you more links if you need it :)


Re 3 the endless battle 😅

PR #1353, opened 2023-09-06
Remove toast message when episode duration diff changes

mchowning:
## Description
It looks like the episode duration diff toast might get shown more than I anticipated, so let's remove it for the time being.

> **Note**
> This PR is targeting `release/7.47`, but I don't think we need to rush out a new beta for this.

## Testing Instructions
1. Play "Monday Morning Podcast 9-4-23" from Bill Burr's Monday Morning Podcast
2. Observe that the logs contain a message about the episode duration changing when you start playing the episode
3. Observe that there is _not_ a toast message shown to the user

## Checklist

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
@@ -1294,20 +1294,12 @@ open class PlaybackManager @Inject constructor(
 
         val durationDiffSeconds = (durationMs - episode.durationMs) / 1000
         if (abs(durationDiffSeconds) > 0) {
-            val notifyUser = abs(durationDiffSeconds) > 30
-            if (notifyUser) {
-                LogBuffer.e(LogBuffer.TAG_PLAYBACK, "The total episode duration has changed significantly ($durationDiffSeconds seconds)")
-                launch(Dispatchers.Main) {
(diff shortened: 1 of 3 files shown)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
```diff
@@ -1294,20 +1294,12 @@ open class PlaybackManager @Inject constructor(
 
         val durationDiffSeconds = (durationMs - episode.durationMs) / 1000
         if (abs(durationDiffSeconds) > 0) {
-            val notifyUser = abs(durationDiffSeconds) > 30
-            if (notifyUser) {
-                LogBuffer.e(LogBuffer.TAG_PLAYBACK, "The total episode duration has changed significantly ($durationDiffSeconds seconds)")
-                launch(Dispatchers.Main) {
-                    val message = application.getString(LR.string.episode_duration_change, durationDiffSeconds)
```
geekygecko, line 1301: Should we remove the string `episode_duration_change` too?

geekygecko (approved):

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
mchowning: Ah, good catch! Removed in https://github.com/Automattic/pocket-casts-android/pull/1353/commits/47dc0234486e9a9d1205877a777d0aa7f2f4e54a.

PR #1355, opened 2023-09-06
Make markPlayed and Unplayed suspend functions

mchowning:
## Description
This reduces how often the episodeManager launches new coroutines.

### 1. markAllAsUnplayed
1. Use multiselect to select a bunch of played episodes
2. Observe that marking them all as unplayed by selecting the mark as unplayed option in the multiselect toolbar still works as expected.

### 2. markAsPlayed
Archive an episode from within the app:
* Now Playing bottom sheet
* Episode details screen

### 3. markAsPlayed from external source
1. Log into an account on your phone
2. Start playing an episode
3. Pause the episode
4. Sync the app
5. Log into the same account on the web
6. Finish playing the episode
7. On your phone, sync the app again
8. Verify that the episode you completed on the web is now archived

### 4. archivePlayedEpisode
1. Start playing an episode
2. Complete playing the episode
3. Verify that the episode is archived

## Checklist

### DIFF
--- a/modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/multiselect/MultiSelectEpisodesHelper.kt
+++ b/modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/multiselect/MultiSelectEpisodesHelper.kt
@@ -190,7 +190,7 @@ class MultiSelectEpisodesHelper @Inject constructor(
 
             episodeManager.markAsUnplayed(list)
             episodeAnalytics.trackBulkEvent(AnalyticsEvent.EPISODE_BULK_MARKED_AS_UNPLAYED, source, list.size)
-            launch(Dispatchers.Main) {
+            withContext(Dispatchers.Main) {
                 val snackText = resources.getStringPlural(selectedList.size, LR.string.marked_as_unplayed_singular, LR.string.marked_as_unplayed_plural)
                 showSnackBar(snackText)
                 closeMultiSelect()
--- a/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
+++ b/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
@@ -614,7 +614,8 @@ class MainActivity :
             episodeManager = episodeManager,
             fragmentManager = supportFragmentManager,
             analyticsTracker = analyticsTracker,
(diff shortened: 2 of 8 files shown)


modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/multiselect/MultiSelectEpisodesHelper.kt
```diff
@@ -190,7 +190,7 @@ class MultiSelectEpisodesHelper @Inject constructor(
 
             episodeManager.markAsUnplayed(list)
             episodeAnalytics.trackBulkEvent(AnalyticsEvent.EPISODE_BULK_MARKED_AS_UNPLAYED, source, list.size)
-            launch(Dispatchers.Main) {
+            withContext(Dispatchers.Main) {
```
mchowning, line 193: This change isn't really important at this point since we just launched a coroutine a few lines earlier, but the fewer times we launch new coroutines the fewer opportunities there are for race conditions, so I included this since it is a small step in the right direction.

geekygecko: Great change. I just read about `withContext` and it makes sense not to kick off another Coroutine.

Mzazi25:
@mchowning with this change, does that mean we can also change the `archivePlayedEpisode`  in the `episodeManager` since we can avoid launching a coroutine

mchowning:
> does that mean we can also change the archivePlayedEpisode in the episodeManager since we can avoid launching a coroutine?

That's right. The reason I moved this PR back to a draft was to make that and a couple of other updates.

mchowning:
While testing this PR, I had the app lock up one time with the following error showing repeatedly in the logs.

```
The connection pool for database '+data+user+0+au_com_shiftyjelly_pocketcasts_debug+databases+pocketcasts' has been unable to grant a connection to thread 52065 (arch_disk_io_0) with flags 0x2 for 96.025 seconds. Connections: 0 active, 1 idle, 3 available. 
```

I haven't been able to reproduce so I don't know if it is a result of the changes in this PR or not, but I marked this PR as a draft until there's some more testing of this PR (or we find this issue on the `main` branch).

mchowning:
Closing this for now since the PR is pretty outdated. We can make these changes in a new PR if we decide to move forward with them.

PR #1357, opened 2023-09-07
Watch app: Remove login by email and password screen 

ashiagr:
## Description
This removes login by email and password in the watch app to meet [WO-P6](https://developer.android.com/docs/quality-guidelines/wear-app-quality#:~:text=Identity,Wear%20OS%20device.) guideline.

> Your app must not ask the user to input a username or password directly on the Wear OS device.

The following login methods will continue to exist:
- Log in on phone (or pass tokens using the data layer)
- Google Sign-In

which complies with the [requirement](https://developer.android.com/training/wearables/apps/auth-wear#auth-methods:~:text=Your%20Wear%20app%20must,with%20an%20iOS%20device.) to have one other login method in addition to the pass token method.

## Testing Instructions

- Install watch app
- Tap "Log in"
- ✅  Notice that there is no "Log in with email" options in the login options screen

#### Regression

Make sure that "Log in with Google" and "Log in on phone" work as before.

## Checklist
- Ensure the linter passes (`./gradlew spotlessApply` to automatically apply formatting/linting)
 
- with different themes


mchowning (approved):
Thanks for taking care of this. I added a commit removing the `LoginWithEmailScreen` since it isn't being used anymore.

Tested and this works well. Code changes are also good. 👍

PR #1362, opened 2023-09-08
Add recommendations to Automotive

geekygecko:
## Description

This change adds the recommendations to Automotive. This feature is already available for Android Auto. I made some code changes and added a test to ensure it still worked correctly. There was a bug that you could have the same episode in the list from the Up Next queue and the top filter.

It's hard to test it from the Automotive emulator as I don't know how to trigger it. The code changes to the suggestion method can be tested using Android Auto in Google Maps.

## Testing Instructions
1. Deploy to a physical device
2. Start playing an episode
3. Add another episode to your Up Next
4. Open Google Maps
5. Setup navigation to a location
6. The Pocket Casts mini player should appear
7. Tap on the mini player in Google Maps
8. Tap "Browse Pocket Debug"
9. ✅ Verify your current episode is first, followed by your Up Next, and then episode from your top filter.

I changed some other methods over to using the same new Coroutine version, to test these do the following.
1. Open a filter
2. Tap on the title to expand the filter option
3. Tap on "Episode Status"
4. Change the setting
5. Tap "Update filter"
6. Tap on "Episode Status" again
7. ✅ Verify the options have persisted
8. Do the same for the filter setting "Duration" and "Download Status"

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackService.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackService.kt
@@ -19,12 +19,14 @@ import androidx.media.MediaBrowserServiceCompat
 import au.com.shiftyjelly.pocketcasts.analytics.FirebaseAnalyticsTracker
 import au.com.shiftyjelly.pocketcasts.localization.BuildConfig
 import au.com.shiftyjelly.pocketcasts.models.db.helper.UserEpisodePodcastSubstitute
+import au.com.shiftyjelly.pocketcasts.models.entity.BaseEpisode
 import au.com.shiftyjelly.pocketcasts.models.entity.Podcast
 import au.com.shiftyjelly.pocketcasts.models.entity.PodcastEpisode
 import au.com.shiftyjelly.pocketcasts.models.to.FolderItem
 import au.com.shiftyjelly.pocketcasts.models.to.SubscriptionStatus
 import au.com.shiftyjelly.pocketcasts.models.type.PodcastsSortType
 import au.com.shiftyjelly.pocketcasts.preferences.Settings
+import au.com.shiftyjelly.pocketcasts.preferences.Settings.MediaNotificationControls.Companion.items
 import au.com.shiftyjelly.pocketcasts.repositories.extensions.id
 import au.com.shiftyjelly.pocketcasts.repositories.notification.NotificationDrawer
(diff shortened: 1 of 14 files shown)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackService.kt
```diff
@@ -396,44 +402,57 @@ open class PlaybackService : MediaBrowserServiceCompat(), CoroutineScope {
...
 
-        Timber.d("Recent item ${upNext.title}")
-        return arrayListOf(AutoConverter.convertEpisodeToMediaItem(this, upNext, parentPodcast))
+    private suspend fun convertEpisodesToMediaItems(episodes: List<BaseEpisode>): List<MediaBrowserCompat.MediaItem> {
+        return episodes.mapNotNull { episode ->
+            // find the podcast
+            val podcast = if (episode is PodcastEpisode) {
+                podcastManager.findPodcastByUuidSuspend(episode.podcastUuid)
+            } else {
+                Podcast(uuid = UserEpisodePodcastSubstitute.substituteUuid, title = UserEpisodePodcastSubstitute.substituteTitle)
```
mchowning, line 451: Not a change for this PR, but this got me thinking of whether we'd be better off having a global object along these lines instead of the `UserEpeisodePodcastSubstutite`:

``` kotlin
class Podcast {
  // ...
  companion object {
    // ...
    val userPodcast = Podcast(
      uuid = "da7aba5e-f11e-f11e-f11e-da7aba5ef11e",    // UserEpisodePodcastSubstitute.substituteUuid
      title = "Custom Episode",                         // UserEpisodePodcastSubstitute.substituteTitle
    )
  }
```

That would avoid us reconstructing this object multiple times (not that we do it that often), and would feel a bit more intuitive. I'd mildly prefer to make a UserEpisodePodcast `object` instead of just making it a value, but that's not possible since `Podcast` is a data object.

Anyway, let me know what you think because I might make that change at some point if you like it. I don't feel strongly though.

mchowning (approved):
Works well and code looks good. Liked the test and really liked getting more of these calls moved over to use coroutines! 👍

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackService.kt
geekygecko: Yes, that would be great! Also I can never remember `UserEpisodePodcastSubstitute` so it takes me a while to find it.

PR #1364, opened 2023-09-08
Moving the user files podcast substitute.

geekygecko:
## Description

This was an excellent suggestion made in another [PR review](https://github.com/Automattic/pocket-casts-android/pull/1362#discussion_r1319970327). As we often use the user file podcast as a Podcast object, this PR creates one for us. It looks much cleaner and stops the repeating code.

## Testing Instructions
There are lots of changes which are simple enough to just look through but it might be worth testing sync.
1. Sign in to a Plus Pocket Casts account
2. Go to Profile -> Files
3. Add a new file
4. Tap the file row
5. Tap "Upload to Cloud"
6. Tap the file row again
7. Tap "Add to Up Next"
8. Go back to the Profile tab and manually refresh
9. Go to another app and confirm the file is in the Up Next.

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackService.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackService.kt
@@ -18,7 +18,6 @@ import android.support.v4.media.session.PlaybackStateCompat
 import androidx.media.MediaBrowserServiceCompat
 import au.com.shiftyjelly.pocketcasts.analytics.FirebaseAnalyticsTracker
 import au.com.shiftyjelly.pocketcasts.localization.BuildConfig
-import au.com.shiftyjelly.pocketcasts.models.db.helper.UserEpisodePodcastSubstitute
 import au.com.shiftyjelly.pocketcasts.models.entity.BaseEpisode
 import au.com.shiftyjelly.pocketcasts.models.entity.Podcast
 import au.com.shiftyjelly.pocketcasts.models.entity.PodcastEpisode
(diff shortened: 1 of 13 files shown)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackService.kt
```diff
@@ -560,8 +555,7 @@ open class PlaybackService : MediaBrowserServiceCompat(), CoroutineScope {
 
     protected suspend fun loadFilesChildren(): List<MediaBrowserCompat.MediaItem> {
         return userEpisodeManager.findUserEpisodes().map {
-            val podcast = Podcast(uuid = UserEpisodePodcastSubstitute.substituteUuid, title = UserEpisodePodcastSubstitute.substituteTitle, thumbnailUrl = it.artworkUrl)
-            AutoConverter.convertEpisodeToMediaItem(this, it, podcast)
```
geekygecko, line 564: The `thumbnailUrl` attribute isn't needed in convertEpisodeToMediaItem as it already has this logic.
```
    fun convertEpisodeToMediaItem(context: Context, episode: BaseEpisode, parentPodcast: Podcast, groupTrailers: Boolean = false, sourceId: String = parentPodcast.uuid): MediaBrowserCompat.MediaItem {
        val localUri = getBitmapUriForPodcast(parentPodcast, episode, context)
        ...

    fun getBitmapUriForPodcast(podcast: Podcast?, episode: BaseEpisode?, context: Context): Uri? {
        val url = if (episode is UserEpisode) {
            // the artwork for user uploaded episodes are stored on each episode
            episode.artworkUrl
        } else {
            podcast?.getArtworkUrl(480)
        }
```

ashiagr (approved):
Works well and is easier to read 👍

PR #1366, opened 2023-09-12
Added Description to find app version

Mzazi25:
## Description

I added a description on the issue template to make it easier for someone new to find the current app version when reporting a bug 
Fixes # N/A

## Checklist
 
- [ ] with different themes

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -48,7 +48,7 @@ body:
           label: Device, Operating system, and Pocket Casts app version
           placeholder: |
               Google Pixel 6 wAndroid 12, Pocket Casts 7.19.0
-          description: Pocket Casts app version can be found from the settings screen under App info
+          description: Pocket Casts app version can be found from the settings screen then scroll down to About
       validations:
           required: false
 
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -48,6 +48,7 @@ body:
           label: Device, Operating system, and Pocket Casts app version
           placeholder: |
               Google Pixel 6 wAndroid 12, Pocket Casts 7.19.0
+          description: Pocket Casts app version can be found from the settings screen under App info
```
mchowning, line 51: Should we change "App info" to "About" since that is the name of the screen under "Settings" in the app that has the version number? Or are you referring to the OS "App Info" screen?

Mzazi25: I was referring to the OS App info 
The settings on the phone 

But the Settings- About on the pocket cast app does the same thing. 

Which option do you think is the easier one?

mchowning: I might leave toward the in-app directions just because I feel like getting to the OS App Info screen can be different on different flavors of Android. Either way can work though. If you stick with the "App Info" direction though, lets add something to indicate that we are talking about the OS settings screen and not the Pocket Casts settings screen.

Mzazi25: Just pushed a commit.
Should be good now

mchowning (approved):
Looks good! Thanks for improving this @Mzazi25 ! 🙇

PR #1372, opened 2023-09-13
Add podcast/episode title and thumbnail to android share sheet preview

### LINKED ISSUE #1281: Title is missing from share
### Description

Title is missing when sharing an episode or a podcast. Only URL is visible and when sharing to Buffer, even URL is not visible.
Could this be as a result of "Swipe to Share"?

1. try to share an episode 
2.See only URL, no title

Android             |  iOS
:-------------------------:|:-------------------------:
 |

_No response_

joenjogu:
## Description
This PR adds a rich share sheet preview when sharing podcasts and podcast episodes. The share sheet now includes the title and a thumbnail (or placeholder) of the podcast

Fixes #1281 

## Testing Instructions
1. Tap on a podcast or episode
2. Tap on the share icon
3. In the share dialog, choose share episode or share current position
4. The Android share sheet shows up with a thumbnail of the podcast and the podcast/ episode title

## Checklist
 
- [X] with different themes

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/images/PodcastImageLoader.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/images/PodcastImageLoader.kt
@@ -57,6 +57,16 @@ open class PodcastImageLoader(
         }
     }
 
+    suspend fun getBitmapSuspend(podcast: Podcast, size: Int): Bitmap? {
+        return try {
+            val request = load(podcast).size(size, size).build()
+            context.imageLoader.execute(request).drawable!!.toBitmap()
+        } catch (e: Exception) {
+            val request = loadNoPodcastCoil().size(size, size).build()
+            context.imageLoader.execute(request).drawable?.toBitmap()
+        }
+    }
+
     fun loadForTarget(podcast: Podcast, size: Int, bitmapListener: coil.target.Target) {
         val request = load(podcast).size(size, size).target(bitmapListener).build()
         context.imageLoader.enqueue(request)
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/SharePodcastHelper.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/SharePodcastHelper.kt
@@ -1,17 +1,28 @@
 package au.com.shiftyjelly.pocketcasts.repositories.podcast
 
+import android.content.ClipData
 import android.content.Context
 import android.content.Intent
+import android.graphics.Bitmap
+import android.net.Uri
+import android.os.Build
+import androidx.lifecycle.lifecycleScope
 import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsEvent
 import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsTrackerWrapper
 import au.com.shiftyjelly.pocketcasts.analytics.SourceView
 import au.com.shiftyjelly.pocketcasts.models.entity.Podcast
 import au.com.shiftyjelly.pocketcasts.models.entity.PodcastEpisode
 import au.com.shiftyjelly.pocketcasts.preferences.Settings
+import au.com.shiftyjelly.pocketcasts.repositories.images.PodcastImageLoader
 import au.com.shiftyjelly.pocketcasts.utils.FileUtil
-import okhttp3.Call
+import au.com.shiftyjelly.pocketcasts.utils.extensions.getActivity
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 import timber.log.Timber
 import java.io.File
+import java.io.FileOutputStream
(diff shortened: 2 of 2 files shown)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/images/PodcastImageLoader.kt
```diff
@@ -57,6 +57,16 @@ open class PodcastImageLoader(
         }
     }
 
+    suspend fun getBitmapSuspend(podcast: Podcast, size: Int): Bitmap? {
+        return try {
+            val request = load(podcast).size(size, size).build()
+            context.imageLoader.execute(request).drawable!!.toBitmap()
```
mchowning, line 63: Is the reason we're using `!!` here instead of `?` like we are on line 66 because if it returns null we'd rather fallback to the default podcast image than return `null`? What would you think about adding a short comment explaining that? It wasn't obvious to me at first glance why we were using `!!`.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/SharePodcastHelper.kt
```diff
@@ -39,17 +48,25 @@ data class SharePodcastHelper(
         } ?: run {
             url = "$host/podcast/${podcast.uuid}"
         }
-        sendText(url)
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
```
mchowning, line 51: I think it would be good to explain why we're doing this version check. I don't feel strongly about the specifics, but something along these lines could work. Alternatively, we could just extract the version check out into a variable named something like `isCapableOfRichContentPreviews`.
```suggestion
        
        // Android 10+ supports rich content previews: https://developer.android.com/training/sharing/send#adding-rich-content-previews
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
```

mchowning (approved):
Thanks so much for contributing to Pocket Casts! This is working well, and the code looks great. This was a tricky task and you did a really nice job.

I only left two small comments about a couple of spots where I think we could add a brief bit of explanation. They're not blockers though, so let me know what you think.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/SharePodcastHelper.kt
joenjogu: This works for me. Let me know If I can commit this suggestion.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/images/PodcastImageLoader.kt
joenjogu: This logic was copied over from the existing methods on lines 36 and 48. I didn't look too deep into it but that seems to be the intended operation - a fallback mechanism. I can add a comment clarifying, but that also means I'd have to do the same for the other two methods.
Alternatively, we could change the logic to use an `if-else` instead of a`try-catch` with a generic exception for control flow. Something like this:
```
val request = load(podcast).size(size, size).build()
val drawable = context.imageLoader.execute(request).drawable
return if (drawable != null) {
        drawable.toBitmap()
    } else {
        val fallbackRequest = loadNoPodcastCoil().size(size, size).build()
        context.imageLoader.execute(fallbackRequest).drawable?.toBitmap()
    }
```

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/SharePodcastHelper.kt
mchowning: Committing that is fine with me. 👍

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/images/PodcastImageLoader.kt
mchowning: I really like your if-else suggestion. If you want to make that change, that would be great.

joenjogu: I'll work on this change.

joenjogu: Done @mchowning 
Please review.

mchowning (approved):
These changes look good. 👍 Thanks @joenjogu ! 

I added https://github.com/Automattic/pocket-casts-android/pull/1372/commits/b3c380800483f30c7a22f75002b12955f606b0f9 to remove a bit of duplication, so please take a look and let me know if you have any thoughts (or feel free to remove that change if you don't like it). Once you've taken a look, I'll get this merged.

Thanks for your contribution! 🙇

joenjogu:
This looks good to me! Thanks for the fix.

PR #1373, opened 2023-09-14
Change ruby version to Ruby 3.2.2

crazytonyli:
### Description
Ruby 2.7.4 has [reached end of life](https://www.ruby-lang.org/en/downloads/branches/) and it's about time that we migrate our Ruby toolings to Ruby 3.

## Risk

The only risk I can see is there is a small chance that release tasks (that rely on release-toolkit) may break on Ruby 3, which we can't test at PR stage. But if that does happen and it's difficult to solve it properly, we can revert back to Ruby 2.7.4 by changing `.ruby-version` file (or revert the only comment in this PR).

## Checklist
 
N/A


iangmaia (approved):

crazytonyli:
@Automattic/pocket-casts-android Hi, can I get a review on this PR please? Thanks!

geekygecko (approved):

PR #1376, opened 2023-09-15
Improved performance of recording a Track event

geekygecko:
## Description

This change reduces the time it takes to record a Track event. It seems like when a member variable is used that has a property accessed by a `get()` each time it's used, the value is recalculated. The variable `paidSubscription` was called multiple times and this calls `settings.cachedSubscriptionStatus`. The cachedSubscriptionStatus seems to be quite slow so now the result is cached.

## Testing Instructions
1. Filter logcat by Tracked
2. ✅ Verify the analytics include the Track event properties.

### DIFF
--- a/modules/services/analytics/src/main/java/au/com/shiftyjelly/pocketcasts/analytics/TracksAnalyticsTracker.kt
+++ b/modules/services/analytics/src/main/java/au/com/shiftyjelly/pocketcasts/analytics/TracksAnalyticsTracker.kt
@@ -23,40 +23,8 @@ class TracksAnalyticsTracker @Inject constructor(
 ) : IdentifyingTracker(preferences) {
     private val tracksClient: TracksClient? = TracksClient.getClient(appContext)
     override val anonIdPrefKey: String = TRACKS_ANON_ID
-    private val paidSubscription: SubscriptionStatus.Paid?
-        get() = settings.cachedSubscriptionStatus.value as? SubscriptionStatus.Paid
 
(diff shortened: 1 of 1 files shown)


modules/services/analytics/src/main/java/au/com/shiftyjelly/pocketcasts/analytics/TracksAnalyticsTracker.kt
```diff
@@ -70,12 +38,12 @@ class TracksAnalyticsTracker @Inject constructor(
...
         Properties defined by the user have precedence over the default ones pre-defined at "event level" */
         val propertiesToJSON = JSONObject(properties)
-        predefinedEventProperties.keys.forEach { key ->
+        predefinedEventProperties.forEach { (key, value) ->
             if (propertiesToJSON.has(key)) {
                 Timber.w("The user has defined a property named: '$key' that will override the same property pre-defined at event level. This may generate unexpected behavior!!")
-                Timber.w("User value: " + propertiesToJSON.get(key).toString() + " - pre-defined value: " + predefinedEventProperties[key].toString())
+                Timber.w("User value: ${propertiesToJSON.get(key)} - pre-defined value: $value")
             } else {
-                propertiesToJSON.put(key, predefinedEventProperties[key])
```
geekygecko, line 78: Using the `predefinedEventProperties` variable causes the get() to be called again.

ashiagr (approved):
LGTM 👍

PR #1381, opened 2023-09-18
On the podcast page reduce the amount of observers

geekygecko:
## Description

There is an issue on the podcast page where the view model keeps hold of the flowable which updates the UI state. In the logs you can see it repeatedly outputting the following "Observing podcast 0fb82b10-55ed-0138-3a41-0e94336f7152 changes". It seems to be an issue when the flowable is set up that the old flowable is still there.  

I have changed the logic to just be Rx rather than using asFlow() and disposed of the previous flowable before creating a new one.

## Testing Instructions
1. Set up the logcat filter `package:au.com.shiftyjelly.pocketcasts.debug "Observing podcast"`
2. Go the Podcasts tab 
3. Open a podcast 
4. ✅ Verify a log line similar to `Observing podcast 0fb82b10-55ed-0138-3a41-0e94336f7152 changes` appears.
5. Tap the settings cog
6. Change the "Skip first time" and click OK
7. Tap back to return to the podcast page
8. Tap on the Profile tab
9. Tap "Refresh now"
10. ✅ Verify there are only 2 log lines similar to above.
11. Repeat the steps of changing the podcast "Skip first" time and tapping "Refresh now"
10. ✅ Verify there are only 2 log lines again.


ashiagr (approved):
Works well in my testing 👍

PR #1387, opened 2023-09-19
Fixing Sentry issues

geekygecko:
### Issue 1. Remove invalid sponsored error message

Since Sentry reports Timber exceptions, there is an issue "Invalid sponsored podcast list found.". I don't believe this is an invalid sponsor and it should say valid sponsor. We don't need this log so I have removed it.

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/viewmodel/DiscoverViewModel.kt
+++ b/modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/viewmodel/DiscoverViewModel.kt
@@ -142,7 +142,6 @@ class DiscoverViewModel @Inject constructor(
             .map { sponsoredPodcast ->
                 loadPodcastList(sponsoredPodcast.source as String)
                     .filter {
-                        Timber.e("Invalid sponsored podcast list found.")
                         it.podcasts.isNotEmpty() && it.listId != null
                     }
                     .map {


modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/type/PodcastsSortType.kt
```diff
@@ -89,7 +92,9 @@ enum class PodcastsSortType(
 }
 
 class PodcastsSortTypeMoshiAdapter : JsonAdapter<PodcastsSortType>() {
-    override fun toJson(writer: JsonWriter, value: PodcastsSortType?) {}
+    override fun toJson(writer: JsonWriter, value: PodcastsSortType?) {
+        writer.value(value?.serverId)
+    }
```
geekygecko, line 97: We never use the PodcastSortType when sending the value to the server but it seems dangerous not to implement it. I have added it just in case we use it in the future.

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/type/PodcastsSortType.kt
```diff
@@ -60,9 +60,12 @@ enum class PodcastsSortType(
         val default = DATE_ADDED_OLDEST_TO_NEWEST
 
         fun fromServerId(serverId: Int?): PodcastsSortType {
+            if (serverId == null) {
+                return default
+            }
             val sortType = values().firstOrNull { it.serverId == serverId }
             if (sortType == null) {
-                LogBuffer.e(LogBuffer.TAG_INVALID_STATE, "Invalid server ${PodcastsSortType::class.java.simpleName}: $serverId")
```
geekygecko, line 65: With Proguard this comes out as "k" so I have put in the class name to make it clear.

PR #1389, opened 2023-09-20
Improve the worker with updates the episode file size and content type

geekygecko:
## Description

As part of the investigation into the Android ANR and the Sentry issue "Unable to check episode file details with a head request", I looked into the UpdateEpisodeDetailsTask. I don't believe there are any issues relating to ANRs but there are some improvements that could be made.

This change makes the following improvements:
- Adds a developer menu option to trigger the worker making it easier to test
- Increases the default timeout from 10 to 20 seconds. This matches iOS and will help with the megaphone calls that seem to take around 10 seconds to complete.
- Removes the retry logic. This isn't in iOS and if the job has failed after 20 seconds there is a small chance that it will succeed on a second or third attempt.
- It stops using `Result.failure()` as from my testing it causes the Work Manager to not run future tasks. 
- Any IO exception is not sent to Sentry as this relies on the user's network and the producer's servers which we can't fix.
- The task can be sent multiple episodes but if one fails because of a timeout all of them would be retried. Now the timeout is caught per episode. 

## Testing Instructions
1. Open the App Inspection window -> Background Task Inspector
2. Subscribe to some podcasts
3. Go to Profile tab -> settings cog -> Developer
4. Tap "Trigger update episode details"
5. ✅ Verify the workers are completing in the Background Task Inspector window
6. To test the timeout, change the UpdateEpisodeDetailsTask REQUEST_TIMEOUT_SECS amount to 2 seconds
7. Tap "Trigger update episode details" again
8. Filter logcat by "UpdateEpisodeDetailsTask"
9. ✅ Verify that an episode timeout is logged but the worker still completes

## Screenshots 

Screenshot 2023-09-20 at 12 44 31 pm

### DIFF
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/EpisodeDao.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/EpisodeDao.kt
@@ -77,14 +77,26 @@ abstract class EpisodeDao {
     @Query("SELECT * FROM podcast_episodes WHERE podcast_id = :podcastUuid ORDER BY UPPER(title) ASC")
     abstract fun findByPodcastOrderTitleAsc(podcastUuid: String): List<PodcastEpisode>
 
+    @Query("SELECT * FROM podcast_episodes WHERE podcast_id = :podcastUuid ORDER BY UPPER(title) ASC")
+    abstract suspend fun findByPodcastOrderTitleAscSuspend(podcastUuid: String): List<PodcastEpisode>
+
     @Query("SELECT * FROM podcast_episodes WHERE podcast_id = :podcastUuid ORDER BY UPPER(title) DESC")
     abstract fun findByPodcastOrderTitleDesc(podcastUuid: String): List<PodcastEpisode>
+
+    @Query("SELECT * FROM podcast_episodes WHERE podcast_id = :podcastUuid ORDER BY UPPER(title) DESC")
+    abstract suspend fun findByPodcastOrderTitleDescSuspend(podcastUuid: String): List<PodcastEpisode>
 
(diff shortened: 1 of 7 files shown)


modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/EpisodeDao.kt
```diff
@@ -77,24 +77,42 @@ abstract class EpisodeDao {
     @Query("SELECT * FROM podcast_episodes WHERE podcast_id = :podcastUuid ORDER BY UPPER(title) ASC")
     abstract fun findByPodcastOrderTitleAsc(podcastUuid: String): List<PodcastEpisode>
 
+    @Query("SELECT * FROM podcast_episodes WHERE podcast_id = :podcastUuid ORDER BY UPPER(title) ASC")
+    abstract suspend fun findByPodcastOrderTitleAscSuspend(podcastUuid: String): List<PodcastEpisode>
+
```
geekygecko, line 82: There is a bit of duplicate code for the suspend version but hopefully we can switch over to just the Coroutine version.

ashiagr (approved):
Thanks for these nice improvements, @geekygecko! 

- Adds a developer menu option to trigger the worker making it easier to test

Works well 

- Increases the default timeout from 10 to 20 seconds. This matches iOS and will help with the megaphone calls that seem to take around 10 seconds to complete.

Good one! 

- Removes the retry logic. This isn't in iOS and if the job has failed after 20 seconds there is a small chance that it will succeed on a second or third attempt.
- It stops using Result.failure() as from my testing it causes the Work Manager to not run future tasks.
- Any IO exception is not sent to Sentry as this relies on the user's network and the producer's servers which we can't fix.
- The task can be sent multiple episodes but if one fails because of a timeout all of them would be retried. Now the timeout is caught per episode.

All these make sense to me

PR #1390, opened 2023-09-20
Queue show notes downloading

geekygecko:
## Description

After doing some debugging with the Background Task Inspector, I noticed that when you download multiple episodes at once, the show notes tasks run in parallel. A single queue that caches the show notes seems to be a better approach. If you download multiple episodes from the same podcast the show notes will be cached with the first call, before the same file would be downloaded multiple times until the cache was populated. 

This PR also stops some reports to Sentry, I will comment on them inline.

## Testing Instructions
1. Add the "package:mine  UpdateShowNotesTask"
2. Subscribe to some podcasts
3. Go to a podcast page
4. Long press an episode
5. Press the toolbar overflow menu and select all
6. Tap download in the toolbar
7. ✅ Verify after the first "Worker completed" line, the time taken is much shorter indicating the cache is being used.

### DIFF
--- a/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastRatingsViewModel.kt
+++ b/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastRatingsViewModel.kt
@@ -15,6 +15,7 @@ import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.stateIn
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
+import retrofit2.HttpException
 import timber.log.Timber
 import java.io.IOException
 import javax.inject.Inject
@@ -62,7 +63,13 @@ class PodcastRatingsViewModel
             try {
                 ratingsManager.refreshPodcastRatings(uuid)
             } catch (e: Exception) {
-                Timber.e(e, "Failed to refresh podcast ratings")
+                val message = "Failed to refresh podcast ratings"
+                // don't report missing rating or network errors to Sentry
+                if (e is HttpException || e is IOException) {
+                    Timber.i(e, message)
+                } else {
+                    Timber.e(e, message)
+                }
             }
         }
     }
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/converter/DateTypeConverter.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/converter/DateTypeConverter.kt
@@ -6,8 +6,8 @@ import java.util.Date
 class DateTypeConverter {
 
     @TypeConverter
-    fun toDate(value: Long?): Date {
-        return if (value == null) Date() else Date(value)
+    fun toDate(value: Long?): Date? {
+        return if (value == null) null else Date(value)
     }
 
     @TypeConverter
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/converter/DateTypeConverter.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/converter/DateTypeConverter.kt
@@ -6,8 +6,8 @@ import java.util.Date
 class DateTypeConverter {
 
     @TypeConverter
-    fun toDate(value: Long?): Date? {
-        return if (value == null) null else Date(value)
+    fun toDate(value: Long?): Date {
+        return if (value == null) Date() else Date(value)
     }
 
     @TypeConverter
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/download/task/DownloadEpisodeTask.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/download/task/DownloadEpisodeTask.kt
@@ -5,6 +5,7 @@ import android.media.MediaExtractor
 import android.media.MediaFormat
 import android.system.ErrnoException
 import android.system.OsConstants
+import android.util.Log
 import androidx.hilt.work.HiltWorker
 import androidx.work.Data
 import androidx.work.ForegroundInfo
@@ -125,13 +126,13 @@ class DownloadEpisodeTask @AssistedInject constructor(
             this.episode = runBlocking { episodeManager.findEpisodeByUuid(episodeUUID!!) } ?: return Result.failure()
 
             if (this.episode.downloadTaskId != id.toString()) {
-                LogBuffer.e(LogBuffer.TAG_BACKGROUND_TASKS, "Mismatched download task id for episode ${episode.title}. Cancelling.")
+                LogBuffer.i(LogBuffer.TAG_BACKGROUND_TASKS, "Mismatched download task id for episode ${episode.title}. Cancelling.")
                 return Result.failure()
             }
 
             if (this.episode.isArchived) {
                 // In case the episode was archived again but the cancellation event hasn't come through yet
-                LogBuffer.e(LogBuffer.TAG_BACKGROUND_TASKS, "Episode ${episode.title} is archived in download task. Cancelling.")
+                LogBuffer.i(LogBuffer.TAG_BACKGROUND_TASKS, "Episode ${episode.title} is archived in download task. Cancelling.")
                 return Result.success(Data.Builder().putBoolean(OUTPUT_CANCELLED, true).build())
             }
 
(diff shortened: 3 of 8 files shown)


modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/converter/DateTypeConverter.kt
```diff
@@ -6,8 +6,8 @@ import java.util.Date
 class DateTypeConverter {
 
     @TypeConverter
-    fun toDate(value: Long?): Date? {
-        return if (value == null) null else Date(value)
+    fun toDate(value: Long?): Date {
+        return if (value == null) Date() else Date(value)
```
geekygecko, line 10: This should stop the following Sentry error.
```
au.com.shiftyjelly.pocketcasts.models.db.dao.EpisodeDao_Impl$44 in call
IllegalStateException
Expected non-null java.util.Date, but it was null.
```
It seems like the database episode has a publish date of null. I can't find how this is possible but this should stop the crash happening when the data is read.

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastRatingsViewModel.kt
```diff
@@ -62,7 +63,13 @@ class PodcastRatingsViewModel
...
                 ratingsManager.refreshPodcastRatings(uuid)
             } catch (e: Exception) {
-                Timber.e(e, "Failed to refresh podcast ratings")
+                val message = "Failed to refresh podcast ratings"
+                // don't report missing rating or network errors to Sentry
+                if (e is HttpException || e is IOException) {
+                    Timber.i(e, message)
+                } else {
+                    Timber.e(e, message)
+                }
```
geekygecko, line 72: This should stop the following Sentry error as I can see the previous call is the ratings endpoint.
```
HttpException
HTTP 404 
```

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/download/task/DownloadEpisodeTask.kt
```diff
@@ -125,13 +126,13 @@ class DownloadEpisodeTask @AssistedInject constructor(
             this.episode = runBlocking { episodeManager.findEpisodeByUuid(episodeUUID!!) } ?: return Result.failure()
 
             if (this.episode.downloadTaskId != id.toString()) {
-                LogBuffer.e(LogBuffer.TAG_BACKGROUND_TASKS, "Mismatched download task id for episode ${episode.title}. Cancelling.")
+                LogBuffer.i(LogBuffer.TAG_BACKGROUND_TASKS, "Mismatched download task id for episode ${episode.title}. Cancelling.")
```
geekygecko, line 129: There are a few issues reported in this class that are being sent to Sentry. If we aren't going address them I believe they should just be the log level of `info`.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/download/task/UpdateShowNotesTask.kt
```diff
@@ -21,24 +29,50 @@ class UpdateShowNotesTask @AssistedInject constructor(
...
+            showNotesManager.downloadToCacheShowNotes(podcastUuid = podcastUuid)
+            info("Worker completed - took ${SystemClock.elapsedRealtime() - startTime} ms")
             Result.success()
         } catch (e: Exception) {
-            LogBuffer.e(LogBuffer.TAG_BACKGROUND_TASKS, e, "Failed to update show notes")
-            if (e !is HttpException) {
-                SentryHelper.recordException(e)
-            }
+            info("Worker failed - took ${SystemClock.elapsedRealtime() - startTime} ms")
+            val logPriority = if (e is HttpException) Log.INFO else Log.ERROR
```
geekygecko, line 64: I wonder if there is a better way to do this. Maybe I should add a common way to add network exceptions which doesn't report to Sentry. 🤔

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/converter/DateTypeConverter.kt
mchowning: Maybe `LogBuffer.log` would be better than just adding the breadcrumb.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/download/task/DownloadEpisodeTask.kt
mchowning: I agree with the other places you've made this change, but I wonder if we should keep this one since if it shot up all of a sudden that might be an issue we'd want to know about. I don't feel strongly though.

Aside from that, do you think it would be worth including the `downloadTaskId` and `id` in the error message. Seems like that would be helpful in tracking down why these are occurring. Just a thought, I don't feel strongly about this either.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/download/task/UpdateShowNotesTask.kt
```diff
@@ -21,24 +29,46 @@ class UpdateShowNotesTask @AssistedInject constructor(
...
     private val podcastUuid = inputData.getString(INPUT_PODCAST_UUID) ?: ""
-    private val episodeUuid = inputData.getString(INPUT_EPISODE_UUID) ?: ""
 
     override suspend fun doWork(): Result {
+        info("Worker started - podcast: $podcastUuid")
+        val startTime = SystemClock.elapsedRealtime()
         return try {
-            showNotesManager.downloadShowNotes(podcastUuid = podcastUuid, episodeUuid = episodeUuid)
+            showNotesManager.downloadToCacheShowNotes(podcastUuid = podcastUuid)
+            info("Worker completed - took ${SystemClock.elapsedRealtime() - startTime} ms")
```
mchowning, line 60: Including the timing here is nice!

modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/log/LogBuffer.kt
```diff
@@ -110,6 +111,14 @@ object LogBuffer {
         addLog(Log.INFO, tag, null, message, *args)
     }
 
+    fun i(tag: String, throwable: Throwable, message: String, vararg args: Any) {
+        addLog(Log.INFO, tag, throwable, message, *args)
+    }
+
+    fun log(priority: Int, tag: String, throwable: Throwable, message: String, vararg args: Any) {
+        addLog(priority, tag, throwable, message, *args)
+    }
```
mchowning, line 120: Instead of adding a  new `log` method, could we get the same effect by renaming `addLog` to `log` and making that public? Certainly not a big deal, but it feels a bit more straightforward.

mchowning (approved):
Really nice work. Just had a couple of comments/questions, but nothing big. This is a great improvement. 🤩

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/converter/DateTypeConverter.kt
geekygecko: Excellent idea to log and return a constant date. I have made this change.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/download/task/DownloadEpisodeTask.kt
geekygecko: I have kept this and added the extra ids in the message. Thanks for the suggestions.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/download/task/UpdateShowNotesTask.kt
```diff
@@ -21,24 +29,46 @@ class UpdateShowNotesTask @AssistedInject constructor(
...
+            LogBuffer.i(LogBuffer.TAG_BACKGROUND_TASKS, "$TASK_NAME - enqueued ${episode.uuid}")
+            val cacheShowNotesData = Data.Builder()
+                .putString(INPUT_PODCAST_UUID, episode.podcastUuid)
+                .build()
+            val workRequest = OneTimeWorkRequestBuilder<UpdateShowNotesTask>()
+                .setInputData(cacheShowNotesData)
+                .addTag(episode.uuid)
+                .setConstraints(constraints)
+                .build()
+            WorkManager.getInstance(context).beginUniqueWork(TASK_NAME, ExistingWorkPolicy.APPEND, workRequest).enqueue()
```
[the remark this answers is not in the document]
geekygecko: The show notes task always returns `Result.success()` so I think this shouldn't matter but it seems like a safer option, I will change it.

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/converter/DateTypeConverter.kt
geekygecko: That log just saved me! It gets called with dates that need to be null. I'm reverting this issue fix.

PR #1397, opened 2023-09-20
Clear multiselect listener when destroying UpNextFragment to avoid memory leak

mchowning:
## Description
This clears the multiselect listener when destroying the UpNextFragment to avoid retaining a reference to the Fragment.

> **Note**
> This PR builds on #1393. Please merge that PR before this one.

## Testing Instructions
1. Enable Leak Canary
2. Open and close the up next queue from the miniplayer
3. Observe that there is no memory leak
4. Generally smoke test the up next queue screens both from the miniplayer and the full screen player to check for regressions.

| Before | After |
| --- | --- |
| <video src="https://github.com/Automattic/pocket-casts-android/assets/4656348/4e568369-8842-49e0-9c1d-a92e49a49995"/> | <video src="https://github.com/Automattic/pocket-casts-android/assets/4656348/4ab21c3e-c577-47f3-9618-5e398737e26d"/> |

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 *   New Feature:
     *   Suggest episodes to play in Automotive
         ([#1362](https://github.com/Automattic/pocket-casts-android/pull/1362)).
+*   Bug Fixes:
+    *   Avoid memory leak when opening Up Next queue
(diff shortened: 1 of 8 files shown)


geekygecko (approved):
Great fix!

PR #1398, opened 2023-09-20
Improve multiselect listener handling

mchowning:
## Description
Have the `PlayerContainerFragment` initiate the addition/removal of the multiSelectHelper's listener as the UpNextFragment is expanded/contracted. This should help make sure that the UpNextFragment on the full screen player always has the right listener attached, regardless of whether a different instance of `UpNextFragment` gets created and destroyed in the meantime.

This avoids the problem where multiselection would not work after swiping up to open the up next screen on the full screen player, if the up next screen had been viewed already by tapping the up next button on either the full screen player or the miniplayer. See the description of #1395 for some more discussion about this.

Note that this PR also contains the changes from #1404 reverting the removal of the creation of new `UpNextFragment`s on the full screen player because that introduced issues opening the up next queue solely using the full screen player button, which could cause issues when testing this. If you think it would be better to leave that change out of this PR and to bring it in via the `release/7.48`, feel free to just remove that commit from this branch.

## Testing Instructions
1. Tap on the Up Next queue button on the miniplayer
2. Close the Up next queue
3. Open the full screen player
4. Open the up next queue by swiping up from the bottom
5. Open multiselect mode
6. ✅ Try select all/below/above/none and make sure they work
7. Dismiss the up next queue
8. Tap on the button to open the up next queue in the full screen player (i.e., do _**not**_ swipe the up next queue open)
9. Dismiss the up next queue
10. Swipe open the up next queue
11. ✅ Try select all/below/above/none and make sure they work

## Screenshots or Screencast 

| Before | After |

Uploading Screen Recording 2023-09-21 at 3.55.17 PM.mov…

## Checklist
 
- [ ] with different themes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,6 @@
 *   Bug Fixes:
     *   Avoid memory leak when opening Up Next queue
         ([#1397](https://github.com/Automattic/pocket-casts-android/pull/1397))
-    *   Improved multiselect handling in Up Next queue
-        ([#1395](https://github.com/Automattic/pocket-casts-android/pull/1392))
     *   Improved the downloading of episode show notes
         ([#1390](https://github.com/Automattic/pocket-casts-android/pull/1390))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+7.49
+-----
+
+*   Bug Fixes:
+    *   Improve multiselect handling in Up Next queue
+        ([#1398](https://github.com/Automattic/pocket-casts-android/pull/1390))
+
 7.48
 -----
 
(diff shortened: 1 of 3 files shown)


CookieyedCodes:
Could we have a show/hide episodes in upnext toggle somewhere on the downloads page (or an upnext filter pill for episodes in/not in upnext -tho that might break play all & should sync🤔) -id personally prefer the latter idea truthfully just because it's permanent improvement 

As fixing this will get rid of one benifit that you can select all your upnext & distinctify what is & what is not in your upnext slightly easier & very much quicker (ie check boxes vs non check boxes are just quicker for my eyes)

mchowning:
> Could we have a show/hide episodes in upnext toggle somewhere on the downloads page...

I can definitely see how that could be handy. Thanks for the suggestion @CookieyedCodes . I've logged it on our internal feature request tracker. 👍

CookieyedCodes:
Please try to prioritise it 😅, I can imagine an icon being the upnext icon with an open/closed eye in the bottom right if it's not a checkbox under the three dot or filter pill, I'm gonna be really missing it as soon as multiselect is fixed, assign it to @Mzazi25 if your all busy & I can help him with it 😉

mchowning:
> Please try to prioritise it... assign it to @Mzazi25 if your all busy & I can help him with it

I don't think we'll be getting to this in the near future, so someone else can definitely work on this. Just please run any proposed changes by @david-gonzalez-a8c as early as possible since he's our expert on the design side of things and we want to make sure Pocket Casts stays easy to use for non-power users. Opening a GitHub discussion about it might be a good way to proceed.

geekygecko (approved):
That worked well for me

Issue #1396, opened 2023-09-20
SearchHistoryViewModel.loadSearchHistory loads infinitely when the search history is opened


emilylaguna:
### Description

When the search history view opens and has results `loadSearchHistory` will be called infinitely. 

### Step-by-step reproduction instructions

1. Open `SearchHistoryViewModel.kt`
2. Go to [line 59](https://github.com/Automattic/pocket-casts-android/blob/93013f4bbd564dacc23dfdbb1a3a0e79302ab0c6/modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/searchhistory/SearchHistoryViewModel.kt#L59C5-L59C5)
3. Add `LogBuffer.i("🚨","loadSearchHistory")`
4. Run the app
5. Go to Discover > Tap Search
6. 💥 If you have search history results, you will see the added log output infinitely 
7. Dismiss the search history, the log will stop
8. Open the search history, and clear it, the log will stop.

### Screenshots or screen recording

https://github.com/Automattic/pocket-casts-android/assets/793774/055c7c79-4cdc-4e81-9540-847050ad03cb

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

Pixel 4, Android 11 and 13. Running 7.47 and main.

PR #1400, opened 2023-09-21
Stop the search history loading infinitely

geekygecko:
## Description

When the search page was opened and there was search history, the history was constantly loaded. This was due to the load being in the Compose render loop. I believe the correct approach is to use a `LaunchedEffect` but it would be good to confirm that. The fix only loads the history once when opening the search page. 

Fixes https://github.com/Automattic/pocket-casts-android/issues/1396

## Testing Instructions
1. Open `SearchHistoryViewModel.kt`
2. Go to [line 59](https://github.com/Automattic/pocket-casts-android/blob/93013f4bbd564dacc23dfdbb1a3a0e79302ab0c6/modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/searchhistory/SearchHistoryViewModel.kt#L59C5-L59C5)
3. Add `LogBuffer.i("🚨","loadSearchHistory")`
4. Run the app
5. Go to Discover > Tap Search
7. Make sure you have a search history term
8. ✅ Verify the log statement is only printed once


ashiagr (approved):
Verified. The log is printed just once now. Thanks for fixing it. 🙇‍♀️

PR #1408, opened 2023-09-22
Fix the podcast settings fragment leak

geekygecko:
## Description

While testing the app with Leak Canary when you open the podcast settings and return to the podcast page it reports a leak. This change fixes this leak. The cause of the issues seems to be setting the toolbar in the activity, I don't think this is needed so I have removed it.

This PR contains some other clean up

## Testing Instructions
1. In base.gradle uncomment `debugImplementation androidLibs.leakCanary`
2. Setup a filter in logcat `package:mine LeakCanary`
3. Go to the Podcasts tab
4. Open a podcast page
5. Tap the settings cog 
6. Wait for the log message and ignore `Found 1 object retained` message from the Podcast Fragment
7. Tap the back button in the toolbar
9. ✅ Verify if there is an object retained log line that there is also a `All retained objects have been garbage collected` shortly after.


modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/podcast/PodcastFragment.kt
```diff
@@ -189,7 +184,7 @@ class PodcastFragment : BaseFragment(), Toolbar.OnMenuItemClickListener, Corouti
     }
 
     private val onUnsubscribeClicked: (successCallback: () -> Unit) -> Unit = { successCallback ->
-        launch {
+        lifecycleScope.launch {
```
geekygecko, line 187: Rather than using the CoroutineScope it's better to launch using the fragment lifecycle scope.

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/podcast/PodcastFragment.kt
```diff
@@ -533,14 +527,6 @@ class PodcastFragment : BaseFragment(), Toolbar.OnMenuItemClickListener, Corouti
     override fun onStart() {
         super.onStart()
 
-        if (shouldCloseOnReturn) {
```
geekygecko, line 536: The variable `shouldCloseOnReturn` is never set to true.

modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/accountmanager/PocketCastsAuthenticatorService.kt
```diff
@@ -10,14 +10,20 @@ import javax.inject.Inject
...
     }
 
+    override fun onDestroy() {
+        super.onDestroy()
+        authenticator = null
+    }
+
     override fun onBind(intent: Intent?): IBinder? {
-        return authenticator.iBinder
+        return authenticator?.iBinder
```
geekygecko, line 27: This fixes the following leak.

Screenshot 2023-09-22 at 4 48 26 pm

geekygecko, line 27: This fixes the following leak.

Screenshot 2023-09-22 at 4 48 26 pm

ashiagr (approved):

Issue #1421, opened 2023-09-28
Play next in queue doesn't remove the playing episode


geekygecko:
### Description

The playback manager `playNextInQueue` method works differently on iOS and Android. On iOS when you ask to play the next episode in the queue, it removes the currently playing episode. But Android puts the current playing episode back on the Up Next queue. It makes more sense to me that the episode is removed from the queue or if you keep saying next episode it will just switch between the two episodes.

### Step-by-step reproduction instructions

1. Add multiple episodes to your Up Next queue
2. Play the current episode
3. Open the full-screen player
4. Long press the skip forward button
5. Tap "Next episode"
6. Open the Up Next
7. ✅ The currently playing episode should of been removed and not put on the Up Next queue

### iOS

https://github.com/Automattic/pocket-casts-android/assets/308331/817f757a-4b1a-4887-83eb-61a4c61718a4

### Android

https://github.com/Automattic/pocket-casts-android/assets/308331/f0557845-5fb7-403c-8a22-f56a1a4da5fe

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

PR #1422, opened 2023-09-28
Remove the playing episode when playing the next in queue

geekygecko:
## Description

The playback manager `playNextInQueue` method works differently on iOS and Android. On iOS when you ask to play the next episode in the queue, it removes the currently playing episode. But Android puts the current playing episode back on the Up Next queue. It makes more sense to me that the episode is removed from the queue or if you keep saying next episode it will just switch between the two episodes.

Fixes https://github.com/Automattic/pocket-casts-android/issues/1421

## Testing Instructions
1. Add multiple episodes to your Up Next queue
2. Play the current episode
3. Open the full-screen player
4. Long press the skip forward button
5. Tap "Next episode"
6. Open the Up Next
7. ✅ The currently playing episode should of been removed and not put on the Up Next queue

## Screencast 

[Screen_recording_20230929_090836.webm](https://github.com/Automattic/pocket-casts-android/assets/308331/5aa20e21-f62c-448a-9381-6776f575dbd7)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 *   Bug Fixes:
     *   Improve multiselect handling in Up Next queue
         ([#1398](https://github.com/Automattic/pocket-casts-android/pull/1390))
+    *   Improve the next episode action to remove the playing episode from the Up Next
+        ([#1422](https://github.com/Automattic/pocket-casts-android/pull/1422))
     *   Fixed bug where theme selection UI was not functioning as intended
         ([#1399](https://github.com/Automattic/pocket-casts-android/pull/1399))
     *   Improve RTL handling in episode search box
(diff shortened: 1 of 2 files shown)


CookieyedCodes:
As long as it doesn't wipe the episode when long pressing the episode in upnext I don't have a problem with this change 🤔 😉

geekygecko:
> As long as it doesn't wipe the episode when long pressing the episode in upnext I don't have a problem with this change 🤔 😉

@CookieyedCodes don't worry long pressing on the Up Next episode won't be affected by this change, it uses `playEpisode` instead of `playNextInQueue `.

CookieyedCodes:
Coolio just checking 😉

ashiagr (approved):

PR #1425, opened 2023-10-01
PCAndroid: Update Fastfile with version model

spencertransier:
## Summary
- This PR integrates the new version and build code models from https://github.com/wordpress-mobile/release-toolkit/pull/512
- Many of the changes here are from breaking out the code within the various Release Toolkit `*_prechecks` actions into the Fastfile itself. Nearly all of the `prechecks` actions were tightly coupled with the existing version methods. To simplify things with the new version model, I took the relevant snippets from the `prechecks` actions and added them directly to the Fastfile. We'll be evaluating those and adding better-named and succinct shared actions to the Release Toolkit in the future. 
- "Release version" refers to either the `1.2` or `1.2.3` (hotfix) version format. 
- "Beta version" refers to the `1.2.3-rc-1` format with the `rc` suffix. So either the "release version" or "beta version" will be used as. the `versionName` in `version.properties` depending on the build.
-  "Build code" is what we're calling the incremented number for each build (in PCAndroid's case, the `versionCode` in the `version.properties` file.). This is because the `versionCode` in Android apps or `BUILD_NUMBER`/`VERSION_LONG` used by A8C iOS apps are formatted differently. So "build code" is the overarching term for all of those. Just wanted to clarify that vocabulary as you review this PR


fastlane/Fastfile
```diff
@@ -74,10 +83,43 @@ end
...
+
+    # Create the release branch
+    UI.message 'Creating release branch...'
+    ensure_git_branch(branch: DEFAULT_BRANCH)
+    Fastlane::Helper::GitHelper.create_branch("release/#{next_release_version}", from: DEFAULT_BRANCH)
+    UI.message "Done! New release branch is: #{git_branch}"
+
+    # Bump the version and build code
+    UI.message 'Bumping beta version and build code...'
+    Fastlane::Helper::GitHelper.ensure_on_branch!('release')
```
jkmassel, line 112: Looks like we use `ensure_git_branch` above, but this here – can we make that consistent?

fastlane/Fastfile
```diff
@@ -74,10 +83,43 @@ end
...
+      Do you want to continue?
+
+    MESSAGE
+
+    # Ask user confirmation
+    UI.user_error!('Aborted by user request') unless options[:skip_confirm] || UI.confirm(confirmation_message)
+
+    # Create the release branch
+    UI.message 'Creating release branch...'
+    ensure_git_branch(branch: DEFAULT_BRANCH)
```
jkmassel, line 106: Can we bump this up with `ensure_git_status_clean`? 

That way all the checks happen before user input is required

jkmassel (changes requested):

fastlane/Fastfile
spencertransier: Addressed in https://github.com/Automattic/pocket-casts-android/pull/1425/commits/29bb7e4e334991f5b6cca3fa550fb74905f0cec2

fastlane/Fastfile
spencertransier: Addressed in https://github.com/Automattic/pocket-casts-android/pull/1425/commits/e39452bdb1e5d5de01fe658931d8845beb97ae5f

fastlane/Fastfile
```diff
@@ -209,29 +372,33 @@ platform :android do
       lint_task: 'lintRelease'
     )
 
+    version = current_release_version
```
ashiagr, line 375: `version` is already defined on line 365. Do we need this again?

ashiagr (approved):
Thanks for the changes and for bringing consistency across the apps! 

I tested if the version name and version code were generated/ formatted correctly by running `new_beta_release` fastlane command locally and found them correct:



I'm approving the PR and will give these changes a try in the upcoming code freeze on Monday.

fastlane/Fastfile
spencertransier: Fixed in https://github.com/Automattic/pocket-casts-android/pull/1425/commits/eb2eb1e6659e1c3f930794454b8a626fd7ed0fa6

PR #1437, opened 2023-10-05
Enable in app review

ashiagr:
## Description

This enables in-app review. It can be turned off using Firebase remote config for the key `in_app_review_enabled`.

> **Warning**
> Targets 7.49

#### Test.1 Feature on
- Test that the in-app review dialog is shown for the testing steps in https://github.com/Automattic/pocket-casts-android/pull/1305

#### Test.2 Feature off

- Turn off the feature in the Firebase remote config for the key `in_app_review_enabled` and publish changes
- Search for `setMinimumFetchIntervalInSeconds(...)` in the code and set to 5s
- Add a `LogBuffer.i(...)` just before line 94 in `StatsViewModel` to print value for `featureFlag.isEnabled(Feature.IN_APP_REVIEW_ENABLED)`
- Install release build
- Open `Profile` -> `Stats`
- Open `Profile` -> `Settings` -> `Help & feedback` -> `Logs` 
- ✅ Notice that the value is `false` for `featureFlag.isEnabled(Feature.IN_APP_REVIEW_ENABLED)`
- Turn on the feature in the Firebase remote config for the key `in_app_review_enabled` and publish changes

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/services/featureflag/src/main/java/au/com/shiftyjelly/pocketcasts/featureflag/providers/FirebaseRemoteFeatureProvider.kt
+++ b/modules/services/featureflag/src/main/java/au/com/shiftyjelly/pocketcasts/featureflag/providers/FirebaseRemoteFeatureProvider.kt
@@ -15,8 +15,12 @@ class FirebaseRemoteFeatureProvider @Inject constructor(
     override fun isEnabled(feature: Feature): Boolean =
         firebaseRemoteConfig.getBoolean(feature.key)
 
-    override fun hasFeature(feature: Feature): Boolean {
-        return false
+    override fun hasFeature(feature: Feature) = when (feature) {
+        Feature.IN_APP_REVIEW_ENABLED -> true
+        Feature.END_OF_YEAR_ENABLED,
+        Feature.SHOW_RATINGS_ENABLED,
+        Feature.ADD_PATRON_ENABLED,
+        Feature.BOOKMARKS_ENABLED -> false
     }
 
     override fun refresh() {
--- a/modules/services/featureflag/src/main/java/au/com/shiftyjelly/pocketcasts/featureflag/Feature.kt
+++ b/modules/services/featureflag/src/main/java/au/com/shiftyjelly/pocketcasts/featureflag/Feature.kt
@@ -40,7 +40,7 @@ enum class Feature(
     IN_APP_REVIEW_ENABLED(
         key = "in_app_review_enabled",
         title = "In App Review",
-        defaultValue = false,
+        defaultValue = true,
         tier = FeatureTier.Free,
         hasDevToggle = true,
     );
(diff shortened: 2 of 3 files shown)


modules/services/featureflag/src/main/java/au/com/shiftyjelly/pocketcasts/featureflag/providers/FirebaseRemoteFeatureProvider.kt
```diff
@@ -15,8 +15,12 @@ class FirebaseRemoteFeatureProvider @Inject constructor(
     override fun isEnabled(feature: Feature): Boolean =
         firebaseRemoteConfig.getBoolean(feature.key)
```
mchowning, line 16: If for some reason this key does not exist in Firebase, this `getBoolean` method will fall back to the default value we give to Firebase when we call `setDefaultsAsync`. We're not doing that currently, which means it would always default to `false`. I'm thinking maybe we should add an entry to our [FirebaseConfig](https://github.com/Automattic/pocket-casts-android/blob/main/modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/config/FirebaseConfig.kt) for this key so it has a default value.

Ideally we would make that happen automatically based on the `Feature` class, but that's a larger change, so perhaps for this PR we just hardcode the key and default value into `FirebaseConfig`?

I made a separate PR targeting `main` (https://github.com/Automattic/pocket-casts-android/pull/1441) that I think helps make it so that when we create a new feature flag it will get the appropriate default value set automatically.

mchowning (approved):
This is working well.

The only change I think we might want is setting a default value in our `FirebaseConfig` object. This is very much an edge case though, so I'll go ahead and approve since I don't think that's a blocker.

modules/services/featureflag/src/main/java/au/com/shiftyjelly/pocketcasts/featureflag/providers/FirebaseRemoteFeatureProvider.kt
ashiagr: It's better to add a default value to `RemoteConfig` to handle a scenario when the key doesn't exist in `Firebase`. 

> perhaps for this PR we just hardcode the key and default value into FirebaseConfig

Added in https://github.com/Automattic/pocket-casts-android/pull/1437/commits/68d7e94d69429472f9ef68ba9493c9f79a7efa65

> I made a separate PR targeting main (https://github.com/Automattic/pocket-casts-android/pull/1441) that I think helps make it so that when we create a new feature flag it will get the appropriate default value set automatically.

That's so nice! I'll shortly take a look.

mchowning: > It's better to add a default value to RemoteConfig to handle a scenario when the key doesn't exist in Firebase.

Can you tell me a bit more about what you mean here?

ashiagr: Oh, I just meant to agree with you to add a default value to [FirebaseConfig](https://github.com/Automattic/pocket-casts-android/blob/main/modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/config/FirebaseConfig.kt).

PR #1441, opened 2023-10-05
Ensure firebase remote feature flags have default values

mchowning:
## Description
This ensures that every feature flag we create with a firebase remote flag also has a default value set in the app. I've broken up the commits to make this easier to review since there's a lot of noise in the diff:

1. https://github.com/Automattic/pocket-casts-android/commit/5605461997ec66753ed84c89d86418a808c12669 is just the changes from https://github.com/Automattic/pocket-casts-android/pull/1437 since that is our only `Feature` that currently has a remote feature flag.
2. https://github.com/Automattic/pocket-casts-android/commit/047f4c1986f2ee5fa4bd9f37a2fbd3e9a44a527a is just moving the `featureflag` package out of the `featureflag` module and to the `utils` module so that we have access to the `Feature` enum when we create the `FirebaseConfig`. No behavior should be changed by this commit, it's almost entirely changes to import statements.
3. https://github.com/Automattic/pocket-casts-android/commit/aa2578968934be174cd274ab58a70e5c87a5d888 is the core change in this PR. It moves the determination for whether a feature has a firebase remote flag to the `Feature` class and uses that new field to ensure that the `FirebaseConfig` always has a default value for every feature that has a Firebase remote feature flag.

## Testing Instructions
Follow the test steps from https://github.com/Automattic/pocket-casts-android/pull/1437 to ensure the feature flag still works as expected.

## Checklist


modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/config/FirebaseConfig.kt
```diff
@@ -1,14 +1,18 @@
...
+    val defaults = mapOf(
         PERIODIC_SAVE_TIME_MS to 60000L,
         PODCAST_SEARCH_DEBOUNCE_MS to 2000L,
         EPISODE_SEARCH_DEBOUNCE_MS to 2000L,
-        CLOUD_STORAGE_LIMIT to 20L,
-    )
+        CLOUD_STORAGE_LIMIT to 10L,
+    ) + Feature.values()
+        .filter { it.hasFirebaseRemoteFlag }
+        .associate { it.key to it.defaultValue }
```
ashiagr, line 17: Tested and it works perfectly!

Note for later: Since `Patron` is enabled in beta, let's ensure that `hasFirebaseRemoteFlag` is correctly set for it when we merge the beta release to `main`.

ashiagr (approved):

Issue #1374, opened 2023-09-14
Switches to streaming on metered connection without data warning


mchowning:
### Description

If you start streaming an episode on wifi, and switch to a metered connection in the middle of the episode, the episode will continue playing without giving any warning even if you have turned on the setting to warn before streaming an episode. That warning will only get shown when you start playback.

This is a problem for users who do not want to use metered data.

Reported in 6813856-zd-a8c.

### Step-by-step reproduction instructions

1. Turn on the setting to warn when streaming an episode.
2. Connect to wifi and a metered connection
3. Start streaming an episode
4. Disconnect from wifi
5. ❗  Observe that the episode continues playing, even if you skip to a part of the episode that is far in the future (and therefore wasn't buffered when on wifi)
6. Pause the episode
7. Tap play to resume the episode
8. Observe that the stream warning is now presented

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

Android Pixel 6 Pro, Android 13, Pocket Casts 7.46.2

mchowning:
In fixing this, we probably want to make sure that for someone who switches on and off of wifi repeatedly (because, for example, they're at the edge of their wifi connection) we don't present them with the stream warning each and every time the connection switches to metered. Seems like just once may be enough.

PR #1447, opened 2023-10-09
Whenever users switches from wifi to mobile data he will get notified…

neeldoshii:
Linked Issue #1374 

## Description

Whenever a user streams podcast on Mobile Data he/she will get a toast warning that its streaming on mobile data. 

## Testing Instructions

1. Turn on the setting to warn when streaming an episode.
2. Connect to wifi and a metered connection
3. Start streaming an episode
4. Disconnect from wifi
5. You will get a toast saying its connected to mobile data. (for log testing filter the log with Check Data Connectivity)

### Improvements 

Needs to be only called when the episode is playing, not when its paused/stopped xyz.

## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/4656348/4ebe4f3f-0cd4-49bd-9152-281fbfc84472

## Checklist
 
- [X] with different themes

### DIFF
--- a/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/PlayerContainerFragment.kt
+++ b/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/PlayerContainerFragment.kt
@@ -1,11 +1,14 @@
 package au.com.shiftyjelly.pocketcasts.player.view
 
 import android.graphics.drawable.Drawable
+import android.net.ConnectivityManager
 import android.os.Bundle
+import android.util.Log
 import android.view.Gravity
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import android.widget.Toast
 import androidx.annotation.StringRes
 import androidx.appcompat.content.res.AppCompatResources
 import androidx.core.view.doOnLayout
@@ -47,15 +50,20 @@ import com.google.android.material.tabs.TabLayoutMediator
 import dagger.hilt.android.AndroidEntryPoint
 import timber.log.Timber
 import javax.inject.Inject
+import org.checkerframework.checker.units.qual.Length
 import au.com.shiftyjelly.pocketcasts.images.R as IR
 import au.com.shiftyjelly.pocketcasts.localization.R as LR
 import au.com.shiftyjelly.pocketcasts.views.R as VR
 
+@Suppress("DEPRECATION")
 @AndroidEntryPoint
 class PlayerContainerFragment : BaseFragment(), HasBackstack {
-    @Inject lateinit var settings: Settings
-    @Inject lateinit var analyticsTracker: AnalyticsTrackerWrapper
-    @Inject lateinit var multiSelectHelper: MultiSelectBookmarksHelper
+    @Inject
+    lateinit var settings: Settings
+    @Inject
+    lateinit var analyticsTracker: AnalyticsTrackerWrapper
+    @Inject
+    lateinit var multiSelectHelper: MultiSelectBookmarksHelper
 
     lateinit var upNextBottomSheetBehavior: BottomSheetBehavior<View>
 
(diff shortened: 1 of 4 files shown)


CookieyedCodes:
Personally I also will say their should be a toggle in settings to not turn on this popup 🤔 😉 (maby advanced settings?), just to prevent pointless pausing/people who are allways on the edge 😉

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/PlayerContainerFragment.kt
```diff
@@ -208,6 +258,35 @@ class PlayerContainerFragment : BaseFragment(), HasBackstack {
...
         )
+
+
+        viewModel.networkType.observe(viewLifecycleOwner) { type ->
+            when (type) {
+
+                ConnectivityManager.TYPE_WIFI -> {
+                    Timber.tag("Check Data Connectivity").d("Connect to wifi 😊")
+                    println("Connect to wifi ")
+
```
mchowning, line 269: Looks like there are some empty lines at the end of each of the when clauses. Can we remove those?

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/PlayerContainerFragment.kt
```diff
@@ -208,6 +258,35 @@ class PlayerContainerFragment : BaseFragment(), HasBackstack {
...
+                else -> {
+                    Timber.tag("Check Data Connectivity").d("Neither Connected to Wifi nor Mobile" +
+                            " Internet")
+
+
+                }
+            }
+        }
+        viewModel.checkNetworkType(requireContext())
```
mchowning, line 289: I'm not sure that the `PlayerContainerFragment` is the right place for this functionality. It doesn't seem like whether we show this should be tied to this view. Instead, I would think we'd want to give the warning anytime playback is happening, so maybe `MainActivity` or `PocketCastsApplication` would make more sense.

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/PlayerContainerFragment.kt
```diff
@@ -208,6 +258,35 @@ class PlayerContainerFragment : BaseFragment(), HasBackstack {
...
+                    println("Connect to wifi ")
+
+                }
+
+                ConnectivityManager.TYPE_MOBILE -> {
+                    //Log Purposes
+                    Timber.tag("Check Data Connectivity").d("Connect to Mobile Internet 😊")
+                    println("Connect to Mobile Internet ")
+                    //Toast when user is using Mobile Internet
+                    Toast.makeText(context, "Using Mobile Data", Toast.LENGTH_SHORT).show()
```
mchowning, line 277: This is showing the text anytime the network type changes, even if playback is not currently happening, whereas [the linked issue](https://github.com/Automattic/pocket-casts-android/issues/1374) is just addressing the situation where someone is actively streaming an episode and they switch to mobile data and they don't get any warning even though they have turned on the setting for showing a warning when streaming mobile data. Basically, I think we only want to respond when the user has that setting turned on and the user is streaming an episode that is switching to use mobile data.

mchowning, line 277: For strings like this that are user facing, we need to use a string resource so that they get translated.

mchowning (changes requested):
Thanks so much for your contribution @neeldoshii ! We really appreciate all your contributions, and I'm very sorry it took so long to get this reviewed.

I left a few comments about the code, but at a higher level, I don't think this is the best way to address https://github.com/Automattic/pocket-casts-android/issues/1374. A toast message isn't likely to be noticed by most users since people often aren't looking at their phone when they listen to podcasts. It also would still feel like the "warn before streaming" setting isn't working as it should since it would have drastically different behavior depending on whether the playback began on the mobile network (pop-up with a warning that requires confirmation before continuing to stream) versus when playback switches to a mobile network (a toast message is shown with no confirmation needed to continue).

I think what we need is to stop playback and display a warning that requires confirmation before continuing (like we do when for someone who starts playback when connected to a mobile network when they have the stream warning setting turned on).

Because I think we need to take a different approach to this, my inclination is to close this PR. If you want to continue working on this, let's open another PR with a different approach. If you'd rather collaborate in this PR to get this to where it needs to be though, let me know, because we can certainly do that.

One other nice-to-have in the future would be to try to avoid reformatting code where you are not making changes. When there are a lot of code formatting changes, it makes it more difficult to find and review the lines that are being changed, and that helps the review process go more quickly.

Let me know how you'd like to close this PR or if you want to work on moving this PR forward. 

Thanks again for contributing to Pocket Casts!

mchowning:
Since you're not working on this, I'm going to go ahead and pick this up this issue myself @neeldoshii. I think I've got an idea for how to handle this well. Thanks so much for opening this PR though because it helped give me an idea for how to approach this.

PR #1448, opened 2023-10-09
Add feature flag for Giving Ratings

mchowning:
## Description
This adds a feature flag for giving ratings and, depending on if that feature flag is enabled, clicking a rating will load the Give Rating Screen.

> **Note**
> This PR builds on https://github.com/Automattic/pocket-casts-android/pull/1446

## Testing Instructions
1. Go to a podcast that has a rating
2. If needed tap, tap the caret so the podcast rating is visible
3. If the "Give Rating" feature flag is enabled, clicking the rating will open a placeholder rating screen. Otherwise, clicking will not affect the UI.

## Checklist


modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/components/ratings/StarRatingView.kt
```diff
@@ -26,12 +26,14 @@ import au.com.shiftyjelly.pocketcasts.compose.theme
...
 import au.com.shiftyjelly.pocketcasts.podcasts.viewmodel.PodcastRatingsViewModel.RatingState
 import au.com.shiftyjelly.pocketcasts.podcasts.viewmodel.PodcastRatingsViewModel.Star
+import au.com.shiftyjelly.pocketcasts.ui.helper.FragmentHostListener
 import au.com.shiftyjelly.pocketcasts.ui.theme.Theme
 import au.com.shiftyjelly.pocketcasts.utils.extensions.abbreviated
 import java.util.UUID
 
 @Composable
 fun StarRatingView(
+    fragmentHostListener: FragmentHostListener,
```
mchowning, line 36: Passing a `FragmentHostListener` into a composable feels a little weird to me, but since this composable should always open the same thing anytime the it is clicked, passing in something like `onClick: () -> Unit` didn't feel right since we would always be passing in the same on click behavior.

CookieyedCodes:
Video is wrong?

ashiagr (approved):
Works well in my testing 👍

mchowning:
> Video is wrong?

Yeah, that wasn't the right video at all. I just removed the video since I don't know where the correct video disappeared to.

PR #1450, opened 2023-10-10
EOY 2023: Intro Story

ashiagr:
| 🎨 Designs: lH66LwxxgG8btQ8NrM0ldx-fi-1599_20385 |
|:---:|

## Description

This updates the visuals for 2023 intro story.

## Testing Instructions

** Test on a real device so you can test the parallax effect.

1. Make sure you're logged in into an account that has a few episodes listened
2. Go to `Profile` > `Settings` > `Beta Features` > enable `End of Year`
3. Go back to `Profile`
4. Tap the EOY image
5. ✅ Check that the Intro Story looks good
6. Tilt your device
7. ✅ The "2023" should move around

## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/1405144/15155d24-d73b-46a4-aef0-f14a040acb7c

Logo color - yet to be updated

## Checklist
 
- [x] with different themes
- [ ] for accessibility with TalkBack <- this is not working, I need to look into it.


modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/parallaxview/ParallaxView.kt
```diff
@@ -0,0 +1,82 @@
...
+@SuppressLint("UnrememberedMutableState")
+@Composable
+private fun Content(
+    modifier: Modifier = Modifier,
+    data: SensorData?,
+    depthMultiplier: Int = 20,
+    content: @Composable (Modifier, BiasAlignment) -> Unit,
+) {
+    val roll by derivedStateOf { (data?.roll ?: 0f) * depthMultiplier }
+    val pitch by derivedStateOf { (data?.pitch ?: 0f) * depthMultiplier }
```
mchowning, line 65: The parallax effect works pretty well, but it seems a bit jittery to me. I wonder if we could smooth it out a bit. Maybe something like looking at the change in roll and pitch from the currently displayed values and only using some percent of that change instead of the full change would work to dampen the change. Or maybe averaging the last two or three values instead of just taking the latest would work. (obviously I'm just brainstorming 🤔). 

Not a blocker.

modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/views/stories/StoryIntroView.kt
```diff
@@ -13,69 +14,122 @@ import androidx.compose.material.Surface
...
+    }
+}
 
-        PodcastLogoWhite()
+@Composable
+private fun BackgroundImage() {
+    Box(
+        contentAlignment = Alignment.Center,
+    ) {
+        ImageTwenty()
```
mchowning, line 65: This seems a bit different from the designs because in the designs the middle numbers numbers (the 0 and the second 2) overlap the text, but we have them here as part of the background, so the text overlaps them.

I think it could be really interesting if we gave each number a slightly different parallax effect so they had different depths (i.e., some were closer and some further away). If you think that sounds interesting @david-gonzalez-a8c  and would be willing to get us separate assets for each number, I'd be interested to see if I could make that look good without having it be too much parallax (or you're more than welcome to Ashita, I just didn't want to add more tasks to your plate 🙂 ).

Here's a quick example of what this could look like with the two groups of numbers we have currently ([code diff](https://gist.github.com/mchowning/c7951f2f00e5b3447b69f5da531b286b)). We'd need to tweak the numbers I'm using for depth if we want some of the numbers to appear in front of the text.


https://github.com/Automattic/pocket-casts-android/assets/4656348/87fa77b1-3edc-4838-8c67-4c89c6e33310

mchowning (approved):
Nice work @ashiagr . I left a few comments, but no blockers, so I'll go ahead and merge this. I'm fine with addressing any updates you want to do based on my comments in follow-up PRs.

modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/views/stories/StoryIntroView.kt
david-gonzalez-a8c: > I think it could be really interesting if we gave each number a slightly different parallax effect so they had different depths (i.e., some were closer and some further away).

That'd be awesome! I modified the [specs](https://www.figma.com/file/lH66LwxxgG8btQ8NrM0ldx/End-of-Year?type=design&node-id=1881%3A2951&mode=design&t=DsQRtiw4ou98ocD2-1) so each number can be exported separately, but you can also find them here:

[2023 svg.zip](https://github.com/Automattic/pocket-casts-android/files/12860233/2023.svg.zip)

/cc @leandroalonso

PR #1453, opened 2023-10-11
Show no ratings text when no ratings

mchowning:
## Description
As discussed internally, this updates our rating view so that it shows no stars with "No ratings" text when there are no ratings (internal ref: pdeCcb-3ic-p2#comment-2707).

## Testing Instructions
1. Go to a podcast that does not have any ratings (i.e., The Big Dip on the Trending section of Discover)
2. If necessary, expand the podcast details so you can see the rating
3. Observe that the rating shows 5 empty stars and the text "No ratings"

## Screenshots or Screencast 

image

## Checklist
- [x] If this is a user-facing change, I have added an entry in CHANGELOG.md
- [X] Ensure the linter passes (`./gradlew spotlessApply` to automatically apply formatting/linting)
- [X] I have considered whether it makes sense to add tests for my changes
- [X] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml`
- [X] Any jetpack compose components I added or changed are covered by compose previews
 
#### I have tested any UI changes...

- [x] with different themes
- [X] with a landscape orientation
- [X] with the device set to have a large display and font size
- [X] for accessibility with TalkBack

PR #1454, opened 2023-10-11
Make PodcastRating::average nullable

mchowning:
## Description
It feels odd to me that our `PodcastRatings` object has a nullable `total` field and a non-null `average` field. I feel like they should either both be nullable or both not be nullable. 

This PR originally updated the `total` field so it is not nullable, and if we didn't get a `total` value back from the backend, it used a `total` of 0, which is what we were already doing elsewhere in the app. I originally went with making them non-null just because we already treat any null values as `0` in the app, so we didn't really need to preserve the null values.

After PR review feedback (https://github.com/Automattic/pocket-casts-android/pull/1454#issuecomment-1756759270) though, I've adopted the alternative approach of making both `average` and `total` nullable. The benefit of this is that the `total` and `average` fields are not required parts of the ratings API response, so this makes our database consistent with that.

This does add in a non-trivial database migration, so if you think adding the migration is too much of a cost to just remove a null field, let me know.

> **Note**
> This PR builds on https://github.com/Automattic/pocket-casts-android/pull/1453

## Testing Instructions
1. Install the app from `main`
2. Open the podcast page for a few podcasts including one with no ratings like "The Big Dig"
3. Install the app from this PR
4. Open the podcast pages for the same podcasts you opened before along with a few new ones and verify that the ratings are displayed.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 7.50
 -----
+
+*   Updates:
+    *   Show if a podcast has no ratings
+        ([#1453](https://github.com/Automattic/pocket-casts-android/pull/1453))
 *   Bug Fixes:
     *   Fixed show notes loading issues
         ([#1436](https://github.com/Automattic/pocket-casts-android/pull/1436))
--- a/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/AppDatabaseTest.kt
+++ b/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/AppDatabaseTest.kt
@@ -143,6 +143,7 @@ class AppDatabaseTest {
                 AppDatabase.MIGRATION_74_75,
                 AppDatabase.MIGRATION_75_76,
                 AppDatabase.MIGRATION_76_77,
+                AppDatabase.MIGRATION_77_78,
             )
             .build()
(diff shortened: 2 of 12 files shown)


ashiagr:
> An alternative approach would be to make both average and total nullable. The benefit of this is that, at least according to swagger, the total and average fields are not required parts of the ratings API response, so this makes our database consistent with that. 

I think it'll be good to keep both fields nullable to match the rating API responses. 

> This does add in a non-trivial database migration, so if you think adding the migration is too much of a cost to just remove a null field, let me know.

I tested the migration and it works well.

ashiagr (approved):

mchowning:
> I think it'll be good to keep both fields nullable to match the rating API responses.

You bet. I made that change in https://github.com/Automattic/pocket-casts-android/pull/1454/commits/dfa696244f6c1722a2b99cdb2e6d19f0411af9d1 and also restructured the logic a bit in the  `PodcastRatingsViewModel` to make it easier to handle nulls and change the behavior here in the future (since how we want to handle podcasts that don't have any/enough ratings still seems unclear).

These are some pretty big changes, so please take another look.

ashiagr (approved):
Works well after testing using updated instructions. Code changes look good too. 👍

PR #1456, opened 2023-10-11
Start migrating the build to version catalogs

geekygecko:
## Description

Google recommends using version catalogs and provides a [migration guide](https://developer.android.com/build/migrate-to-catalogs#groovy). 
> Gradle version catalogs enable you to add and maintain dependencies and plugins in a scalable way. Using Gradle version catalogs makes managing dependencies and plugins easier when you have [multiple modules](https://developer.android.com/topic/modularization). Instead of hardcoding dependency names and versions in individual build files and updating each entry whenever you need to upgrade a dependency, you can create a central version catalog of dependencies that various modules can reference in a type-safe way with Android Studio assistance.

I have tried this migration before and failed, so I'm only migrating small parts of the app at a time in this attempt.

### DIFF
--- a/automotive/build.gradle
+++ b/automotive/build.gradle
@@ -58,9 +58,9 @@ dependencies {
     implementation androidLibs.ktx
     implementation androidLibs.design
     implementation androidLibs.constraintLayout
-    implementation libs.aboutLibrariesCore
-    implementation libs.aboutLibrariesCompose
-    testImplementation libs.junit
+    implementation androidLibs.aboutLibrariesCore
+    implementation androidLibs.aboutLibrariesCompose
+    testImplementation androidLibs.junit
     androidTestImplementation androidLibs.junitExt
     androidTestImplementation androidLibs.testEspressoCore
 
--- a/base.gradle
+++ b/base.gradle
@@ -60,7 +60,7 @@ android {
     }
 
     composeOptions {
-        kotlinCompilerExtensionVersion project.versionComposeCompiler
+        kotlinCompilerExtensionVersion libs.versions.compose.kotlin.compiler.get()
     }
 
     testOptions {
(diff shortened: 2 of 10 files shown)


automotive/build.gradle
```diff
@@ -58,9 +58,9 @@ dependencies {
     implementation androidLibs.ktx
     implementation androidLibs.design
     implementation androidLibs.constraintLayout
-    implementation libs.aboutLibrariesCore
-    implementation libs.aboutLibrariesCompose
-    testImplementation libs.junit
+    implementation androidLibs.aboutLibrariesCore
+    implementation androidLibs.aboutLibrariesCompose
+    testImplementation androidLibs.junit
```
geekygecko, line 63: Version catalogs use the variable libs so for this migration I'm going to change all the existing dependencies to use our `androidLibs` variable.

base.gradle
```diff
@@ -209,78 +209,77 @@ dependencies {
...
+    implementation androidLibs.rxRelay
+    implementation androidLibs.retrofit
+    implementation androidLibs.retrofitMoshi
+    implementation(androidLibs.retrofitProtobuf) {
         exclude group: 'com.google.protobuf', module: 'protobuf-java'
     }
-    implementation libs.retrofitRxJava
-    implementation libs.moshi
-    implementation libs.moshiAdapters
-    implementation libs.kotlin
```
geekygecko, line 235: I created a new Android project and it seems like you no longer define Kotlin as a dependency and let the Gradle plugin do this.

build.gradle
```diff
@@ -1,36 +1,11 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 buildscript {
-    ext {
-        kotlin_version = '1.7.10'
-        // KSP releases need to match the Kotlin version - https://github.com/google/ksp/releases
-        ksp_version = "$kotlin_version-1.0.6"
-        hilt_version = '2.43.2'
```
geekygecko, line 7: These versions are now in the `libs.versions.toml` file.

build.gradle
```diff
@@ -1,36 +1,11 @@
...
-        mavenLocal()
-
-        maven {
-            url 'https://a8c-libs.s3.amazonaws.com/android' 
-            content {
-                includeGroup "com.automattic.android"
-                includeGroup "com.automattic.android.publish-to-s3"
-            }
-        }
-    }
```
geekygecko, line 20: From the blank Android project and examples I have seen online it seems like you define this in your `settings.gradle` file instead.

gradle/libs.versions.toml
```diff
@@ -0,0 +1,30 @@
...
+# compose-kotlin-compiler = Compose Compiler Version (can be found in composeOptions -> kotlinCompilerExtensionVersion)
+# kotlin = Kotlin Version
+#
+# 2. Upgrade KSP version to match Kotlin version. https://github.com/google/ksp/releases
+
+[versions]
+# Android Gradle Plugin - https://developer.android.com/studio/releases/gradle-plugin
+android-application = "8.1.0"
+# @keep
+compose-kotlin-compiler = "1.3.0"
```
geekygecko, line 15: This requires a keep annotation as this isn't referenced below so when upgrading it gets removed.

build.gradle
```diff
@@ -1,59 +1,29 @@
...
-        classpath 'com.google.android.gms:oss-licenses-plugin:0.10.5'
-        // Hilt dependency injection
-        classpath "com.google.dagger:hilt-android-gradle-plugin:$hilt_version"
-    }
-
-    // Force proguard to the new version to fix "proguard.ParseException: Use of generics not allowed for java type at '<1>_<2>_<3>JsonAdapter' in line 31 of file '/Users/philip/.gradle/caches/transforms-2/files-2.1/d5ca2d039e1d32e9ba6bfd6a67df5151/META-INF/proguard/moshi.pro'"
-    configurations.all {
-        resolutionStrategy {
-            force 'net.sf.proguard:proguard-gradle:6.1.1'
-        }
```
geekygecko, line 40: I thought it was worth cleaning up this file while I was here. Proguard is up to 6.2.2 so I don't believe we need this anymore. I have tested the parsing of JSON in a release build with the user sync and everything seems fine.

mchowning (approved):
Changes look good. Thanks for tackling this and breaking the work up into smaller chunks. I tested both debug and release builds for each of our app modules and saw no issues. 👍

app/build.gradle
```diff
@@ -1,12 +1,16 @@
...
 plugins {
-    id 'io.sentry.android.gradle'
+    alias libs.plugins.android.application
+    alias libs.plugins.kotlin.android
+    alias libs.plugins.kapt
+    alias libs.plugins.ksp
+    alias libs.plugins.hilt
+    alias libs.plugins.sentry
+    alias libs.plugins.google.services
+    id 'kotlin-parcelize'
```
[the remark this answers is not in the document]
geekygecko: I didn't think it was possible. Thanks for the branch! I will merge it in.

modules/modules.gradle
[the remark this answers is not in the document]
geekygecko: That is a good idea. I will do that.

Issue #1424, opened 2023-09-30
Next chapter button ring not showing progress


b-aaron:
### Description

The next chapter button normally shows the progress within the chapter by filling the ring around it. This is not happening

### Step-by-step reproduction instructions

- Play a Podcast with chapters
- Advance within the chapter
- See that the next chapter button does not change
- Look in chapter tab, how far the chapter has advanced
- This does not get represented on the next chapter button

### Screenshots or screen recording

https://github.com/Automattic/pocket-casts-android/assets/72910336/f2772b43-9eff-46e0-ba73-f15899fa9123

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

Android 13
OneUI 5.1
Pocket Casts v7.48-rc-3

pachlava:
Adding a priority label to better align with the rules for MTTR metric.
See pdcxQM-2Su-p2

I'm using `meduim priority` judging from the issue description. Please correct me if I'm wrong.

PR #1461, opened 2023-10-12
Fix chapter progress circle

geekygecko:
## Description

The next chapter on the full-screen player has a progress circle around it. This was broken recently when I upgraded the chapters tab to Compose. This change fixes that and cleans up some of the chapter code.

Fixes https://github.com/Automattic/pocket-casts-android/issues/1424

## Testing Instructions

1. Play a Podcast with chapters (for example [Upgrade](https://pca.st/upgrade))
2. Advance within the chapter
3. ✅ Verify the full screen player next chapter button progress is working

## Screencast 

https://github.com/Automattic/pocket-casts-android/assets/308331/c768e980-d675-418d-94a5-3a817f6e1ee9

### DIFF
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/Chapter.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/Chapter.kt
@@ -5,14 +5,12 @@ import kotlin.math.roundToInt
 
 data class Chapter(
     val title: String,
-    val url: HttpUrl?,
     var startTime: Int,
     var endTime: Int,
-    val imagePath: String?,
-    val mimeType: String?,
-    var played: Boolean = false,
-    var index: Int = 0,
-    var progress: Float = 0.0f
+    val url: HttpUrl? = null,
+    val imagePath: String? = null,
+    val mimeType: String? = null,
+    var index: Int = 0
 ) {
 
     val isImagePresent: Boolean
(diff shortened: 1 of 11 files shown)


modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/Chapter.kt
```diff
@@ -5,14 +5,12 @@ import kotlin.math.roundToInt
...
 data class Chapter(
     val title: String,
-    val url: HttpUrl?,
     var startTime: Int,
     var endTime: Int,
-    val imagePath: String?,
-    val mimeType: String?,
-    var played: Boolean = false,
-    var index: Int = 0,
-    var progress: Float = 0.0f
```
geekygecko, line 15: It didn't feel right storing the `progress` and `played` state in the chapter object as they can change regularly based on the playback position.

CookieyedCodes:
Chapters might be currently on the current rc, will this also fix it? Someone on Reddit pointed it out to me 🤔

Also as the chapter code is being looked at any further thoughts for RSS timetags to "chapter conversation" for non chaptered podcasts?

geekygecko:
@CookieyedCodes 

> Chapters might be currently on the current rc, will this also fix it? Someone on Reddit pointed it out to me 🤔

This bug is fixed in 7.50 and the open beta version will be out early this week.

> Also as the chapter code is being looked at any further thoughts for RSS timetags to "chapter conversation" for non chaptered podcasts?

I did notice a podcast on the weekend which had timestamps in the show notes but no chapters. It would be a handy feature. When we are adding Podlove and Podcast Index chapters we should do this. Thanks for the reminder, again. One day it will happen. 😄

CookieyedCodes:
I'll keep an eye out ❤️❤️

ashiagr (approved):
Works well in my testing 👍

PR #1462, opened 2023-10-12
Finish converting the root build.gradle to version catalogs

geekygecko:
## Description

This change moves the remaining plugins in the root build.gradle to the libs.versions.toml file. 

## Testing Instructions
1. Run ./gradlew assemble
2. Deploy app/build/outputs/apk/release/app-release.apk to a device
3. ✅ Verify the build is successful and the runs.
4. Login to a sync account with podcasts 
5. ✅ Verify the podcast and episode sync data is working. 
6. ✅ Verify on the Profile tab that the refresh has been successful.
7. ✅ Verify there are no exceptions in logcat

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,17 +2,7 @@
 buildscript {
     // Gradle Plugins
     dependencies {
-        // Google Services - https://developers.google.com/android/guides/google-services-plugin
-        classpath "com.google.gms:google-services:4.3.14"
-        // Open source licenses plugin
-        classpath 'com.google.android.gms:oss-licenses-plugin:0.10.5'
-    }
-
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -2,17 +2,7 @@
...
-        classpath "com.google.gms:google-services:4.3.14"
-        // Open source licenses plugin
-        classpath 'com.google.android.gms:oss-licenses-plugin:0.10.5'
-    }
-
-    // Force proguard to the new version to fix "proguard.ParseException: Use of generics not allowed for java type at '<1>_<2>_<3>JsonAdapter' in line 31 of file '/Users/philip/.gradle/caches/transforms-2/files-2.1/d5ca2d039e1d32e9ba6bfd6a67df5151/META-INF/proguard/moshi.pro'"
-    configurations.all {
-        resolutionStrategy {
-            force 'net.sf.proguard:proguard-gradle:6.1.1'
-        }
```
geekygecko, line 15: I thought it was worth cleaning up this file while I was here. Proguard is up to 6.2.2 so I don't believe we need this anymore. I have tested the parsing of JSON in a release build with the user sync and everything seems fine.

PR #1470, opened 2023-10-15
Fix Play button color not updating after changing filter

### LINKED ISSUE #1435: Changing the color of a Filter does not immediately update play icon color
### Description

After changing the color of a filter, when you return to the list of episodes for the filter, the UI does not update the play button with the new color you have selected for the filter. Only after you back out of the filter and reopen it will the play icon have the new color.

1. Open the episode list for a filter
2. Open the filter settings from the three dot menu
3. Change the color of the filter
4. Exit the filter settings
5. ❗  Observe that the play icon by each episode still has the previous filter color
6. Exit the filter screen
7. Reopen the filter
8. Observe that the play icon now has the new filter color properly applied

https://github.com/Automattic/pocket-casts-android/assets/4656348/cc7cb8a1-20de-4adc-ba51-df8c92b9cb89

Samsung A13 running Android 13. Pocket Casts pre-7.50 build

thesohelshaikh:
## Description

`PlayButton.setButtonType()` is used to update the button appearance. This method did an early return without checking if the color was updated.

To fix the issue a check was added to check if the previous color is same as current color, if no then we update the button. To keep track of the previous color an additional variable was added. PlayButton now keeps track of the previous selected color.

Fixes #1435 

## Testing Instructions
1. Open the episode list for a filter
2. Open the filter settings from the three dot menu
3. Change the color of the filter
4. Exit the filter settings
5. With the change the play button should be updated with the selected color

## Checklist
 
- [x] with different themes

### DIFF
--- a/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/components/PlayButton.kt
+++ b/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/components/PlayButton.kt
@@ -36,6 +36,7 @@ class PlayButton @JvmOverloads constructor(
     private var buttonType: PlayButtonType = PlayButtonType.PLAY
     private var episodeUuid: String? = null
     private var podcastUuid: String? = null
+    private var buttonColor: Int? = null
     private var fromListUuid: String? = null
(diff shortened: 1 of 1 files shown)


thesohelshaikh:
@ashiagr @mchowning Requesting for review.

ashiagr (approved):
Thanks for the contribution! I confirm that the issue is fixed. 🎉 
I just added a changelog.

PR #1474, opened 2023-10-16
Conver ChangeEmailFragment to Jetpack Compose

### LINKED ISSUE #753: Convert remaining screens to use Jetpack Compose
We really like [Jetpack Compose](https://developer.android.com/jetpack/compose), but we're in an unfortunate in-between place where we have a fair number of screens written in compose, while most of the app is still using Views. Let's work on converting the remaining screens in the app to use Jetpack Compose.

Keep in mind that we do not have to convert entire screens at a time, the [interoperability APIs](https://developer.android.com/jetpack/compose/interop/interop-apis) allow us to use things like a `ComposeView` to convert just a part of a `View` to use compose.

### In Progress
- `ChangeEmailFragment` @RyanKoech

RyanKoech:
# WIP - suggestions and reviews welcome

## Description
- Created `ChangeEmailFragmentPage` which is Jetpack compose version of the `fragment_change_email.xml` which were consequently deleted in the account feature module
- Updated the `ChangeEmailViewModel` to user `MutableState` instead of `MutableLiveData`
- Extracted logic for `RowButton` into `BaseRowButton` which can be reused for example in this case to create `RowLoadingButton`. *{ Main difference between `RowButton` and `RowLoadingButton` is on the `leadingIcon` implementation as the previous one on `RowButton` could not facilitate `CircularProgressIndicator` needed in the `RowLoadingButton` which would have needed more refactoring to support it. I also lacked context as to the current `RowButton` implementation }*
- Add placeholder parameters to `EmailAndPasswordFields` components to make them more flexible like in this PRs case.

Fixes #753 

## Checklist
 
- [ ] with different themes


RyanKoech:
@mchowning @ashiagr @geekygecko or any other code reviewers please review

mchowning:
👋 @RyanKoech ! Thanks so much for contributing to Pocket Casts! We're just backed up with work right now (one of the disadvantages of being a small team), but I promise we'll get to this as soon as we can.

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/ChangeEmailFragmentPage.kt
```diff
@@ -0,0 +1,193 @@
...
+                TextH50(
+                    modifier = Modifier
+                        .padding(end = 8.dp),
+                    text = stringResource(LR.string.profile_current_email),
+                    color = MaterialTheme.theme.colors.primaryText02,
+                )
+                TextH50(
+                    modifier = Modifier
+                        .weight(1f)
+                        .basicMarquee(),
```
ashiagr, line 129: Thanks for taking care of long email display by adding this marque effect! 😍

modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/buttons/RowButton.kt
```diff
@@ -48,6 +51,88 @@ fun RowButton(
...
+        } else null,
+        onClick = onClick@{
+            if (isLoading) return@onClick
+            onClick()
+        },
+    )
+}
+
+@Composable
+fun BaseRowButton(
```
ashiagr, line 121: Nitpick: Can we make it private?

modules/features/account/src/main/res/layout-car/fragment_change_email.xml
```diff
@@ -1,147 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
ashiagr, line 2: Have we checked that the screen is not displayed in the car? I belive they're not but can you please confirm by checking on Automotive app and Android Auto [DHU](https://developer.android.com/training/cars/testing/dhu#running-dhu) since the layouts were available earlier?

modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/buttons/RowButton.kt
```diff
@@ -37,6 +39,79 @@ fun RowButton(
...
+                        .padding(4.dp),
+                    tint = textColor,
+                )
+            }
+        } else null,
+        onClick = onClick,
+    )
+}
+@Composable
+fun RowLoadingButton(
```
ashiagr, line 84: I noticed previews for `RowButton` were replaced with `RowLoadingButton` in the previews here:

https://github.com/Automattic/pocket-casts-android/pull/1474/commits/a0ee61ff61869191e44066e2b153a7b818a7c7a7#diff-37cb97d49ec540989482b482c81a88357f6a92c95bb309eac181a6a5a7222fd4L79-R185

I reverted them while resolving the conflict as we would like to keep the `RowButton` previews. Can we add a preview for just the `RowLoadingButton`?

ashiagr (approved):
Thank you so much for this work! 🙇‍♀️ We really appreciate the effort. 

I have tested the change email functionality after resolving the merge conflict. 
It works well as per testing instructions. I also tested that server error messages are displayed correctly.

I have shared a minor suggestion and a question. As these are non-blocking, I'll go ahead and approve the PR but not merge it yet.

modules/features/account/src/main/res/layout-car/fragment_change_email.xml
RyanKoech: Yes, I did test on Automotive App and Android Auto and that screen was not being used, but I'll just do another round to confirm.

modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/buttons/RowButton.kt
RyanKoech: I'll add that too!

RyanKoech: Done

modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/buttons/RowButton.kt
RyanKoech: Kept it public since I have moved the three buttons to their own files

modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/buttons/RowButton.kt
ashiagr: Thanks! I just enabled isLoading in the previews so that loading icon can be seen in Profile -> Settings -> Developer -> Showkase -> Button. It looks great!

RyanKoech: Awesome. Thanks!

PR #1468, opened 2023-10-13
Add component for tapping and swiping on stars

mchowning:
## Description
This adds a swipeable star component for giving ratings

## Testing Instructions
1. Open up a podcst with ratings
2. Tap on the rating to open up the Give ratings view
3. ✅ Tapping on a star animates to that number of stars being filled (no half values). You can tap on the left side of the first star to select 0 stars.
4. ✅ Dragging over the stars with your finger will update the stars based on where you drag (it should reflect exactly where your finger is and not animate to your finger's location), and when you release your finger, the filled star state will animate to the nearest half-star.

> **Warning**
> This component is NOT accessible currently. I will address that in a follow-up PR.

## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/4656348/645c4a0d-db21-4df5-b7e6-1da5d10e844d

## Checklist
- [X] If this is a user-facing change, I have added an entry in CHANGELOG.md
- [X] Ensure the linter passes (`./gradlew spotlessApply` to automatically apply formatting/linting)
- [X] I have considered whether it makes sense to add tests for my changes
- [X] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml`
- [X] Any jetpack compose components I added or changed are covered by compose previews
 
#### I have tested any UI changes...

- [X] with different themes
- [X] with a landscape orientation
- [X] with the device set to have a large display and font size
- [ ] for accessibility with TalkBack ⚠️

PR #1476, opened 2023-10-16
Make rating stars component accessible

mchowning:
## Description
This updates the rating stars component to be accessible. This is a relatively minimal approach in that using TalkBack you are only able to select full stars. It would be good to improve this, but I think it'll be a bit more complicated to add in the half stars, and I didn't want to go too far down that rabbit hole at this point since I'll be updating this component in the future to possibly not allow 0 or half star ratings and to accept a star rating parameter so users can update their rating. When I work on those I'll see if it make sense to improve this behavior as well.

## Testing Instructions
1. Go to an episode with a rating and tap on the stars to open the Add Rating screen
2. Observe that the stars component still works as before (see https://github.com/Automattic/pocket-casts-android/pull/1468)
3. Turn on TalkBack
4. Observe that you can select the stars with TalkBack either by switching between all the interaction points on the page or by directly double tapping on one of the stars.
5. Selecting one of the stars should fill the stars for that full value.
6. Turn off TalkBack
7. Observe that the behavior has returned to the behavior you observed in Step 2

## Checklist

### DIFF
--- a/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/components/ratings/SwipeableStars.kt
+++ b/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/components/ratings/SwipeableStars.kt
@@ -1,7 +1,5 @@
 package au.com.shiftyjelly.pocketcasts.podcasts.view.components.ratings
 
-import android.content.Context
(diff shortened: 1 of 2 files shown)


ashiagr (commented):
Checked with talkback on/off and it works well 👍

ashiagr (approved):

PR #1477, opened 2023-10-17
Starting migrating libraries to the version catalog

geekygecko:
## Description

As part of our tech debt, this PR gives a small example of how the libraries will be migrated to the version catalog. If this is approved I will raise some more PRs to complete this.

## Testing Instructions
1. Full clean and build the app
2. ✅ Verify the app still runs

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -13,11 +13,14 @@
 android-application = "8.1.0"
 # @keep
 compose-kotlin-compiler = "1.3.0"
+firebase = "30.5.0"
 kotlin = "1.7.10"
 ksp = "1.7.10-1.0.6"
 hilt = "2.43.2"
 hilt-compiler = "1.0.0"
 google-services = "4.3.14"
+lifecycle = "2.6.0"
+paging = "3.1.1"
 sentry = "3.10.0"
 
 [libraries]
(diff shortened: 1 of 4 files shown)


gradle/libs.versions.toml
```diff
@@ -26,10 +29,56 @@ hilt-compiler =  { module = "androidx.hilt:hilt-compiler", version.ref = "hilt-c
...
+# Paging
+paging-runtime = { module = "androidx.paging:paging-runtime", version.ref = "paging" }
+paging-rxjava2 = { module = "androidx.paging:paging-rxjava2", version.ref = "paging" }
+
+# Individual dependencies
+
+browser = "androidx.browser:browser:1.4.0"
+cardview = "androidx.cardview:cardview:1.0.0"
+cast = "com.google.android.gms:play-services-cast-framework:21.2.0"
+constraintlayout = "androidx.constraintlayout:constraintlayout:2.1.4"
```
geekygecko, line 58: Rather than listing all the library versions above, if they aren't grouped or important I think we should leave them here. I tried this a previous time and the size of this grew very quickly and was hard to read.

gradle/libs.versions.toml
```diff
@@ -26,10 +29,56 @@ hilt-compiler =  { module = "androidx.hilt:hilt-compiler", version.ref = "hilt-c
...
+lifecycle-runtime = { module = "androidx.lifecycle:lifecycle-runtime-ktx", version.ref = "lifecycle" }
+lifecycle-runtime-compose = { module = "androidx.lifecycle:lifecycle-runtime-compose", version.ref = "lifecycle" }
+lifecycle-viewmodel = { module = "androidx.lifecycle:lifecycle-viewmodel-ktx", version.ref = "lifecycle" }
+lifecycle-viewmodel-compose = { module = "androidx.lifecycle:lifecycle-viewmodel-compose", version.ref = "lifecycle" }
+
+# Paging
+paging-runtime = { module = "androidx.paging:paging-runtime", version.ref = "paging" }
+paging-rxjava2 = { module = "androidx.paging:paging-rxjava2", version.ref = "paging" }
+
+# Individual dependencies
```
mchowning, line 53: Nitpick, but the word "Individual" threw me for a second because it briefly made me think the other dependencies were not individual. Maybe some other word like "Other" makes more sense here. I definitely do NOT feel strongly about this.

```suggestion
# Other dependencies
```

mchowning (approved):
This is looking great! 🤩

gradle/libs.versions.toml
geekygecko: Thanks for the suggestion, I struggled with naming that it. Other is great, I will use that.

Issue #1463, opened 2023-10-13
Pocket Casts Playback: 2023


ashiagr:
## Stories design

- [x] Intro Story (https://github.com/Automattic/pocket-casts-android/pull/1450)
- [x] Total episodes / shows (https://github.com/Automattic/pocket-casts-android/pull/1471)
- [x] Most listened show (https://github.com/Automattic/pocket-casts-android/pull/1480)
- [x] Top 5 shows (https://github.com/Automattic/pocket-casts-android/pull/1481)
- [x] Top categories (https://github.com/Automattic/pocket-casts-android/pull/1482)
- [x] Total time listening to podcasts (https://github.com/Automattic/pocket-casts-android/pull/1495)
- [x] Longest episode (https://github.com/Automattic/pocket-casts-android/pull/1497)
- [x] Epilogue (https://github.com/Automattic/pocket-casts-android/pull/1501)

## Plus

- [x] Listening hours comparison story (https://github.com/Automattic/pocket-casts-android/pull/1502)
- [x] Completion rate story (https://github.com/Automattic/pocket-casts-android/pull/1506)
- [x] Skip "Plus" stories (https://github.com/Automattic/pocket-casts-android/pull/1512)
- [x] Plus Upsell story (https://github.com/Automattic/pocket-casts-android/pull/1516)
- [x] Plus purchase flow (https://github.com/Automattic/pocket-casts-android/pull/1517)
- [x] Add logic to sync 2022 data (https://github.com/Automattic/pocket-casts-android/pull/1522)

## Entry point

- [x] Update the modal entry point (https://github.com/Automattic/pocket-casts-android/pull/1503)
- [x] Update the Profile entry point (https://github.com/Automattic/pocket-casts-android/pull/1503)

## Refinements
- [x] Update listened time font size (https://github.com/Automattic/pocket-casts-android/pull/1495#discussion_r1376655773) (https://github.com/Automattic/pocket-casts-android/pull/1528)
- [x] Fix listening time can get cut off on huge increase/ decrease (https://github.com/Automattic/pocket-casts-android/pull/1502#pullrequestreview-1706653781) (https://github.com/Automattic/pocket-casts-android/pull/1524)
- [x] Update the Pocket Casts watermark when exporting (https://github.com/Automattic/pocket-casts-android/pull/1524)

## Tracks

- [x] Double check Tracks (https://github.com/Automattic/pocket-casts-android/pull/1523)
[...]

PR #1479, opened 2023-10-17
EOY Intro screen tweaks

mchowning:
| 📘 Part of: https://github.com/Automattic/pocket-casts-android/issues/1463 | 🎨 Designs: lH66LwxxgG8btQ8NrM0ldx-fi-1599_20385 |
|:---:|:---:|

## Description
This follows up on https://github.com/Automattic/pocket-casts-android/pull/1450 and applies a few tweaks to the EOY intro screen:
1. It applies a slight parallax to each background number separately. This didn't end up making quite as big of a difference as I had hoped it would, but I do think it's a nice, small improvement. I didn't spend a ton of time tweaking the depthMultipliers for the different numbers, so there's quite possibly some room for improvement in the numbers we use there.
2. It positions the numbers using a constraint layout and allows two of the numbers to overlap like they do in the designs. The constraint layout seemed to give me more consistent positioning of the numbers relative to the center text (i.e., it does a better job of not overlapping the text too much).
3. It smooths out the sensor data that we use to apply parallax changes. Before it seemed like the parallax was fairly jumpy to me, and this seems better now. I am doing more work smoothing out the sensor data to accomplish this, but the performance still seems pretty solid on my low-end test device, and it looks really good on my Pixel 6 Pro. We can easily turn some of this smoothing off though, if it turns out to be an issue.
4. I updated the text on this screen from "End of Year Review" to "Pocket Casts Playback" based on the updated designs.

## Testing Instructions
** Test on a real device so you can test the parallax effect.

1. Make sure you're logged in into an account that has a few episodes listened
2. Go to Profile > Settings > Beta Features > enable End of Year
3. Go back to Profile
5. Tap the EOY image
6. ✅ Check that the Intro Story looks good
7. Tilt your device
8. ✅ The text and the numbers in the background should all move with a parallax effect.

## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/4656348/98752898-5448-4ccd-b9d5-eccd95a5ec7d

## Checklist


mchowning:
For next year, I think it might be nice to see if we could reposition the numbers so that the larger numbers are in front of (overlap) the text and the smaller numbers are behind it. I think that would make the parallax effect look a lot better since it would more naturally reflect distance compared to how we currently have the smaller numbers in front of the text and the larger numbers behind it, cc: @david-gonzalez-a8c .

ashiagr:
Looks really nice! Thanks for the improvements. 🙏

ashiagr (approved):

PR #1484, opened 2023-10-18
Migrating libraries to the version catalog

geekygecko:
## Description

This is the final part of the tech debt migration to move to a version catalog that Google recommends.

## Testing Instructions
1. Build and run the app
2. ✅ Verify there are no new warnings in the build output.
3. ✅ Verify there are no exceptions in logcat.
4. Login to an active sync account.
5. ✅ Verify the podcast and episode sync data is working. 
6. ✅ Verify the listen history is populated. (Check the Work Manager jobs are successful.)
7. ✅ Verify on the Profile tab that the refresh has been successful.
8. ✅ Verify there are no exceptions in logcat.


mchowning (approved):
Changes look good and I smoke tested release builds for the phone, automotive, and watch apps without observing any issues. Great work on this!

PR #1485, opened 2023-10-19
Remove the Gradle Versions Plugin

geekygecko:
## Description

With the new version catalog I believe we can remove the extra Gradle plugin to tell us what to update. Android Studio is doing a good job.

Screenshot 2023-10-19 at 9 26 01 pm

There is another Gradle plugin that helps upgrade libs.versions.toml files but maybe we should try without it first.
https://github.com/littlerobots/version-catalog-update-plugin

## Testing Instructions
Not much to test apart from building the app and making sure it runs.


mchowning (approved):
Thanks for cleaning this up. 🙇

PR #1487, opened 2023-10-19
Request exact alarm permission if it is missing

### LINKED ISSUE #1486: Not requesting SCHEDULE_EXACT_ALARM permission
### Description

Android now requires that user's explicitly grant permission for exact alarms and on an Android 14 emulator, I'm seeing that we are not able to sleep podcasts because of the missing permission.

We're currently catching the exception and asking the user to grant the alerms & reminders permission, but the way to do that is not at all obvious (to me at least). On stock Android you need to go to the Android system settings → `Apps` → `Special app access` → `Alarms & reminders` and then update the permission for Pocket Casts to "allowed".

It would be much better if we fired off the appropriate intent for the user to update this setting.

1. Fresh install the app on an Android 14 device
2. Play a podcast
3. Try to set up a sleep timer
4. Observe the toast message
5. Observe the sleep timer does not start and a toast message is displayed saying "Unable to start timer. Please allow 'Alarms & Reminders' permission".

_No response_

Google Pixel 7 emulator running Android 14

mchowning:
## Description
This updates the app so that instead of just presenting a Toast message asking the user to grant permission for exact alarms, we also prompt them with the screen for updating that permission. This is important because I'm seeing on the Android 14 emulator that the permission is not being granted by default and it is not particularly easy to find the place to update this permission for an app.

> **Note**
> In my testing, the permission was getting automatically granted on Android 13 devices, which surprised me (after reading [the developer documentation](https://developer.android.com/about/versions/14/changes/schedule-exact-alarms) I thought the permission would be denied by default on Android 13 as well, but apparently not). That's better for our users though, so I'm not complaining.

Fixes #1486 

### 1. Android 14
1. Fresh install the app on an Android 14 device
2. Try to start a sleep timer from the Full Screen player
3. Observe that you are taken to the settings screen for granting the alarm permission and also shown the toast message indicating we need the permission.
4. Grant the permission
5. Try setting a sleep timer again
6. Observe that the sleep timer works

### 2. Android <14
1. Fresh install the app on a device running Android <14
2. Try to start a sleep timer from the Full Screen player
3. Observe that the sleep timer works

## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/4656348/be1ca3df-ae2d-4f33-b6bc-7ca4fc02a3bf

## Checklist

### DIFF
--- a/modules/services/repositories/src/main/AndroidManifest.xml
+++ b/modules/services/repositories/src/main/AndroidManifest.xml
@@ -2,6 +2,8 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
 
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
+    <uses-permission android:name="android.permission.SCHEDULE_EXACT_ALARM" />
+
     <application
         android:usesCleartextTraffic="true">
         <provider
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SleepTimer.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SleepTimer.kt
@@ -5,6 +5,7 @@ import android.app.PendingIntent
 import android.content.Context
 import android.content.Intent
 import android.os.Build
+import android.provider.Settings.ACTION_REQUEST_SCHEDULE_EXACT_ALARM
 import android.text.format.DateUtils
 import android.widget.Toast
 import au.com.shiftyjelly.pocketcasts.utils.log.LogBuffer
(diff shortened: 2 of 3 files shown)


modules/services/repositories/src/main/AndroidManifest.xml
```diff
@@ -2,6 +2,8 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
 
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
+    <uses-permission android:name="android.permission.SCHEDULE_EXACT_ALARM" />
```
mchowning, line 5: I don't think we really have to have this since this is taken care of by the app module, but it is probably a good idea to include it here as well and removes a lint warning from Android Studio.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SleepTimer.kt
```diff
@@ -48,16 +49,23 @@ class SleepTimer @Inject constructor(@ApplicationContext private val context: Co
...
-            LogBuffer.e(LogBuffer.TAG_CRASH, e, "Unable to start sleep timer.")
-            if (Build.VERSION.SDK_INT >= 31 && !alarmManager.canScheduleExactAlarms()) {
-                Toast.makeText(context, LR.string.player_sleep_timer_start_failed, Toast.LENGTH_LONG).show()
-            }
+        return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S && !alarmManager.canScheduleExactAlarms()) {
+            Toast.makeText(context, LR.string.player_sleep_timer_start_failed, Toast.LENGTH_LONG).show()
+            context.startActivity(
+                Intent(ACTION_REQUEST_SCHEDULE_EXACT_ALARM)
+                    // Because we're not launching this from an activity context, we must add the FLAG_ACTIVITY_NEW_TASK flag
+                    .addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
```
mchowning, line 57: Instead of adding this flag we could pass an `Activity` instance (or a `startActivity: (Intent) -> Unit` function from an activity) to this method and then start this intent without the `FLAG_ACTIVITY_NEW_TASK` flag, but I don't think adding the flag will cause any harm in this circumstance and it keeps the code a bit cleaner so I went with this approach. Let me know if you think we're better off avoiding the flag.

ashiagr (approved):
Works well in my testing 👍

Issue #1499, opened 2023-10-30
Still signing in with Account Activity UI in a few places


mchowning:
### Description

We have migrated signing the user into the app to use our `OnboardingFlowComposable`, but there are still a few places in the app where we're using `AccountActivity`. In addition to not providing the full onboarding flow, the AccountActivity does not allow the user to sign in with Google. We should remove any remaining places where the user is prompted to sign in with the AccountActivity.

### Step-by-step reproduction instructions

1. Tap on the EOY upgrade banner
2. Observe that the sign in screen does not offer the option of signing in with Google.

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

Samsung A13, Pocket Casts 7.50 development build, Android 13.

PR #1500, opened 2023-10-30
Direct EOY banner to use onboarding flow

mchowning:
## Description
This makes it so that the EOY banner uses the onboarding flow instead of the old AccountActivity sign in screen.

Related to #1499 (but does not entirely fix that issue since we still have `AccountActivity` code in other places).

## Testing Instructions
1. Fresh install the app
2. Tap on the EOY banner on the profile screen
3. ✅ Observe that you are taken to the OnboardingFlow sign in screen (easy to tell because this screen has a Google Sign In option whereas the old AccountActivity screen does not provide that option).
4. Sign in
5. ✅ Observe that after sign in you are taken to the EOY stories
6. Exit the EOY stories
7. Tap on the EOY banner again
8. ✅  Observe that you are taken to the EOY stories

## Video

https://github.com/Automattic/pocket-casts-android/assets/4656348/3c861f97-a72c-46bd-8272-21658ad00eb7

## Checklist

### DIFF
--- a/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
+++ b/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
@@ -652,8 +652,7 @@ class MainActivity :
             showStories(StoriesSource.fromString(source))
         } else {
             viewModel.waitingForSignInToShowStories = true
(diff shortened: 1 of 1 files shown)


ashiagr (approved):
Works as described, thanks for this!

PR #1503, opened 2023-10-31
Playback 2023: Entry point

ashiagr:
| 📘 Part of: #1463 | 🎨 Designs: lH66LwxxgG8btQ8NrM0ldx-fi-1599_20385 |
|:---:|:---:|

## Description
Updates the logic and visuals for the Playback 2023 modal, profile badge, and profile card.

## Testing Instructions
1. Start the app with an account that has a few episodes listened
2. ✅ A modal should be shown
3. ✅  Tapping on "View My 2023" shows the stories
4. Dismiss it
5. ✅ There should be a badge on the "Profile" tab
6. Tap it
7. ✅ The badge should disappear
8. ✅ Check that there's a card on Profile prompting you to see your 2023
9. Tap it
10. ✅ Playback 2023 stories should start

If you want to reset the modal being shown and badge, go to Profile > Settings > Developer > Reset modal/profile badge

## Screenshots or Screencast 

Profile Entry | Profile Dot | Modal Entry
----|----|----
 |  | 

Tablet Profile Entry | Tablet Modal Entry 
-------|----
tablet_lt | tablet 

## Checklist
 
- [x] with different themes


modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/developer/DeveloperPage.kt
```diff
@@ -129,6 +128,19 @@ private fun TriggerUpdateEpisodeDetails(
...
+    onClick: () -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    SettingRow(
+        primaryText = "Reset modal/profile badge",
+        secondaryText = "Reset modal and profile badge for end of year",
+        icon = rememberVectorPainter(Icons.Outlined.EditCalendar),
+        modifier = modifier.clickable { onClick() }
+    )
+}
```
mchowning, line 142: Good idea! 👍

modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/Settings.kt
```diff
@@ -370,8 +370,8 @@ interface Settings {
     val sendCrashReports: UserSetting<Boolean>
     val linkCrashReportsToUser: UserSetting<Boolean>
 
-    fun setEndOfYearShowBadge2022(value: Boolean)
-    fun getEndOfYearShowBadge2022(): Boolean
+    fun setEndOfYearShowBadge2023(value: Boolean)
+    fun getEndOfYearShowBadge2023(): Boolean
 
     fun setEndOfYearModalHasBeenShown(value: Boolean)
     fun getEndOfYearModalHasBeenShown(): Boolean
```
mchowning, line 377: This is a minor thing, but I find it a bit confusing that for the modal we're storing whether it has been shown and for the badge we're storing whether it should be shown. It kind of feels like their opposite each other because if one of those settings returns true that means we show it, but for the other setting returns true that means we don't show it. If possible, I'd lean toward making them both return whether the badge/modal should be shown. Definitely not a blocker at all, and may be just my personal preference, so don't feel like you need to change it. It's very possibly not worth making the change at this point, so I'm more sharing just as food for thought.

app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
```diff
@@ -376,7 +376,7 @@ class MainActivity :
                             VR.id.navigation_profile -> {
                                 if (settings.getEndOfYearModalHasBeenShown()) {
                                     binding.bottomNavigation.removeBadge(VR.id.navigation_profile)
-                                    settings.setEndOfYearShowBadge2022(false)
+                                    settings.setEndOfYearShowBadge2023(false)
                                 }
```
mchowning, line 380: It seems that if the screen opens to the profile view, the badge is shown but tapping on the badge doesn't remove it. Instead you have to navigate away and then come back. Not a big deal or a blocker, but a mild nuisance.


https://github.com/Automattic/pocket-casts-android/assets/4656348/7e8af749-5b26-4490-9bbf-e89ffb0f60a2

modules/services/localization/src/main/res/values/strings.xml
```diff
@@ -1570,10 +1570,11 @@
     <string name="end_of_year_launch_modal_title">Your Year in Podcasts</string>
     <string name="end_of_year_launch_modal_image_title">Year in Podcasts</string>
     <string name="end_of_year_launch_modal_summary">See your top podcasts, categories, listening stats and more. Share with friends and shout out your favorite creators!</string>
-    <string name="end_of_year_launch_modal_primary_button_title">View My 2022</string>
+    <string name="end_of_year_launch_modal_primary_button_title">View My 2023</string>
     <string name="end_of_year_launch_modal_secondary_button_title">Not Now</string>
     <string name="end_of_year_prompt_card_title" translatable="false">@string/end_of_year_launch_modal_title</string>
-    <string name="end_of_year_prompt_card_summary">See your top podcasts, categories, listening stats and more.</string>
+    <string name="end_of_year_prompt_card_summary">See your listening stats, top podcasts, and more.</string>
+    <string name="end_of_year_pocket_casts_playback">Pocket Casts Playback</string>
```
mchowning, line 1577: David asked that we not translate this, so I think we want this to be marked as not translatable. 

Related: https://github.com/Automattic/pocket-casts-android/pull/1490

mchowning (approved):
Looks good. Just noted a few minor issues.

modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/Settings.kt
ashiagr: Yeah, it could be confusing, addressed in https://github.com/Automattic/pocket-casts-android/pull/1503/commits/ec7708f7c4d418575740bf90ad5fdbaaae0b37ab

modules/services/localization/src/main/res/values/strings.xml
ashiagr: Removed in https://github.com/Automattic/pocket-casts-android/pull/1503/commits/97ba624d8c70c0b187c31a94c1c556a7b3691b30

app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
ashiagr: Fixed in https://github.com/Automattic/pocket-casts-android/pull/1503/commits/2c84eceeb87f716128972250a9e55cd895898e

https://github.com/Automattic/pocket-casts-android/assets/1405144/66cc90a2-621b-4d53-9c7e-dc52fe05a41c

ashiagr:
Thanks for the review,  @mchowning! They were valid comments and addressed them all.
I'll merge this one to proceed to the next one, feel free to re-look when you can.

Issue #1509, opened 2023-11-01
App is inconsistent in whether it resets the  LocalPlayer after downloading an episode that is being streamed


mchowning:
### Description

I've noticed that when downloading a playing episode (i.e., if you start playing a new episode and you have auto download episodes added to up next turned on), once the download is complete, we usually, but not always reset the player.

I have only briefly looked into the issue, and I don't have time to look into it further, so I'm filing this issue as a reminder to follow-up on this when we have more time. At this point, it is not clear to me why we reset the player sometimes and not others.

It seems like either:
1. We should be resetting the player, and perhaps there are some negative consequences when we don't reset the player (although it's not clear to me right now what those are). I thought that maybe we were continuing to stream the podcast in that scenario, but that doesn't seem to be the case.
2. We do not need to be resetting the player every time we download and we can avoid doing that, which would be nice because it would let us avoid having the audio pause for approximately 1 second after the download completes while the player is being reset, or
3. There's a good reason we only need to reset the player some of the time, and I just haven't identified yet.

### Step-by-step reproduction instructions

1. Turn on the setting to auto download podcasts added to up next
2. Start playing a podcast that hasn't been downloaded
3. Observe that once the download completes, often, the logs show that we're resetting the player and there is an ~1 second pause in the audio. But observe that this does not always happen.

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

* Samsung A13, Android 13, Pocket Casts pre-release 7.52 build
* Pixel 4 emulator, Android 13, Pocket Casts pre-release 7.52 build

mchowning:
In my testing, it _seems_ that we either reset or do not reset the player repeatedly. In other words, if we reset the player on the first podcast you play, then it is more likely that we will keep resetting the player on subsequent podcasts you play. Likewise, if the app does not reset the player for a given podcast, it seems that will usually lead to subsequent podcasts not resetting the player as well. 

Basically, it does not appear that whether or not a given podcast resets the player when the download completes is independent of whether we reset the player on previous podcasts. Or, to put it another way, it seems like a given behavior tends to happen repeatedly, until it doesn't, and then the opposite behavior will happen multiple times in a row.

PR #1510, opened 2023-11-01
Avoid audio skip back when a streaming episode is downloaded

### LINKED ISSUE #1491: Playback skips back a half-second, once, every playback, about 30 seconds in
### Description

When playing back a podcast on PocketCasts, the app repeats about half a second of the podcast. This happens consistently for every podcast, about 30 seconds or so into the podcast (the exact time differs).

> If you want to know what it sounds like, it sounds a little like little like this.

This occurs on _every_ podcast, whether it has dynamic advertising or not.

1. Have playback speed set to 1.3x // trim silence set to "Mad Max" // volume boost set on
2. Do NOT auto-download the podcast
3. Find an episode to listen to, and press play. After a short time listening to the podcast, it will skip slightly backwards, and you'll hear a word or two repeated.

This is 100% reproducible, as above. It does NOT happen when an episode has been auto-downloaded.

What I think is happening is this: when you hit play, PocketCasts 'streams' the audio file, but also starts downloading a copy of the audio file to storage. When it has finished downloading it - which for my cell network takes about 30 seconds - it switches state from 'streaming' the audio file, to playing the audio file back from the download. However, it has failed to take into account the time saved by trimming silence (or maybe playback speed), and starts playing the downloaded file at the wrong place. It results in a second or two being repeated (and it might well be the case that this is the total amount of audio skipped in trim silence).

It's happened for the last three years of using PocketCasts on my Android phone; and it's only now that I've thought it might be worth reporting this as a bug.

(Interestingly, during a podcast with dynamically-inserted preroll ads, you can hear the skip happening within the ad.)

_No response_

Google Pixel, with Android 12/13/14, and all recent Pocket Casts versions.

But, specifically for today, Google Pixel 6 Pro, Android 14, Pocket Casts version 7.50-rc-1 (9157)

mchowning:
## Description
This fixes an issue where the download of an episode that is being streamed could cause playback to jump back a second or so.

When the app downloads an episode that is currently being streamed we usually (but [not always](https://github.com/Automattic/pocket-casts-android/issues/1509)), reset the player. When doing this, the new player was loading the last position of the episode that was saved to the database, which would often be out-of-date by a second or so.

This fixes that by (1) ensuring that we get the current playing position at the last possible moment before resetting the player, (2) updating the database with that playing position, and (3) ensuring that we fetch the latest episode data from the database before restoring the playback position.

It would be nice if we didn't have to manually refetch the episode from the db and if instead the `UpNextQueue` would always provide an up-to-date version of the currentEpisode, but that is a larger refactor that I don't have time to take on right now.

Note that this issue does not seem to occur _every_ time that we finish downloading an episode that is streaming because in some circumstances, we do not reset the player. I have not had time to isolate why that is, so I filed https://github.com/Automattic/pocket-casts-android/issues/1509 to track it.

Fixes #1491 

## Testing Instructions

1. Fresh install of the app
2. Begin playing an episode. Note that although the steps to reproduce in #1491 include setting various playback effects, I can reproduce the issue without those playback effects).
3. Start downloading the currently playing episode. You can either initiate this download manually or let the app start the download automatically by turning on the setting to download episodes added to the Up Next queue.
4. Observe that once the episode is downloaded, _if the player gets reset_ (which does not always happen, see [#1509](https://github.com/Automattic/pocket-casts-android/issues/1509)), the audio will pause for ~1 second and then resume as the player gets reset, and the audio should NOT jump backwards by ~1 second (i.e., there should not hear any audio repeated both before the pause and after the pause). Note that you can check the logs for "Resetting player" to confirm whether the player was reset.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 7.52
 -----
 
+*   Bug Fixes:
+    *   Avoid brief audio skip back when a streaming episode is downloaded
+        ([#1510](https://github.com/Automattic/pocket-casts-android/pull/1510))
 
 7.51
 -----
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
@@ -1602,18 +1602,36 @@ open class PlaybackManager @Inject constructor(
             }
         }
 
-        val chromeCastConnected = castManager.isConnected()
-        val currentPosition = player?.getCurrentPositionMs()
-        if (isPlayerSwitchRequired() || isPlayerResetNeeded(episode, sameEpisode, chromeCastConnected)) { // Don't create a player if we aren't playing because it will start to buffer
+        // We want to make sure we get the current position at the last possible moment before changing/resetting the player
+        val currentPositionMs = if (
+            isPlayerSwitchRequired() ||
+            isPlayerResetNeeded(episode, sameEpisode, castManager.isConnected())
+        ) {
+            // Don't create a player if we aren't playing because it will start to buffer
             if (play) {
+
                 Timber.d("Resetting player")
+                val playerPositionMs = player?.getCurrentPositionMs()
+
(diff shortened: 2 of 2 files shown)


ashiagr (approved):
LGTM

Hats off for the detective work! 🕵️

Issue #1511, opened 2023-11-02
Fix change email handling


ashiagr:
### Description

When a user navigates away from the `Change Email Address` screen while `/user/change_email` request is in progress, HTTP failed error is received and the new email address is not reflected in the app even though it gets changed on the server. It is not updated even after a forced refresh. The only way to see the new email is to log out and log in with the new email.

Failed response:
```
 I  --> POST https://api.pocketcasts.com/user/change_email
 I  Content-Type: application/json; charset=UTF-8
 I  Content-Length: 86
 I  Authorization: Bearer ACCESS_TOKEN
 I  {"email":"xxx@automattic.com","password":"xxx","scope":"mobile"}
 I  --> END POST (86-byte body)
 I  <-- HTTP FAILED: java.io.IOException: Canceled
```

### Reproduce behavior when email change is not reflected in the app

1. Login with an email address
2. Go to `Profile` -> `Account`
3. On the `Pocket Casts Account` screen, tap `Change email address`
4. On the `Change Email Address` screen, enter new email and confirm password
5. Tap `Confirm`
6. Quickly navigate away from the screen while the request is in progress
7. Notice that on the `Pocket Casts Account` screen, the email address is still the old one
8. On the `Profile` tab, tap `Refresh now`
9. Go back to `Pocket Casts Account` screen by tapping `Account` button on `Profile` tab
10. Notice that the email address is still the old one 

### Check if the email address is changed

11. Continue from above
12. Logout
13. Login with the new email address
14. Notice that the login is successful and the account has all data from the previous account confirming that the change email request was completed on the server 

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

suwasto:
In my testing, it seems that the ```disposables.clear()``` inside ```onCleared()```  in the ```ChangeEmailViewModel``` called so that the network request cancelled by client. the request already accepted and processed by server but the response lost due cancelation of http request on disposable cleared. Since we don't have any other network request in the ```ChangeEmailViewModel```  other than ```changeEmail()``` function, how about we remove ```disposable.clear()``` inside ```onCleared()``` method and move it after the request finished ex: ```doOnSuccess``` and ```subscribeBy(onError{})``` ?, will it cause memory leaks?

Another alternative I can think of is move the changeEmail to the activity view model ```AccountActivity``` so that the request is not canceled on navigate away from ```ChangeEmailFragment```

PR #1514, opened 2023-11-02
Migrating RxJava to Coroutines 

Mzazi25:
## Description

This PR migrates `emailChange` on the `SyncManager` to implement Coroutines.  Since the function was returning a `Single`, I migrated to use a suspend function. 
Also on the `ChangeEmailViewModel` , I changed `changeEmailState` to use a stateflow instead of livedata. I did this change since its a good step in the right direction to implement coroutines. 
I would be glad to proceed with more migrations to coroutines if this PR is accepted. 

Fixes # 
N/A
## Testing Instructions
1. Go to Profile 
2. Tap on Account
3. Navigate to Change Email Address
4. Confirm all inputs work as before just better this time with Coroutines 😄 

## Checklist
 
- [ ] with different themes


ashiagr (changes requested):
Thanks, @Mzazi25! You're on the right track, let's just handle two more cases and we should be ready to merge this PR:

### No internet crash

1. Navigate to `Change Email Address`
2. Disable internet
3. Add new email, pwd
4. Tap Confirm
5. App crashes 💥 

### Navigating away from the screen while the request is not complete 

1. Repeat steps 1,3,4 from above
2. Quickly navigate away from the screen while the request is not complete
3. Notice that the request is canceled with logs:

```
--> POST https://api.pocketcasts.com/user/change_email
...
 I  <-- HTTP FAILED: java.io.IOException: Canceled
```

4. Return to the `Profile` tab,
5. Notice that you still see the old email
6. Log out
7. Log in with old email
8. Notice you see "Email not found" error

This was noticed in https://github.com/Automattic/pocket-casts-android/issues/1511 and fixed in the linked PR but it seems we're back to the old behavior.

Mzazi25:
@ashiagr let me look further into this and will make necessary adjustments. 
Thanks for this update, I hadn't tested these scenario.

ashiagr (approved):

PR #1522, opened 2023-11-06
Playback 2023: Sync 2022 and 2023 listening history

ashiagr:
| 📘 Part of: #1463 |
|:---:|

This PR adds logic for syncing 2022 and 2023 data.

#### Test.1 Plus user  

1. Clean install debug build
2. Make sure you're logged in with a plus account that has a few episodes listened to this year and last year
3. Open stories
4. ✅ You should see in logs `End of Year: Downloading listening history for year xxx` for both 2022, 2023
6. Go to the "Year over Year" story
7. Notice you do not see any upsell button
8. Tap next
9. ✅ You should see the completion rate story

#### Test.2 Free user - Purchase from stories (release build) - triggers purchase flow

Prerequisites
- Enable EoY Year for release build
- Release build
- License test account 

 _(See Test.2 (Optional) if you want to test it on debug build)_

1. Clean install release build
2. Make sure you're logged in to a license account as a free user that has a few episodes listened to this year and last year
3. Open stories
4. Go to the "Year over Year" story
5. Purchase Plus or Patron
6. ✅ You should see a screen with a spinner, meaning your 2022 data is syncing
7. After it finishes, you should see the "Year over Year" story
8. Tap next
9. ✅ You should see the completion rate story

#### Test.3 (Optional) Free user - Purchase from stories (debug build)  - does not trigger purchase flow but you can test syncing

Prerequisites
- Apply [this patch](https://github.com/Automattic/pocket-casts-android/files/13265366/test_syncing.patch)

1. Clean install debug build
2. Make sure you're logged in with a free account that has a few episodes listened to this year and last year
3. Open stories
4. ✅ You should see in logs `End of Year: Downloading listening history for year xxx` for 2023 only
5. Wait for 2023 sync to complete
6. Go to the "Year over Year" story
7. Add Plus plan to the account using admin console
8. Tap upsell button
9. ✅ You should see in logs `End of Year: Downloading listening history for year xxx` for 2023, 2022. Since 2023 data is already synced at step 6, 2022 data should be synced now.
10. ✅ You should see a screen with a spinner, meaning your 2022 data is syncing
11. After it finishes, you should see the "Year over Year" story
12. Tap next
13. ✅ You should see the completion rate story

https://github.com/Automattic/pocket-casts-android/assets/1405144/6d13ba17-ff2b-482c-8d3b-532acd9e72cb

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/StoriesViewModel.kt
+++ b/modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/StoriesViewModel.kt
@@ -68,6 +68,7 @@ class StoriesViewModel @Inject constructor(
 
     private val currentStoryIsPlus: Boolean
         get() = stories.value[currentIndex].plusOnly
+    private var currentUserTier: UserTier = UserTier.Free
     private var manuallySkipped = false
 
     init {
@@ -81,12 +82,17 @@ class StoriesViewModel @Inject constructor(
             val onProgressChanged: (Float) -> Unit = { progress ->
                 mutableState.value = State.Loading(progress)
             }
-            endOfYearManager.downloadListeningHistory(onProgressChanged = onProgressChanged)
-            stories.value = endOfYearManager.loadStories()
             combine(
                 subscriptionManager.freeTrialForSubscriptionTierFlow(Subscription.SubscriptionTier.PLUS),
                 settings.cachedSubscriptionStatus.flow
             ) { upsellState, _ ->
+                currentUserTier = settings.userTier
+                val lastUserTier = (state.value as? State.Loaded)?.userTier
+                if (lastUserTier == this@StoriesViewModel.currentUserTier) return@combine
+
+                endOfYearManager.downloadListeningHistory(onProgressChanged = onProgressChanged)
+                stories.value = endOfYearManager.loadStories()
+
                 updateState(upsellState)
                 if (state.value is State.Loaded) start()
             }.stateIn(this)
@@ -108,7 +114,7 @@ class StoriesViewModel @Inject constructor(
                     xStartOffsets = List(numOfStories) { getXStartOffsetAtIndex(it) },
                     widths = storyLengthsInMs.map { it / totalLengthInMs.toFloat() },
                 ),
-                userTier = settings.userTier,
+                userTier = currentUserTier,
                 freeTrial = freeTrial,
             )
         }
@@ -242,7 +248,7 @@ class StoriesViewModel @Inject constructor(
     }
 
     private fun isPaidUser(): Boolean {
-        val currentState = state.value as State.Loaded
+        val currentState = state.value as? State.Loaded ?: return false
         return currentState.userTier != UserTier.Free
     }
 
--- a/modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/StoriesViewModel.kt
+++ b/modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/StoriesViewModel.kt
@@ -68,7 +68,6 @@ class StoriesViewModel @Inject constructor(
 
     private val currentStoryIsPlus: Boolean
         get() = stories.value[currentIndex].plusOnly
-    private var currentUserTier: UserTier = UserTier.Free
     private var manuallySkipped = false
 
     init {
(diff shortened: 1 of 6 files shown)


modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/StoriesViewModel.kt
```diff
@@ -108,7 +114,7 @@ class StoriesViewModel @Inject constructor(
                     xStartOffsets = List(numOfStories) { getXStartOffsetAtIndex(it) },
                     widths = storyLengthsInMs.map { it / totalLengthInMs.toFloat() },
                 ),
-                userTier = settings.userTier,
+                userTier = currentUserTier,
```
mchowning, line 117: Instead of storing `currentUserTier` as a field, could we pass it as an argument in `updateState`? It looks like we only call `updateState` from one place (where we also set the `currentUserTier` field, so I think this would be an easy way to avoid having a field variable. Feel free to leave as-is if you prefer though.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/endofyear/EndOfYearManagerImpl.kt
```diff
@@ -59,48 +71,79 @@ class EndOfYearManagerImpl @Inject constructor(
         hasEpisodesPlayedUpto(YEAR, TimeUnit.MINUTES.toSeconds(EPISODE_MINIMUM_PLAYED_TIME_IN_MIN)) &&
             FeatureFlag.isEnabled(Feature.END_OF_YEAR_ENABLED)
 
-    /**
-     * Download the year's listening history.
-     */
+    private var synced: Double = 0.0
+    private var syncTotal: Double = 0.0
```
mchowning, line 75: Did you mean to put these above the `isEligibleForStories` function?

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/endofyear/EndOfYearManagerImpl.kt
```diff
@@ -59,48 +71,79 @@ class EndOfYearManagerImpl @Inject constructor(
...
             onProgressChanged(0.2f)
             val history = response.history ?: return
             historyManager.processServerResponse(
                 response = history,
                 updateServerModified = false,
+                updateSyncTotal = { syncTotal += it },
                 onProgressChanged = {
-                    onProgressChanged(0.2f + (it * 0.8f))
+                    synced += 1
+                    val progress = min((0.2f + (synced / syncTotal) * 0.8f), 0.95).toFloat()
```
mchowning, line 124: Just double checking, but this `min` call means that the number will never get over `0.95`, right? It might be worth a comment explaining why we want that.

mchowning, line 124: If it's feasible, it would be nice if this math happened inside the history manager (or inside a helper class) and instead of having both the `updateSyncTotal: (Float) -> Unit)` and `onProgressChanged: () -> Unit` parameters, the history manager took care of the calculations itself and just accepted a single `onProgressChanged: (percentChange: Float) -> Unit` parameter. That's certainly not a blocker though.

mchowning (approved):
Working perfectly, and the code looks good. I just noted a few non-blocking code changes I thought you might want to consider. This can be merged as-is though.

modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/StoriesViewModel.kt
ashiagr: Nice pick! Suggestion applied in d0ffd6c9ee1a10863613ab86f8862b9bd2947fbe

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/endofyear/EndOfYearManagerImpl.kt
ashiagr: Yes 🤦‍♀️. These are not needed in `EndOfYearManagerImpl` anymore as I moved them to `HistoryManager` in https://github.com/Automattic/pocket-casts-android/pull/1522/commits/e5d3b1e0b921b520556b300aabb2d06a3d518ebc.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/endofyear/EndOfYearManagerImpl.kt
ashiagr: Added a comment here: https://github.com/Automattic/pocket-casts-android/pull/1522/commits/e5d3b1e0b921b520556b300aabb2d06a3d518ebc#diff-d4ead7d8903c1b9b64124a86579385d509abee41144951e6ed7f0a2b1f32a2efR78-R80

ashiagr: You're right. I moved the calculation to `HistoryManager` in https://github.com/Automattic/pocket-casts-android/pull/1522/commits/e5d3b1e0b921b520556b300aabb2d06a3d518ebc.

PR #1526, opened 2023-11-07
Enable bookmarks with Patron early access in 7.52

mchowning:
## Description
This enables bookmarks for entering Patron early access in `7.52` and for full release in `7.53`.

## Testing Instructions

I think it's worth double-checking that our logic is hitting all the right states for the right releases now that we know the release version. Using release builds with the following versions (specified in [versions.properties](https://github.com/Automattic/pocket-casts-android/blob/main/version.properties)), you should see the following behavior for bookmarks:

| App version | Not signed in | Free          | Plus          | Patron    |
| ---| ---- | ----- | ------ | ------|
| `7.52-rc-1`   | Plus upsell   | Plus upsell   | Available     | Available |
| `7.52`        | Patron upsell | Patron upsell | Patron upsell | Available |
| `7.53-rc-1`   | Plus upsell   | Plus upsell   | Available     | Available |
| `7.53`        | Plus upsell   | Plus upsell   | Available     | Available |

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 7.52
 -----
 
+* New Features:
+    *   Bookmarks entering Patron early access with full release for Plus users in 7.53
+        ([#1526](https://github.com/Automattic/pocket-casts-android/pull/1526))
 *   Bug Fixes:
     *   Avoid brief audio skip back when a streaming episode is downloaded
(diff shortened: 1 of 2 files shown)


ashiagr (approved):
LGTM

I also enabled remote feature flag for the feature.

PR #1535, opened 2023-11-10
Fixes Multiselect: Does not work in some cases(#1507)

### LINKED ISSUE #1507: Multiselect: Does not work in some cases
### Description

If I select an episode of a podcast from the "Up Next" list, the multi-select function doesn't work. However, if I back out to the main podcast list and select an episode from there, the multi-select function works perfectly.

1. Tap on the Up Next List
2. Tap on any episode in the list
3. Tap on the podcast page link from the episode info
4. Long press on an episode in the Podcast Page when it loads to select it
5. Long press on the episode again to bring up the Select all menu

At this point, selecting any of the three options listed has no effect. Additionally, selecting "Select All" does not select anything.

https://github.com/Automattic/pocket-casts-android/assets/16253818/bc838aae-48ed-4c45-80d5-c5963199d079

Samsung Fold 4, Android 13, Pocket Casts v7.50

AnuragDevMastermind:
## Description
The problem with current navigation system is that if there are two fragment A and B and we are changing fragment from fragment A to fragment B using `navigator.addFragment` function of MainActivity then onDestroy of A is running after onCreate of B which is not expected 

From the above observation i found that
- The `onDestroy` of `UpNextFragment` is nullifying the multiselect listener of `PodcastFragment` when navigating from `UpNextFragment` to `PodcastFragment`.
- when navigating from `PodcastFragment`  to `PodcastFragment`, when `podcastUuid` is different, is also making the multiselect listener as null

## Solution
Whenever `onRowLongPress` is invoked I am attaching the multiselect listeners again in `PodcastFragment`

Fixes #1507

## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/134276544/d7fef0fa-a5b1-47f9-8b08-2a58ee56a637

## Checklist
 
- [ ] with different themes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
         ([#1518](https://github.com/Automattic/pocket-casts-android/pull/1518))
     *   Fix Chapter Length Calculation Mismatch
         ([#1525](https://github.com/Automattic/pocket-casts-android/pull/1525))
+    *   Fix multiselect not working in some cases
+        ([#1535](https://github.com/Automattic/pocket-casts-android/pull/1535))
 7.51
 -----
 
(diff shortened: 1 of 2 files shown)


ashiagr (approved):

CookieyedCodes:
I think fixing this also fixed much of my swipe lag ❤️❤️@mchowning

devlopersabbir (commented):
seems looks like great work :)

PR #1546, opened 2023-11-15
Bookmarks - Fix user bookmarks sync when feature is disabled

ashiagr:
## Description
This fixes "user bookmark" sync when bookmark feature is disabled. (p1699988968426399-slack-C02A333D8LQ)

If a user bookmark is synced using a build with the feature enabled, then syncing fails on a build where the feature is disabled and gives this error:

```
SyncProcess: Sync failed
com.squareup.moshi.JsonDataException: Expected a name but was BEGIN_OBJECT at path $.result.changes[0].fields
    at com.squareup.moshi.i.m(SourceFile:96)
    at au.com.shiftyjelly.pocketcasts.servers.sync.update.SyncUpdateResponseParser.m(SourceFile:20)
    at au.com.shiftyjelly.pocketcasts.servers.sync.update.SyncUpdateResponseParser.t(SourceFile:50)
    at au.com.shiftyjelly.pocketcasts.servers.sync.update.SyncUpdateResponseParser.fromJson(SourceFile:127)
 ```

This PR skips parsing user bookmark when the feature is disabled.

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/sync/update/SyncUpdateResponseParser.kt
+++ b/modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/sync/update/SyncUpdateResponseParser.kt
@@ -91,10 +91,10 @@ class SyncUpdateResponseParser(
             "UserFolder" -> readFolder(reader, response)
             "UserPodcast" -> readPodcast(reader, response)
             "UserEpisode" -> readEpisode(reader, response)
-            "UserBookmark" -> if (!featureFlagWrapper.isEnabled(Feature.BOOKMARKS_ENABLED)) {
-                reader.skipValue()
-            } else {
+            "UserBookmark" -> if (featureFlagWrapper.isEnabled(Feature.BOOKMARKS_ENABLED)) {
                 readBookmark(reader, response)
+            } else {
+                reader.skipValue()
             }
             null -> throw Exception("No type found for field")
             else -> reader.skipValue()
--- a/modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/sync/update/SyncUpdateResponseParser.kt
+++ b/modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/sync/update/SyncUpdateResponseParser.kt
@@ -91,7 +91,11 @@ class SyncUpdateResponseParser(
             "UserFolder" -> readFolder(reader, response)
             "UserPodcast" -> readPodcast(reader, response)
             "UserEpisode" -> readEpisode(reader, response)
-            "UserBookmark" -> readBookmark(reader, response)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by pocketcasts** [comment]
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-android/milestone/39">7.52 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

File: modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/sync/update/SyncUpdateResponseParser.kt [hunk 1]
```diff
@@ -91,7 +91,11 @@ class SyncUpdateResponseParser(
             "UserFolder" -> readFolder(reader, response)
             "UserPodcast" -> readPodcast(reader, response)
             "UserEpisode" -> readEpisode(reader, response)
-            "UserBookmark" -> readBookmark(reader, response)
+            "UserBookmark" -> if (featureFlagWrapper.isEnabled(Feature.BOOKMARKS_ENABLED)) {
+                readBookmark(reader, response)
+            } else {
+                reader.skipValue()
+            }
```
mchowning, line 98: I would be inclined to include keep the feature check if-else logic in the `readBookmark` method myself like it was before. I don't think we'd ever want to read the bookmark if the feature was disabled, and that would keep the different branches of this when statement more consistent. Definitely not a big deal though. Moving it here is fine if you prefer.

mchowning (approved):
This works well. If I'm understanding correctly though, users on old versions of the app (with bookmarks disabled) will still have syncing fail if they have added bookmarks to a user episode, and this PR just helps us if we have to turn bookmarks off in the future for some reason, right?

I tested this using the following steps:

1. Installed `7.52-rc-1`
2. Created a bookmark on a user uploaded episode
3. Updated the code for `7.52-rc-1` so bookmarks feature was turned off
4. Installed the modified `7.52-rc-1` build with that change
5. Observing that sync failed (reproducing the bug)
6. Installed build from this PR, but updated so that the bookmarks feature is turned off
7. Observed that sync succeeded (so the fix works)
8. Installed an unmodified build from this PR
9. Confirmed that the bookmark on the user episode shows up, and that clicking on the bookmark moves playback to the appropriate position in the episode

I'm not merging since the PR is tagged with `do not merge`.

ashiagr:
>  users on old versions of the app (with bookmarks disabled) will still have syncing fail if they have added bookmarks to a user episode, and this PR just helps us if we have to turn bookmarks off in the future for some reason, right?

That's correct.

PR #1547, opened 2023-11-15
Merge release/7.52-rc-2 to main

ashiagr:
Binary has been submitted for Google review.


.configure
```diff
@@ -1,7 +1,7 @@
 {
   "project_name": "pocketcasts-android",
   "branch": "trunk",
-  "pinned_hash": "d3e500e4dc78d4097c4fa119565c69333bc1720d",
+  "pinned_hash": "d2b5c069b68218b0dd86c75e0e06b594d36b97ec",
```
mchowning, line 4: Do we know why this needed to change? Looks like the last time we changed it was about half a year ago when we made changed something about the google-services file. I wonder if there's any chance this could have some relationship to the wear os issue that got reported on our most recent submission. 🤔

ashiagr: Yes, that change was not in `7.52-rc-1` so it shouldn't have caused the issue. I pulled the latest secrets but it wasn't necessary since we made no changes. The pinned-hash just points to the latest commit from mobile-secrets.

mchowning (approved):

PR #1548, opened 2023-11-15
Fix file deletion when the episode is marked as played

### LINKED ISSUE #1248: Automatic delete settings for files do not work when an episode is marked as played

rodrigoborgesdeoliveira:
## Description
This PR fixes an issue where a file would not be deleted (while the "delete local file" setting is enabled) when marking the episode as played, either individually or through the multiple selection option.

- Fixes #1248

## Proposed solution
When marking a single episode or multiple episodes as played, call the `UserEpisodeManager.deletePlayedEpisodeIfReq` method for each file. This is a straightforward solution, with minimal changes.

Currently it seems that the same "mark as played" action is handled in a couple of different places, so further improvements can be made by having a "mark as played" method in the user episode manager, for example, and using it everywhere we need to do the actions resulting from marking an episode as played.

## Testing Instructions
See #1248.

## Screenshots or Screencast 
See #1248.

## Checklist


ashiagr (approved):
Nice job! This works well in my testing 👍  

> further improvements can be made by having a "mark as played" method in the user episode manager, for example, and using it everywhere we need to do the actions resulting from marking an episode as played.

Makes sense to me. This can be taken up in a separate PR. Another way to address it could be to add a `markAsPlayed` in user manager and invoke it from episode manager's `markAsPlayed` similar to how it is done for `markAllAsPlayed` 

```
userEpisodeManager.markAllAsPlayed(episodes.filterIsInstance<UserEpisode>(), playbackManager)
```

It can be done in this PR itself if you wish. I'll approve this PR but wait to merge it in case you want to make this change. 

P.S.> I updated changelog to include it in v7.54. Sorry about the delay in reviewing it. 🙏

rodrigoborgesdeoliveira:
Thank you, @ashiagr. I'll address the improvements in a separate PR later this week.

PR #1544, opened 2023-11-14
Bookmarks - Fix add bookmark not showing at times

ashiagr:
## Description

Fixes add bookmark not showing at times.

** This also adds a small delay when playing at a bookmark position when the selected episode is loading or switching from a different one.

#### Test.1. Add bookmark
1. Install release build from `release/7.51` branch
2. Play an episode and open full-screen player
3. From the bottom shelf, open more option
4. Rearrange a few shelf items
5. Install release build from this branch
6. Play an episode and open full-screen player
7.  From the bottom shelf, open more option
8. ✅ Notice you see the Add bookmark option

#### Test.2 Confirm bookmark position for archived episode

1. Continue from above
2. Add a bookmark for the selected episode
3. (Optionally) Archive it
4. Play an episode from another podcast
5. Select the previous episode to which bookmark was added
6. Tap the bookmark play button for the episode
7. ✅ Notice that the bookmark episode is selected and the bookmark position is correctly seeked

## Checklist
- [ ] If this is a user-facing change, I have added an entry in CHANGELOG.md
- [ ] Ensure the linter passes (`./gradlew spotlessApply` to automatically apply formatting/linting)
- [ ] I have considered whether it makes sense to add tests for my changes
- [ ] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml`
- [ ] Any jetpack compose components I added or changed are covered by compose previews
 
#### I have tested any UI changes...

- [ ] with different themes
- [ ] with a landscape orientation
- [ ] with the device set to have a large display and font size
- [ ] for accessibility with TalkBack

PR #1550, opened 2023-11-16
Use suspend method when playing bookmarks

mchowning:
## Description
This updates the call to play a bookmark to use the suspended `playNowSync` function instead of the asynchronous `playNow` function. This helps make sure that the episode is loaded before we update the playback position. 

With this change I don't think we need the manual delay anymore, but let me know what you think @ashiagr since you're more familiar. We could leave the delay in if you think that is needed.

Without this change, I was seeing that playing a bookmark from an archived episode had a race condition where one of three things would occur:

1. The bookmarked episode is loaded and the playback position on the episode is properly set. This is the happy path.
2. The bookmarked episode is loaded, but the playback position is not set. As a result, the episode is played from the beginning.
3. The bookmark playback position is set on the currently playing episode _before the bookmarked episode is loaded_. I only saw this happen a couple of times. This could cause the currently playing episode to be completed and archived if the bookmark position is beyond the end of the episode.

This was reported in the [Pocket Casts Beta Slack](https://pocketcastsbeta.slack.com/archives/C013F9S9C2W/p1699972221471819?thread_ts=1699923175.314329&cid=C013F9S9C2W), and builds on the changes in https://github.com/Automattic/pocket-casts-android/pull/1544

I do think it is probably still technically possible to get in a race condition scenario because even though `playNowSync` is a suspended method, it launches a new coroutine in the `onAdd` callback that it passes to the `upNextQueue.playNow` method. Changing that would be tricky and have wide-ranging impact, and I didn't see this causing any race condition issues in my testing, so I don't think it makes sense to changing that now.

> [!CAUTION] 
> This PR is targeting the `release/7.52` branch.

### 1. Avoids playback race conditions 
Run through these steps a few times. Since the problem this is fixing is a race condition, it does not consistently occur.

1. Add some bookmarks to an episode
2. Archive that episode
3. [optional] Force close and restart the app (this makes it take longer to reload the bookmarked episode)
4. Start playing some other episode
5. Go to the bookmarks for this episode on the podcast screen or on the episode screen (run through these test steps with both screens)
6. Tap play on one of the bookmarks for the episode you archived
7. Observe that the bookmarked episode is loaded and the correct playback position is set

### 2. Playing bookmarks from full screen player still works

Tap the play button on a few bookmarks in the full-screen player to verify that this still works as before.

## Checklist

### DIFF
--- a/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/viewmodel/BookmarksViewModel.kt
+++ b/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/viewmodel/BookmarksViewModel.kt
@@ -33,7 +33,6 @@ import dagger.hilt.android.lifecycle.HiltViewModel
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.cancelChildren
-import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
@@ -43,8 +42,6 @@ import kotlinx.coroutines.flow.flatMapLatest
 import kotlinx.coroutines.flow.stateIn
 import kotlinx.coroutines.launch
 import javax.inject.Inject
-
-private const val DELAY_IN_MS = 300L
 
 @HiltViewModel
 class BookmarksViewModel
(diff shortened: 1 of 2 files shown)


pocketcasts:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pocketcasts:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-android/milestone/39">7.52 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

ashiagr:
This works much better!

> I do think it is probably still technically possible to get in a race condition scenario because even though playNowSync is a suspended method, it launches a new coroutine in the onAdd callback that it passes to the upNextQueue.playNow method. Changing that would be tricky and have wide-ranging impact, and I didn't see this causing any race condition issues in my testing, so I don't think it makes sense to changing that now.

I agree, let's not make any changes to that part now. 

> With this change I don't think we need the manual delay anymore, but let me know what you think @ashiagr since you're more familiar. We could leave the delay in if you think that is needed.

It's ok to remove the delay. I could not reproduce the race condition without it.

ashiagr (approved):
This works much better. I wasn't able to reproduce race condition in my testing

PR #1561, opened 2023-11-21
Fetch latest episode url before playing episode

### LINKED ISSUE #272: Patreon podcasts won't play on Android Automotive and Chromecast
A user in 5514912-zen reported that Patreon podcasts won't play on Android Automotive and Chromecast.

On Chromecasting an episode from a Patreon podcast, 
> The pocket casts app launches, shows the album art but after a few seconds goes back to the splash screen.

The logs contain following error:
> can't startForeground as the notification is null

##### This could be a device specific issue as the user is able to play the episode on nest mini but could not on nest hub.

mchowning:
## Description
This has the app check the server for updated episode URLs before playing an episode. This avoids a problem where patreon podcasts would fail to play because the app had an expired URL. This was particularly problematic on the Automotive app because on the Automotive app there was no way to update the URL for an existing episode (refresh would just pull in new episodes). This could also be a problem on the phone app though as the only way to update the URL for an existing episode is to go to the screen for the podcast, which automatically updates all the episodes automatically (but this isn't made apparent to the user in any way).

I initially thought about only getting the URL after we see playback fails, but iOS is fetching all the episode data for a podcast before playback starts without problems apparently, and it turns out that the Android app used to fetch episode data before playback as well. It's not clear why that functionality was removed, and I don't want to download all the episode data before every playback when all we need at that point is the URL, so I'm using a more specific endpoint that only returns the episode's URL and nothing else. In the future, we can consider when/how we want to do a better job of more generally updating all the episodes for a podcast.

Fixes #272 

> [!NOTE]
> Because this fix touches our core playback logic, I have it targeting the `main` branch instead of being a betafix this late in the release process. Let me know if you feel differently.

## Testing Instructions

You'll need to use the information for a private Patreon podcast to test this. Relvant information for a private patreon podcast is in [this google doc](https://docs.google.com/document/d/1HWjWlkiofoEwytbQ5ger4rmefAv5FRI6uvDD6qK42Xc/edit).

### 1. Testing an Expired URL

Please perform the following tests on both the phone app and an Android Automotive build

#### 1.a. Verify the test episode works normally and fails when the expired URL is injected when building from `main`

1. Fresh install a version of the app build **from the `main` branch**
3. Subscribe to the private patreon podcast from the [google doc](https://docs.google.com/document/d/1HWjWlkiofoEwytbQ5ger4rmefAv5FRI6uvDD6qK42Xc/edit)
4. Verify that you can stream the "“Apps for Noteworthy Events” episode from that podcast without any issues
5. Begin playing a different episode and pause playback
6. Inject an expired URL for the "Apps for Noteworthy Events" episode into the app's database by following the directions from the [google doc](https://docs.google.com/document/d/1HWjWlkiofoEwytbQ5ger4rmefAv5FRI6uvDD6qK42Xc/edit)
7. Do NOT go to the podcasts screen on your phone after injecting the expired URL because that could cause the URL to be updated.
8. Attempt to play the "Apps for Noteworthy Events" episode again and observe that playback now fails

#### 1.b. Verify that this PR fixes playback when there is an expired url
1. Fresh install a version of the app build from **THIS PR**
2. Subscribe to the private patreon podcast from the [google doc](https://docs.google.com/document/d/1HWjWlkiofoEwytbQ5ger4rmefAv5FRI6uvDD6qK42Xc/edit)
3. Inject an expired URL for the "Apps for Noteworthy Events" episode into the app's database by following the directions from the [google doc](https://docs.google.com/document/d/1HWjWlkiofoEwytbQ5ger4rmefAv5FRI6uvDD6qK42Xc/edit)
4. Do NOT go to the podcasts screen on your phone after injecting the expired URL because that could cause the URL to be updated.
5. Attempt to play the "Apps for Noteworthy Events" episode and observe that the app fetches and uses the new URL (you can see that in the logs) and playback succeeds.

### 2. Other Scenarios to smoke test

Because this change touches our core playback logic, I think it makes sense to verify that other playback scenarios are still working as expected.

#### 2.a. Phone

1. Test a recent episode that has a working download URL that does not need to be updated (the most recent episode from the test Patreon podcast will work here)
7. Verify that you can do the following without crashing the app or seeing any other unexpected behavior:
    1. Trying to stream an episode in airplane mode
    2. Playing a downloaded episode when online
    8. Playing a downloaded episode when not online
    9. Playing user episode not uploaded to cloud and only stored locally
    10. Playing a user episode uploaded to cloud and not stored locally

#### 2.b. Automotive

1. Test a recent episode that has a working download URL that does not need to be updated (the most recent episode from the test Patreon podcast will work here)
2. Verify that you can do the following without crashing the app or seeing any other unexpected behavior:
    1. Playback works for user episode uploaded to cloud and not stored locally
    2. Playing an episode when not online doesn't crash

## Checklist

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
@@ -1503,19 +1503,29 @@ open class PlaybackManager @Inject constructor(
         if (episode.isFinished) {
             episodeManager.markAsNotPlayed(episode)
         }
-        // check we have the latest episode url in the background
-        if (episode is PodcastEpisode) {
-            updateEpisodeUrl(episode)
-        }
 
-        if (episode is UserEpisode && episode.serverStatus == UserEpisodeServerStatus.UPLOADED) {
-            try {
-                val playbackUrl = userEpisodeManager.getPlaybackUrl(episode).await()
-                episode.downloadUrl = playbackUrl
-            } catch (e: Exception) {
-                onPlayerError(PlayerEvent.PlayerError("Could not load cloud file ${e.message}"))
-                removeEpisode(episode, source = sourceView)
-                return
+        // make sure we have the latest episode url
+        when (episode) {
+            is PodcastEpisode -> {
+                if (!episode.isDownloaded) {
+                    val newDownloadUrl = episodeManager.updateDownloadUrl(episode)
+                    if (newDownloadUrl != null && newDownloadUrl != episode.downloadUrl) {
+                        Timber.i("Updating url for PodcastEpisode playback")
+                        episode.downloadUrl = newDownloadUrl
+                    }
+                }
+            }
+            is UserEpisode -> {
+                if (episode.serverStatus == UserEpisodeServerStatus.UPLOADED) {
+                    try {
+                        val newDownloadUrl = userEpisodeManager.getPlaybackUrl(episode).await()
+                        episode.downloadUrl = newDownloadUrl
+                    } catch (e: Exception) {
+                        onPlayerError(PlayerEvent.PlayerError("Could not load cloud file ${e.message}"))
+                        removeEpisode(episode, source = sourceView)
+                        return
+                    }
+                }
             }
         }
 
@@ -1689,34 +1699,6 @@ open class PlaybackManager @Inject constructor(
             is UserEpisode -> podcastManager.buildUserEpisodePodcast(episode)
             else -> null
         }
-    }
-
-    /**
-     * Check we have the latest episode url in the background.
-     */
-    private fun updateEpisodeUrl(episode: PodcastEpisode) {
-        if (episode.isDownloaded) {
-            return
-        }
-
-        // TODO
-//        val updateEpisodeTask = UpdateEpisodeTask(podcastCacheServerManager, episodeManager)
-//        val episodeSingle = updateEpisodeTask.download(episode.podcastUuid, episode.uuid)
-//        val disposable = episodeSingle
-//                .subscribeOn(Schedulers.io())
-//                .subscribe(
-//                        { episodeUpdated ->
-//                            val currentEpisode = getCurrentEpisode()
-//                            if (currentEpisode != null &&
-//                                    currentEpisode.uuid == episodeUpdated.uuid &&
-//                                    currentEpisode.downloadUrl != episodeUpdated.downloadUrl) {
-//                                upNextQueue.reloadCurrentEpisode()
-//                                forcePlayerSwitch = true
-//                            }
-//                        },
-//                        { throwable -> Timber.e(throwable, "Failed to update episode url") }
-//                )
-//        disposables.add(disposable)
     }
 
     @Suppress("DEPRECATION")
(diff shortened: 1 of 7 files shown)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
```diff
@@ -1503,19 +1503,29 @@ open class PlaybackManager @Inject constructor(
...
+                    try {
+                        val newDownloadUrl = userEpisodeManager.getPlaybackUrl(episode).await()
+                        episode.downloadUrl = newDownloadUrl
+                    } catch (e: Exception) {
+                        onPlayerError(PlayerEvent.PlayerError("Could not load cloud file ${e.message}"))
+                        removeEpisode(episode, source = sourceView)
+                        return
+                    }
+                }
             }
```
mchowning, line 1529: Apart from moving this handling for user episodes into the when statement, the actual handling for use episode's should be unchanged. I just thought using a when statement here was nice since it helps show highlight that we're updating the URLs here differently for user episodes and podcast episodes.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
```diff
@@ -1691,34 +1701,6 @@ open class PlaybackManager @Inject constructor(
...
-//                            if (currentEpisode != null &&
-//                                    currentEpisode.uuid == episodeUpdated.uuid &&
-//                                    currentEpisode.downloadUrl != episodeUpdated.downloadUrl) {
-//                                upNextQueue.reloadCurrentEpisode()
-//                                forcePlayerSwitch = true
-//                            }
-//                        },
-//                        { throwable -> Timber.e(throwable, "Failed to update episode url") }
-//                )
-//        disposables.add(disposable)
```
mchowning, line 1719: You don't by chance remember why we removed this do you @geekygecko ? This is from around 5 years ago, so I doubt you'll have any recollection, but I just wanted to make sure there wasn't anything we should be watching out for in this PR.

mchowning:
> Be careful this doesn't remove starred podcasts from device, currently putting a star of an episode will keep that episode arround even if the RSS source is lost

Thanks for calling this out. I don't believe it will cause any problems for that because if we fail to fetch an updated URL we just use the old one.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
geekygecko: Sorry, I can't remember why we removed it. But it's not called in iOS either now so there must have been a reason. I wonder if it was too slow calling it before we played the episode.

mchowning: No worries.

> it's not called in iOS either now so there must have been a reason

iOS does seem to update the URL before playing an episode [here](https://github.com/Automattic/pocket-casts-ios/blob/trunk/podcasts/PlaybackActionHelper.swift#L93) I believe.

ashiagr (approved):
Excellent investigation and a fix! 🕵️ 

I could reproduce the issue reliably with the given steps on both Phone and Automotive on the main branch and I confirm that changes in this branch updates the URL and fixes the issue.

PR #1577, opened 2023-11-30
Update billing lib version to 6.0.1

ashiagr:
## Description

This bumps billing lib version from `5.2.1` to `6.0.1` and makes necessary changes. 
Internal ref: pdeCcb-3t4-p2

[Release notes - v6.0.1](https://developer.android.com/google/play/billing/release-notes#6-0-1)

> Added new [ReplacementMode](https://developer.android.com/reference/com/android/billingclient/api/BillingFlowParams.SubscriptionUpdateParams.ReplacementMode) enum to replace [ProrationMode](https://developer.android.com/reference/com/android/billingclient/api/BillingFlowParams.ProrationMode).

I replaced the deprecated enum in 8efbbe9c0c923bf4902e0737aff28fa52aeb9509

> [!Warning]
> Not ready to merge since I need to check if everything is still good backend side

## Testing Instructions

Prerequisites
- Release build
- License test account

1. Login with a free account
2. Subscribe to Plus
3. ✅  Notice that the subscription succeeds
4. Go to profile
5. Tap Upgrade to Patron
6. ✅  Notice that the upgrade proceeds as expected with the correct replacement mode.
 

## Checklist
 
- [ ] with different themes

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -25,7 +25,7 @@ aboutlibraries = "10.5.0"
 accompanist = "0.30.1"
 # Android Gradle Plugin - https://developer.android.com/studio/releases/gradle-plugin
 android-application = "8.1.0"
-billing = "5.2.1"
+billing = "6.0.1"
 coil = "2.2.1"
 compose = "1.4.2"
 # @keep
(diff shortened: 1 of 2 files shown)


mchowning (approved):
Changes make sense, and the test steps worked perfectly. I also tested upgrading directly to Patron and that worked just as expected. 👍

PR #1589, opened 2023-12-06
Add transaction annotation for large result select episodes queries

ashiagr:
## Description

> [!Warning]
> Targets 7.53

Part of #1572

This adds `@Transaction` annotation for large result select episodes queries.

This is a sample error seen in logs when we were able to reproduce the error:
```
Failed to read row 638, column 0 from a window with 638 rows, 35 columns
 E  Failed to read row 638, column 1 from a window with 638 rows, 35 columns
 E  Failed to read row 638, column 2 from a window with 638 rows, 35 columns
 E  Crash: Fatal crash.
 E  java.lang.IllegalStateException: Expected non-null java.util.Date, but it was null.
 E  	at au.com.shiftyjelly.pocketcasts.models.db.dao.EpisodeDao_Impl$56.call(EpisodeDao_Impl.java:9496)
 E  	at au.com.shiftyjelly.pocketcasts.models.db.dao.EpisodeDao_Impl$56.call(EpisodeDao_Impl.java:9432)
 E  	at io.reactivex.internal.operators.maybe.MaybeFromCallable.subscribeActual(MaybeFromCallable.java:46)
 E  	at io.reactivex.Maybe.subscribe(Maybe.java:4290)
```

Here is a recommended solution:

https://issuetracker.google.com/issues/73859920#comment2
> Any time the size goes above 2mb, you need to run the query in a transaction, otherwise database might change as the Framework loads more pages in.
> Just annotate the query w/ @Transaction. Room cannot fix this as we have no way to change that 2mb limit nor how Cursor pages in big results.

Since the PR targets a frozen branch, I've added the annotation to the most important methods:

// Order by published date methods
`findByPodcastOrderPublishedDateAsc`
`findByPodcastOrderPublishedDateAscSuspend`
`findByPodcastOrderPublishedDateDesc`
`findByPodcastOrderPublishedDateDescSuspend`

// Order by duration methods
`findByPodcastOrderDurationAsc`
`findByPodcastOrderDurationAscSuspend`
`findByPodcastOrderDurationDesc`
`findByPodcastOrderDurationDescSuspend`
`observeByPodcastOrderDurationAsc`
`observeByPodcastOrderDurationDesc`

// Order by title methods
`observeByPodcastOrderTitleAsc`
`observeByPodcastOrderTitleDesc`

// Downloaded/ downloading methods
`observeDownloadedEpisodes`
`observeDownloadingEpisodes`
`observeDownloadingEpisodesIncludingFailed`
[...]

PR #1611, opened 2023-12-13
Remove unused episode query methods

mchowning:
## Description
Removing some unused episode query methods. Doing this as a follow-up to https://github.com/Automattic/pocket-casts-android/pull/1589 so that we don't have risky methods that aren't using transactions laying around.

## Testing Instructions

If CI is green, this should be good.

## Checklist


ashiagr:
Whoa! That's a sizeable no. of queries ... thanks for getting rid of these!

ashiagr (approved):

PR #1619, opened 2023-12-14
Remove usage of GlobalScope

mchowning:
## Description
This replaces our usage of GlobalScope in the app with an applicationScope. I'm not sure that we're experiencing any direct negative effects from how we use GlobalScope, but it's usage is [strongly discouraged](https://medium.com/androiddevelopers/coroutines-patterns-for-work-that-shouldnt-be-cancelled-e26c40f142ad#:~:text=%E2%9D%8C-,GlobalScope,-There%20are%20multiple) and it feels like the kind of thing that could eventually cause us problems. If nothing else, testing can be more difficult with a hard-coded `GlobalScope`.

There are definitely further improvements we can do here, but my hope is that this is a step in the right direction.

One change I made in this PR (in https://github.com/Automattic/pocket-casts-android/commit/ba4f5669565c5f9b9c1ab6e7d00839a466faef8c) is to provide the application scope in a lower module (the repository) instead of the top-level app modules. I did this just because it makes the DI setup much simpler if we don't have to inject something into low-level modules that is provided by high-level modules. That's certainly doable, but it's a bit more complicated (among other things, the watch, phone, and automotive apps each have to indepenently provide the dependency), so I thought that keeping this simple made more sense. If we need to start providing the application scope at the app-level for some reason in the future, it will be easy enough to make that change then.

## Testing Instructions

No specific test steps for this, just generally smoke test the app making sure you do various actions with playback, syncing, and downloading.

## Checklist

### DIFF
--- a/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/PodcastManagerTest.kt
+++ b/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/PodcastManagerTest.kt
@@ -20,6 +20,8 @@ import au.com.shiftyjelly.pocketcasts.servers.podcast.PodcastCacheServerManager
 import au.com.shiftyjelly.pocketcasts.servers.refresh.RefreshServerManager
 import au.com.shiftyjelly.pocketcasts.utils.Optional
 import io.reactivex.Single
+import kotlinx.coroutines.CoroutineScope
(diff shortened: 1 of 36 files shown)


ashiagr (approved):
I smoke-tested the app for refresh, auth, playback, sync and found them working. 👍

PR #1627, opened 2023-12-16
Prevents playbutton in files episode list from changing back to pause on upnext

### LINKED ISSUE #397: Files episode list play/pause icon gets stuck
### Description

When playing a local file, the file's listing in the Files list gets a pause icon that persists even after the file has finished playing. Tapping that icon pauses playback of whatever got loaded from the up next queue, but the icon does not update to reflect the new status (should update to be a play icon).

1. Add a very short local file
2. Make sure you an episode in the Up Next queue
3. Go to Profile tab -> Files
4. Tap on the row for the short local file
5. Tap on the play button
6. Let the local file complete playing in the full screen player so that the next episode is loaded from the Up Next queue
7. Back out of the full screen player
8. 🔴  Observe that the local file row has a pause icon even though the local file is not playing
9. Tap the local file's pause button
10. 🔴 Observe that playback is stopped (even though it wasn't the local file that was playing)
11. 🔴  Observe that even though playback has stopped, the local file still has a pause button (that doesn't seem to do anything)

https://user-images.githubusercontent.com/4656348/195361420-4f906139-08ea-4f6f-bff0-3d76c3be3675.mp4

Pixel 3, Android 12, Pocket Casts `7.24-rc-1`

jmorton138:
## Description
When playing a local file, on completion, the play button would breifly get set to played, but then would change back to a pause button when the player switched to the podcast episode up next in the queue. 

This issue only seemed to occur when going from a user episode to a podcast episode. If the episode up next was also a user episode, the playbutton didn't get switched back to a pause button.

The fix I've implemented works, however it doesn't solve the root of the problem which is that the playback state of the user episode gets set back to playing after it's marked as completed. My hypothesis is that whatever is updating the playback state of the upnext episode to playing is doing the same to the played local file.

I'm happy to look into this further to try and solve the issue this bug is stemming from, but I would need some guidance and direction on how player events and episode switching works as I wasn't able to quite figure it out.

Fixes #397

## Testing Instructions

1. Add a very short local file
2. Make sure you an episode in the Up Next queue
3. Go to Profile tab -> Files
4. Tap on the row for the short local file
5. Tap on the play button
6. Let the local file complete playing in the full screen player so that the next episode is loaded from the Up Next queue
7. Back out of the full screen player
8. 🔴 Observe that the local file row has a played icon even
9. Tap the local file's pause button
10. 🔴 Observe that playback is stopped 
11. 🔴 Observe that playback has stopped, and the local file has a played button

## Checklist
 
- [x] with different themes


MiSikora:
Hey @jmorton138, sorry for the no response for such a long time. I merged changes from our `main` to you branch. I'll run the CI checks and merge it once everything is green. Thank you for the submission!

MiSikora (approved):

PR #1629, opened 2023-12-17
Using compose-bom

mchowning:
## Description
Adding the compose-bom to help make our upgrades easier. When the compose BOM was created, Google [noted](https://android-developers.googleblog.com/2022/10/whats-new-in-jetpack-compose.html#:~:text=A%20BOM%20is%20a%20Maven%20module%20that%20declares%20a%20set%20of%20libraries%20with%20their%20versions.%20It%20will%20greatly%20simplify%20the%20way%20you%20define%20Compose%20library%20versions%20in%20your%20Gradle%20dependencies%20block%2C%20especially%20now%20that%20we%20moved%20the%20various%20Jetpack%20Compose%20libraries%20to%20independent%20versioning%20schemes) that it would "greatly simplify the way you define Compose library versions in your Gradle dependencies block, especially now that we [moved the various Jetpack Compose libraries to independent versioning schemes](https://android-developers.googleblog.com/2022/06/independent-versioning-of-Jetpack-Compose-libraries.html)." It doesn't seem like nearly as much of an improvement as that implies (at least not yet) though because almost all of the compose libraries are still on the same version. So this isn't going to make upgrades as much easier as I had hoped, but it still seemed like it was probably a worthwhile improvement to add. Let me know if you disagree though since it doesn't make a big difference now.

I picked a BOM that should not update any dependencies, so there should be no change in the dependency versions that the app uses.

## Testing Instructions
If CI is green we should be good, but just as an extra test I did a quick smoke-test of all the apps.

## Checklist

### DIFF
--- a/base.gradle
+++ b/base.gradle
@@ -171,6 +171,8 @@ dependencies {
     implementation libs.bundles.sentry
     implementation libs.bundles.work
 
+    implementation platform(libs.compose.bom)
+    androidTestImplementation platform(libs.compose.bom)
     implementation platform(libs.firebase.bom)
     implementation platform(libs.sentry.bom)
 
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -17,7 +17,7 @@
 #
 # 4. Check the Compose and Accompanist versions are compatible.
 #    https://github.com/google/accompanist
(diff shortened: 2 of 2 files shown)


ashiagr (approved):
LGTM 👍

PR #1636, opened 2023-12-18
Prevent wear chips from getting cut off by round watch edges

mchowning:
## Description
This PR is incorporating the `ScalingLazyColumnState.Factory` that was added relatively recently to [horologist](https://github.com/google/horologist). We're adding this because Google has rejected our latest submissions because some of the chips at the top and bottom of the screen cannot be scrolled fully onto the screen.

I cannot recreate this with any of the "top" chips on a scrollable screen, but I can recreate the issue on the bottom chips on some screens. Using `responsive` avoids the issue in every place I've observed it.

Although I haven't observed the issue at the top of the screen at all in my testing, Google's rejections included a screenshot that appeared near the top of the screen, so for any screen with large elements/chips at the top, I am making sure that the default of `true` for `firstItemIsFullWidth` is used in the `responsive` function. It seems unnecessary in my testing, but setting that doesn't really do any harm apart from possibly adding a small bit of extra padding to the top of the screen.

Ideally we would just update our version of horologist, but we need to fix this on the release branch, and updating horologist requires updating our compileSdk and a lot of other dependencies (kotlin, ksp, etc.) (PRs: [1](https://github.com/Automattic/pocket-casts-android/pull/1633), [2](https://github.com/Automattic/pocket-casts-android/pull/1634)), so my thinking is that it would be better to just manually bring in this specific change that we need into the release branch. We can merge the large dependency updates to `main` so we have more time to test and uncover any issues with that.

## Testing Instructions

For these tests, use the device with the smallest possible screen (i.e., the emulator with the small screen), and make sure that the device is using the highest possible text size and has bold text turned on (see [android dev documentation about clipping](https://developer.android.com/design/ui/wear/guides/behaviors-and-patterns/clipping).

Check every screen in the app to verify that both (1) the top element on the screen is not clipped when the screen is loaded or when you scroll all the way to the top and (2) the bottom element on the screen is not clipped when you scroll to the bottom of the screen.

The only screens where I have observed problems are the screens listing podcast episodes where I saw that the bottom chips can be clipped when the podcast name takes two lines:
1. Up Next Screen
2. Podcast screen 
3. Downloads
4. List of episodes in a FIlter
5. Files

Here is a list of the scrollable screens I found in the app. It would be good to double-check that there are no issues with clipping on the top or bottom of these:
1. Initial needs plus screen
2. Select how to sign on screen
3. Watch list screen
4. Now Playing Screen
6. Effects screen
7. Adjust volume screen
8. Up Next Screen
[...]

PR #1634, opened 2023-12-18
Update horologist version (and libraries that matches this version)

ashiagr:
## Description

This updates horologist lib to ~the latest `v0.5.14`~ the latest `0.4.x` version of `0.4.17` from the current `v0.4.10`. Below essential libraries are also upgraded to match this version:

`kotlin`: `1.7.0` to `1.9.0`
`ksp`: `1.7.10-1.0.6` to `1.9.0-1.0.13`
`compose-kotlin-compiler`: `1.3.0` to `1.5.2 `(refers [Compose Compiler to Kotlin Compatibility Map](https://developer.android.com/jetpack/androidx/releases/compose-kotlin#:~:text=1.5.2,1.9.0))
`hilt`: `2.43.2` to `2.49`
`hilt-compiler`: `1.0.0` to `1.1.0`
`media3`: `1.0.0` to `1.2.0`

In addition, we had to update the compileSdk.

#### Release notes

Horologist: https://github.com/google/horologist/releases
Kotlin: https://github.com/JetBrains/kotlin/releases?page=3
Kotlin Compiler: https://developer.android.com/jetpack/androidx/releases/compose-compiler
Hilt Android: https://mvnrepository.com/artifact/com.google.dagger/hilt-android, https://github.com/google/dagger/releases
Hilt Compiler: https://developer.android.com/jetpack/androidx/releases/hilt

## Testing Instructions
1. Smoke test the phone, automotive, and especially the watch app for any regressions since this is updating quite a few dependencies.
2. Redo the testing steps for https://github.com/Automattic/pocket-casts-android/pull/1636 to make sure that horologist is properly keeping the chips fully visible.
3. Verify that all screens which are scrollable display a scrollbar as you scroll them.

## Checklist
- [x] If this is a user-facing change, I have added an entry in CHANGELOG.md
- [x] Ensure the linter passes (`./gradlew spotlessApply` to automatically apply formatting/linting)
- [x] I have considered whether it makes sense to add tests for my changes
- [x] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml`
- [x] Any jetpack compose components I added or changed are covered by compose previews
 
#### I have tested any UI changes...

- [x] with different themes
- [x] with a landscape orientation
- [x] with the device set to have a large display and font size
- [x] for accessibility with TalkBack

PR #1641, opened 2023-12-21
Update horologist version (and libraries that matches this version) (targeting `main`)

mchowning:
## Description
This PR is a follow-up to https://github.com/Automattic/pocket-casts-android/pull/1634. Instead of targeting the `release/7.54` branch, this PR targets `main` because we were able to extract just the changes we needed for the release and add only those changes to the release branch in https://github.com/Automattic/pocket-casts-android/pull/1636.

In addition, this PR uses horologist version `0.4.17` instead of `0.5.14` because the `0.4.X` versions of horologist use stable androidx APIs whereas `0.5.x` versions use alpha androidx APIs.

This PR also reverts the changes from https://github.com/Automattic/pocket-casts-android/pull/1636 because those changes should not be needed anymore.

> [!WARNING]
> Reverting https://github.com/Automattic/pocket-casts-android/pull/1636 also removed [this change](https://github.com/Automattic/pocket-casts-android/pull/1636#issuecomment-1862192925) you added because you saw the "Contact Support" chip slightly cut off. I didn't observe that issue on this branch so I did not re-add that change, but if you still see the chip being cut off (or you think it's just prudent to add in that extra padding), feel free to add that change back in.

## Testing Instructions

Smoke test the phone, automotive, and especially the watch app for any regressions since this is updating quite a few dependencies.

Also redo the testing steps for https://github.com/Automattic/pocket-casts-android/pull/1636 to make sure that horologist is properly keeping the chips fully visible.

## Checklist
 
- [X] with different themes


ashiagr (approved):
👋 

I reverted media3 changes as switching to horologist `0.4.17` did not require those changes. I smoke-tested the apps and found them working. 

For the watch app:


I tested the clipping issue and found no issues. Below are a few screenshots after the revert commits

Podcast Episodes | Up Next | Contact Support
------| ----|-----
podcast_episodes | upnext | contact support


I also tested the following scenario that you pointed out here: https://github.com/Automattic/pocket-casts-android/pull/1634#issuecomment-1867226987

> The one issue I am seeing on the watch is that if you start playback without an available output device (i.e., no Bluetooth headphones), and then you back out of the selection screen, the app displays as if it is playing even though it is not playing.

and found it does not exist, see the video below:

https://github.com/Automattic/pocket-casts-android/assets/1405144/a8c7167a-c4a6-4525-b6f8-8d89ca8078c0

Issue #44, opened 2022-07-06
Clean up old Java code


geekygecko:
There is still some Java code in the project. It would be good to check if it's still needed and convert it over to Kotlin.

- [x] [EpisodeManagerTest.java](https://github.com/Automattic/pocket-casts-android/blob/61fe9bf02411dbb158ec838d2fc4834099e99a4b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/EpisodeManagerTest.java)
- [x] [FileStorage.java](https://github.com/Automattic/pocket-casts-android/blob/61fe9bf02411dbb158ec838d2fc4834099e99a4b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/file/FileStorage.java)
- [x] [FileUtil.java](https://github.com/Automattic/pocket-casts-android/blob/61fe9bf02411dbb158ec838d2fc4834099e99a4b/modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/FileUtil.java)
- [x] [OldDataOpenHelper.java](https://github.com/Automattic/pocket-casts-android/blob/61fe9bf02411dbb158ec838d2fc4834099e99a4b/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/OldDataOpenHelper.java)
- [x] [PCBottomSheetBehavior.java](https://github.com/Automattic/pocket-casts-android/blob/61fe9bf02411dbb158ec838d2fc4834099e99a4b/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/helper/PCBottomSheetBehavior.java)
- [x] [ShiftyCustomAudio.java](https://github.com/Automattic/pocket-casts-android/blob/61fe9bf02411dbb158ec838d2fc4834099e99a4b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/ShiftyCustomAudio.java)
- [x] [ShiftyTrimSilenceProcessor.java](https://github.com/Automattic/pocket-casts-android/blob/61fe9bf02411dbb158ec838d2fc4834099e99a4b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/ShiftyTrimSilenceProcessor.java)
[...]

PR #1671, opened 2024-01-08
Delete EpisodeManagerTest

MiSikora:
## Description
This is a part of Java to Kotlin migration outlined in #44. This PR deletes [EpisodeManagerTest.java](https://github.com/Automattic/pocket-casts-android/blob/61fe9bf02411dbb158ec838d2fc4834099e99a4b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/EpisodeManagerTest.java). There are couple of reasons for deletion instead of migration.

1. The code was commented out since at least 23rd of June 2022.
2. The test class lived in the actual production code space most likely due to an error in migration.
3. After removing comments some of the classes like `DatabaseManager` no longer exist.

## Testing Instructions

Not much to test. Make sure that CI passes.

## Checklist
 
- [ ] with different themes


mebarbosa (approved):
LGTM

PR #1677, opened 2024-01-08
Player: Episode artwork - add image ratio constraint

ashiagr:
## Description
As part of this PR: https://github.com/Automattic/pocket-casts-android/pull/1665 we added a `Scale.FIT` while loading episode artwork to prevent large bitmap OOM error. 

`coil.size.Scale.FIT`

> Fit the image to the view so that both dimensions (width and height) of the image will be equal to or less than the corresponding dimension of the view.

With the `Scale.FIT`, the width and height of the image should be equal to or less than the corresponding dimension of the view. However, the image spreads beyond the view width.

This is also reported here:

https://pocketcastsbeta.slack.com/archives/C013F9S9C2W/p1704702825434209 

To fix this, I added an aspect ratio constraint for the image view. I think it is safe to include this constraint as it is also applied to the landscape layout loading the same image URL.

Before applying the fix I also tried adding [precision.EXACT](https://rb.gy/pds2o9) but it made no difference. 

Let me know if you can think of a better way to address it. 

## Testing Instructions
1. Fresh install the app
2. Open episode https://pca.st/e3LI in the app
3. Open player
4. Notice that the image has 1:1 aspect ratio

## Screenshots or Screencast 

Before | After
----|-----
image| image

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/features/player/src/main/res/layout/adapter_player_header.xml
+++ b/modules/features/player/src/main/res/layout/adapter_player_header.xml
@@ -36,6 +36,7 @@
                 android:visibility="@{viewModel.isPodcastArtworkVisible()}"
                 android:clickable="true"
                 android:scaleType="fitCenter"
+                app:layout_constraintDimensionRatio="1:1"
                 app:layout_constraintBottom_toTopOf="@+id/episodeTitle"
                 app:layout_constraintEnd_toEndOf="parent"
                 app:layout_constraintStart_toStartOf="parent"


pocketcasts:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

MiSikora (approved):
Yes, I think this the correct solution. Works as expected. 👍

ashiagr:
Thanks! I'll change the target and bring this change to release/7.55-rc-2.

PR #1682, opened 2024-01-09
Streamline FileStorage

MiSikora:
## Description

This is a follow-up to #1668. It improves migrated code. Some highlights:
* Long methods got separated into smaller named methods.
* Use suspending methods instead of deprecated sync methods to interact with `EpisodeManager`.
* `FileStorage` is no longer a singleton as it is not stateful.

## Testing Instructions

Generally, perform tests related to downloading podcasts. You should test this on devices pre and post Android 10. On Android 10+ you can simply test downloading, going into airplane mode, and checking if listening works as expected. On earlier Android versions you can check if you can move downloaded files between different storage choices. [This comment](https://github.com/Automattic/pocket-casts-android/pull/1668#pullrequestreview-1809946229) can be helpful with possible test scenarios.

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/file/FileStorage.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/file/FileStorage.kt
@@ -14,6 +14,8 @@ import java.io.File
 import java.io.IOException
 import javax.inject.Inject
 import javax.inject.Singleton
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.reactive.asFlow
 import timber.log.Timber
 
 @Singleton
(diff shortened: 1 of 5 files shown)


mebarbosa:
Hey @MiSikora 
I have only one real device with Android 11. Do you think it's ok to test the file moving using emulator? I am not 100% confident

MiSikora:
Yes, I tested this on an emulator. But feel free to ping someone else for testing with a real device if you think it's worth it.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/file/FileStorage.kt
```diff
@@ -158,7 +160,7 @@ open class FileStorage @Inject constructor(
         }
     }
 
-    fun moveStorage(oldDir: File, newDir: File, episodesManager: EpisodeManager) {
+    suspend fun moveStorage(oldDir: File, newDir: File, episodesManager: EpisodeManager) {
```
mebarbosa, line 163: Nice!

mebarbosa (approved):
LGTM

PR #1684, opened 2024-01-10
Android14 - Specify foreground service types

ashiagr:
## Description

This PR specifies [foreground service types at runtime](https://developer.android.com/develop/background-work/background-tasks/persistent/how-to/long-running#specify-foreground-service-types-runtime) when calling `setForeground()` or `setForegroundAsync()` which is required if the app targets Android 14.

This is needed at two places, while
1. downloading episodes
2. importing OPML 

To get this working, I also had to [declare the foreground service type in the Android manifest](https://developer.android.com/develop/background-work/background-tasks/persistent/how-to/long-running#declare-foreground-service-types-manifest) for the long-running worker.

## Testing Instructions

Prerequisites: Change `targetSdkVersion` set to 34 in `dependencies.gradle.kts`

#### Test.1 - Download Episode

1. Install the app
2. Open any podcast -> episode
3. Tap `Download` button for the episode
4. ✅ Notice that the download task succeeds without any crash

** `setForegroundAsync(...)` is not called from `DownloadEpisodeTask` for the Wear OS app, so these changes do not apply to the Wear OS app.

```
if (!Util.isWearOs(context)) {
   setForegroundAsync(createForegroundInfo())
}
```

It would also be good to test the episode download on the Wear OS app.

#### Test.2 - Import OPML

1.  Install the app
2. Go to `Profile` -> `Settings` -> `Import & export OPML`
3. Tap `Save file` to export the OPML file to your storage
4. Tap `Select file` and import podcast from the OPML file saved above
6. ✅ Notice that OPML task succeeds without any crash

** Import/ export OPML feature is only available for phone app.

## Checklist
 
- [ ] with different themes

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -20,6 +20,7 @@
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
     <uses-permission android:name="android.permission.BLUETOOTH"/>
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE"/>
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE_DATA_SYNC" />
     <uses-permission android:name="com.android.vending.CHECK_LICENSE"/>
     <uses-permission android:name="com.android.vending.BILLING" />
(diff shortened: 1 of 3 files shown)


mebarbosa (approved):
LGTM 💯

PR #1685, opened 2024-01-10
Merge release 7.55-rc-2 into main

ashiagr:
- Updated `wpmreleasetoolkit  9.2.0  ->  9.3.0 (Latest)`
- Includes #1677


version.properties
```diff
@@ -1,2 +1,2 @@
-versionName=7.55-rc-1
-versionCode=9181
\ No newline at end of file
+versionName=7.55-rc-2
+versionCode=9184
```
ashiagr, line 2: My GitHub token expired and had to retrigger the beta build twice because of which version code is incremented from 9181 to 9184. I manually updated versionName to get the correct one on the next build.

mchowning (approved):

PR #1687, opened 2024-01-10
Add tests for FileUtil

MiSikora:
## Description

This is a preparation PR for migrating [`FileUtil`](https://github.com/Automattic/pocket-casts-android/blob/61fe9bf02411dbb158ec838d2fc4834099e99a4b/modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/FileUtil.java) from Java to Kotlin. Since it is a utility class that is used in multiple areas of code I find it will be easier to verify correctness through automated tests rather than through manual checking of every possible functionality that uses this code.

Some tests seem a bit redundant like `copy()`/`copyFile()`/`copyDirectory()` but I did it to make migration easier. In the migration PR I'll probably make some of these function private and delete tests for them while also migrating code using it to use only a single `copy()` function.

I didn't test for nullability because at quick glace I can ignore it as Kotlin code already uses non null parameters. And if it doesn't I'll make it work using `?.` operator. I also didn't test Android specific stuff related to `Uri`. Since it is permission related and code is quite simple I'll just `go for it`™ during the migration.

## Testing Instructions

Code review my changes.

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/services/utils/src/test/java/au/com/shiftyjelly/pocketcasts/utils/FileUtilTest.kt
+++ b/modules/services/utils/src/test/java/au/com/shiftyjelly/pocketcasts/utils/FileUtilTest.kt
@@ -118,6 +118,29 @@ class FileUtilTest {
         assertEquals("Files contents differ", file1.snapshot(), file2.snapshot())
     }
 
+    @Test
+    fun `directory is copied using copy function`() {
+        val dir1 = tempDir.newFolder().apply {
+            File(this, "file1").writeRandomBytes(100)
+            File(this, "file2").writeRandomBytes(200)
+            val innerDir = File(this, "inner").also(File::mkdirs)
+            File(innerDir, "file3").writeRandomBytes(300)
+        }
+        val dir2 = tempDir.newFolder()
+
+        FileUtil.copy(dir1, dir2)
+
+        val comparisonPairs = listOf(
+            File(dir1, "file1") to File(dir2, "file1"),
+            File(dir1, "file2") to File(dir2, "file2"),
+            File(File(dir1, "inner"), "file3") to File(File(dir2, "inner"), "file3"),
+        )
+        comparisonPairs.forEach { (srcFile, dstFile) ->
+            assertFalse("Original file has no content", srcFile.snapshot().size == 0)
+            assertEquals("Files contents differ", srcFile.snapshot(), dstFile.snapshot())
+        }
+    }
+
     private fun File.writeRandomBytes(count: Int) {
         sink().buffer().use { it.write(Random.nextBytes(count)) }
     }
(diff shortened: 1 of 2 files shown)


modules/services/utils/src/test/java/au/com/shiftyjelly/pocketcasts/utils/FileUtilTest.kt
```diff
@@ -0,0 +1,371 @@
...
+
+    @Test
+    fun `keep nomedia file when deleting directory contents`() {
+        listOf(".nomedia", ".NOMEDIA", ".nOmEdIa").forEach { name ->
+            val dir = tempDir.newFolder()
+            val file = File(dir, name).also(File::createNewFile)
+
+            FileUtil.deleteDirectoryContents(dir.path)
+
+            assertTrue("nomedia file $file does not exists", file.exists())
```
mchowning, line 61: nitpick: exists → exist

modules/services/utils/src/test/java/au/com/shiftyjelly/pocketcasts/utils/FileUtilTest.kt
```diff
@@ -0,0 +1,371 @@
...
+    fun `get file name without extension for a file with no dots`() {
+        val file = tempDir.newFile("hello")
+
+        val fileName = FileUtil.getFileNameWithoutExtension(file)
+
+        assertEquals("hello", fileName)
+    }
+
+    @Test
+    fun `get file name without extension for a directory with a single dot`() {
```
mchowning, line 273: Is this test executing the same code as the `get file name without extension for a file with a single dot` test? I see the variable names are different, but I'm having trouble seeing what else is different. 

Was this maybe supposed to be `tempDir.newFolder` instead of `tempDir.newFile`? Same question for the next two tests.

modules/services/utils/src/test/java/au/com/shiftyjelly/pocketcasts/utils/FileUtilTest.kt
```diff
@@ -0,0 +1,371 @@
...
+
+            File(this, "inner3").also(File::mkdirs).apply {
+                File(this, "file6").writeRandomBytes(100000)
+                File(this, "file7").writeRandomBytes(1000000)
+            }
+        }
+
+        val size = FileUtil.folderSize(dir)
+
+        assertEquals(1111111, size)
```
mchowning, line 353: That's nice how you got the `1111111` total so if the size is wrong it's easy to see where the error happened. 👍

mchowning (approved):
Great idea to add these tests. This is great!

I noted one small issue, but once that is addressed this looks good to me.

modules/services/utils/src/test/java/au/com/shiftyjelly/pocketcasts/utils/FileUtilTest.kt
MiSikora: Right. That's what I get for copying and pasting. 😅

PR #1693, opened 2024-01-11
Add remaining named settings that currently sync to new named settings sync call

mchowning:
## Description
This PR builds on https://github.com/Automattic/pocket-casts-android/pull/1690. That PR added the first named setting to sync using the new call, and this PR adds the remaining named settings that are currently being synced with the new call. With this change, the app is syncing the same data regardless of whether the sync setting feature flag is turned on or off.

This also adds a few more deprecation annotations for methods that we can remove when we remove the feature flag for this feature.

## Testing Instructions

For all these tests you will need to:
1. Install the app on two devices
2. Log into the same Pocket Casts account on both devices
3.  Turn on sync settings under Profile → ⚙️ → Beta features → Settings Sync

### 1. Skip back setting

1. Device 1: go to Profile → ⚙️ → General, and update the skip back time to a new value
2. Device 1: tap the refresh button on the profile screen
3. Device 2: tap the refresh button on the profile screen
4. Device 2: verify that the skip back time on this device is the new value you set on device 1

### 2. Newsletter

1. Device 1: go to Profile → Account, and tap switch the "Pocket Casts newsletter" toggle.
2. Device 1: tap the refresh button on the profile screen
3. Device 2: tap the refresh button on the profile screen
4. Device 2: verify that the newsletter toggle has the new value you set on device 1

### 3. Podcast sort

1. Device 1: Go to the podcast tab and tap on the three-dot overflow menu button
2. Device 1: Change how the podcasts are sorted
3. Device 1: tap the refresh button on the profile screen
4. Device 2: tap the refresh button on the profile screen
5. Device 2: go to the podcast tab, tap the three-dot overflow menu, and confirm the sort type is the new value you set on device 1

### 4. Free gift acknowledgement

I don't know a good way to test this without code changes, so this test requires a special build of the app that includes an added call to

``` kotlin
settings.freeGiftAcknowledged.set(value = [true/false], needsSync = true)
```

in `PocketCastApplication::onCreate`.

1. Device 1: Install a build of the app that sets freeGiftAcknowledged to --true-- in `PocketCastsApplication::onCreate` by adding the above line of code
2. Device 2: Install a build of the app that sets freeGiftAcknowledged to --false-- in `PocketCastsApplication::onCreate` by adding the above line of code
3. Device 1: force stop the app and reopen the app
4. Device 2: Wait a few seconds, force stop the app, and reopen the app
6. Device 2: Tap the refresh button on the profile screen
[...]

PR #1698, opened 2024-01-12
Improve handling of whether a setting should be synced

mchowning:
## Description
This PR improves how we determine/set whether a setting should be synced.

Previously, when updating a value and saying it did not need to be synced with something like `.set(value = 1, needsSync = false)`, we were leaving the modifiedAt value (which indicates sync is needed if it is not `null`). This meant that if you called `.set(value = 1, needsSync = true)` and then immediately called (`.set(value = 2, needsSync = false)` the setting of the value to `2` would not overwrite the modifiedAt value and the setting would be synced during the next incremental sync. 

This is a problem, but I think it's just a theoretical problem at this point because I can't think of any way a user could actually get into this scenario. Still improving this handling feels like the thing to do because it feels like a gotcha that will catch us at some point. For that reason, this PR updates this so that calling the method with `needsSync = false` sets the `modifiedAt` value to `null`, ensuring that the setting will not be synced. 

In addition to avoiding the possible syncing of values that should not have been synced, this change also means that we no longer need `.doesNotNeedSync` to be called to ensure that nothing gets synced from a particular setting because the set method's `needsSync` parameter now takes care of this.

> [!WARNING]
> This PR builds on top of https://github.com/Automattic/pocket-casts-android/pull/1693.

### 1. Handles settings that should be synced
1. Ensure that the settings sync feature flag is toggled on
1. Tap the refresh button on the Profile tab to make sure all settings are up-to-date
2. Tap the account button on the Profile tab
3. Change the newsletter toggle
4. Tap the refresh button on the Profile tab
5. Check the logs and verify that the call to the `/named_settings/update` endpoint includes the newsletter toggle change, i.e., it includes something like `"changed_settings":{"marketingOptIn":{"value":true,"modified_at":"2024-01-12T16:52:57.552Z"}`
6. Tap the refresh button the Profile tab again
7. Verify that the call to the `/named_settings/update` endpoint does not include any changes, i.e., it includes `"changed_settings: {}"`

### 2. Handles settings that should not be synced

Currently I don't think there's a good way to take a setting that can be synced and force a change that we do not want to get synced without signing out of the app so, for this test, you need to build the app with this change so that toggling the newsletter should no longer get synced:

``` diff
diff --git a/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/AccountDetailsViewModel.kt b/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/AccountDetailsViewModel.kt
index 7bea787cb..99661eabc 100644
--- a/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/AccountDetailsViewModel.kt
+++ b/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/AccountDetailsViewModel.kt
@@ -107,7 +107,7 @@ class AccountDetailsViewModel
             AnalyticsEvent.NEWSLETTER_OPT_IN_CHANGED,
             mapOf(SOURCE_KEY to NewsletterSource.PROFILE.analyticsValue, ENABLED_KEY to isChecked),
         )
-        settings.marketingOptIn.set(isChecked, needsSync = true)
+        settings.marketingOptIn.set(isChecked, needsSync = false)
     }
 
     override fun onCleared() {
```

1. Build and install the app with ⬆️that⬆️  change.
2. Ensure that the settings sync feature flag is toggled on
3. Tap the refresh button on the Profile tab to make sure all settings are up-to-date
4. Tap the account button on the Profile tab
5. Change the newsletter toggle
6. Tap the refresh button on the Profile tab
7. Verify that the call to the `/named_settings/update` endpoint does not include any changes, i.e., it includes `"changed_settings: {}"`

## Checklist

### DIFF
--- a/modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/UserSetting.kt
+++ b/modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/UserSetting.kt
@@ -14,14 +14,10 @@ abstract class UserSetting<T>(
 
     private val modifiedAtKey = "${sharedPrefKey}ModifiedAt"
 
-    fun getModifiedAtServerString(): String? = sharedPrefs.getString(modifiedAtKey, null)
+    private fun getModifiedAtServerString(): String? = sharedPrefs.getString(modifiedAtKey, null)
 
     fun getModifiedAt(): Instant? = tryOrNull {
         getModifiedAtServerString()?.let { Instant.parse(it) }
-    }
-
-    fun doesNotNeedSync() {
-        updateModifiedAtServerString(null)
     }
 
     // Returns the value to sync if sync is needed. Returns null if sync is not needed.
@@ -58,31 +54,26 @@ abstract class UserSetting<T>(
 
     protected abstract fun persist(value: T, commit: Boolean)
 
-    fun set(value: T, commit: Boolean = false, needsSync: Boolean = false) {
+    open fun set(value: T, commit: Boolean = false, needsSync: Boolean = false) {
         persist(value, commit)
         _flow.value = value
-
-        if (needsSync) {
-            updateModifiedAtServerString()
-        }
+        val modifiedAt = if (needsSync) Instant.now().toString() else null
+        updateModifiedAtServerString(modifiedAt)
     }
 
-    private fun updateModifiedAtServerString(modifiedAt: String? = Instant.now().toString()) {
-        val valueToPersist = if (modifiedAt == null) {
-            // Allow persisting of nulls
-            null
-        } else {
+    // A null parameter reflects a setting that does not need to be synced
+    private fun updateModifiedAtServerString(modifiedAt: String?) {
+        if (modifiedAt != null) {
             val parsable = tryOrNull { Instant.parse(modifiedAt) } != null
             if (!parsable) {
+                // Only persist the string if it is parsable
                 LogBuffer.e(LogBuffer.TAG_INVALID_STATE, "Cannot set invalid modified at server string: $modifiedAt")
                 return
             }
-            // Only persist the string if it is parsable
-            modifiedAt
         }
 
         sharedPrefs.edit().run {
-            putString(modifiedAtKey, valueToPersist)
+            putString(modifiedAtKey, modifiedAt)
             apply()
         }
     }
@@ -299,6 +290,7 @@ abstract class UserSetting<T>(
     ) {
         override fun get(): T = initialValue
         override fun persist(value: T, commit: Boolean) {}
+        override fun set(value: T, commit: Boolean, needsSync: Boolean) {}
     }
 }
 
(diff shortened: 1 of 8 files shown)


modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/sync/SyncSettingsTask.kt
```diff
@@ -30,15 +30,6 @@ class SyncSettingsTask(val context: Context, val parameters: WorkerParameters) :
...
 
-            listOf(
-                settings.freeGiftAcknowledged,
-                settings.marketingOptIn,
-                settings.podcastsSortType,
-                settings.skipBackInSecs,
-                settings.skipForwardInSecs,
-            ).forEach {
-                it.doesNotNeedSync()
-            }
```
mchowning, line 41: Getting rid of the need to do this feels like a big win because it's only a matter of time until someone added a new setting that needed to be synced and forgot to include it here (which would have meant the app would have sent values that had just been synced from the server back to the server on the next sync call).

modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/UserSetting.kt
```diff
@@ -61,28 +57,23 @@ abstract class UserSetting<T>(
     fun set(value: T, commit: Boolean = false, needsSync: Boolean = false) {
         persist(value, commit)
         _flow.value = value
-
-        if (needsSync) {
-            updateModifiedAtServerString()
-        }
+        val modifiedAt = if (needsSync) Instant.now().toString() else null
+        updateModifiedAtServerString(modifiedAt)
```
mchowning, line 61: The change here to set the modified at value to `null` if the setting does not need sync is the core change in this PR.

modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/UserSetting.kt
```diff
@@ -58,31 +54,26 @@ abstract class UserSetting<T>(
...
 
-    fun set(value: T, commit: Boolean = false, needsSync: Boolean = false) {
+    open fun set(value: T, commit: Boolean = false, needsSync: Boolean = false) {
         persist(value, commit)
         _flow.value = value
-
-        if (needsSync) {
-            updateModifiedAtServerString()
-        }
+        val modifiedAt = if (needsSync) Instant.now().toString() else null
```
MiSikora, line 60: It would be nice if `UserSettings` somehow accepted `Clock` so we can control the time to make this part more testable. Not a priority for this PR but something to keep in mind.

modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/sync/SyncSettingsTask.kt
MiSikora: Yes, that's super nice!

MiSikora (approved):
LGTM

PR #1703, opened 2024-01-15
Add new sync call and sync currently-syncing podcast settings with it

mchowning:
## Description

This PR takes the first step in adding a new call to "/user/sync/update", which will allow us to sync the user's podcast settings. This new call will take the place of the current "/sync/update" call. That means that the first step is to get the new call syncing everything that the current call syncs. This includes the currently syncing settings for:

- podcasts
- episodes
- filters
- folder
- bookmarks
- stats

This PR adds the podcast specific syncing data to the new call. The other settings (episodes, filters, etc.) will be added in separate PRs.

> [!NOTE]
> This PR is targetting `update/sync-podcast-settings` instead of `main` because the new sync call does not yet have all the functionality of the old sync call, and I don't want it to appear like we have bugs in debug builds (where the sync settings feature flag is on by default) until I get the new sync call caught up. I don't think it'll take long to get caught up, so it should just be a few days before we're able to merge `update/sync-podcast-settings` to `main`.

> [!WARNING]
> The diff on this PR is very large because it is using protobuf and adds all the protobuf generated classes for this sync call (generated using the `scripts/update_protobuf.sh` script). I added all those classes in the first commit (a9707810d6e73b04c86e7772236ca38e9f762361), so you can review the substantive changes in this PR much more easily by looking at the other commit.

## Testing Instructions

For all these tests you will need to:
1. Install the app on two devices
2. Log into the same Pocket Casts account on both devices
3.  Turn on sync settings under Profile → ⚙️ → Beta features → Settings Sync

### 1. Syncs subscribed & date added
1. Device 1:
    1. Subscribe to a podcast
    2. Tap the refresh button on the Profile tab
    3. Using the database inspector, note the value in the podcast table for this podcast's `added_date` column
2. Device 2:
    1. Tap the refresh button on the Profile tab
    2. Observe that you are now subscribed to the podcast you subscribed to on Device 1
    4. Using the database inspector, note the value in the podcast table for this podcast's `added_date` column is the value that you noted on Device 1 after refreshing

### 2. Syncs skip first & skip last
1. Device 1:
    1. Go to the podcast screen for a subscribed podcast
    2. Tap the ⚙️ to view the podcast's settings
    5. Change the "Skip first" and "Skip last" values
    6. Tap the refresh button on the Profile tab 
2. Device 2:
    1. Tap the refresh button on the Profile tab
    2. Confirm that the podcast's settings reflect the "Skip first" and "Skip last" settings that were synced from Device 1

[...]

PR #1708, opened 2024-01-16
Adding incremental sync of episodes with the new endpoint

mchowning:
## Description
This is a follow-up to #1703 that updates the new sync call (`/user/sync/update`) to sync episode information. This new call will take the place of the current "/sync/update" call. 

Before this PR, the information that was being synced in the old call that the new call is also syncing is for:

- podcasts

The information that was not _yet_ being synced by the new call is:

- episodes
- folder
- filters
- bookmarks
- stats

This PR is adding syncing of **episode** data to the new call.

## Testing Instructions

For all these tests you will need to:
1. Install the app on two devices
2. Log into the same Pocket Casts account on both devices
3.  Turn on sync settings under Profile → ⚙️ → Beta features → Settings Sync

### 1. Syncs playing status
1. Device 1:
    1. Select an episode you have not interacted with before and mark that episode as played from the episode details screen.
    2. Tap the refresh button on the Profile tab
2. Device 2:
    1. Tap the refresh button on the Profile tab
    2. Verify that the episode details screen for the episode you marked as played on Device 1 shows the episode as played

### 2. Syncs played up to
1. Device 1:
    1. Begin playing an episode
    2. Skip to somewhere in the middle of the episode. Remember this episode and the position playback is at in the episode
    3. Begin playing some other episode
    5. Tap the refresh button on the Profile tab
2. Device 2:
    1. Tap the refresh button on the Profile tab
    2. Begin playing the episode that you remembered the playback position for
    3. Observe that it begins playing at the same position the episode had on Device 1

### 3. Syncs star status
1. Device 1:
    1. Star an episode
    5. Tap the refresh button on the Profile tab
4. Device 2:
    1. Tap the refresh button on the Profile tab
    4. Verify that the episode you starred on Device 1 is now starred
    7. Unstar the episode
    1. Tap the refresh button on the Profile tab
5. Device 1:
    1. Tap the refresh button on the Profile tab
    4. Verify that the episode is no longer starred

### 4. Syncs archive status
1. Device 1:
    1. Archive an episode
    2. Tap the refresh button on the Profile tab
2. Device 2:
    1. Tap the refresh button on the Profile tab
    2. Verify that the episode you archived on Device 1 is now archived
    5. Unarchive the episode
    1. Tap the refresh button on the Profile tab
5. Device 1:
    1. Tap the refresh button on the Profile tab
    2. Verify that the episode is no longer archived

### 5. Syncs duration

There isn't a good way to manipulate this from the app's UI, so we'll manipulate and inspect the databases in this test.

1. Device 1:
    1. Using the database inspector, change the `duration` field to be different and the `durationModified` field to be the current seconds since the epoch for one episode of a podcast you are subscribed to
    2. Tap the refresh button on the Profile tab
2. Device 2:
    1. Tap the refresh button on the Profile tab
    2. Verify that the episode now reflects the changed duration from Device 1 on the podcast screen

## Checklist

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/PodcastSyncProcess.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/PodcastSyncProcess.kt
@@ -39,6 +39,7 @@ import au.com.shiftyjelly.pocketcasts.utils.featureflag.FeatureFlagWrapper
 import au.com.shiftyjelly.pocketcasts.utils.log.LogBuffer
 import com.google.protobuf.boolValue
 import com.google.protobuf.int32Value
+import com.google.protobuf.int64Value
 import com.google.protobuf.stringValue
 import com.google.protobuf.timestamp
 import com.pocketcasts.service.api.Record
@@ -47,6 +48,7 @@ import com.pocketcasts.service.api.int32Setting
 import com.pocketcasts.service.api.podcastSettings
 import com.pocketcasts.service.api.record
 import com.pocketcasts.service.api.syncUpdateRequest
+import com.pocketcasts.service.api.syncUserEpisode
 import com.pocketcasts.service.api.syncUserPodcast
 import io.reactivex.Completable
 import io.reactivex.Maybe
@@ -148,7 +150,8 @@ class PodcastSyncProcess(
         }
 
     private suspend fun performIncrementalSyncSuspend(lastModified: String) {
-        val syncUpdateRequest = getSyncUpdateRequest(lastModified)
+        val episodesToSync = episodeManager.findEpisodesToSync()
+        val syncUpdateRequest = getSyncUpdateRequest(lastModified, episodesToSync)
         val protobufResponse = syncManager.userSyncUpdate(syncUpdateRequest)
         val syncUpdateResponse = SyncUpdateResponse.fromProtobufSyncUpdateResponse(protobufResponse)
         processServerResponse(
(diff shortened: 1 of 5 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">No reviewers have been set for this PR yet. Please request a review from <strong>@ Automattic/pocket-casts-android</strong>.</td>
    </tr>
  </tbody>
</table>

</p>

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/PodcastSyncProcess.kt
```diff
@@ -931,44 +937,85 @@ class PodcastSyncProcess(
...
+                        val folderUuid = podcast.folderUuid
+                        value = if (folderUuid.isNullOrEmpty()) {
+                            Folder.homeFolderUuid
+                        } else {
+                            folderUuid
+                        }
+                    }
+
+                    isDeleted = boolValue { value = !podcast.isSubscribed }
+                    subscribed = boolValue { value = podcast.isSubscribed }
```
mchowning, line 961: This does seem a bit weird, so I double-checked and confirmed it's also what [iOS is doing](https://github.com/Automattic/pocket-casts-ios/blob/a081912bba9b143a547c4023175f3d5e643c336f/Modules/Server/Sources/PocketCastsServer/Public/Sync/SyncTask%2BLocalChanges.swift#L17-L18):

``` swift
podcastRecord.isDeleted.value = !podcast.isSubscribed()
podcastRecord.subscribed.value = podcast.isSubscribed()
```

mebarbosa (approved):
Nice work! 👍🏽 

I left just one comment but is not a blocker

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/PodcastSyncProcess.kt
mchowning: A comment here is a good idea. I've asked for some clarification around why both values exist (p1705498431040889-slack-C029BMW150R), and I'll add a comment in a separate PR when I get that clarification.

PR #1713, opened 2024-01-17
Add incremental sync of folder data with new endpoint

mchowning:
## Description
This follows-up on https://github.com/Automattic/pocket-casts-android/pull/1708 by updating the new sync call (/user/sync/update) to sync folder information. This new call will take the place of the current "/sync/update" call once it is fully implemented.

Before this PR, the information that was being synced in the old call that the new call is also syncing is for:

- podcasts
- episodes

The information that was not yet being synced by the new call is:

- folder
- filters
- bookmarks
- stats

This PR is adding syncing of folder data to the new call.

## Testing Instructions

For these tests you will need to:
1. Install the app on two devices
2. Log into the same Pocket Casts account on both devices
3.  Turn on sync settings under Profile → ⚙️ → Beta features → Settings Sync

### 1. Syncs folder state
1. Device 1:
    1. Go to the Podcast tab and tap the button to create a new folder (you'll need to be using a Plus or Patron account)
    2. Select some podcasts to include in the folder
    3. Give the folder a name
    4. Select a color other than the default color for the folder
    5. Tap "Save folder"
    6. You should be taken to the folder's screen
    7. Tap the three-dot overflow menu and and change the sort option for how podcasts are sorted in the folder
    8. Exit out of the folder screen to the top-level Podcasts tab screen
    9. Drag the folder to a different position
    10. Tap the refresh button on the Profile tab
2. Device 2:
    1. Tap the refresh button on the Profile tab
    2. Verify that the folder you created on device 1 is present with the correct:
        1. folder name
        2. folder color
        3. sort type for podcasts within the folder
        4. sort position on the top-level podcasts screen
    3. Delete the folder:
        1. Open the folder
        2. using the three-dot menu to access the "Edit folder" screen
        3. Tap "Delete folder"
    4. Tap the refresh button on the Profile tab
3. Device 1:
    1. Tap the refresh button on the Profile tab
    2. Go to the Podcasts tab and verify that the folder is no longer there

### 2. Syncs date folder was added

I don't think there is any way to see this from the app's UI, so we'll inspect the databases in this test.

1. Device 1:
    1. Add a new folder
    2. Tap the refresh button on the Profile tab
2. Device 2: tap the refresh button on the Profile tab
3. Use the database inspector to verify that the rows for the folder you added (in the `folders` table) have the same value in the `added_date` column on both devices.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

MiSikora (approved):
Works as described.

PR #1717, opened 2024-01-18
Add upload of stats data with the new sync endpoint

mchowning:
## Description
This follows-up on https://github.com/Automattic/pocket-casts-android/pull/1715 by updating the new sync call (/user/sync/update) to sync stats information. This new call will take the place of the current "/sync/update" call once it is fully implemented.

Before this PR, the information that was being synced in the old call that the new call is also syncing is for:

- podcasts
- episodes
- folder
- filters
- bookmark

The stats/device info was not yet being synced by the new call, and that is what this PR adds.

This stats/device sync is a bit different from the other kinds of sync because although we send the info when we make the call, that data doesn't come back in the call's response if it is different on the server. Instead, we make a separate stats call to obtain stats data. I've asked about the reason for that difference here: p1705597902387879-slack-C029BMW150R.

> [!NOTE]
> This PR is building on top of #1715. It would be good to merge that PR before we merge this one.

### 1. Listening Time
1. Confirm that the Profile → Stats screen is showing the same stats on both devices
2. Device 1: Open the Profile → Stats screen and leave it open
3. Device 2:
    1. Play a bit of an episode
    2. Verify that the listening time at the top of the Stats screen is now different from what is shown on Device 1
    3. Tap the refresh button on the Profile screen
4. Device 1:
    1. Note that the listening time at the top of the Stats screen is not the same as Device 2
    2. Exit and return to the Stats screen
    3. Verify that the listening time at the top of the Stats screen is now the same as Device 2

### 2. Skipping
1. Confirm that the Profile → Stats screen is showing the same stats on both devices
2. Device 1: Open the Profile → Stats screen and leave it open
3. Device 2:
    1. Begin playing an episode and skip forward a few times
    2. Verify that the "Skipping" section on the Stats screen is now different from what is shown on Device 1
    3. Tap the refresh button on the Profile screen
4. Device 1:
    1. Note that the "Skipping" value is not the same as Device 2
    2. Exit and return to the Stats screen
    3. Verify that the "Skipping" value is now the same as Device 2

### 3. Variable speed
1. Confirm that the Profile → Stats screen is showing the same stats on both devices
2. Device 1: Open the Profile → Stats screen and leave it open
3. Device 2:
    1. From the full-screen player, turn the playback speed to 2x
    2. Play the episode for a few seconds
    3. Verify that the "Variable speed" section on the Stats screen is now different from what is shown on Device 1
    4. Tap the refresh button on the Profile screen
4. Device 1:
[...]

PR #1720, opened 2024-01-19
Add modified value to podcast settings

mchowning:
## Description
This PR builds on top of https://github.com/Automattic/pocket-casts-android/pull/1717 and fixes a bug where synced settings could get overwritten by subsequent syncs.

I initially thought we wouldn't need to start saving and sending modified times with these events, but then I realized that not doing that (and instead just sending the request time as the time the setting was modified) does not work well because it means that a device that does not have the current settings from the server can overwrite the settings on the server. At a high level, this is the kind of thing that can happen:
1. Change `setting1` on device1 and sync it
2. Before syncing device2, change `setting2` on device2, then sync device2
3. the sync of `device2` essentially "reverts" the update to setting1 that was synced from `device1` because device2 sends all its settings.

This is not a new issue—the same thing can happen with the old sync call—but now that we're going to be syncing a lot more settings the chances that this will happen and the harm it could cause are significantly higher, so I think we need to send the modifedAt times to avoid the issue. 

This PR is adding those modifiedAt times for the two podcast settings that are currently being synced ("skip first" and "skip last"). We'll need to do this same kind of handling for the other podcast settings that we sync.

> [!WARNING]
> This PR does not ensure that a user's settings are synced properly after an upgrade (i.e., when we don't have a `modifiedAt` time in the db). I am using a default `modifiedAt` time of 0 in this PR, but that doesn't work when a user upgrades because a setting synced with a modified time of `0` seems to get discarded by the server. This is something we need to address before this feature is released.

### 1. Migration
1. Device 1: install a build without the changes on this PR (for example, you could install from the `update/sync-podcast-settings` branch)
3. Device 1:
    1. Create a new account and log in
    2. Subscribe to a podcast
    3. Tap the refresh button on the profile tab
    5. Install a build from this PR
    6. Tap the refresh button on the profile tab
    7. ✅ Observe that the app does not crash or show any other problems because of a failed db migration

### 2. Syncing conflicting settings

> [!NOTE]
> When testing this, note that the podcast settings screen does not react to settings changes, so you will need to make sure you reload that screen to see any settings changes reflected.

1. Install a build from this PR on two devices
2. Log into the same account on both devices
3. Ensure the two devices are fully synced up by tapping the refresh button on the profile screen on both devices
1. Device 1:
    1. Go to the settings for a subscribed podcast (⚙️ on the page for the podcast) and change the "skip first" setting (do not change "skip last")
    4. Tap the refresh button on the profile tab
2. Device 2:
    1. Go to the settings for the same podcast (⚙️ on the page for the podcast) and change the "skip last" setting (do not change "skip first"). The state we want to get into here is where the server has a new value for one setting ("skip first") and locally we have a new value for a different setting ("skip last")
    2. Tap the refresh button on the profile tab
    5. ✅ Verify that the "skip first" setting has been updated to match Device 1 and the "skip last" setting has not changed
3. Device 1:
    1. Tap the refresh button on the profile tab
    2. ✅  Verify that the "skip first" setting has not changed and that the "skip last" setting has been updated to match Device 2. In other words, Device 1 and Device 2 should now have identical "skip first" and "skip last" settings.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #1729, opened 2024-01-23
And incremental sync for folders, filters and podcasts settings

MiSikora:
## Description

This PR merges all of the changes related to incremental sync that Matt's been working on. [Here is the PR list](https://github.com/Automattic/pocket-casts-android/issues?q=label%3A%22sync+settings%22+is%3Aclosed+created%3A%3E%3D2024-01-15+merged%3A%3C%3D2024-01-23). All of them has been tested and work. This is only a sync with `main`.

Syncing is hidden behind a feature flag that is enabled by default only for debug builds so it should be safe to merge it.

https://github.com/Automattic/pocket-casts-android/blob/46d9385a74045b565df12867185afab05de3cda8/modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/featureflag/Feature.kt#L30-L37

https://github.com/Automattic/pocket-casts-android/blob/46d9385a74045b565df12867185afab05de3cda8/modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/sync/SyncSettingsTask.kt#L24-L29

https://github.com/Automattic/pocket-casts-android/blob/46d9385a74045b565df12867185afab05de3cda8/modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/sync/SyncSettingsTask.kt#L44-L48

## Testing Instructions

I do not expect anyone to test all of this. It has already been tested during separate PRs. Just make sure that CI passes.

## Checklist
 
- [ ] with different themes


MtndaleRedneck:
What is all this? Will this support syncing podcast settings across devices so you can finally switch to a new phone without resetting up everything?

MiSikora:
> What is all this? Will this support syncing podcast settings across devices so you can finally switch to a new phone without resetting up everything?

Yes, this is the idea. Syncing most settings between devices. But it's a work in progress.

MtndaleRedneck:
> > What is all this? Will this support syncing podcast settings across devices so you can finally switch to a new phone without resetting up everything?
> 
> Yes, this is the idea. Syncing most settings between devices. But it's a work in progress.

Finally!! Will there be the ability to move all your downloads also? Don't understand why it shouldn't be possible to move all the app data over from one phone to another when setting up a new phone. Got a new phone 2 years ago and lost a ton of downloads I couldn't even re-download as some of them such as old Joe Rogan's didn't even exist as rss feeds anymore. You transfer all your other data to your new phone. Why doesn't pocket casts data move? 

Was debating getting the s24 but loosing all my podcasts was a big factor in me not getting it.

CookieyedCodes:
@MtndaleRedneck from my understanding -please correct me if you changed your mind too :)- auto downloads will be one of the few things that won't be synced when I asked Matt (this is mainly to make new devices fresh -add to upnext + notifications will be synced tho- & along with the reason that if the RSS source is gone the uuid will mean nothing), because of this tho adding everything to upnext before new devices will be my prefence (along with folder backup) & if their are episodes that get caught due to lossed rss I would upload the to files (maby this could be automated tho 🤔)

tho I think their should be an option to sync the auto downloads list in advanced settings? imo

MtndaleRedneck:
@CookieyedCodes we argued about this on slack already but I'm still not understanding why this should be so hard to implement. When I get a new phone I use Samsung Smart Switch and transfer all my 400gb+ of data to my new phone. I log into WhatsApp for example and everything is there. All my old chats. Why can't Pocket Casts do the same? Why should any data be lost? This doesn't need to be an online thing at all. The data is already downloaded somewhere on my phone. Move it to the new phone, log in, and Bob's your uncle.

MtndaleRedneck:
@CookieyedCodes dont know what uuid is, but I'm not asking for an online data sync at all. All these podcasts including the Joe Rogan's which have been downloaded already exist as data downloaded onto my old phone. I'm connecting my new phone to my old phone directly with a USB-C cable. Just transfer the data directly over.

CookieyedCodes:
@MtndaleRedneck I'm aware 😅, fair enough, I'll try to explain it, when pocketcasts downloads it converts the file  it to a string of letters & numbers(this is sting is uuid this uuid is tied to the share episode functionality & I would say is helping with sync), this uuid linked to the RSS file, if the RSS no longer exists pocketcasts can't place it so the file errors, you can copy the folder structure from one device to another but from my understanding but if the RSS isn't their pocketcasts will say nope (hence why I use cloud files)

MtndaleRedneck:
@CookieyedCodes think I finally get what you mean. Any devs who can weigh in on this? I understand with the way the system is currently implemented why it can't be instituted. But does it need to be that way? Any way around it? I have 100+gb of podcasts downloaded that I really don't want to lose next time I upgrade my phone.

MiSikora:
@MtndaleRedneck There are multiple things that I'll try to answer to the best of my abilities.

> Will there be the ability to move all your downloads also?

Unfortunately, no. This is only about syncing settings between devices. Things like podcasts configuration, filters, theming and so on.

> Don't understand why it shouldn't be possible to move all the app data over from one phone to another when setting up a new phone. (...) Why doesn't pocket casts data move?

Simply put, it's not something that has been prioritized for the product. We have an internal feature request/idea board that we use to evolve the product. Choosing what we do is a mix of demand, necessity, complexity, product vision, resources, and many other different factors.

> I log into WhatsApp for example and everything is there. All my old chats. Why can't Pocket Casts do the same? Why should any data be lost? This doesn't need to be an online thing at all.

You conflate two things here. WhatsApp stores those things online and syncs your device's data with the cloud. Communicators like WhatsApp deal with much less data than media apps like Pocket Casts. Having downloaded podcasts stored somewhere on our servers wouldn't be manageable.

Maybe there's a feature in WhatsApp that allows you to backup your data offline and then import it. I'm not sure off the top of my head but let's just say there is. Pocket Casts doesn't have it due to mentioned reasons of demand, complexity, and so on.

> I'm connecting my new phone to my old phone directly with a USB-C cable. Just transfer the data directly over.

It's really not as simple as copying files from one device to another. @CookieyedCodes mentioned some of the edge cases but in general Android with newer and newer versions is less and less friendly when it comes to file access. And there's the complexity of making sure that transferred files didn't lose integrity and are still understood by the app. I hope this somewhat cleared the picture.

Having said all of that. I'll add your vote/comments to our internal feature request board. In the future you can use our [forum](https://forums.pocketcasts.com/) or [contact us](https://support.pocketcasts.com/contact-us/) directly if you have any feature requests.

mebarbosa (approved):

Issue #1739, opened 2024-01-25
Enable settings sync between devices


MiSikora:
This issue is for tracking synced settings.

```[tasklist]
### Podcast settings
- [ ] `notification`: This one is a little bit tricky because it needs to be coordinated somehow with syncing global settings for notifications.
- [ ] `autoArchive`: No point in syncing ATM because `autoArchivePlayed`, `autoArchiveInactive`, and `autoArchiveEpisodeLimit` are missing on the server.
```

```[tasklist]
### Named settings
- [ ] `notifications: `settings.notifyRefreshPodcast`. See podcast `notification` setting.
- [ ] `theme` - `Settings.theme`.
- [ ] `useEmbeddedArtwork` - `Settings.useEmbeddedArtwork`.
- [ ] `trimSilence`: `Settings.globalPlaybackEffects`. Shared with `playbackSpeed` and `volumeBoost`.
- [ ] `playbackSpeed`:`Settings.globalPlaybackEffects`. Shared with `trimSilence` and `volumeBoost`.
- [ ] `volumeBoost`: `Settings.globalPlaybackEffects`. Shared with `trimSilence` and `playbackSpeed`.
- [ ] `badges`: `Settings.podcastBadgeType`.
- [ ] `rowAction`: `Settings.streamingMode`.
- [ ] `upNextSwipe`: `Settings.upNextSwipe`
- [ ] `episodeGrouping`: `Settings.podcastGroupingDefault`.
- [ ] `showArchived`: `Settings.showArchivedDefault`.
- [ ] `mediaActions`: `Settings.customMediaActionsVisibility`.
- [ ] `mediaActionsOrder`: `Settings.mediaControlItems`.
- [ ] `keepScreenAwake`: `Settings.keepScreenAwake`.
- [ ] `openPlayer`: `Settings.openPlayerAutomatically`.
- [ ] `intelligentResumption`: `Settings.intelligentPlaybackResumption`.
- [ ] `playUpNextOnTap`: `Settings.tapOnUpNextShouldPlay`.
- [ ] `notificationActions`: `Settings.newEpisodeNotificationActions`.
- [ ] `playOverNotifications`: `Settings.playOverNotification`.
- [ ] `hideNotificationOnPause`: `Settings.hideNotificationOnPause`.
- [ ] `autoUpNextLimit`: `Settings.autoAddUpNextLimit`.
[...]

PR #1759, opened 2024-01-30
Sync episode grouping setting

MiSikora:
## Description

One of the settings listed for sync project #1739. This PR adds syncing to the global settings for episode grouping.

### ⚠️ Warning ⚠️ 

It can be confusing that sync setting does not apply automatically to podcasts that are on a device. This is by design. The setting on a synced device applies only to new podcasts. If users want to apply the grouping on a synced device they need to re-trigger the setting.

3. D1: Go to `Profile`/`Settings (cog icon)`/`General`/`Podcast episode grouping`.
4. D1: Change the value of the setting. You can dismiss the popup about applying the setting to available podcasts.
5. D1: Sync the device in the `Profile`.
6. D2: Sync the device in the `Profile`.
7. D2: Subscribe to a new podcast.
8. D2: Open that podcast and check if episode grouping is what you set on D1.
9. Perform this test for each grouping mode.

---

3. D1: Go to `Profile`/`Settings (cog icon)`/`General`/`Podcast episode grouping`.
4. D1: Change the value of the setting. You can dismiss the popup about applying the setting to available podcasts.
5. D1: Sync the device in the `Profile`.
6. D2: Sync the device in the `Profile`.
7. D2: Go to `Profile`/`Settings (cog icon)`/`General`/`Podcast episode grouping`.
8. D2: Confirm already selected grouping.
9. D2: Confirm that you want to apply changes to available podcasts.
10. D2: Open a podcast that you are already subscribed to.
11. D2: Check if the grouping is correctly applied.

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/PodcastGrouping.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/PodcastGrouping.kt
@@ -9,21 +9,35 @@ import au.com.shiftyjelly.pocketcasts.models.to.PodcastGrouping.Season.getSeason
 import au.com.shiftyjelly.pocketcasts.models.type.EpisodesSortType
 import au.com.shiftyjelly.pocketcasts.localization.R as LR
 
-sealed class PodcastGrouping(@StringRes val groupName: Int, val sortFunction: ((PodcastEpisode) -> Int)?) {
+sealed class PodcastGrouping(
+    @StringRes val groupName: Int,
+    val serverId: Int,
+    val sortFunction: ((PodcastEpisode) -> Int)?,
+) {
     companion object {
         val All
             get() = listOf(None, Downloaded, Unplayed, Season, Starred)
+
+        fun fromServerId(id: Int) = All.find { it.serverId == id } ?: None
     }
 
     abstract fun groupTitles(index: Int, context: Context): String
 
-    object None : PodcastGrouping(LR.string.none, null) {
+    data object None : PodcastGrouping(
+        groupName = LR.string.none,
+        serverId = 0,
+        sortFunction = null,
+    ) {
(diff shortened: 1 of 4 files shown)


CookieyedCodes:
Will the local grouping settings be synced under the sync whole sync project?

MiSikora:
It looks like the setting is missing in the project. Thanks for asking. I'll pass it on.

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/PodcastGrouping.kt
```diff
@@ -9,21 +9,35 @@ import au.com.shiftyjelly.pocketcasts.models.to.PodcastGrouping.Season.getSeason
...
+sealed class PodcastGrouping(
+    @StringRes val groupName: Int,
+    val serverId: Int,
+    val sortFunction: ((PodcastEpisode) -> Int)?,
+) {
     companion object {
         val All
             get() = listOf(None, Downloaded, Unplayed, Season, Starred)
+
+        fun fromServerId(id: Int) = All.find { it.serverId == id } ?: None
```
Mukuljangir372, line 21: We should replace `find` with `firstOrNull` for reducing the unneccesary iterations.

MiSikora: What unnecessary iterations?

```kotlin
@kotlin.internal.InlineOnly
public inline fun <T> Iterable<T>.find(predicate: (T) -> Boolean): T? {
    return firstOrNull(predicate)
}
```

https://github.com/JetBrains/kotlin/blob/09f2f365452396a0b6b835b1b7862c5d1b33dc69/libraries/stdlib/common/src/generated/_Collections.kt#L165-L168

ashiagr (approved):

PR #1773, opened 2024-01-31
Sync warn on metered network setting

MiSikora:
## Description

One of the settings listed for sync project #1739. This PR adds syncing to the global setting for warning on metered networks. This setting can be changed in multiple ways: Phone app and WearOS app.

## Testing prerequisites

To test wear OS you need either a physical device or prepare the app so you can use it with the wearOS emulator.

Build the *release* version of the app with this condition inverted.

```diff
--- a/modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/featureflag/Feature.kt
+++ b/modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/featureflag/Feature.kt
@@ -30,7 +30,7 @@ enum class Feature(
     SETTINGS_SYNC(
         key = "settings_sync",
         title = "Settings Sync",
-        defaultValue = BuildConfig.DEBUG,
+        defaultValue = !BuildConfig.DEBUG,
         tier = FeatureTier.Free,
         hasFirebaseRemoteFlag = false,
         hasDevToggle = true,

```

## Testing Instructions

Each test requires two devices to check syncing. Sync process looks as follows.

1. Make sure that both devices are synced before starting the test by refreshing apps in the profile.
2. Change the setting on the device 1.
3. Refresh settings on the device 1.
4. Refresh settings on the device 2.
5. Check settings on the device 2.

### Phone app

The setting is placed in `Profile`/`Settings (cog icon)`/`Storage & data use`/`Warn before using data`. When this setting is enabled and you are using metered network you should be able to trigger these popups when playing or downloading an episode.

| Download | Playback |
| - | - |
| Screenshot_20240131_133726 | Screenshot_20240131_133705 |

> [!note]
> Playback popup won't appear again for the same episode once it's been played. However it will stop playing if during the playback you'll switch from a not metered to a metered network.

### WearOS app

The setting is placed in `Settings`/`Metered data warning`. When this setting is enabled and you are using metered network you should be able to trigger these popups when playing or downloading an episode.

| Download | Playback |
| - | - |
| Screenshot_20240131_133253 | Screenshot_20240131_132730 |

## Checklist
 
- [ ] with different themes

### DIFF
--- a/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/settings/SettingsViewModel.kt
+++ b/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/settings/SettingsViewModel.kt
@@ -59,16 +59,24 @@ class SettingsViewModel @Inject constructor(
                     _state.update { it.copy(refreshState = refreshState) }
                 }
         }
+        viewModelScope.launch {
+            settings.warnOnMeteredNetwork.flow.collectLatest { warnOnMeteredNetwork ->
+                _state.update { it.copy(showDataWarning = warnOnMeteredNetwork) }
+            }
+        }
+        viewModelScope.launch {
+            settings.backgroundRefreshPodcasts.flow.collectLatest { refreshInBackground ->
+                _state.update { it.copy(refreshInBackground = refreshInBackground) }
+            }
+        }
     }
 
     fun setWarnOnMeteredNetwork(warnOnMeteredNetwork: Boolean) {
-        settings.warnOnMeteredNetwork.set(warnOnMeteredNetwork, needsSync = false)
-        _state.update { it.copy(showDataWarning = warnOnMeteredNetwork) }
+        settings.warnOnMeteredNetwork.set(warnOnMeteredNetwork, needsSync = true)
     }
 
     fun setRefreshPodcastsInBackground(isChecked: Boolean) {
         settings.backgroundRefreshPodcasts.set(isChecked, needsSync = false)
-        _state.update { it.copy(refreshInBackground = isChecked) }
     }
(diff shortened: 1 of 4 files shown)


wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/settings/SettingsViewModel.kt
```diff
@@ -59,16 +59,24 @@ class SettingsViewModel @Inject constructor(
                     _state.update { it.copy(refreshState = refreshState) }
                 }
         }
+        viewModelScope.launch {
+            settings.warnOnMeteredNetwork.flow.collectLatest { warnOnMeteredNetwork ->
+                _state.update { it.copy(showDataWarning = warnOnMeteredNetwork) }
+            }
+        }
```
MiSikora, line 66: This change was made so that the setting switch changes position automatically after refreshing.

wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/settings/SettingsViewModel.kt
```diff
@@ -59,16 +59,24 @@ class SettingsViewModel @Inject constructor(
...
+        viewModelScope.launch {
+            settings.backgroundRefreshPodcasts.flow.collectLatest { refreshInBackground ->
+                _state.update { it.copy(refreshInBackground = refreshInBackground) }
+            }
+        }
```
MiSikora, line 71: We don't plan on syncing this setting but I made this change for completeness.

PR #1780, opened 2024-02-01
Sync theme setting

MiSikora:
## Description

One of the settings listed for sync project #1739. This PR adds syncing of the global settings for the app theme.

## Testing Instructions

2. On both devices disable `Profile`/`Setting (cog icon)`/`Appearance`/`Use Android Light/Dark Mode`. This setting is not synced and might be problematic during the test.
3. Make sure that both devices are synced before starting the test by refreshing apps in the profile.
4. D1: Go to `Profile`/`Setting (cog icon)`/`Appearance`.
5. D1: Change the theme.
6. D1: Sync the device in the `Profile`.
7. D2: Sync the device in the `Profile`.
8. D2: Notice that theme changes. Screen might jump because applying theme in Pocket Casts recreates the Activity.
9. Test this with different themes.

## Checklist
 
- [ ] with different themes


mebarbosa (approved):
LGTM

PR #1796, opened 2024-02-06
dependabot.yml

AbdhilahiRWabwire:
# Dependabot

- dependabot.yml - Automated dependency updates built into GitHub
 
- https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
 
- https://github.com/dependabot


MiSikora:
Hey @AbdhilahiRWabwire thanks for the contribution. However, I don't think it is something we can just add. We'd need to make a dependencies audit, and probably update some of them manually beforehand.

AbdhilahiRWabwire:
Dependabot can handle all of that. Dependabot will submit all of the audits, reports && submit the pull requests. Yes, it is that easy.

MiSikora:
No, for example, it won't know if introducing a dependency update causes a bug. I'm not opposed to automating dependency updates process. But the first step shouldn't be to slap it on the repository, get tons of updates, and hope for the best. We need to do the manual groundwork first.

AbdhilahiRWabwire:
@MiSikora I can run it in my fork first and submit the fixes for you if you like

MiSikora:
@AbdhilahiRWabwire I appreciate the effort but this is something that I'd prefer we do on our own at this stage. Having one PR with all dependencies updated is not something that we can reasonably test and verify.

AbdhilahiRWabwire:
@MiSikora I wasn't finished yet, I was going to run gradle builds and unit tests but I understand if you want handle it within your team.

PR #1798, opened 2024-02-06
Use last sync time as a default modified at timestamp

### LINKED ISSUE #1762: Use last sync timestamp for unknown timestamps
Using `0` for unknown timestamps is rejected by the server. Last sync seems to be the best option to allow syncing on the feature launch.

MiSikora:
## Description

This PR uses last sync timestamp as a default modification timestamp when the value of a setting modification timestamp is unknown. We do this because we want to make sure that all settings sync when the feature is enabled. Otherwise users would have to re-trigger all of the settings one by one.

Closes #1762 

## Testing Instructions

2. D1: Install the app from c86132a18ae6eebe67b65ccde33f46e9173c7dd1.
3. D1: Sign up with a new account.
4. D1: Subscribe to any podcast.
5. D1: Go to this podcast settings.
6. D1: Enable notifications for this podcast.
7. D1: Set custom playback effects for this podcast.
8. D1: Go to the global settings in the profile.
9. D1: Go to `General`.
10. D1: Change `Row action` setting.
11. D1: Change `Up Next swipe` setting.
12. D1: Change `Podcast episode grouping` setting.
13. D1: Change `Media notification controls` setting`.
14. D1: Go back to general settings and open `Notifications`.
15. D1: Change `Play over notifications` setting.
16. D1: Sync the device in the profile.
17. D1: Close the app.
18. D1: Update the app with the one from this branch.
19. D1: Sync the device in the profile.
20. D1: Make sure that all of the settings that you changed are still applied.
21. D2: Install the app from this branch.
22. D2: Sign in with the same account as in step 3.
23. D2: Sync the device in the profile.
24. D2: Check if all of the setting that you changed applied (both podcast and global settings).

## Checklist
 
- [ ] with different themes

### DIFF
--- a/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/preferences/UserSettingTest.kt
+++ b/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/preferences/UserSettingTest.kt
@@ -2,6 +2,7 @@ import android.content.Context
 import android.content.SharedPreferences
 import androidx.test.platform.app.InstrumentationRegistry
 import au.com.shiftyjelly.pocketcasts.preferences.UserSetting
+import java.time.Instant
 import junit.framework.TestCase.assertEquals
 import junit.framework.TestCase.assertNotNull
 import junit.framework.TestCase.assertNull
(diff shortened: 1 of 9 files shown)


mebarbosa (approved):
LGTM 👍🏽

PR #1821, opened 2024-02-12
Sync system theme settings

MiSikora:
## Description

One of the settings listed for sync project #1739. This PR adds syncing to the global settings for system theme controls.

## Testing Instructions

> [!note]
> You need to test this with staging because changes aren't deployed to production, yet.
> 
> To test on staging use the `debug` variant instead of the `debugProd` one.

3. D1: Go to `Profile`/`Settings (cog icon)`/`Appearance`.
4. D1: Toggle `Use Android Light/Dark Mode` to `OFF`.
5. D1: Sync the device in the `Profile`.
6. D2: Sync the device in the `Profile`.
7. D2: Notice that the app doesn't follow Android's Light/Dark Mode.
8. D2: Go to `Profile`/`Settings (cog icon)`/`Appearance`.
9. D2: Set `Classic` theme (this will trigger default light theme setting).
10. D2: Set `Dark contrast` theme (this will trigger default dark theme setting).
11. D2: Toggle `Use Android Light/Dark Mode` to `ON`.
12. D2: Sync the device in the `Profile`.
13. D1: Sync the device in the `Profile`.
14. D1: Notice that the app follows Android's Light/Dark Mode and uses `Classic` and `Dark contrast` as themes.

## Checklist
 
- [ ] with different themes

### DIFF
--- a/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
+++ b/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
@@ -420,7 +420,7 @@ class MainActivity :
 
         mediaRouter = MediaRouter.getInstance(this)
 
-        ThemeSettingObserver(theme, settings.theme, settings.useSystemTheme, this).observeThemeChanges()
+        ThemeSettingObserver(this, theme, settings.themeReconfigurationEvents).observeThemeChanges()
     }
 
     private fun resetEoYBadgeIfNeeded() {
--- a/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
+++ b/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
@@ -420,7 +420,7 @@ class MainActivity :
 
         mediaRouter = MediaRouter.getInstance(this)
 
-        ThemeSettingObserver(theme, settings.theme, this).observeThemeChanges()
+        ThemeSettingObserver(theme, settings.theme, settings.useSystemTheme, this).observeThemeChanges()
     }
 
     private fun resetEoYBadgeIfNeeded() {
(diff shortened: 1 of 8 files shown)


ashiagr:
After toggling `Use Android Light/Dark Mode` to ON and syncing both devices, the toggle remained OFF on the second device. I installed the debug variant on both devices and the sync settings feature was enabled. Am I missing something?

https://github.com/Automattic/pocket-casts-android/assets/1405144/5dcb75f8-b147-464d-9544-4d3e948861cc

MiSikora:
Interesting, I haven't checked with the actual toggle. I'll see what's going on. Thanks for testing that.

MiSikora:
The behavior is more complicated than I thought. This led to unexpected behavior when applying themes within our application. The root of this problem can be traced back to the following block of code:

https://github.com/Automattic/pocket-casts-android/blob/9d139388f92902365cdf68526cb68069e206a696/modules/services/ui/src/main/java/au/com/shiftyjelly/pocketcasts/ui/theme/Theme.kt#L215-L235

This particular section of code was designed to ensure the deactivation of our light/dark mode setting whenever there was a mismatch with the selected theme. However, the complexity of the issue necessitated a slight modification in our approach.

1. I've adopted a more holistic approach towards theme changes. Now, whenever a theme-related setting is altered, we conduct a thorough check and subsequently trigger a corresponding event to ensure all changes are reflected accurately.
2. I identified and rectified a bug within the Theme class where it erroneously relied on a cached version of `activeTheme` rather than the persisted one. This flaw went unnoticed previously as the device itself was the sole source of truth regarding the theme.

These modifications may slightly alter the theme behavior, particularly in scenarios where the system's light/dark mode is set to follow the device settings, and these settings vary across devices. However, this adjustment ensures a more consistent and intuitive user experience, aligning the theme behavior across different devices.

ashiagr (approved):
I have tested that `Use Android Light/Dark Mode` is now properly synced and themes are appropriately set.

Thanks so much for discovering and fixing an existing bug! 🥇

PR #1833, opened 2024-02-14
Update JDK to 17

MiSikora:
## Description

While working on automating proto generation in #1831 I encountered a problem with Java compatibility. This PR switches our Java target and compatibility versions to 17. I had to fix some of the tests because we were mocking final classes and it is no longer possible with newer versions of JDK. There are some discussions on Mockito project about it with potential solutions like switching to MockK. But I think it is much simpler, and better, to not mock unnecessary objects.

## Testing Instructions

CI checks should be enough. You can smoke test the apps.

## Checklist
 
- [ ] with different themes

### DIFF
--- a/wear/build.gradle.kts
+++ b/wear/build.gradle.kts
@@ -66,7 +66,6 @@ android {
     }
 
     kotlinOptions {
-        jvmTarget = "1.8"
         // Allow for widescale experimental APIs in Alpha libraries we build upon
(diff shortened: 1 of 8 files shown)


wear/build.gradle.kts
```diff
@@ -66,7 +66,6 @@ android {
     }
 
     kotlinOptions {
-        jvmTarget = "1.8"
```
MiSikora, line 69: This is taken from `base.gradle`.

ashiagr (approved):

PR #1834, opened 2024-02-14
Use correct setting when changing auto play preference

MiSikora:
## Description

We had a bug on automotive. We updated a wrong setting when `Auto play` preference was changed.

## Testing Instructions

1. Sign in to Automotive app.
2. Go to Settings.
3. Toggle `Auto Play` setting.
4. Go back to the main screen.
5. Go back to Settings.
6. `Auto Play` setting should keep its value.

## Checklist
 
- [ ] with different themes


mebarbosa (approved):
It's working as expected!

PR #1842, opened 2024-02-15
Fix/sync skip last

MiSikora:
## Description

There was a bug with a podcast skip last setting where we didn't use a correct value for mapping.

## Testing Instructions

2. Sign in with an account a podcast added.
3. Make sure that both devices are synced before starting the test by refreshing apps in the profile.
4. D1: Go to `Podcasts`/`Select a podcast`/`Settings (cog icon)`.
5. D1: Change the value of `Skip last`.
6. D1: Sync the device in the `Profile`.
7. D2: Sync the device in the `Profile`.
8. D2: Check that the podcast from step 4 has `Skip last` value synced.

## Checklist
 
- [ ] with different themes


mebarbosa (approved):
LGTM

Issue #1807, opened 2024-02-08
Deselecting Chapters


ashiagr:
Add the ability to deselect chapters and sync them between devices.

### Basics

- [x] Feature Flag (https://github.com/Automattic/pocket-casts-android/pull/1811)
- [x] Update the chapter UI to include way to choose which chapters to play (https://github.com/Automattic/pocket-casts-android/pull/1827)
- [x] While playing an episode and coming to the end of a chapter, skip to the next available chapter position (https://github.com/Automattic/pocket-casts-android/pull/1850)
- [x] Add skip chapters header (https://github.com/Automattic/pocket-casts-android/pull/1874)
- [x] Update design https://github.com/Automattic/pocket-casts-android/pull/1886

### Paid Feature

- [x] Add the locked UI if the feature is not available (https://github.com/Automattic/pocket-casts-android/pull/1880)
- [x] Show the upsell when tapped  (https://github.com/Automattic/pocket-casts-android/pull/1880)
- [x] Enable the feature for Patron under early access  (https://github.com/Automattic/pocket-casts-android/pull/1880)
- [x] Then Plus for full access  (https://github.com/Automattic/pocket-casts-android/pull/1880)

### Syncing

- [x] Send the selections to the server (https://github.com/Automattic/pocket-casts-android/pull/1916)
- [x] Apply incoming selections from the server (https://github.com/Automattic/pocket-casts-android/pull/1916)

### Watch

- ~~Skip deselected chapters on the watch when tapping the chapter next/previous buttons~~  - Chapter next/previous buttons not available in Wear OS app

### Other

- [x] Add What’s New (https://github.com/Automattic/pocket-casts-android/pull/1896) (need to update conditions)
- [x] Add the feature to the plus features list (https://github.com/Automattic/pocket-casts-android/pull/1895)
- [x] Tracks (https://github.com/Automattic/pocket-casts-android/pull/1887)
- [x] Disable for user files (https://github.com/Automattic/pocket-casts-android/pull/1939)

Feedback

- Our chapter titles in the player can be either 1 line or 2 lines which causes the content to jump. We also blink the skipped chapter artwork. (https://github.com/Automattic/pocket-casts-android/pull/1850#pullrequestreview-1884807457)

CookieyedCodes:
Woop looking forward to this :), just a reminder I also would like RSS time tags to be converted to chapters as well under this project 😅, as I still find them more frequently used then RSS chapters & MP3 chapters 🤔

[...]

PR #1850, opened 2024-02-16
Deselect Chapters: Skip to selected chapter

ashiagr:
Part of #1807 

## Description
Handles skipping chapters when they are deselected. 
(It doesn't add select at least one chapter condition yet.)

#### Test.1 Skipping from chapter controls 

1. Play an episode having chapters from https://pca.st/upgrade
2. On the full player, go to "Chapters" tab and deselect some chapters
3. Go back to the "Now Playing Tab"
4. Use the chapter controls to move forward/backwards 
5. ✅ Notice that the chapters you unselected are not played

#### Test.2 Skipping from scrubbing playback position on the seekbar

1. Play an episode having chapters from https://pca.st/upgrade
2. On the full player, go to "Chapters" tab and deselect some chapters
3. Go back to the "Now Playing Tab"
4. Tap on the seek bar at different positions
6. ✅ Notice that the chapters you unselected are not played

#### Test.3 Skipping on chapter end

1. Play an episode having chapters from https://pca.st/upgrade
2. On the full player, go to "Chapters" tab and deselect some chapters
3. Go back to the "Now Playing Tab"
4. Let the current chapter complete
6. ✅ Notice that the next selected chapter is played or playback is seeked to end of the last chapter if there were no selected chapters left

**Important**

In iOS, there is a test case (https://github.com/Automattic/pocket-casts-ios/pull/1426) to check skipping for a podcast (`Clublife`) where the last chapter doesn't end the episode in iOS. 

In Android, we set the end time of the last chapter to episode duration [here](https://github.com/Automattic/pocket-casts-android/blob/main/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt#L1452).  So the behavior is different from iOS where in Android, skipping to the next chapter when there are no unselected chapters after current chapter, seeks to the end of last chapter which is the episode duration.

I'll discuss it with the iOS team about which behavior to keep for consistency and address it in a future PR if needed.

## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/1405144/17c17801-53b4-4708-a9eb-0767b1189e56

## Checklist
 
- with different themes

### DIFF
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/Chapters.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/Chapters.kt
@@ -11,8 +11,8 @@ data class Chapters(private val items: List<Chapter> = emptyList()) {
     private val selectedItems: List<Chapter>
         get() = items.filter { it.selected }
 
-    val lastChapter: Chapter
-        get() = items[items.size - 1]
+    val lastChapter: Chapter?
+        get() = items.getOrNull(items.size - 1)
 
     fun getNextSelectedChapter(timeMs: Int): Chapter? {
         val currentTimeFinal = if (timeMs < 0) 0 else timeMs
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/Chapters.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/Chapters.kt
@@ -8,9 +8,12 @@ data class Chapters(private val items: List<Chapter> = emptyList()) {
     val size: Int
         get() = items.size
 
-    fun getNextChapter(timeMs: Int): Chapter? {
+    private val selectedItems: List<Chapter>
+        get() = items.filter { it.selected }
+
+    fun getNextSelectedChapter(timeMs: Int): Chapter? {
         val currentTimeFinal = if (timeMs < 0) 0 else timeMs
-        for (chapter in items) {
+        for (chapter in selectedItems) {
             if (chapter.startTime > currentTimeFinal) {
                 return chapter
             }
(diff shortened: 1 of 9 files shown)


modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/Chapters.kt
```diff
@@ -8,23 +8,29 @@ data class Chapters(private val items: List<Chapter> = emptyList()) {
     val size: Int
         get() = items.size
 
-    fun getNextChapter(timeMs: Int): Chapter? {
+    private val selectedItems: List<Chapter>
+        get() = items.filter { it.selected }
+
+    val lastChapter: Chapter
+        get() = items[items.size - 1]
```
MiSikora, line 15: Shouldn't this be `Chapter?` with `items.getOrNull()`?

MiSikora (commented):
All tests passed. I left one comment that I think should be fixed before merging.

I noticed one UI bug that isn't related to these changes. Our chapter titles in the player can be either 1 line or 2 lines which causes the content to jump. We also blink the skipped chapter artwork. See video below.

[Screen_recording_20240216_114803.webm](https://github.com/Automattic/pocket-casts-android/assets/30936061/b4bf7e14-ff90-4400-9dad-ca8feb5d1dce)

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/Chapters.kt
ashiagr: Yes! updated in https://github.com/Automattic/pocket-casts-android/pull/1850/commits/f8db0177162546e025f46b67832b3d2ab739e3a0

MiSikora (approved):

ashiagr:
Thanks for the review!

I added this feedback to our main project task:

> Our chapter titles in the player can be either 1 line or 2 lines which causes the content to jump. We also blink the skipped chapter artwork.

PR #1854, opened 2024-02-17
App option to toggle showing podcast artwork embedded in show notes

### LINKED ISSUE #1709: Add setting to toggle episode artwork from podcast feed

In pull request #1599, we began showing episode artwork from the podcast feed via show notes endpoint. This overrides podcast artwork and displays episode artwork if available in the podcast feed without giving an option that allows users to switch back to viewing the standard podcast artwork.

7595144-zd-a8c
https://forums.pocketcasts.com/forums/topic/disable-episode-artwork

Add a setting to toggle between episode artwork from podcast feed and standard podcast artwork 

A similar setting is available for embedded artwork under `Profile` -> `Settings` -> `Appearance` that toggles between embedded artwork in the downloaded episode (if available - e.g. https://pca.st/7hesqiu3) and standard podcast artwork. The new setting can be added below it.

owenlejeune:
## Description
Recently, a change was made to show artwork that is embedded in a podcast episode's show notes by default.  This adds a toggle to appearance settings that functions similarly to the existing toggle for embedded artwork.

Fixes #1709

## Testing Instructions
1. Begin playing an episode that includes artwork embedded in show notes (eg. https://pcast.pocketcasts.net/podcast/ede41160-9eeb-012f-3e7d-525400c11844)
2. Observe that the default podcast artwork is shown in both the miniplayer and fullscreen player.
3. Navigate to Settings -> Appearance and toggle "Use episode artwork"
4. Observe that embedded artwork is show in fullscreen and miniplayer (playback state may need to be toggled for miniplayer to be updated)

## Screenshots or Screencast 
Screenshot 2024-02-17 at 11 22 16
Screenshot 2024-02-17 at 11 22 19
Screenshot 2024-02-17 at 11 22 32
Screenshot 2024-02-17 at 11 22 28

## Checklist
 
- [x] with different themes


owenlejeune:
@MiSikora I'm happy to make any changes to the strings if you have better suggestions as well

CookieyedCodes:
I'd say instead of calling it episode artwork I would call it RSS episode artwork 🤔🤔 and call Embedded artwork MP3 artwork 🤔🤔, my two cents and thus exclude the extra line 😉

owenlejeune:
@CookieyedCodes that's not a bad idea, I agree that it might make sense to change the copy for "embedded" artwork to distinguish between RSS and MP3 artwork

owenlejeune:
Updated the copy to better distinguish between MP3 and RSS artwork
Screenshot 2024-02-18 at 13 01 27

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/MiniPlayer.kt
```diff
@@ -219,12 +220,12 @@ class MiniPlayer @JvmOverloads constructor(context: Context, attrs: AttributeSet
         button.contentDescription = if (isPlaying) stringPause else stringPlay
     }
 
-    private fun loadArtwork(podcast: Podcast?, episode: BaseEpisode) {
+    private fun loadArtwork(podcast: Podcast?, episode: BaseEpisode, useEpisodeArtwork: UserSetting<Boolean>) {
```
MiSikora, line 223: `useEpisodeArtwork` should be `Boolean` instead of `UserSettings<Boolean>`.

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/MiniPlayer.kt
```diff
@@ -285,9 +286,9 @@ class MiniPlayer @JvmOverloads constructor(context: Context, attrs: AttributeSet
     }
 
     companion object {
-        private fun getEpisodeArtwork(episode: BaseEpisode): Artwork {
+        private fun getEpisodeArtwork(episode: BaseEpisode, useEpisodeArtwork: UserSetting<Boolean>): Artwork {
```
MiSikora, line 289: Same here. Use `Boolean` instead of `UserSetting<Boolean>`.

modules/features/settings/src/main/res/layout/fragment_settings_appearance.xml
```diff
@@ -251,6 +251,56 @@
                 app:layout_constraintStart_toStartOf="@+id/lblUseEmbeddedArtwork"
                 app:layout_constraintTop_toBottomOf="@+id/lblUseEmbeddedArtwork" />
 
+            <View
```
MiSikora, line 254: I believe this setting should be above already existing `Use embedded MP3 artwork` because RSS artwork takes precedence over it.

modules/services/localization/src/main/res/values/strings.xml
```diff
@@ -1268,7 +1268,9 @@
     <string name="settings_use_dark_up_next_details">When enabled the Up Next will always use the dark theme, or will match the current theme when disabled</string>
     <string name="settings_use_dynamic_colors_widget">Use dynamic colors for widget</string>
     <string name="settings_use_dynamic_colors_widget_details">When enabled, widget colors will change based on the device theme colors.</string>
-    <string name="settings_use_embedded_podcast_artwork">Use embedded podcast artwork</string>
+    <string name="settings_use_embedded_podcast_artwork">Use embedded MP3 artwork</string>
```
MiSikora, line 1271: I'm not sure if `MP3 artwork` is the best term here. I'd prefer `file artwork` because we support other audio types as well and they might have embedded artworks as well.

modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/SettingsImpl.kt
```diff
@@ -544,6 +544,12 @@ class SettingsImpl @Inject constructor(
         sharedPrefs = sharedPreferences,
     )
 
+    override val useEpisodeArtwork = UserSetting.BoolPref(
+        sharedPrefKey = "useEpisodeArtwork",
+        defaultValue = false,
```
MiSikora, line 549: The default value should be `true`. This is currently enabled for all of the users.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/UpNextQueue.kt
```diff
@@ -40,7 +41,7 @@ interface UpNextQueue {
         object Empty : State()
 
         // Loaded state includes the current episode as episode and the episodes in up next in queue. Queue does not include the currently playing episode
-        data class Loaded(val episode: BaseEpisode, val podcast: Podcast?, val queue: List<BaseEpisode>) : State()
+        data class Loaded(val episode: BaseEpisode, val podcast: Podcast?, val queue: List<BaseEpisode>, val useEpisodeArtwork: UserSetting<Boolean>) : State()
```
MiSikora, line 44: This cannot be a part of the `UpNextQueue` state. They aren't related to each other.

modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/sync/NamedSettings.kt
```diff
@@ -57,6 +57,7 @@ data class ChangedNamedSettings(
     @field:Json(name = "autoShowPlayed") val autoShowPlayed: NamedChangedSettingBool? = null,
     @field:Json(name = "autoPlayLastListUuid") val autoPlayLastSource: NamedChangedSettingString? = null,
     @field:Json(name = "useEmbeddedArtwork") val useEmbeddedArtwork: NamedChangedSettingBool? = null,
+    @field:Json(name = "useEpisodeArtowork") val useEpisodeArtwork: NamedChangedSettingBool? = null,
```
MiSikora, line 60: The setting will be synced at some point but this needs server changes first before we can support on the client side so its better to remove it for now.

modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/sync/SyncSettingsTask.kt
```diff
@@ -436,6 +437,11 @@ class SyncSettingsTask(val context: Context, val parameters: WorkerParameters) :
                         setting = settings.useEmbeddedArtwork,
                         newSettingValue = (changedSettingResponse.value as? Boolean),
                     )
+                    "useEpisodeArtwork" -> updateSettingIfPossible(
```
MiSikora, line 440: Likewise here. This needs server support first.

modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/Settings.kt
```diff
@@ -329,6 +329,7 @@ interface Settings {
     val autoDownloadUpNext: UserSetting<Boolean>
 
     val useEmbeddedArtwork: UserSetting<Boolean>
+    val useEpisodeArtwork: UserSetting<Boolean>
```
MiSikora, line 332: Can you change the name of this property and all others where you reference it as `episodeArtwork` to `rssArtwork`? This will be a bit clearer and be consistent with the setting translation.

MiSikora (changes requested):
Thanks for the contribution! I left some comments. There's also two bugs that I found.

1. Mini player artwork doesn't update with the setting. You can see that when the player is paused toggling the setting doesn't affect the mini player artwork. When the player is playing it works sometimes. I assume it has something to do with the state but I haven't investigated it.

https://github.com/Automattic/pocket-casts-android/assets/30936061/4836e642-2233-487c-ad5b-0db46ab9bde9

2. Notification center always uses RSS artwork.

https://github.com/Automattic/pocket-casts-android/assets/30936061/1094b934-21af-460b-acd2-918432a46b69

owenlejeune:
@MiSikora I address the comments you made. One note is that the notification artwork doesn't seem to be reflected immediately if a podcast is playing, but is respected if the podcast is paused. If a podcast is playing when this setting is changed, the change will be reflected when the user changes the playback state. I'm not sure why it's not reflected immediately. I suspect this is not a setting that users will be changing frequently so I'm not sure if that's a blocker or not.

https://github.com/Automattic/pocket-casts-android/assets/20568035/3c157bab-0fa4-4b83-81b7-b9813f0d7922

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/MediaSessionManager.kt
```diff
@@ -171,6 +176,15 @@ class MediaSessionManager(
             .doOnNext { updateUpNext(it) }
             .subscribeBy(onError = { Timber.e(it) })
             .addTo(disposables)
+        settings.useRssArtwork.flow.asObservable(coroutineContext)
```
MiSikora, line 179: We don't use RxJava for new code. This should be implemented using `Flow` from coroutines. And since my previous suggestion mentions combining this stream with `upNextQueue` stream you'll need to convert `upNextQueue` to `Flow` as well.

modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/sync/SyncSettingsTask.kt
```diff
@@ -436,6 +436,11 @@ class SyncSettingsTask(val context: Context, val parameters: WorkerParameters) :
                         setting = settings.useEmbeddedArtwork,
                         newSettingValue = (changedSettingResponse.value as? Boolean),
                     )
+                    "useRssArtwork" -> updateSettingIfPossible(
+                        changedSettingResponse = changedSettingResponse,
+                        setting = settings.useRssArtwork,
+                        newSettingValue = (changedSettingResponse.value as? Boolean),
+                    )
```
MiSikora, line 443: This should be removed since we don't have syncing on the server yet.

MiSikora:
> I address the comments you made. One note is that the notification artwork doesn't seem to be reflected immediately if a podcast is playing, but is respected if the podcast is paused. If a podcast is playing when this setting is changed, the change will be reflected when the user changes the playback state. I'm not sure why it's not reflected immediately. I suspect this is not a setting that users will be changing frequently so I'm not sure if that's a blocker or not.

Yeah, this is perfectly fine. 👍

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/MediaSessionManager.kt
owenlejeune: Any thoughts on how to convert `upNextQueue` to a `Flow`? I implemented a new method in `UpNextQueue`
```kotlin
fun getChangesFlowWithLiveCurrentEpisode(episodeManager: EpisodeManager, podcastManager: PodcastManager): Flow<State> {
    return changesObservable.asFlow().debounce(100).flatMapLatest { state ->
        if (state is State.Loaded) {
            if (state.podcast != null) {
                episodeManager.observeEpisodeByUuid(state.episode.uuid)
                    .combine<BaseEpisode, Podcast, State>(podcastManager.observePodcastByUuidFlow(state.podcast.uuid).distinctUntilChanged { t1, t2 -> t1.isUsingEffects == t2.isUsingEffects }) { episode, podcast ->
                        State.Loaded(episode, podcast, state.queue)
                    }
                    .catch { emit(State.Empty) }
            } else {
                episodeManager.observeEpisodeByUuid(state.episode.uuid)
                    .map<BaseEpisode, State> { State.Loaded(it, state.podcast, state.queue) }
                    .catch { emit(State.Empty) }
            }
        } else {
            flowOf(state)
        }
    }
}
```
and modified `MediaSessionManager#startObserving()` to include
```kotlin
playbackManager.upNextQueue.getChangesFlowWithLiveCurrentEpisode(episodeManager, podcastManager)
      .distinctUntilChanged { stateOne, stateTwo ->
          UpNextQueue.State.isEqualWithEpisodeCompare(stateOne, stateTwo) { episodeOne, episodeTwo ->
              episodeOne.uuid == episodeTwo.uuid && episodeOne.duration == episodeTwo.duration
          }
      }
      .combine(settings.useRssArtwork.flow) { state, useRssArtwork ->
          Pair(state, useRssArtwork)
      }
      .onEach {
          updateUpNext(it.first, it.second)
      }
      .catch { Timber.e(it) }
```
However, this causes the media notification to never appear. I implemented this same method and flow combining in `MainActivity#setupPlayerViews` in place of the `upNextQueueObservable` block but this causes the mini player to never be shown.

I did have to add new methods `observePodcastByUuidFlow` to `PodcastManager` and `observeByUuidFlow` to `PodcastDao`, but these methods are nearly identical to the existing `observePodcastByUuid` and `observeByUuid` methods respectively, but returning a `Flow` instead of a `Flowable`.

MiSikora: I can't tell why a solution doesn't work without seeing it. The simplest thing to do is to convert final result of `upNextQueue` in `MediaSessionManager` with `asFlow()`. This way we can incrementally change each component to a `Flow` and not necessarily in this PR.

owenlejeune: Ah! I was missing a call to `.launchIn(this)`. All seems to be working as expected now, thanks.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/UpNextQueue.kt
```diff
@@ -115,3 +158,20 @@ fun Observable<UpNextQueue.State>.containsUuid(uuid: String): Observable<Boolean
         }
     }
 }
+
+fun <T, R> Flow<T>.switchMap(mapper: (value: T) -> Flow<R>): Flow<R> {
```
MiSikora, line 162: I don't think you need to introduce a new operator. `flatMapLatest()` acts effectively the same. And if you need to introduce a new operator it should be in a different module with tests.

MiSikora (changes requested):
I won't be able to test it until Monday but code-wise it looks good.

MiSikora:
The miniplayer is now broken.

Miniplayer artwork doesn't toggle.

https://github.com/Automattic/pocket-casts-android/assets/30936061/3caf6c6e-ea93-41cf-9030-bf875ff80d6b

Miniplayer doesn't switch artwork at all. Even when playing a different podcast.

https://github.com/Automattic/pocket-casts-android/assets/30936061/cc597c9e-15a6-4d17-9cc9-2d655592c3b8

Miniplayer doesn't close.

https://github.com/Automattic/pocket-casts-android/assets/30936061/47f8da59-a8b0-4006-b359-d8b54851d453

Another issue is with chapter artwork. I'm not 100% if this is should happen but I personally would expect to this toggle disabling also chapter artwork but it doesn't.

https://github.com/Automattic/pocket-casts-android/assets/30936061/14adb7b5-f282-44be-83c2-0608d4b03df8

CookieyedCodes:
From my understanding the toggle only applys to the next podcast so it makes sence the artwork cache is updated for the miniplayer when it changes episodes 🤔

Miniplayer should close tho hmmmm

Well chapter artwork is embedded in the audio file so while it probably should disable it would be a completely different toggle in my opinion

MiSikora:
> From my understanding the toggle only applys to the next podcast so it makes sence the artwork cache is updated for the miniplayer when it changes episodes 🤔

Sorry, I don't understand your point. Are you agreeing with me that it is an issue that the miniplayer doesn't update the artwork when a different episode/podcast is played?

> Well chapter artwork is embedded in the audio file so while it probably should disable it would be a completely different toggle in my opinion

If that's the case we should leave this behavior as is. 👍

CookieyedCodes:
Regarding the confusion, from my understanding the miniplayer artwork caches when the episode loaded not during the episode (same with the player from my understanding, which is probably harder to fix tbh) just saying why it may be happening 🤔 🤔

owenlejeune:
@MiSikora that's really odd, in my testing I'm pretty sure I was seeing the mini player artwork being updated with that toggle as expected. I'll take another look sometime today

owenlejeune:
@MiSikora I was able to repro your issues. Looks like this was a Coroutine Context issue. Flows in MainActivity were launched using Dispatchers.Default while the mini player uses Dispatchers.Main. It looks like this combined flow is the only member of MainActivity using the coroutine context so I changed it to Dispatchers.Main and I'm not longer seeing any issues with the miniplayer.

app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
```diff
@@ -752,22 +753,23 @@ class MainActivity :
...
+            .onEach { (upNextQueue, useRssArtwork) ->
+                binding.playerBottomSheet.setUpNext(
+                    upNext = upNextQueue,
+                    theme = theme,
+                    shouldAnimateOnAttach = !showMiniPlayerImmediately,
+                    useRssArtwork = useRssArtwork
+                )
+            }
+            .catch { Timber.e(it) }
+            .launchIn(this)
```
MiSikora, line 772: Oh, I completely missed that part in the review. We shouldn't be using implicit `CoroutineScope` in the activities. In this particular instance it would cause memory leaks.

Instead please use `repeatOnLifecycle` API. It would look something like this. 

```kotlin
lifecycleScope.launch {
  repeatOnLifecycle(Lifecycle.State.STARTED) {
    <your_flow>.collect()
  }
}
```

app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
```diff
@@ -752,22 +753,23 @@ class MainActivity :
...
-            binding.playerBottomSheet.setUpNext(
-                upNext = upNext,
-                theme = theme,
-                shouldAnimateOnAttach = !showMiniPlayerImmediately,
-            )
+        val upNextQueueChanges = playbackManager.upNextQueue.getChangesFlowWithLiveCurrentEpisode(episodeManager, podcastManager)
+        val useRssArtworkChanges = settings.useRssArtwork.flow
+
+        combine(upNextQueueChanges, useRssArtworkChanges) { upNextQueue, useRssArtwork ->
+            Timber.i("useRssArtwork combined, $useRssArtwork")
```
MiSikora, line 760: A debug log.

app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
```diff
@@ -235,7 +236,7 @@ class MainActivity :
     private val bottomSheetQueue: MutableList<(() -> Unit)?> = mutableListOf()
 
     override val coroutineContext: CoroutineContext
-        get() = Dispatchers.Default
+        get() = Dispatchers.Main
```
MiSikora, line 239: Don't change this `coroutineContext`. It will have impact on other coroutines launched in the `MainActivity` which might be doing some heavy operations.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/UpNextQueue.kt
```diff
@@ -10,6 +10,18 @@ import io.reactivex.Completable
...
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.catch
+import kotlinx.coroutines.flow.combine
+import kotlinx.coroutines.flow.debounce
+import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.flatMapLatest
+import kotlinx.coroutines.flow.flowOf
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.rx2.asFlow
+import timber.log.Timber
```
MiSikora, line 24: Unnecessary import. Please run `./gradlew spotlessApply`.

MiSikora (changes requested):

MiSikora (approved):
I am not able to reproduce the mentioned issue.

MiSikora (approved):
LGTM! Thanks for this PR.

You need to sync with the `main` branch and move your changelog entry to the `7.59` release.

owenlejeune:
@MiSikora sync and CHANGELOG update are complete

CHANGELOG.md
```diff
@@ -27,6 +29,7 @@
         ([#1846](https://github.com/Automattic/pocket-casts-android/pull/1846))
     *   Fixed a bug where podcast setting screen could become unresponsive.
         ([#1845](https://github.com/Automattic/pocket-casts-android/pull/1845))
+*   Updates:
```
MiSikora, line 32:
```suggestion
```

PR #1856, opened 2024-02-19
Deselect Chapters - Persist deselected chapter indexes

ashiagr:
Part of: https://github.com/Automattic/pocket-casts-android/issues/1807
## Description

This saves deselected chapter indexes into the database similar to iOS (https://github.com/Automattic/pocket-casts-ios/pull/1431).

## Testing Instructions

These are podcasts with chapters you can use to test:

* Clublife: https://pca.st/rLzZCv
* Upgrade: https://pca.st/upgrade

#### Test.1 Migration

1. Checkout commit c92bfdaa
2. Open the app
3. Checkout code from this branch
4.  Open the app
5.  ✅ App should not crash

#### Test.2 Chapter selection

1. Launch the app
2. Play any episode with chapters
3. Deselect some chapters
4. Force close and reopen the app
5. ✅ Your chapter list should reflect your previous selection
6. Toggle between episode with chapters and change their selections
7. ✅ Your chapter selection should be correctly reflected

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/BaseEpisode.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/BaseEpisode.kt
@@ -41,7 +41,7 @@ sealed interface BaseEpisode {
     var playedUpToModified: Long?
     var playErrorDetails: String?
     var downloadErrorDetails: String?
-    var deselectedChapters: ChapterIndices?
+    var deselectedChapters: ChapterIndices
 
     // temporary variables
     var playing: Boolean
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/BaseEpisode.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/BaseEpisode.kt
@@ -41,7 +41,7 @@ sealed interface BaseEpisode {
     var playedUpToModified: Long?
     var playErrorDetails: String?
     var downloadErrorDetails: String?
-    var deselectedChapters: String?
+    var deselectedChapters: ChapterIndices?
 
     // temporary variables
     var playing: Boolean
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/BaseEpisode.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/BaseEpisode.kt
@@ -41,6 +41,7 @@ sealed interface BaseEpisode {
     var playedUpToModified: Long?
     var playErrorDetails: String?
     var downloadErrorDetails: String?
+    var deselectedChapters: String?
 
     // temporary variables
     var playing: Boolean
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/ChapterIndices.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/ChapterIndices.kt
@@ -3,7 +3,7 @@ package au.com.shiftyjelly.pocketcasts.models.entity
 data class ChapterIndices(private val delegate: List<Int> = emptyList()) : List<Int> by delegate {
     companion object {
         fun fromString(value: String?): ChapterIndices {
-            val list = value?.split(",")?.map { it.toInt() } ?: emptyList()
+            val list = value?.split(",")?.mapNotNull { it.toIntOrNull() } ?: emptyList()
             return ChapterIndices(list)
         }
 
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/ChapterIndices.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/ChapterIndices.kt
@@ -1,0 +1,12 @@
+package au.com.shiftyjelly.pocketcasts.models.entity
+
+class ChapterIndices(delegate: List<Int> = emptyList()) : List<Int> by delegate {
+    companion object {
+        fun fromString(value: String?): ChapterIndices {
+            val list = value?.split(",")?.map { it.toInt() } ?: emptyList()
+            return ChapterIndices(list)
+        }
+
+        fun toString(indices: ChapterIndices) = indices.takeIf { it.isNotEmpty() }?.joinToString(",")
+    }
+}
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/ChapterIndices.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/ChapterIndices.kt
@@ -1,6 +1,6 @@
 package au.com.shiftyjelly.pocketcasts.models.entity
 
-class ChapterIndices(delegate: List<Int> = emptyList()) : List<Int> by delegate {
+data class ChapterIndices(private val delegate: List<Int> = emptyList()) : List<Int> by delegate {
     companion object {
         fun fromString(value: String?): ChapterIndices {
             val list = value?.split(",")?.map { it.toInt() } ?: emptyList()
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/ChapterIndices.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/ChapterIndices.kt
@@ -1,12 +1,3 @@
 package au.com.shiftyjelly.pocketcasts.models.entity
 
-data class ChapterIndices(private val delegate: List<Int> = emptyList()) : List<Int> by delegate {
-    companion object {
-        fun fromString(value: String?): ChapterIndices {
-            val list = value?.split(",")?.mapNotNull { it.toIntOrNull() } ?: emptyList()
-            return ChapterIndices(list)
-        }
-
-        fun toString(indices: ChapterIndices) = indices.takeIf { it.isNotEmpty() }?.joinToString(",")
-    }
-}
+data class ChapterIndices(private val delegate: List<Int> = emptyList()) : List<Int> by delegate
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/PodcastEpisode.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/PodcastEpisode.kt
@@ -59,6 +59,7 @@ data class PodcastEpisode(
     @ColumnInfo(name = "download_task_id") override var downloadTaskId: String? = null,
     @ColumnInfo(name = "last_archive_interaction_date") var lastArchiveInteraction: Long? = null,
     @ColumnInfo(name = "image_url") var imageUrl: String? = null,
+    @ColumnInfo(name = "deselected_chapters") override var deselectedChapters: String? = null,
 ) : BaseEpisode, Serializable {
 
     sealed class EpisodeType {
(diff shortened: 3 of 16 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/BaseEpisode.kt
```diff
@@ -41,6 +41,7 @@ sealed interface BaseEpisode {
     var playedUpToModified: Long?
     var playErrorDetails: String?
     var downloadErrorDetails: String?
+    var deselectedChapters: String?
```
MiSikora, line 44: Using `String` for keeping deselected chapter indices is a bit troublesome. Now, whenever we want to access this we have to know internal mapping using the comma, and we have to keep in check data integrity because we can put whatever in a `String`.

I'd like to propose a simple class like this one.

```kotlin
class ChapterIndices(delegate: List<Int>) : List<Int> by delegate
```

This way we can write a single mapper for `Room` to understand how to store this data and outside of it we can use it as a list and have access to `Int` representing indices.

MiSikora (changes requested):
Feature works as expected but I left some comments that I believe could improve the code quality.

I have also one more question. Since we're going to release syncing soon do you know if this is planned to be synced as well?

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/BaseEpisode.kt
Mukuljangir372, line 44: I agree with @MiSikora for this approach. I would say, We should further scale it to more simpler approach. In @MiSikora approach, It will be hard to remove or adding elements directly using SQL query and will lead to decrease in performance. 

Instead, We should create a separate table having for tracking the selection or deselection under episode_id, chapter_id keys in table as composition primary key as Room supports this.

MiSikora: > It will be hard to remove or adding elements directly using SQL query

This part is true. Without database normalization we won't be able to access things directly through SQL statements. I don't know if it is something that we need for deselected chapters. Chapters themselves are already baked into file metadata and aren't stored in our database. It's hard for me to imagine a scenario where we need to use direct queries only on chapters and not on episodes.

> and will lead to decrease in performance.

This is hard to tell without benchmarks. Pocket Casts conflated storage and domain models. Having a separate table requires more queries to execute when we want to read episode metadata. Same thing goes for updates.

---

Having said that I'm fine with this solution as well if you @ashiagr prefer to implement it. But it is not without its cons.

CookieyedCodes: Theirs a feature that I'm requesting for this project that RSS time tags will be converted to chapters, mainly because I see them more then embedded chapters. Hopefully it can be added easily enough 🤔🤔

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/UserEpisode.kt
```diff
@@ -46,6 +46,7 @@ data class UserEpisode(
     @ColumnInfo(name = "tint_color_index") var tintColorIndex: Int = 0,
     @ColumnInfo(name = "has_custom_image") var hasCustomImage: Boolean = false,
     @ColumnInfo(name = "upload_task_id") var uploadTaskId: String? = null,
+    @ColumnInfo(name = "deselected_chapters") override var deselectedChapters: String? = null,
```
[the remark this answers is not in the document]
ashiagr: That's correct. It should be `@ColumnInfo`.

I also tested that user episode can have chapters:

https://github.com/Automattic/pocket-casts-android/assets/1405144/89849de9-9ab3-40ec-b493-6de26b632fc4

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/BaseEpisode.kt
ashiagr: I converted the return type to `ChapterIndices` in https://github.com/Automattic/pocket-casts-android/pull/1856/commits/a2a8f1b5b8608dce4666e525436f45007a460bdd and https://github.com/Automattic/pocket-casts-android/pull/1856/commits/4b41ae5872d29caddb45ea84f38a3f4cce481f2b. It now looks more type-safe 👍

ashiagr: I went ahead with a simple conversion from `String` to `ChapterIndices` for current purposes. It wouldn't be hard to update the structure when the need comes.

Thanks for all the comments. I appreciate it. 🙂

ashiagr:
There were two dangermattic errors:  https://github.com/Automattic/pocket-casts-android/pull/1856#issuecomment-1952103197 about PR size and `ChapterIndicesConverter` tests. 

- PR size is large because it includes a large file, a generated JSON `86.json`.
- `ChapterIndicesConverter` reuses conversions from `ChapterIndices` which is tested in https://github.com/Automattic/pocket-casts-android/pull/1856/commits/0ae31ce5bb7010e29988c67ae96133b1577041fe.

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/PodcastEpisode.kt
```diff
@@ -59,6 +59,7 @@ data class PodcastEpisode(
     @ColumnInfo(name = "download_task_id") override var downloadTaskId: String? = null,
     @ColumnInfo(name = "last_archive_interaction_date") var lastArchiveInteraction: Long? = null,
     @ColumnInfo(name = "image_url") var imageUrl: String? = null,
+    @ColumnInfo(name = "deselected_chapters") override var deselectedChapters: ChapterIndices? = ChapterIndices(),
```
MiSikora, line 62: If we use `ChapterIndicies()` as a default value I think we can make the field not nullable and use an empty string as a default database entry.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/EpisodeManagerImpl.kt
```diff
@@ -1098,4 +1099,18 @@ class EpisodeManagerImpl @Inject constructor(
 
             return@withContext newDownloadUrl ?: episode.downloadUrl
         }
+
+    override suspend fun selectChapterIndexForEpisode(chapterIndex: Int, episode: PodcastEpisode) {
+        val deselectedChapterIndices = episode.deselectedChapters ?: emptyList()
+        if (!deselectedChapterIndices.contains(chapterIndex)) return
+        episode.deselectedChapters = ChapterIndices(deselectedChapterIndices.toMutableList().apply { remove(chapterIndex) })
```
MiSikora, line 1106: Similarly to what you did in `deselectChapterIndexForEpisode` you can use minus operator here.

```suggestion
        episode.deselectedChapters = ChapterIndices(deselectedChapterIndices - chapterIndex)
```

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/UserEpisodeManager.kt
```diff
@@ -611,6 +615,20 @@ class UserEpisodeManagerImpl @Inject constructor(
     override suspend fun markAllAsUnplayed(episodes: List<UserEpisode>) {
         episodes.map { it.uuid }.chunked(500).forEach { userEpisodeDao.markAllUnplayed(it, System.currentTimeMillis()) }
     }
+
+    override suspend fun selectChapterIndexForEpisode(chapterIndex: Int, episode: UserEpisode) {
+        val deselectedChapterIndices = episode.deselectedChapters ?: emptyList()
+        if (!deselectedChapterIndices.contains(chapterIndex)) return
+        episode.deselectedChapters = ChapterIndices(deselectedChapterIndices.toMutableList().apply { remove(chapterIndex) })
```
MiSikora, line 622: Minus operator can be used here as well.

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/ChapterIndices.kt
```diff
@@ -0,0 +1,12 @@
+package au.com.shiftyjelly.pocketcasts.models.entity
+
+data class ChapterIndices(private val delegate: List<Int> = emptyList()) : List<Int> by delegate {
+    companion object {
+        fun fromString(value: String?): ChapterIndices {
+            val list = value?.split(",")?.map { it.toInt() } ?: emptyList()
```
MiSikora, line 6: We could consider mapping `toIntOrNull()` for safety.

```suggestion
            val list = value?.split(",")?.mapNotNull { it.toIntOrNull() } ?: emptyList()
```

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
```diff
@@ -938,6 +938,29 @@ open class PlaybackManager @Inject constructor(
 
     fun toggleChapter(select: Boolean, chapter: Chapter) {
         launch {
+            getCurrentEpisode()?.let { episode ->
+                when (episode) {
+                    is PodcastEpisode -> {
+                        episodeManager.findByUuid(episode.uuid)?.let {
```
MiSikora, line 944: If we have an `episode` from `getCurrentEpisode()` why do we need to fetch it again with `find()`? Are there any caching issues?

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/ChapterIndices.kt
```diff
@@ -0,0 +1,12 @@
+package au.com.shiftyjelly.pocketcasts.models.entity
+
+data class ChapterIndices(private val delegate: List<Int> = emptyList()) : List<Int> by delegate {
+    companion object {
```
MiSikora, line 4: I don't have a strong opinion on this but I wonder if we shouldn't move this logic from the companion object directly to the database converter. I don't think we should expect chapters to be strings anywhere outside of the context of the database. But like I said, I don't have strong feelings about it.

MiSikora (approved):
LGTM! I left some comments but they don't block the feature.

I want to follow up on syncing though. Since we're going to release syncing soon do you know if chapter selection is planned to be synced as well?

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/ChapterIndices.kt
ashiagr: Np, I updated it in https://github.com/Automattic/pocket-casts-android/pull/1856/commits/ddfe1bd879af3da85d958b41ccd3a580557a0ee2

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
ashiagr: I was trying a different approach earlier where this was needed. I removed it in https://github.com/Automattic/pocket-casts-android/pull/1856/commits/6138d59cc877e7f10596ef606a886910131585ae it doesn't seem to be needed now.

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/PodcastEpisode.kt
ashiagr: This worked as well in https://github.com/Automattic/pocket-casts-android/pull/1856/commits/e87c4de8f81eaf7cada328851774b83e02390725

ashiagr:
> I want to follow up on syncing though. Since we're going to release syncing soon do you know if chapter selection is planned to be synced as well?

I think it will be needed but the feature is still in development so we can wait on this util we have more clarity.

PR #1868, opened 2024-02-22
Sync background refresh setting

MiSikora:
## Description

Add syncing of the global setting to refresh podcasts in the background.

## Testing Instructions

> [!note]
> This should be tested with staging. Use `debug` variant instead of `debugProd`.

3. D1: Go to `Profile`/`Settings (cog icon)`/`Storage & data use`.
4. D1: Toggle `Background refresh` setting `OFF`.
5. D1: Sync the device in the Profile.
6. D2: Sync the device in the Profile.
7. D2: Go to `Profile`/`Settings (cog icon)`/`Storage & data use`.
8. D2: `Background refresh` setting should be `OFF`.
9. D2: Toggle `Background refresh` setting `ON`.
10. D2: Sync the device in the Profile.
11. D1: Sync the device in the Profile.
12. D1: Go to `Profile`/`Settings (cog icon)`/`Storage & data use`.
13. D1: `Background refresh` setting should be `ON`.

You can test this with Wear as well but it might be hard setting up the staging environment. The change in `wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/settings/SettingsViewModel.kt` should be enough to confirm that it works.

## Checklist
 
- [ ] with different themes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 * Bug Fixes:
     *   Fixed an issue where bookmarks did not play when episodes where filtered out due to search queries.
         ([#1857](https://github.com/Automattic/pocket-casts-android/pull/1857))    
+    *   Fixed an issue where bookmarks on description could become unresponsive.
+        ([#1873](https://github.com/Automattic/pocket-casts-android/pull/1873))    
 
 7.58
 -----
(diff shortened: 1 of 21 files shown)


mebarbosa (approved):
LGTM

PR #1874, opened 2024-02-22
Deselect Chapters -  Add skip chapter header

ashiagr:
Part of #1807 

## Description
Toggle the chapter selection/deselection.

(This is not the final design)

These are podcasts with chapters you can use to test:

* Upgrade: https://pca.st/upgrade
* Clublife: https://pca.st/rLzZCv

#### Feature Flag ON

2. Run the app
4. Play any episode with chapters
5. Go to the `Chapters` tab
6. ✅ You should see a header indicating the number of chapters and a button to "Skip chapters"
7. Tap "Skip chapters"
8. ✅ You can now select/deselect chapters
9. Tap "Done"
10. ✅ The number of chapters and hidden chapters should be correctly updated and hidden chapters don't appear on the listing
11. Leave some chapters deselected

#### Feature Flag OFF
13. Go to `Profile` > `Settings` > Beta Features > `Deselect Chapters` and turn it off
14. Kill the app and restart
15. Open `Chapters` tab again
16. ✅ All chapters should be there and there shouldn't be a header

** I also made changes to not skip chapters if feature flag is off in https://github.com/Automattic/pocket-casts-android/pull/1874/commits/4d10992fd8b6d061c65989ede096b8ef9407ec8. I've tried to cover them in tests, it would be good to test it as well.

## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/1405144/f7714748-4089-4cdc-aa86-410ce0c6863c

Large font

## Checklist
 
- [x] with different themes

### DIFF
--- a/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/chapters/ChaptersPage.kt
+++ b/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/chapters/ChaptersPage.kt
@@ -37,7 +37,7 @@ fun ChaptersPage(
         modifier = modifier
             .background(backgroundColor)
             .fillMaxSize()
-            .padding(top = 16.dp),
+            .padding(top = if (FeatureFlag.isEnabled(Feature.DESELECT_CHAPTERS)) 0.dp else 16.dp),
     ) {
         if (FeatureFlag.isEnabled(Feature.DESELECT_CHAPTERS)) {
             item {
--- a/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/chapters/ChaptersPage.kt
+++ b/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/chapters/ChaptersPage.kt
@@ -1,5 +1,6 @@
 package au.com.shiftyjelly.pocketcasts.player.view.chapters
 
+import androidx.compose.foundation.ExperimentalFoundationApi
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.padding
@@ -14,14 +15,20 @@ import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.unit.dp
 import au.com.shiftyjelly.pocketcasts.compose.theme
 import au.com.shiftyjelly.pocketcasts.models.to.Chapter
+import au.com.shiftyjelly.pocketcasts.utils.featureflag.Feature
+import au.com.shiftyjelly.pocketcasts.utils.featureflag.FeatureFlag
 
+@OptIn(ExperimentalFoundationApi::class)
 @Composable
 fun ChaptersPage(
     lazyListState: LazyListState,
     chapters: List<ChaptersViewModel.ChapterState>,
+    totalChaptersCount: Int,
     onSelectionChange: (Boolean, Chapter) -> Unit,
     onChapterClick: (Chapter, Boolean) -> Unit,
     onUrlClick: (String) -> Unit,
+    onSkipChaptersClick: (Boolean) -> Unit,
+    isTogglingChapters: Boolean,
     backgroundColor: Color,
     modifier: Modifier = Modifier,
 ) {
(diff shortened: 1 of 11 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please add tests for class <code>TextToggle</code> (or add <code>unit-tests-exemption</code> label to ignore this).</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/chapters/ChaptersPage.kt
```diff
@@ -30,14 +37,26 @@ fun ChaptersPage(
...
+            }
+        }
         itemsIndexed(chapters, key = { _, state -> state.chapter.index }) { index, state ->
             ChapterRow(
                 state = state,
+                isTogglingChapters = isTogglingChapters,
                 onSelectionChange = onSelectionChange,
                 onClick = { onChapterClick(state.chapter, state is ChaptersViewModel.ChapterState.Playing) },
                 onUrlClick = { onUrlClick(state.chapter.url.toString()) },
+                modifier = Modifier.animateItemPlacement(),
```
ashiagr, line 59: `animateItemPlacement()` is an experimental modifier property. I've tested it on API 24 emulator.

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/Chapters.kt
```diff
@@ -1,6 +1,12 @@
 package au.com.shiftyjelly.pocketcasts.models.to
 
-data class Chapters(private val items: List<Chapter> = emptyList()) {
+import au.com.shiftyjelly.pocketcasts.utils.featureflag.Feature
+import au.com.shiftyjelly.pocketcasts.utils.featureflag.FeatureFlagWrapper
+
+data class Chapters(
+    private val items: List<Chapter> = emptyList(),
+    private val featureFlagWrapper: FeatureFlagWrapper = FeatureFlagWrapper(),
```
MiSikora, line 8: It feels weird to have a `data class` contain something like a `FeatureFlagWrapper`. I'm not sure why we need this class. Feature flag values can be mocked in tests with providers like here.

https://github.com/Automattic/pocket-casts-android/blob/c502a6267b58b305cd708d955e5f977976ae9986/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/repositories/sync/PodcastSyncProcessTest.kt#L65-L73

I might not have enough context but if this is something that we do often we could just create a provider like this and reuse it in tests.

```kotlin
class InMemoryFeatureProvider : ModifiableFeatureProvider {
    private val features = mutableMapOf<Feature, Boolean>()

    override val priority = 0

    override fun setEnabled(feature: Feature, enabled: Boolean) {
        features[feature] = enabled
    }

    override fun hasFeature(feature: Feature) = true

    override fun isEnabled(feature: Feature) = features[feature] ?: feature.defaultValue
}
```

Possibly even as a rule that initializes and clears providers so we don't need to register it every time.

MiSikora (approved):
LGTM. I left one comment related to the feature flags design but I think it's ok to merge if you don't agree.

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/Chapters.kt
ashiagr: Love the idea! I applied it at https://github.com/Automattic/pocket-casts-android/pull/1874/commits/eac5ca1e8f37370859cc6aa058de7a02435ab15a.

I'll check about creating a rule and address it in another PR.

PR #1875, opened 2024-02-23
Add navigation to podcast from the player screen

MiSikora:
## Description

This PR adds navigation from the `Now Playing` tab to a podcast directly.

Closes #1644 

## Testing Instructions

1. Play an episode without any chapters.
2. Open the player.
3. Tap on the podcast title.
5. The app should navigate to the podcast.

## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/30936061/7519db18-77b8-44b3-8322-e540f6c3d140

## Checklist

- [x] with different themes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 7.59
 -----
+*   Updates:
+    *   Navigate to a podcast when a podcast title in the player tapped.
+        ([#1875](https://github.com/Automattic/pocket-casts-android/pull/1875))
 * Bug Fixes:
     *   Fixed an issue where bookmarks did not play when episodes where filtered out due to search queries.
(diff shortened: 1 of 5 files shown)


david-gonzalez-a8c:
Thanks for your work on this @MiSikora 😍 

I'd like us to change the color in the podcast title so it's grey like before and remove the chevron. We [discussed](https://github.com/Automattic/pocket-casts-ios/pull/1333#issuecomment-1884790656) a bit when this was implemented on iOS and I believe this interaction is standard enough that we don't need the extra affordance of the chevron and the color, keeping the view a bit cleaner (plus having parity with iOS).

Thank you!

mebarbosa (approved):
It worked with all themes 👍🏽

PR #1881, opened 2024-02-26
Sync podcast settings when signing in

MiSikora:
## Description

The sync process did not account for signing in where we sync podcasts with the `/user/podcast/list` endpoint and not the `user/sync/update` endpoint. This resulted in a bug where most of the podcast settings could not be synced until changed again.

### New API

1. Be subscribed to a podcast and change some of the new settings.
    - Enable podcast notifications.
    - Skip first.
    - Playback effects.
    - Auto archive.
2. Sync the device in the Profile.
3. Install the app on a second device.
4. Sign in on the second device with the same account.
5. Podcasts settings should be synced.

### Old API

1. Disable `Settings Sync` flag in the the `Beta features`.
2. Be subscribed to a podcast and change some of the old settings.
    - Skip first.
    - Skip last.
3. Install the app on a second device.
4. Disable `Settings Sync` flag in the the `Beta features`.
5. Sign in on the second device with the same account.
6. Podcasts settings should be synced.

## Checklist
 
- [ ] with different themes


modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/sync/SyncServerManager.kt
```diff
@@ -154,19 +160,8 @@ open class SyncServerManager @Inject constructor(
...
 
-    fun getHomeFolder(token: AccessToken): Single<PodcastListResponse> =
-        server.getPodcastList(addBearer(token), buildBasicRequest()).map { response ->
-            response.copy(podcasts = removeHomeFolderUuid(response.podcasts), folders = response.folders)
-        }
-
-    private fun removeHomeFolderUuid(podcasts: List<PodcastResponse>?): List<PodcastResponse>? =
-        podcasts?.map { podcast ->
-            if (podcast.folderUuid != null && podcast.folderUuid == Folder.homeFolderUuid) {
-                podcast.copy(folderUuid = null)
```
MiSikora, line 165: Because Google's protoc implementation doesn't allow to set `null` value here I had to move it to `PodcastSyncProcess` when applying properties to podcasts.

ashiagr (approved):
LGTM

PR #1878, opened 2024-02-25
Configure app to cache episodes when playing

mebarbosa:
## Description
- This is part of the Playback Skipping project: Prevent the playback position from skipping while listening to a streaming episode by fully downloading it
- The goal of this PR is to prevent the unexpected playback skipping that our users reported. According to some investigations we did in the past, we believe this is related to ads insertion so our approach to fix this is to implement a cache system by triggering a download when the user begins streaming an episode switching from the streaming episode to the downloaded episode once the download completes.
- Project thread internal reference: `pdeCcb-4ux-p2`
- Playback skipping section from: `pdeCcb-4gS-p2`
- Slack thread: `p1708533677908569-slack-C05RR9P9RAT`
- #1865

> [!note]
> While you are testing, you might encounter the existing issue https://github.com/Automattic/pocket-casts-android/issues/86 related to the downloading percentage. I won't address it here as it's not within the scope of this task

> [!note]
> To test this PR I will to have the local `CACHE_PLAYING_EPISODE` feature flag enabled.

### Notes
- Have the feature flag enabled to be able to test it
- To see if the playback if switching from streaming to downloaded file I put a log to print the value of `episodeLocation`. I will noticed that when you tap to play an episode it will print the URL of the episode first and then the path to the downloaded episode:

https://github.com/Automattic/pocket-casts-android/blob/c24f8591be2c747d3e3acc04712de04a9f025a2e/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/LocalPlayer.kt#L184

### Play Episode
1. Play an episode.
2. Wait for it to finish downloading.
3. Ensure the app switches from streaming to downloading the file.
4. Go to Profile -> Downloads.
5. Confirm that the podcast is listed in the download section.
6. Play another episode.
7. The first episode should be deleted, and the current one playing starts downloading

### Queue
1. Play an episode.
2. Wait for it to finish downloading.
3. Ensure the app switches from streaming to downloading the file.
4. Go to Profile -> Downloads.
5. Confirm that the podcast is listed in the download section.
6. Add some episodes to the queue.
7. Seek the currently playing episode to the end and wait for the next one from the queue to start playing.
8. When the next episode starts playing, the previous episode should be removed from the downloads screen (cache), and the current episode should start downloading.
9. Add more episodes to the Up Next queue.
10. Go to Profile -> Settings -> General -> Toggle "Play up next episode on tap" ON.
11. Go to the Up Next queue.
[...]

Issue #86, opened 2022-07-11
% of downloading reset to 0% and back to correct value after switching from one layout to another


mchowning:
This can also happen when scrolling a view off and back onto the screen.

### Steps to Reproduce

1. Download an episode from the **Filters* tab.
2. Observe the downloading percentage. It should increment.
3. While the episode is still downloading, switch to a different tab and back to the original position.

**Observations**
- Download percentage in Step 3 keeps resetting to 0% after going back to the **Filters** view.

https://user-images.githubusercontent.com/4656348/178306295-8f38c6de-cf8a-407c-a8e2-f79b17c067a4.mp4

**Test Build**
- 7.7 (4913)
- Reproduced on pre-release 7.20.0 build

joashrajin:
+1 here: 8046547-zd-a8c

https://github.com/Automattic/pocket-casts-android/assets/16253818/d459e454-d27e-4aff-b3da-3914d55a316b

PR #1888, opened 2024-02-27
[Playback skipping] Smooth transition from streaming to downloaded episode

mebarbosa:
## Description
- As was reported [here](https://github.com/Automattic/pocket-casts-android/pull/1878#pullrequestreview-1900099251), the system to switch from streaming to downloaded episode was resulting in jumping a second or two back.
- This is caused by an random playback restart that we tracked here: https://github.com/Automattic/pocket-casts-android/issues/1509
- This PR removes the unnecessary playback restart for the caching episode case to implement a better transition between streaming and downloaded the episode.

> [!note]
> While you are testing, you might encounter the existing issue https://github.com/Automattic/pocket-casts-android/issues/86 related to the downloading percentage. I won't address it here as it's not within the scope of this task

> [!note]
> To test this PR I will to have the local `CACHE_PLAYING_EPISODE` feature flag enabled.

## Testing Instructions
- Repeat the tests described in https://github.com/Automattic/pocket-casts-android/pull/1878
- You can skip the `Charging Phone`, `Only on Unmetered WiFi` and `Automotive and Watch` tests if you wanted
- You should see a better transition after download finishes

## Screenshots or Screencast 
- See the video reported here when we were reseting the player for cached episodes: https://github.com/Automattic/pocket-casts-android/pull/1878#pullrequestreview-1900099251

After this improvement the transition will look like this:

[Screen_recording_20240227_135712.webm](https://github.com/Automattic/pocket-casts-android/assets/42220351/4fa23525-b4e3-4052-8e74-c365d5cb18b6)

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
@@ -1890,7 +1890,7 @@ open class PlaybackManager @Inject constructor(
         widgetManager.updateWidget(podcast, play, episode)
 
         if (play) {
-            if (sameEpisode && currentPositionMs != null) {
+            if (sameEpisode && currentPositionMs != null && !isCached) {
                 player?.seekToTimeMs(currentPositionMs)
             }
             play(sourceView, posUpdatedOnPlayerReset)
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
@@ -1803,6 +1803,9 @@ open class PlaybackManager @Inject constructor(
             }
         }
 
+        val podcastEpisode = episode as? PodcastEpisode
+        val isCached = podcastEpisode?.isAutomaticallyCached ?: false && FeatureFlag.isEnabled(Feature.CACHE_PLAYING_EPISODE)
+
         var posUpdatedOnPlayerReset = false
         // We want to make sure we get the current position at the last possible moment before changing/resetting the player
         val currentPositionMs = if (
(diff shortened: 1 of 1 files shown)


MiSikora (changes requested):
While the jump is a bit smaller it is still noticeable and present.

Additionally there is a bug which I can't reproduce on `main`. When you select a different episode after the current one is downloaded it can jump back to playing the previous episode.

You can see both bugs in the video. The wrong episode playback is present around `00:15` and `00:35`.

https://github.com/Automattic/pocket-casts-android/assets/30936061/c6c79d34-6567-4742-b398-21725494840f

mebarbosa:
> While the jump is a bit smaller it is still noticeable and present.
> 
> Additionally there is a bug which I can't reproduce on `main`. When you select a different episode after the current one is downloaded it can jump back to playing the previous episode.
> 
> You can see both bugs in the video. The wrong episode playback is present around `00:15` and `00:35`.
> 
>  video.mp4

Hey @MiSikora . I was able to remove the ~1 second jumps. You can check here: [d35260b](https://github.com/Automattic/pocket-casts-android/pull/1888/commits/d35260b7b976e3c0a612d57b8c0e75347c8902fa)

I was not able to reproduce this issue you reported but I believe this is related to switching episodes. Could you test it again?  [ed0281f](https://github.com/Automattic/pocket-casts-android/pull/1888/commits/ed0281fd61bdbaf5e50f1197c9efe366ea3fd725)

MiSikora:
I can confirm that the jumping issue is gone. However the issue of switching episodes still persists.

And after the latest changes there is one more issue. The playback doesn't switch to the downloaded episode which if I assume correctly was the main counteraction to prevent skipping due to ads. It is downloaded and when I manually resume playback it starts from the file but it doesn't do it automatically. You can test it like with these steps.

1. Play an episode.
2. Let it download.
3. Disable Internet connection.
4. Seek to some timestamp in the episode.
5. Playback stops and you have to resume it manually.

https://github.com/Automattic/pocket-casts-android/assets/30936061/0c794351-beef-46a5-812c-656b9766fc2d

mebarbosa:
Closing this PR since we decided to explore the Exoplayer cache option due these issues

PR #1905, opened 2024-03-05
Revert custom cache system implementation

mebarbosa:
## Description
- This PR reverts #1878

>[!note]
> I am not going to delete the `automatically_cached` column from `podcast_episodes` table on this PR

## Testing Instructions
- Play some episodes and make sure we are not caching them anymore

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/services/model/schemas/au.com.shiftyjelly.pocketcasts.models.db.AppDatabase/89.json
+++ b/modules/services/model/schemas/au.com.shiftyjelly.pocketcasts.models.db.AppDatabase/89.json
@@ -1,0 +1,1624 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 89,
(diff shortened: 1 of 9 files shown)


MiSikora:
Shouldn't this revert also the `automatically_cached` property form the `PodcastEpisode` and revert the migration below?

https://github.com/Automattic/pocket-casts-android/blob/3865d2ac7beb041590c52405158d0ae88056f8ee/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/AppDatabase.kt#L630-L637

You can use auto migration like this one.

https://github.com/Automattic/pocket-casts-android/blob/3865d2ac7beb041590c52405158d0ae88056f8ee/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/AppDatabase.kt#L558-L562

MiSikora (approved):

Issue #1865, opened 2024-02-21
Playback Skipping


mebarbosa:
Prevent the playback position from skipping while listening to a streaming episode by fully downloading it.

```[tasklist]
### Tasks
- [x] Add feature flag
- [x] Trigger a download when the user begins streaming an episode and switch to downloaded episode
- [x] Free up disk space by deleting the downloaded file when the episode ends or stopped
- [x] Revert: #1878
- [x] Implement cache using Exoplayer cache
```

mebarbosa:
Investigation internal reference: `pdeCcb-4Gh-p2`

PR #1910, opened 2024-03-06
Configure app to cache episodes when playing using Exoplayer Cache

mebarbosa:
## Description
- This is to implement episodes caching using exoplayer instead of creating our own cache system https://github.com/Automattic/pocket-casts-android/pull/1878
- For reference: https://developer.android.com/media/media3/exoplayer/network-stacks#caching

This is part of #1865

> [!note]
> To test this PR I will to have the local `CACHE_PLAYING_EPISODE` feature flag enabled.

## Testing Instructions

- Do playback regression tests such as play episodes, downloaded episodes and uploaded episodes
- Try to apply playback effects
- During my tests, I observed an improvement when I switched from one episode to another that I had played before, thanks to the cache. 👌
- You can use the device explore see the cache downloaded by Exoplayer. `Android studio -> Device explore -> /data/data/au.com.shiftyjelly.pocketcasts.debug/cache/podcasts-cache`

image

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SimplePlayer.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SimplePlayer.kt
@@ -294,18 +294,17 @@ class SimplePlayer(val settings: Settings, val statsManager: StatsManager, val c
             }
         }
 
-        val cacheDataSourceFactory: CacheDataSource.Factory? =
-            simpleCache?.let {
-                CacheDataSource.Factory()
-                    .setCache(it)
-                    .setUpstreamDataSourceFactory(httpDataSourceFactory)
-            }
+        val sourceFactory = simpleCache?.let {
+            CacheDataSource.Factory()
+                .setCache(it)
+                .setUpstreamDataSourceFactory(httpDataSourceFactory)
+        } ?: dataSourceFactory
 
         val mediaItem = MediaItem.fromUri(uri)
         val source = if (isHLS) {
-            HlsMediaSource.Factory(cacheDataSourceFactory ?: dataSourceFactory)
+            HlsMediaSource.Factory(sourceFactory)
         } else {
-            ProgressiveMediaSource.Factory(cacheDataSourceFactory ?: dataSourceFactory, extractorsFactory)
+            ProgressiveMediaSource.Factory(sourceFactory, extractorsFactory)
         }.createMediaSource(mediaItem)
 
         player.setMediaSource(source)
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SimplePlayer.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SimplePlayer.kt
@@ -16,8 +16,12 @@ import androidx.media3.common.Player
 import androidx.media3.common.Tracks
 import androidx.media3.common.VideoSize
 import androidx.media3.common.util.UnstableApi
+import androidx.media3.database.StandaloneDatabaseProvider
 import androidx.media3.datasource.DefaultDataSource
 import androidx.media3.datasource.DefaultHttpDataSource
+import androidx.media3.datasource.cache.CacheDataSource
+import androidx.media3.datasource.cache.LeastRecentlyUsedCacheEvictor
+import androidx.media3.datasource.cache.SimpleCache
 import androidx.media3.exoplayer.DefaultLoadControl
 import androidx.media3.exoplayer.ExoPlayer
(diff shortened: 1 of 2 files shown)


mebarbosa:
Tagging @ashiagr since it's a important area of the app

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SimplePlayer.kt
```diff
@@ -265,13 +282,32 @@ class SimplePlayer(val settings: Settings, val statsManager: StatsManager, val c
...
+
         val mediaItem = MediaItem.fromUri(uri)
         val source = if (isHLS) {
-            HlsMediaSource.Factory(dataSourceFactory).createMediaSource(mediaItem)
+            HlsMediaSource.Factory(cacheDataSourceFactory ?: dataSourceFactory)
         } else {
-            ProgressiveMediaSource.Factory(dataSourceFactory, extractorsFactory)
-                .createMediaSource(mediaItem)
-        }
+            ProgressiveMediaSource.Factory(cacheDataSourceFactory ?: dataSourceFactory, extractorsFactory)
```
MiSikora, line 308: The `cacheDataSourceFactory ?: dataSourceFactory` can be extracted to something like

```kotlin
val sourceFactory = simpleCache?.let {
     CacheDataSource.Factory()
        .setCache(it)
        .setUpstreamDataSourceFactory(httpDataSourceFactory)
} ?: dataSourceFactory
```

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SimplePlayer.kt
```diff
@@ -59,6 +68,9 @@ class SimplePlayer(val settings: Settings, val statsManager: StatsManager, val c
 
     private var videoChangedListener: VideoChangedListener? = null
 
+    private var databaseProvider: StandaloneDatabaseProvider? = null
+    private var simpleCache: SimpleCache? = null
```
MiSikora, line 72: I'm not sure but it might be simpler to make these properties `lazy` and then access them only when the feature flag is enabled instead of relying on `run()` and nullability.

MiSikora (approved):
Yeah, I can confirm that it improves playback start speed when switching between recent episodes. Caching works quite well there.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SimplePlayer.kt
mebarbosa: Yeah. I liked it! Here: [464ddb5](https://github.com/Automattic/pocket-casts-android/pull/1910/commits/464ddb5eecde733ea45057708b2c3cfe18c72cd0)

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SimplePlayer.kt
MiSikora: I was thinking of something like this.

```kotlin
private val databaseProvider by lazy { StandaloneDatabaseProvider(context) }
private val simpleCache by lazy {
    SimpleCache(
        File(context.cacheDir, "podcasts-cache"),
        LeastRecentlyUsedCacheEvictor(MAX_DEVICE_CACHE_SIZE_BYTES.toLong()),
        databaseProvider,
    )
}

val sourceFactory = if (FeatureFlag.isEnabled(Feature.CACHE_PLAYING_EPISODE) && location is EpisodeLocation.Stream) {
    CacheDataSource.Factory()
            .setCache(simpleCache)
            .setUpstreamDataSourceFactory(httpDataSourceFactory)
} else {
    dataSourceFactory
}

if (FeatureFlag.isEnabled(Feature.CACHE_PLAYING_EPISODE)) {
    databaseProvider.close()
    simpleCache.release()
}
```

mebarbosa: I would need to investigate deeper, but when I change to `lazy` I got this error after tapping on `play` button:




I am getting this after applying this change. If I revert them it works as expected

mebarbosa: I like the change and it's simple, but for some reason it's not working as expected

MiSikora: Ok, then let's ignore it and move on with your solution. 👍

PR #1926, opened 2024-03-12
Use activity instead of fragment manager for multi-selection

MiSikora:
## Description

As a part of hack week I had to refactor multi-selection tool to use `Activity` instead of `FragmentManager`. This enables more navigation options from multi-selection context menu. I think it would nice to land it regardless of my project.

## Testing Instructions

Smoke test mult-selection in different contexts: podcast episodes, bookmarks, user files.

## Checklist
 
- [ ] with different themes


ashiagr (approved):
LGTM

PR #1932, opened 2024-03-13
Player update release time out

ashiagr:
## Description

This increases player release timeout from default `500ms` to `750ms` using remote config field `player_release_time_out_ms`.

Reports in Sentry include a variety of devices with 53% low-end devices in the last 90 days for this error:
[Search for `error.value:"Player release timed out."` in Discover]

It would be interesting to increase player release time-out slightly as suggested [here](https://rb.gy/7hav6r) and monitor to see if it makes any difference. 

## Testing Instructions

1. `setMinimumFetchIntervalInSeconds` to 5L (5 secs) in `FirebaseModule`
2. Add a log statement at the end of `PocketCastsApplication.onCreate()` that prints `settings.getPlayerReleaseTimeOutMs()`
3. Install release build
4. Launch the app
5. ✅ Notice in logs that the release time out is 750.
6. Play a few episodes.  ✅ Notice that they play fine.
7. Uninstall the apk

## Checklist N/A
 
- [ ] with different themes


mebarbosa (approved):
LGTM. 

Awesome idea to use remote config for this 👍🏽

PR #1599, opened 2023-12-08
Display artwork from podcast feed via show notes endpoint on full-screen player and episodes screens

mchowning:
## Description
This adds support for displaying the episode-specific artwork that is sometimes included in a podcast's RSS feed. This information is available to the app from the call to retrieve the show notes that was [implemented already](https://github.com/Automattic/pocket-casts-android/pull/1083).

Fixes #528

Look at why the podcast artwork briefly displays before the episode artwork on the episode screen

### 1. Displays episode artwork
1. Fresh install the app
3. Open up the podcast screen for "Darknet Diaries"
4. Open the episode details page for an episode
5. ✅  Observe that episode-specific  artwork is displayed instead of the podcast artwork (the first time you do this the podcast artwork may show for a split-second before the episode artwork url is available—that appears to only happen the one time though, visiting other episodes later will not show that issue).
6. Begin playing that episode
7. Go to the full-screen player
8. ✅ Observe that the episode-specific artwork is displayed in the full-screen player
9. Stop playback
10. Turn on airplane mode
11. Open the episode screen for a different episode of Darknet Diaries
12. ✅ Observe that the podcast's default artwork is displayed (we have the URL for all of the podcast's episode artwork on the device at this point, but we can't retrieve the image from the url because of airplane mode)
13. Attempt to play this episode (it will fail because of airplane mode)
14. Open the full-screen player
15. ✅  Observe that the podcast artwork is displayed
16. Go back and open the episode screen from the episode where you saw the episode artwork earlier
17. ✅ Observe that you can see the (cached) episode artwork
18. Attempt to play this episode (it will fail because of airplane mode)
19. Open the full-screen payer
20. ✅ Observe that you can see the (cached) episode artwork
21. Turn off airplane mode
22. Without going to the episode screen for the episode (i.e., from the podcast screen), play an episode you haven't played yet.
23. Without going to the episode screen for the episode, open the full-screen player
24. ✅ Observe that you can see the episode artwork
25. Attach the network inspector in Android Studio
26. Subscribe to Darknet diaries
27. Confirm that the app uses the cache and the `episodes_full_#####.json` call gets a 304 response with no body.

### 2. Artwork for user files still works
1. Begin playing a user file with an associated image
5. Verify that the file image is shown in the full-screen player

### 3. Show notes are still fetched with downloads
1. Fresh install the app
6. Open the podcast page for Darknet Diaries
7. Turn on airplane mode
8. Open an episode and observe you do not see the show notes
9. Back out to the podcast screen
[...]

PR #1943, opened 2024-03-14
Display RSS artwork in all possible places

### LINKED ISSUE #1608: Display episode artwork in more places in the app
In https://github.com/Automattic/pocket-casts-android/pull/1599 we added episode artwork to the full-screen player and the episode details screen, and we now want to use it in more places in the app. The current suggestion is to add it to the:

MiniPlayer | Bookmark row (Podcast screen) | Notification | Widget
---|----|---|---
miniplayer | bookmark | widget | widget1

@ashiagr  brought this up [here](https://github.com/Automattic/pocket-casts-android/pull/1599#pullrequestreview-1774364673), and @david-gonzalez-a8c shared his thoughts [here](https://github.com/Automattic/pocket-casts-android/pull/1599#issuecomment-1852267611). @b-aaron [said](https://github.com/Automattic/pocket-casts-android/pull/1599#issuecomment-1850851452) they would like to have a setting for whether to use the episode artwork in the notification.

MiSikora:
## Description

This PR redesigns how we create request for images. Previously we had `PodcastImageLoader` with multitude of different methods and it wasn't clear when to use which one. It was also hard extend this class to support handling different artworks. I removed that class and introduced `PocketCastImageRequestFactory` with a clearer and simpler API.

Episodes shown in the discovery feature do not show RSS artwork because server does not return image URL for episodes there.

Closes #1608 

## Testing Instructions

Test the screens that are using RSS artwork and see if they respect the `Use embedded RSS artwork` appearance setting. Also smoke test the rest of the app including user files, discovery, and any other place that you can think of.

### Android

| Episode | Up Next | Filters | 
| - | - | - |
| android-episode | android-up-next | android-filters |

| Downloads | Bookmarks | Widget |
| - | - | - |
| android-downlods | android-bookmarks | android-widget |

| Player | Mini player | Media notification |
| - | - | - |
| android-player | android-mini-player | android-notification-center |

| History | Starred |
| - | - |
| history | starred |

In theory notifications should use this as well but I'm not able to test it properly because our developer testing action clears the episodes which means that don't have image URLs when fetched.

### Wear OS

| Now Playing | Up Next | Filters | Downloads |
| - | - | - | - |
| wear-now-playing | wear-up-next | wear-filters | wear-downloads |

### Automotive

No changes to the automotive codebase and it still works as before. It is a bit finicky to test because there might be some delay due to image URL availability in episode entities.

## Checklist
 
- [x] with different themes


CookieyedCodes:
Can we keep the upnext feed clear of RSS artwork 😅, mainly because to me it makes more sense that the upnext should be for sorting feed priority 😅

MiSikora:
I don't have a strong opinion on it but single sample preferences aren't the best way to decide here. If enough people get upset we can always change it but for now I'd prefer to keep it.

MiSikora:
@david-gonzalez-a8c Can you weigh in on whether we shouldn't support episode artwork on some of the screens? And if yes, in which ones?

CookieyedCodes:
> I don't have a strong opinion on it but single sample preferences aren't the best way to decide here. For example @b-aaron wanted it https://github.com/Automattic/pocket-casts-android/pull/1599#issuecomment-1935826324. For my use case it also makes sense to use RSS artwork. If enough people get upset or let us know beforehand we can always change it but for now I'd prefer keeping it.

Ya understandable, I'm actually fine with it being everywhere but upnext, so maby an setting toggle spefically for upnext 🤔 🤔

MiSikora:
> maby an setting toggle spefically for upnext

I'd prefer not showing RSS artwork at all in Up Next than having a separate setting. It's too specific and would go in the direction of turning our settings menu into a plane control panel. 😅

david-gonzalez-a8c:
> @david-gonzalez-a8c Can you weigh in on whether we shouldn't support the episode artwork on some of the screens? And if yes, on which ones?

Besides the ones shown, I can think of:

- Starred
- Listening history
- Search results

I don't have a strong opinion, but I think listening history and starred might make sense and search results not so much. What do you think?

Also, it would be great coordinating this with a similar effort on iOS so we have parity.

MiSikora:
> I don't have a strong opinion, but I think listening history and starred might make sense and search results not so much. What do you think?

Yes, starred and listening history are already handled. I forgot to add it in the PR description. Search uses podcast's artwork because we don't have episode URLs when browsing on that screen and I agree that it would be better to not use it there.

> Also, it would be great coordinating this with a similar effort on iOS so we have parity.

iOS right now doesn't support episode artwork at all because there were some memory issues. But I'll let the iOS folks know.

mebarbosa (changes requested):
Great job!! As a user, I also prefer to display the RSS artwork

- I smoke-tested the app, and it worked pretty well. However, I noticed one thing related to our widget:
<details><summary>Widget test details</summary>

1. If I add the widget with the RSS artwork toggle OFF,
2. then turn the toggle ON,
3. the image in the widget is not updating, even after refreshing it. 

The only way to update it is by actually switching to another podcast
</details> 

- I also have a suggestion to add this to `Recent searches`:

<details><summary>Recent searches screenshot</summary>
<p>

Screenshot_20240315_092024


</p>
</details>

MiSikora:
> Widget test details

Yes, I forgot to mentioned that. Right now it updates in a similar way to the notification center. You need to pause and resume an episode for the setting to take effect. Doesn't widget artwork refresh for you when you pause and play the episode? Changing that to be live would require far more drastic redesign of our widget handling and I'm not sure I want to introduce it in this PR.

> I also have a suggestion to add this to `Recent searches`:

As I mentioned it is not possible to have it in the Discovery page. Server doesn't provide the image URL there.

mebarbosa (approved):

b-aaron:
Seeing this implemented now (which I am really enjoying), I really would love to be able to see all the episode covers within the main Episodes feed of a podcast. Can that be implemented as well?
I think it would be great, if there was a toggle within the podcast settings to show them, so that podcasts, which don't use custom episode covers show the same cover all over.

CookieyedCodes:
One thing I'm getting concerned about with is the amount of custom artwork in the app is the amount of storage its going to take up on my device, can we get some controls for this & even a format that could compress the data even further 😅, I have nearly 2000 feeds & I want to cut down on photos that takeing up my storage 🤔

MiSikora:
@b-aaron Can you create a feature request for that? It probably won't have a high priority but at least it will be tracked this way.

@CookieyedCodes Image cache size is limited.

CookieyedCodes:
@MiSikora I assume it is, what's it set to artwork is like taking 6gbs 😅

MiSikora:
It depends on a device and its storage state. But it is between 10MB and 250MB.

CookieyedCodes:
Hmmmmm I don't know why my app says it's taking up 29Gb where's on google play is taking up 35Gb 🤔🤔

CookieyedCodes:
I in fairness did assume it but my basis is somewhat in good grounding, when I had podcast addict as a pocketcasts backup 12 GBs out of 20glGBs taken by podcast addict of my phone at the time were being taken up with just artwork (no downloads, 5gbs were backups & the remainder was feeds I assume), I had about 1600+ feeds at time I think 

( if you think it's files 99.5% of my files I keep on my pocketcasts cloud so I have no idea what's taking up that missing 6 GBs 😅)

PR #1945, opened 2024-03-14
Update categories pills UI

mebarbosa:
## Description
- Implements the action to update the category pills UI
- Moved `CategoriesListRowRedesignAdapter` to another file
- Added icons
- Created `CategoryPill` to wrap `DiscoverCategory`

This is part of the categories redesign. You can find references and figma here in our internal reference: `pdeCcb-4tv-p2`.

> [!note]
> The behavior to open categories view in the same discover view is not implemented yet. So clicking on some category may not provide the best experience

## Testing Instructions
1. `git apply` [categories-redesign.patch](https://github.com/Automattic/pocket-casts-android/files/14606106/categories-redesign.patch)
2. The categories should always be displayed in the following order: `True Crime, Comedy, Culture, History, Fiction, Technology`
3. Click on any category
4. The selected category must be updated to background `blue`, text color `white` ✅
5. The other categories must be hidden ✅
6. `All categories` button must be updated to X button ✅
7. Click on `X` button
8. The categories filter must be cleaned ✅
9. Click on `All categories` button
10. The modal with all categories must be opened with the arrow icon updated ✅

## Screenshots or Screencast 
[Screen_recording_20240314_143823.webm](https://github.com/Automattic/pocket-casts-android/assets/42220351/691c81c0-3695-47d5-8674-9003c2272927)

## Checklist
 
- [x] with different themes

cc: @david-gonzalez-a8c


modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/view/CategoriesListRowRedesignAdapter.kt
```diff
@@ -0,0 +1,127 @@
...
+import au.com.shiftyjelly.pocketcasts.discover.databinding.CategoryPillBinding
+import au.com.shiftyjelly.pocketcasts.servers.model.DiscoverCategory
+import au.com.shiftyjelly.pocketcasts.servers.model.NetworkLoadableList
+
+@SuppressLint("NotifyDataSetChanged")
+class CategoriesListRowRedesignAdapter(
+    private val onCategoryClick: (NetworkLoadableList) -> Unit,
+    private val onAllCategoriesClick: (() -> Unit) -> Unit,
+    private val categories: List<CategoryPill>,
+) : RecyclerView.Adapter<CategoriesListRowRedesignAdapter.CategoriesRedesignViewHolder>() {
```
Mukuljangir372, line 22: Do you think we should consider to use ListAdapter rather than using RecyclerView.Adapter for better performance?

modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/view/CategoriesListRowRedesignAdapter.kt
```diff
@@ -0,0 +1,127 @@
...
+        fun bind(category: CategoryPill, context: Context) {
+            if (category.discoverCategory.id == DiscoverCategory.ALL_CATEGORIES_ID) {
+                setUpAllCategoriesAndClear(context, category)
+            } else {
+                setUpCategories(context, category)
+            }
+        }
+        private fun setUpAllCategoriesAndClear(context: Context, category: CategoryPill) {
+            if (category.isSelected) {
+                binding.categoryName.visibility = View.GONE
```
Mukuljangir372, line 48: Do you think we should use `isVisibility = false` here for better readable code?

mebarbosa: Yeah. Looks better with `isVisibility `. I've just updated the code

modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/view/CategoriesListRowRedesignAdapter.kt
mebarbosa: I agree with you. The code is already updated. Thank you for your suggestion!

modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/view/CategoriesListRowRedesignAdapter.kt
```diff
@@ -0,0 +1,151 @@
...
+import androidx.recyclerview.widget.ListAdapter
+import androidx.recyclerview.widget.RecyclerView
+import au.com.shiftyjelly.pocketcasts.discover.R
+import au.com.shiftyjelly.pocketcasts.discover.databinding.CategoryPillBinding
+import au.com.shiftyjelly.pocketcasts.servers.model.DiscoverCategory
+import au.com.shiftyjelly.pocketcasts.servers.model.NetworkLoadableList
+
+private val CATEGORY_REDESIGN_DIFF = object : DiffUtil.ItemCallback<CategoryPill>() {
+    override fun areItemsTheSame(oldItem: CategoryPill, newItem: CategoryPill): Boolean {
+        return oldItem == newItem
```
MiSikora, line 21: This should compare by ID.

modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/view/CategoriesListRowRedesignAdapter.kt
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    override fun areContentsTheSame(oldItem: CategoryPill, newItem: CategoryPill): Boolean {
+        return oldItem.hashCode() == newItem.hashCode()
```
MiSikora, line 25: I think this comes from the old code but we shouldn't compare object's equality using hash codes. Hash code collision is a thing. And while it probably doesn't really matter for this use case because we have very few categories I'd prefer avoiding it because someone might copy/paste it in instances when it does matter.

modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/view/CategoriesListRowRedesignAdapter.kt
```diff
@@ -0,0 +1,151 @@
...
+    }
+}
+
+@SuppressLint("NotifyDataSetChanged")
```
MiSikora, line 29: I have a general question about this class. The main benefit of using `ListAdapter` is that it handles diffing and computes output automatically. Using `notifyDataSetChanged()` on it is counterproductive. The way `ListAdapter` works best is that whenever an event happens it should pass it to the higher layers until the change reaches a view model. There, view model does necessary computations and updates items to be displayed by the adapter using `submitList()`.

This design where we keep internally `currentCategories` as a mutable list inside `submitList()` doesn't utilize this and makes it hard to test.

MiSikora (changes requested):
There is UI bug on my Pixel 6 where text in pills isn't centered.

Screenshot_20240315-142917

I have also some concerns about the design of the `CategoriesListRowAdapter`.

modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/view/CategoriesListRowRedesignAdapter.kt
mebarbosa: I copied from an exiting implementation of this 😝 

[377f966](https://github.com/Automattic/pocket-casts-android/pull/1945/commits/377f966417f4a9444966a282e4ad9c9e42afa6f7)

modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/view/CategoriesListRowRedesignAdapter.kt
mebarbosa: I agree. [377f966](https://github.com/Automattic/pocket-casts-android/pull/1945/commits/377f966417f4a9444966a282e4ad9c9e42afa6f7)

mebarbosa:
> 06b0615

@MiSikora Could you check if they are centralized now? [06b0615](https://github.com/Automattic/pocket-casts-android/pull/1945/commits/06b06152cb1b087c60382aef3039e2e6f71d9088)

MiSikora:
Yes, they are centered now correctly. 👍

modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/view/CategoriesListRowRedesignAdapter.kt
mebarbosa: I agree and I was not proud to use the `notifyDataSetChanged `. I tried this new implementation and I think it's better: [516c2ce](https://github.com/Automattic/pocket-casts-android/pull/1945/commits/516c2ced7cdb1676ad5ddcb32967e6f08eacaa49)

Thanks

MiSikora (approved):
LGTM

PR #1964, opened 2024-03-20
Fetch chapters from standard tags

MiSikora:
## Description

This PR adds support for fetching and storing episode chapters. There are two ways we can retrieve chapters. They are either directly available under `chapters` field or we have to make a network request to an URL under `chapters_url` field.

The rules behind client processing are as follows:

1. Process and save chapters for all episodes when they are directly available under `chapters` field.
2. Process and save chapter only for requested episode when chapters are available through `chapters_url` field. It would be too costly to execute hundreds of additional network requests for all `chapter_url` fields.
3. If both `chapters` and `chapters_url` are present skip `chpaters_url` processing if `chapters` is not empty.

### `chapters`

1. Subscribe to a podcast with chapters from Podlove like [this one](https://pca.st/chaosradio).
2. Open an episode from this podcast.
3. Open app inspector and execute `SELECT * FROM episode_chapters` query on our database.
4. You should see results similar to the ones below.

Screenshot 2024-03-20 at 09 21 03

### `chapters_url`

1. Subscribe to a podcast with chapters from Podcast Index like [this one](https://pca.st/changelog).
2. Open [this episode](https://pca.st/n2c345cd).
3. Open app inspector and execute `SELECT * FROM episode_chapters WHERE episode_uuid IS  'c34c9e87-12c2-458c-b70c-b4e6e5c33ece'` query on our database.
4. You should see results similar to the ones below.
5. You can verify if chapters for any other episodes of this podcast are not stored.

Screenshot 2024-03-20 at 09 19 52

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/di/RepositoryModule.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/di/RepositoryModule.kt
@@ -155,5 +155,5 @@ abstract class RepositoryModule {
     abstract fun provideRatingsManager(ratingsManagerImpl: RatingsManagerImpl): RatingsManager
 
     @Binds
-    abstract fun bindChapterManager(chapterManagerImpl: ChapterManagerImpl): ChapterManager
+    abstract fun provideChapterManager(chapterManagerImpl: ChapterManagerImpl): ChapterManager
 }
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/di/RepositoryModule.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/di/RepositoryModule.kt
@@ -20,6 +20,8 @@ import au.com.shiftyjelly.pocketcasts.repositories.playback.PlayerFactory
 import au.com.shiftyjelly.pocketcasts.repositories.playback.PlayerFactoryImpl
 import au.com.shiftyjelly.pocketcasts.repositories.playback.UpNextQueue
 import au.com.shiftyjelly.pocketcasts.repositories.playback.UpNextQueueImpl
+import au.com.shiftyjelly.pocketcasts.repositories.podcast.ChapterManager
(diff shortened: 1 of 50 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please add tests for class <code>ChapterManagerImpl</code> (or add <code>unit-tests-exemption</code> label to ignore this).</td>
    </tr>
  </tbody>
</table>

</p>

CookieyedCodes:
I been pushing for RSS time tags to be converted to chapters, just a reminder to consider that could be another form of chapters 😉

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/di/RepositoryModule.kt
```diff
@@ -151,4 +153,7 @@ abstract class RepositoryModule {
     @Binds
     @Singleton
     abstract fun provideRatingsManager(ratingsManagerImpl: RatingsManagerImpl): RatingsManager
+
+    @Binds
+    abstract fun bindChapterManager(chapterManagerImpl: ChapterManagerImpl): ChapterManager
```
mebarbosa, line 158:
```suggestion
    abstract fun provideChapterManager(chapterManagerImpl: ChapterManagerImpl): ChapterManager
```

I personally prefer `bind` instead of `provide`, but since all methods of this class use `provide` what do you think to change it?

MiSikora: I don't care either way. I used bind because it actually binds but sounds good.

mebarbosa (changes requested):
Nice job! ~I had one problem with the last test. I could not get the results like you described:~
It worked as expected! As we checked, I subscribed to another podcast. 

image

mebarbosa (approved):

MiSikora:
> I been pushing for RSS time tags to be converted to chapters, just a reminder to consider that could be another form of chapters 😉

Thanks, I forwarded it to our folks.

PR #1969, opened 2024-03-22
Add alphabetical sort order for podcast episodes

MiSikora:
## Description

Web has alphabetical sort ordering for episodes. We also have it in the code but never used it. This PR adds it to the UI.

Also this PR improves the alphabetical sorting order for English titles and matches it with Web. Instead of doing dummy sorting we exclude `a`, `an`, and `the` articles from sorting. So for example `The Big Bad Wolf` isn't treated as staring with `The` but with `Big`.

## Testing Instructions

1. Subscribe to a podcast.
2. Sort episodes by `A-Z`.
3. Check that the episodes are sorted alphabetically. English articles at the beginning of titles shouldn't be included in sorting.
4. Change sorting order to `Z-A`.
5. Check that the episodes are sorted alphabetically in a reversed order. English articles at the beginning of titles shouldn't be included in sorting.

## Screenshots or Screencast 

| Settings | A-Z sorting |
| - | - |
| Screenshot_20240322-141144 | Screenshot_20240322-141152 |

## Checklist
 
- [ ] with different themes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 *   Updates:
     *   Display episode's RSS artwork in more places in the app.
         ([#1943](https://github.com/Automattic/pocket-casts-android/pull/1943))
+    *   Add alphabetical sort order for podcast episodes.
+        ([#1968](https://github.com/Automattic/pocket-casts-android/pull/1968))
 
 7.60
 -----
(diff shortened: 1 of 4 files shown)


CookieyedCodes:
I think in an earlier version it actually was used, removed between 7.0-7.3 or something like that, tbh I don't have much use tho I would love a sort option on the podcast page for sort A-Z but folder first & filters to have an option to group by podcast  😅 (or even better folder filter pill with the option to sort by folder order, that would enable episode stacking, which will help fill the void until playlist functionality & enhance folders, filters & plus 😉)

PR #1987, opened 2024-03-27
Consolidate embedded episode artwork settings

### LINKED ISSUE #1983: Embedded artwork: consolidate options into a single one
## Goal

To have a single option for the user to opt-in to embedded artwork under Settings > Appearance.

Suggestion:

> Use episode artwork (off/on)
> Some shows have custom artwork for certain episodes. Enable this option and Pocket Casts will display them instead of the show’s artwork.

MiSikora:
## Description

This PR consolidates file artwork and RSS artwork into a single setting. It also improves handling file artwork by using it if possible in all of the same places that we use RSS artwork. There are some limitation though. For example it isn't supported in the media notifications.

Closes #1983

### Migration

1. Clean install the app from b78b3acb38b23650acfddfb619747ed6e23a06cd.
2. Set the `Use embedded RSS artwork` setting or the `Use embedded file artwork` setting on.
3. Update the app from the `task/consolidate-embedded` branch.
4. Go to `Profile`/`Settings`/`Appearance`.
5. Check the `Use episode artwork` setting. If any of the settings from the 2nd step was enabled it should be enabled too.
6. Repeat this test for different combinations of settings.

### Syncing

Check that the `Use episode artwork` setting syncs.

### RSS Artwork 

Smoke test displaying RSS artwork from #1943.

### File Artwork

1. Enable the `Use episode artwork` setting.
2. Subscribe to [Upgrade](https://pca.st/upgrade) podcast and play [this episode](https://pca.st/7hesqiu3).
3. Check that the app displays embedded artwork in most places. There are some limitations that I mentioned in the description.
4. There might be small delay (~10 seconds) before the artwork takes effect.
5. Disable the `Use episode artwork` setting. App should display podcast artwork.

| File artwork | Podcast artwork |
| - | - |
| Screenshot_20240327_232628 | Screenshot_20240327_232643 |

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/jobs/VersionMigrationsJob.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/jobs/VersionMigrationsJob.kt
@@ -19,6 +19,7 @@ import au.com.shiftyjelly.pocketcasts.repositories.podcast.EpisodeManager
 import au.com.shiftyjelly.pocketcasts.repositories.podcast.PodcastManager
 import au.com.shiftyjelly.pocketcasts.repositories.sync.SyncManager
 import au.com.shiftyjelly.pocketcasts.utils.FileUtil
+import au.com.shiftyjelly.pocketcasts.utils.Util
 import au.com.shiftyjelly.pocketcasts.utils.log.LogBuffer
 import dagger.hilt.android.AndroidEntryPoint
 import javax.inject.Inject
@@ -158,6 +159,10 @@ class VersionMigrationsJob : JobService() {
         if (previousVersionCode < 6362) {
             upgradeTrimSilenceMode()
         }
+
+        if (previousVersionCode < 9204) {
+            consolidateEmbeddedArtworkSettings(applicationContext)
+        }
     }
 
     private fun removeOldTempPodcastDirectory() {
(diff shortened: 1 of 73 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/jobs/VersionMigrationsJob.kt
```diff
@@ -158,6 +159,10 @@ class VersionMigrationsJob : JobService() {
         if (previousVersionCode < 6362) {
             upgradeTrimSilenceMode()
         }
+
+        if (previousVersionCode < 9204) {
```
mebarbosa, line 163: Do you think we should extract this number to a variable?

mebarbosa (approved):
It worked as expected! I've just left one minor comment, but it's not a blocker

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/jobs/VersionMigrationsJob.kt
MiSikora: I don't think it would be helpful. It doesn't have any semantic meaning except for being used for the migration purposes. But do you have an idea for the name?

MiSikora, line 163:
```suggestion
        if (previousVersionCode < 9207) {
```

PR #2022, opened 2024-04-09
Bookmark share

ashiagr:
## Description
This adds a share option for bookmarks.

## Testing Instructions

- Start the app using a user with a Patron Account on a device
- If you still don't have any Bookmarks please create some on the account on multiple podcasts
- Go a podcast with bookmarks
- Open Bookmarks tab for the podcast
- Tap on the options button
- Select one bookmark
- Tap on Share
- Copy the link
- Open the link that you copied from the other device on device browser
- Tap on Open In Pocket Casts
- Check that the link is opened on the correct Podcast -> Episode -> Time location
- Repeat the test above on the other locations where Bookmarks list are shown
    - Player full screen, tab bookmarks
    - Episode detail, tab bookmarks  

## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/1405144/72ccd3dd-6f7b-4167-af4e-7138ce2fcd7f

## Checklist
- [ ] If this is a user-facing change, I have added an entry in CHANGELOG.md
- [x] Ensure the linter passes (`./gradlew spotlessApply` to automatically apply formatting/linting)
- [ ] I have considered whether it makes sense to add tests for my changes
- [ ] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml`
- [ ] Any jetpack compose components I added or changed are covered by compose previews
 
#### I have tested any UI changes...

- [ ] with different themes
- [ ] with a landscape orientation
- [ ] with the device set to have a large display and font size
- [ ] for accessibility with TalkBack

PR #2026, opened 2024-04-10
Profile bookmarks - Display & Sort

ashiagr:
## Description
This displays bookmarks on the Profile screen and allows sorting them.

## Testing Instructions

1. Start the app with an user that has Plus or Patron active
3. Go to profile
4. Scroll down and check if you see a Bookmarks option
5. Tap on Bookmarks
6. Check if the list of your bookmarks is seen ( if you never set a bookmark you should see an empty view that says how to set it)
7. Tap the bookmark play button and check if it starts playing the bookmark episode at the bookmark timestamp
8. Check if sort option works
    - Tap on the more button on the top right
    - Tap on Sort By
    - Switch between different sort orders and check if it works correctly
9. Check if delete works
     - Tap on more button again
     - Tap on select bookmarks
     - Select some bookmarks
     - Tap on the delete icon ( trash button)
     - Confirm the deletion
     - Check that the delete happens
10. Check if edit works
     - Tap on more button again
     - Tap on select bookmarks
     - Select one bookmark
     - Tap on the edit button
     - Change the name of the bookmark
     - Confirm
     - Check if the name updates correctly
11. Check if share works
     - Tap on more button again
     - Tap on select bookmarks
     - Select one bookmark
     - Tap on the share button
     - Share bookmark as per steps in https://github.com/Automattic/pocket-casts-android/pull/2022#top
12. Check event tracking
    - Interact with bookmarks on the profile (change name, delete)
    - Check if tracks events have the `source` parameters set to `profile`  

### Test without Plus

1. Start the app with an user without Plus or Patron
2. Go to profile
3. Scroll down and check if you see a Bookmarks option
4. Tap on Bookmarks
5. Check if it shows a text that allows to redirect to the Plus purchase flow

## Screenshots or Screencast 
out

35. ## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/bookmark/BookmarksPage.kt
+++ b/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/bookmark/BookmarksPage.kt
@@ -45,7 +45,7 @@ import au.com.shiftyjelly.pocketcasts.localization.R as LR
 
 @Composable
 fun BookmarksPage(
-    episodeUuid: String,
+    episodeUuid: String?,
     backgroundColor: Color,
     textColor: Color,
     sourceView: SourceView,
@@ -116,7 +116,7 @@ private fun Content(
     Box(
         modifier = Modifier
             .background(color = backgroundColor)
-            .padding(bottom = 28.dp),
+            .padding(bottom = if (sourceView == SourceView.PROFILE) 0.dp else 28.dp),
     ) {
         when (state) {
             is UiState.Loading -> LoadingView()
@@ -177,9 +177,10 @@ private fun BookmarksView(
             )
         }
         items(state.bookmarks, key = { it }) { bookmark ->
+            val episode = state.episodes.firstOrNull { bookmark.episodeUuid == it.uuid }
             BookmarkRow(
-                bookmark = bookmark,
-                episode = state.episode,
+                bookmark = bookmark.copy(episodeTitle = episode?.title ?: ""),
+                episode = episode,
                 isMultiSelecting = { state.isMultiSelecting },
                 isSelected = state.isSelected,
                 onPlayClick = onPlayClick,
@@ -196,7 +197,7 @@ private fun BookmarksView(
                     SourceView.PLAYER -> TimePlayButtonColors.Player(textColor = textColor)
                     else -> TimePlayButtonColors.Default
                 },
-                showIcon = false,
+                showIcon = state.showIcon,
                 useEpisodeArtwork = state.useEpisodeArtwork,
             )
         }
@@ -222,9 +223,11 @@ private fun BookmarksPreview(
                         title = "Funny bit",
                     ),
                 ),
-                episode = PodcastEpisode(
-                    uuid = "",
-                    publishedDate = Date(),
+                episodes = listOf(
+                    PodcastEpisode(
+                        uuid = "",
+                        publishedDate = Date(),
+                    ),
                 ),
                 isMultiSelecting = false,
                 useEpisodeArtwork = false,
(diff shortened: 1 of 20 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          3 Errors
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please add tests for class <code>ProfileBookmark</code> (or add <code>unit-tests-exemption</code> label to ignore this).</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please add tests for class <code>BookmarksSortTypeForProfile</code> (or add <code>unit-tests-exemption</code> label to ignore this).</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please add tests for class <code>UserSettingPref</code> (or add <code>unit-tests-exemption</code> label to ignore this).</td>
    </tr>
  </tbody>
</table>

</p>

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/bookmark/BookmarksPage.kt
```diff
@@ -177,9 +177,10 @@ private fun BookmarksView(
             )
         }
         items(state.bookmarks, key = { it }) { bookmark ->
+            val episode = state.episodes.firstOrNull { bookmark.episodeUuid == it.uuid }
```
MiSikora, line 180: I'm a bit worried by performance here. We iterate over episodes in each composition and in the `items` loop. Would it make sense to have the state hold `Map<String, BaseEpisode>` instead of a `List<BaseEpisode>`. This way lookups by ID will be much faster.

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/BookmarkDao.kt
```diff
@@ -94,6 +95,36 @@ abstract class BookmarkDao {
...
+
+    // Sorts the array by podcast name, episodes release date, and the bookmarks timestamp
+    @Query(
+        """SELECT bookmarks.*, podcasts.title as podcastTitle, podcast_episodes.title as episodeTitle, podcast_episodes.published_date as publishedDate
+            FROM bookmarks
+            LEFT JOIN podcast_episodes ON bookmarks.episode_uuid = podcast_episodes.uuid
+            LEFT JOIN podcasts ON bookmarks.podcast_uuid = podcasts.uuid
+            WHERE deleted = :deleted
+            ORDER BY
+            CASE WHEN podcasts.title is NULL then 1 else 0 end, UPPER(podcasts.title) ASC,
```
MiSikora, line 119: It'd be nice to escape English articles, similarly to what we do here.

https://github.com/Automattic/pocket-casts-android/blob/c9127eb0bfe5c76ea157697904d1fdbcfd64f1dd/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/EpisodeDao.kt#L68-L84

MiSikora (approved):
Good work! Everything works as expected. I left two comments with potential improvements.

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/BookmarkDao.kt
MiSikora: I can also recommend running the query through a formatter like [this one](https://codebeautify.org/sqlformatter) if it gets too messy. 😅

MiSikora:
Ah! Also one thing I missed. Should this be mentioned in the changelog?

ashiagr:
> Ah! Also one thing I missed. Should this be mentioned in the changelog?

I also need to include the search field. I'll add to the changelog when the feature is finalized. I'll look into the rest of the feedback.

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/bookmark/BookmarksPage.kt
ashiagr: I agree, addressed it here: https://github.com/Automattic/pocket-casts-android/pull/2026/commits/be82b2638589b0466c364a541a17e5e751139cde

PR #2028, opened 2024-04-10
Profile bookmarks - Sync Sort Option

ashiagr:
## Description
Adds profile bookmarks sort option sync support between devices.

## Testing Instructions

- Install develop build variant
- Start the app on two different devices (D1, D2) with the same account that needs to have Plus active
- D1: Go to Profile -> Bookmarks
- D1: Set a sort option
- D1: Go back to the profile screen and tap Refresh now
- D2: Go to the profile screen and tap Refresh now
- D2: Open Profile -> Bookmarks
- D2: Ensure that the sort option active is the same as in the D1 device.

## Checklist
 
- [ ] with different themes


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/model/BookmarksSortTypeForProfile.kt
```diff
@@ -26,7 +26,7 @@ enum class BookmarksSortTypeForProfile(
     PODCAST_AND_EPISODE(
         labelId = R.string.podcast_and_episode,
         key = SORT_TYPE_PODCAST_AND_EPISODE,
-        serverId = 2,
+        serverId = 4, // Matches the serverId for the same sort type in iOS app
```
MiSikora, line 29: AFAIK Web uses `2` as a value.

ashiagr: I checked iOS implementation [here](https://github.com/Automattic/pocket-casts-ios/blob/6a41c5a94dde471f4193593081a5a656ccc6254c/Modules/DataModel/Sources/PocketCastsDataModel/Public/Enums.swift#L79)  which sets podcast and episode sort type value as 4. I'll cross-check again tomorrow and update.

mebarbosa (approved):
I've tested and it's working as expected

I will leave this as approved since we are waiting to confirm the `serverId` value. Let me know in case you want me to test it again

ashiagr:
Updated `serverId` for the sort order to 2 (p1712900885888749-slack-C06EHRAPW12)

PR #2029, opened 2024-04-10
Save embedded chapters in the database

### LINKED ISSUE #2018: Save embedded chapters in the database
Now that we have a table for storing chapters we can save embedded chapters in our database. This has two main benefits:

1. Single source of truth for chapters.
2. Improved offline handling. Once chapters are stored in the database we can access them without analyzing the media track for things like displaying in other places than player. It would still require for the file to be played at least once but it is an improvement.

Chapters should not be stored if the current chapters count for an episode is greater or equal to what we about to save.

MiSikora:
## Description

This PR makes `ChaptersManager` a single source of truth for chapters. Previously we juggled between embedded and remote chapters during playback. Now chapters are always saved to the database and database provides it to the playback manager.

Closes #2018 

## Testing Instructions

Test chapters and chapters selection. You should be able to select chapters, skip chapters, and so on.

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/Chapters.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/Chapters.kt
@@ -1,6 +1,5 @@
 package au.com.shiftyjelly.pocketcasts.models.to
 
-import au.com.shiftyjelly.pocketcasts.models.entity.BaseEpisode
 import au.com.shiftyjelly.pocketcasts.utils.featureflag.Feature
 import au.com.shiftyjelly.pocketcasts.utils.featureflag.FeatureFlag
 import kotlin.time.Duration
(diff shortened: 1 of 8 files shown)


modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/Chapters.kt
```diff
@@ -82,19 +81,14 @@ data class Chapters(
...
-                else -> chapter
-            }
-        },
-    )
-
-    fun updateDeselectedState(currentEpisode: BaseEpisode?) = copy(
-        items = items.map { chapter ->
-            chapter.copy(selected = currentEpisode?.deselectedChapters?.contains(chapter.index) == false)
-        },
-    )
```
MiSikora, line 99: This two functions are no longer necessary because all chapters normalization happens inside `ChapterManager` when fetching chapters.

ashiagr (approved):
LGTM. Thanks for this!

PR #2033, opened 2024-04-11
Do not clear named setting modification timestamp

MiSikora:
## Description

`needsSync` property on `UserSetting` lost its meaning when we switched to always syncing all settings due to legacy reasons. This PR changes behavior of modification timestamp so it never gets cleared - only not updated.

## Testing Instructions

Smoke test named settings syncing.

## Checklist
 
- [ ] with different themes


MiSikora:
You can ignore changes from ae5685736d81204ab1e8beb866cb39710340ab69. It got into this PR because #2035 got merged prematurely

mebarbosa (approved):

PR #2035, opened 2024-04-11
 Update settings only if remote timestamp is greater than the local one

MiSikora:
## Description

This PR addresses two things.

1. Update local setting values only if remote timestamp is greater than the local one. Previously it was greater or equal.
2. Use `EPOCH +1 millisecond` as a default timestamp when syncing. This is to prevent a scenario where users sync their devices with settings changed before we started tracking modification timestamp, and then they switch to a new device. This way we compare local `EPOCH` to `EPOCH +1` which will update.

## Testing Instructions

Do some regressions tests for named settings syncing.

## Checklist
 
- [ ] with different themes


ashiagr (approved):
LGTM!

PR #2036, opened 2024-04-11
Update podcast setting only if remote timestamp is greater than local one

MiSikora:
## Description

This PR addresses couple of things.

1. Update podcast settings **only if** remote timestamp is greater than the local one. Previously we updated it always. I'm not sure why it was designed this way. But new approach follows the pattern that we have for named settings.
2. Use `EPOCH +1` logic for updates. This is described in #2035.
3. Fix some of the syncing bugs.
   1. `showArchivedModified` was updating to `showNotificationsModified`.
   2. `autoArchive` was sending modification timestamp for `overrideGlobalEffects`.

## Testing Instructions

Do some regression tests for syncing podcast settings. Focus especially on trim mode, auto archive, and showing archived.

## Checklist
 
- [ ] with different themes


mebarbosa (approved):
It looks good to me

PR #2037, opened 2024-04-11
Profile bookmarks - Add search

ashiagr:
## Description
This adds search to the profile bookmarks screen.

## Testing Instructions
1. Start the app with a user that has Plus or Patron active and a few bookmarks
3. Go to profile
4. Scroll down and check if you see a Bookmarks option
5. Tap on Bookmarks
6. Notice that the screen now has a search field
7. Search for a bookmark or episode title 
8. Notice that search results are filtered based on searched term or display a no bookmark view

## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/1405144/f7209ff9-86b7-4b88-b3ac-00b5fcd8a226

## Checklist
 
- [x] with different themes

### DIFF
--- a/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/bookmark/BookmarksPage.kt
+++ b/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/bookmark/BookmarksPage.kt
@@ -37,6 +37,7 @@ import au.com.shiftyjelly.pocketcasts.models.entity.Bookmark
 import au.com.shiftyjelly.pocketcasts.models.entity.PodcastEpisode
 import au.com.shiftyjelly.pocketcasts.models.type.SyncStatus
 import au.com.shiftyjelly.pocketcasts.player.view.bookmark.components.HeaderRow
+import au.com.shiftyjelly.pocketcasts.player.view.bookmark.components.NoBookmarksInSearchView
 import au.com.shiftyjelly.pocketcasts.player.view.bookmark.components.NoBookmarksView
 import au.com.shiftyjelly.pocketcasts.player.view.bookmark.components.UpsellView
 import au.com.shiftyjelly.pocketcasts.player.viewmodel.BookmarksViewModel
@@ -176,7 +177,6 @@ private fun BookmarksView(
                     text = state.searchText,
                     placeholder = stringResource(LR.string.search),
                     onTextChanged = onSearchTextChanged,
-                    onSearch = {},
                     modifier = Modifier
                         .fillMaxWidth()
                         .padding(horizontal = 16.dp, vertical = 8.dp)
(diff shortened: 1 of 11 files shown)


modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/bookmark/BookmarksPage.kt
```diff
@@ -155,26 +164,48 @@ private fun BookmarksView(
...
+            state.bookmarks.isEmpty()
+        ) {
+            item { NoBookmarksInSearchView(onActionClick = { onSearchTextChanged("") }) }
+        } else {
+            item {
+                val title = stringResource(
+                    id = if (state.bookmarks.size > 1) {
+                        LR.string.bookmarks_plural
+                    } else {
+                        LR.string.bookmarks_singular
```
MiSikora, line 198: I realize this is already in our app but shouldn't this be handles with [plural strings](https://developer.android.com/guide/topics/resources/string-resource#Plurals) and not regular strings? Otherwise it doesn't translate well to different languages that have different grammar rules for different quantities.

MiSikora (approved):
Works as expected.

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/bookmark/search/BookmarkSearchHandler.kt
```diff
@@ -0,0 +1,43 @@
...
+class BookmarkSearchHandler @Inject constructor(
+    private val bookmarkManager: BookmarkManager,
+) {
+    private var searchTerm = ""
+    private val searchQueryFlow = MutableStateFlow("")
+    private val noSearchResult = SearchResult("", null)
+
+    @OptIn(ExperimentalCoroutinesApi::class)
+    suspend fun getBookmarkSearchResultsFlow() = searchQueryFlow.flatMapLatest { searchTerm ->
+        if (searchTerm.isNotEmpty()) {
```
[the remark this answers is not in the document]
ashiagr: Yes, here you go: https://github.com/Automattic/pocket-casts-android/pull/2037/commits/a115ab3e15bf6bd1792d8a9da3b9901971585db5

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/bookmark/BookmarksPage.kt
ashiagr: I don't know the full context but it could be due to GlotPress not supporting plurals yet. I'll leave it for now and we can discuss it internally.

b-aaron:
When sorting by Podcast & Episode, custom files don't get split up for each file, so all of them are just sorted by timestamp, jumbled together.

Don't really have time to open an Issue right now, so just commenting here.

PR #2041, opened 2024-04-11
Add widget previews

MiSikora:
## Description

Add preview images for widgets.

## Testing Instructions

Open widgets panel and check that the Pocket Casts app has widget previews.

## Screenshots or Screencast 

Screenshot_20240411_184031

## Checklist
 
- [ ] with different themes


mebarbosa (approved):
It looks good to me. 

Just to double check, for the medium player, we don't have the Pocket Casts icon in the top right corner like we do for the Large Player, right?

MiSikora:
Yes, there shouldn't be such an icon

PR #2046, opened 2024-04-12
Adjust corner size to 6 dp for widgets

MiSikora:
## Description

Some last minute changes for widgets consistency that David wanted to include in widgets and widget previews. This uses `6dp` as a corner radius. It also doesn't use Coil's transformation for corner calculations because it isn't handled well for small sizes. Instead we delegate to Compose.

## Testing Instructions

You can smoke test if episode cover corners are rounded correctly. But code review should be enough.

## Checklist
 
- [ ] with different themes


mebarbosa (approved):
LGTM 💯

Issue #2047, opened 2024-04-12
Sleep Timer: End of episode option is not being preserved when switching to another episode from Up Next


mebarbosa:
### Description

If we select the 'End of Episode' option from the 'Sleep Timer' and then switch to another episode from the Up Next list, this sleep timer configuration should be retained for the newly selected episode, but if we switch to another episode the app removes the sleep timer configuration

### Step-by-step reproduction instructions

1. Add at least two episodes to Up Next
2. Have `Play up next episode on tap` enabled
3. Set `End of Episode` in `Sleep timer`
4. Open the Sleep timer again
5. You will see the `End of episode` option was set ✅
6. Open Up next queue
7. Tap on another episode
8. Open the Sleep timer again
9. You will see the `End of episode` option is not set ❌

### Screenshots or screen recording

[Screen_recording_20240412_151938.webm](https://github.com/Automattic/pocket-casts-android/assets/42220351/25328d89-ed9b-4d24-a1f6-a7bbccd0b205)

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

Issue #2045, opened 2024-04-12
Play/Pause button is not getting updated in mini player after Sleep time finishes


mebarbosa:
### Description

When the episode pauses because of sleep time, the play/pause button is not getting updated in mini player. 

### Step-by-step reproduction instructions

1. Play an episode
2. Set a time in Sleep Timer
3. Close the now playing
4. Wait until the episode is paused
5. Notice that the play/pause button does not update after the episode is paused ❌

### Screenshots or screen recording

[Screen_recording_20240412_065642.webm](https://github.com/Automattic/pocket-casts-android/assets/42220351/3a34c8b0-ab98-49cc-bd15-513099eaae93)

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

mebarbosa:
Closing this issue because I can't reproduce anymore in `7.62-rc-3`

PR #2048, opened 2024-04-12
Sleep Timer: automatically restart if playback is resumed within 5 minutes

mebarbosa:
## Description
This PR implements a mechanism to restart the sleep timer whenever the user restarts playback within 5 minutes.
- The sleep timer will restart with the same configuration as before same time initially set or end of episode
- If playing after more than 5 minutes, the sleep timer should not restart

While I was working on this, I have found two issues that I reported here:
- https://github.com/Automattic/pocket-casts-android/issues/2045
- https://github.com/Automattic/pocket-casts-android/issues/2047

### Restarting timer for a given amount of time

1. Run the app
2. Add 5 to 10 podcasts to your Up Next
3. Play something
4. Open the full player, tap the sleep timer icon (zZz), select X minutes (You can select 1 minute using the last option)
5. After X minutes the playback should stop
6. Tap play again
7. ✅ The sleep timer should restart with X minutes you set
8. ✅ Ensure `🔵 Tracked: player_sleep_timer_restarted ["time": [time here]]` is tracked

### Not restarting timer
- You can change the value of `SleepTimer.MIN_TIME_TO_RESTART_SLEEP_TIMER_IN_MINUTES` to 1 so you don't need to wait for 5 minutes

1. Play something
4. Open the full player, tap the sleep timer icon (zZz), select X minutes (You can select 1 minute using the last option)
5. After X minutes the playback should stop
6. Wait for 5 minutes. (or the time you set)
7. Tap play again
9. ✅ The sleep timer should not restart

### Restarting timer for end of episode

1. Play something
4. Open the full player, tap the sleep timer icon (zZz), select end of episode
5. Drag the scrubber close to the end of the episode and wait till it's finished
7. Tap play again
8. ✅ The next episode should play and sleep timer should restart for end of episode
9. ✅ Ensure `🔵 Tracked: player_sleep_timer_restarted ["time": "end_of_episode"]` is tracked

### Canceling

1. Play something
4. Open the full player, tap the sleep timer icon (zZz), select any timer
5. Open the timer sheet again and cancel the timer
7. Pause and play again
9. ✅ The timer should not restart

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SleepTimer.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SleepTimer.kt
@@ -30,7 +30,7 @@ class SleepTimer @Inject constructor(
 
     private var sleepTimeMs: Long? = null
     private var lastSleepAfterTimeInMinutes: Int? = null
-    private var lastTimeHasFinishedInMillis: Long = 0
+    private var lastTimeHasFinishedInMillis: Long? = null
     private var lastEpisodeUuidAutomaticEnded: String? = null
 
     fun sleepAfter(minutes: Int, onSuccess: () -> Unit) {
@@ -57,16 +57,18 @@ class SleepTimer @Inject constructor(
         createAlarm(time.timeInMillis)
     }
     fun restartSleepTimerIfApplies(currentEpisodeUuid: String, isSleepTimerRunning: Boolean, onRestartSleepAfterTime: () -> Unit, onRestartSleepOnEpisodeEnd: () -> Unit) {
-        val diffTimeInMillis = System.currentTimeMillis() - lastTimeHasFinishedInMillis
-        val diffInMinutes = diffTimeInMillis / (1000 * 60) // Convert to minutes
+        lastTimeHasFinishedInMillis?.let { lastTimeHasFinished ->
+            val diffTimeInMillis = System.currentTimeMillis() - lastTimeHasFinished
+            val diffInMinutes = diffTimeInMillis / (1000 * 60) // Convert to minutes
 
-        if (diffInMinutes < MIN_TIME_TO_RESTART_SLEEP_TIMER_IN_MINUTES && !lastEpisodeUuidAutomaticEnded.isNullOrEmpty() && currentEpisodeUuid != lastEpisodeUuidAutomaticEnded) {
-            onRestartSleepOnEpisodeEnd()
-            analyticsTracker.track(PLAYER_SLEEP_TIMER_RESTARTED, mapOf(TIME_KEY to END_OF_EPISODE_VALUE))
-        } else if (diffInMinutes < MIN_TIME_TO_RESTART_SLEEP_TIMER_IN_MINUTES && lastSleepAfterTimeInMinutes != null && !isSleepTimerRunning) {
-            lastSleepAfterTimeInMinutes?.let {
-                analyticsTracker.track(PLAYER_SLEEP_TIMER_RESTARTED, mapOf(TIME_KEY to it * 60)) // Convert to seconds
-                sleepAfter(it, onRestartSleepAfterTime)
+            if (diffInMinutes < MIN_TIME_TO_RESTART_SLEEP_TIMER_IN_MINUTES && !lastEpisodeUuidAutomaticEnded.isNullOrEmpty() && currentEpisodeUuid != lastEpisodeUuidAutomaticEnded) {
+                onRestartSleepOnEpisodeEnd()
+                analyticsTracker.track(PLAYER_SLEEP_TIMER_RESTARTED, mapOf(TIME_KEY to END_OF_EPISODE_VALUE))
+            } else if (diffInMinutes < MIN_TIME_TO_RESTART_SLEEP_TIMER_IN_MINUTES && lastSleepAfterTimeInMinutes != null && !isSleepTimerRunning) {
+                lastSleepAfterTimeInMinutes?.let {
+                    analyticsTracker.track(PLAYER_SLEEP_TIMER_RESTARTED, mapOf(TIME_KEY to it * 60)) // Convert to seconds
+                    sleepAfter(it, onRestartSleepAfterTime)
+                }
             }
         }
     }
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SleepTimer.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SleepTimer.kt
@@ -20,6 +20,9 @@ class SleepTimer @Inject constructor(@ApplicationContext private val context: Co
 
     companion object {
         private var sleepTimeMs: Long? = null
+        private var lastSleepAfterTime: Int = 0
+        private var lastTimeHasFinished: Long = 0
+        private const val MIN_TIME_TO_RESTART_SLEEP_TIMER_IN_MINUTES = 5
     }
 
     fun sleepAfter(mins: Int, onSuccess: () -> Unit) {
@@ -28,6 +31,7 @@ class SleepTimer @Inject constructor(@ApplicationContext private val context: Co
             add(Calendar.MINUTE, mins)
         }
         if (createAlarm(time.timeInMillis)) {
+            lastSleepAfterTime = mins
             onSuccess()
         }
     }
@@ -42,6 +46,16 @@ class SleepTimer @Inject constructor(@ApplicationContext private val context: Co
             add(Calendar.MINUTE, mins)
         }
         createAlarm(time.timeInMillis)
+    }
+    fun restartSleepTimerIfApplies(onSuccess: () -> Unit) {
+        val diffTimeInMillis = System.currentTimeMillis() - lastTimeHasFinished
+        val diffInMinutes = diffTimeInMillis / (1000 * 60) // Convert to minutes
+
+        if (diffInMinutes < MIN_TIME_TO_RESTART_SLEEP_TIMER_IN_MINUTES && lastSleepAfterTime != 0) {
+            sleepAfter(lastSleepAfterTime, onSuccess)
+        } else {
+            cancelAutomaticSleepTimerRestart()
+        }
     }
 
     private fun createAlarm(timeMs: Long): Boolean {
@@ -61,6 +75,7 @@ class SleepTimer @Inject constructor(@ApplicationContext private val context: Co
             return try {
                 alarmManager.setExactAndAllowWhileIdle(AlarmManager.RTC_WAKEUP, timeMs, sleepIntent)
                 sleepTimeMs = timeMs
+                lastTimeHasFinished = timeMs
                 true
             } catch (e: Exception) {
                 LogBuffer.e(LogBuffer.TAG_CRASH, e, "Unable to start sleep timer.")
@@ -71,7 +86,7 @@ class SleepTimer @Inject constructor(@ApplicationContext private val context: Co
 
     fun cancelTimer() {
         getAlarmManager().cancel(getSleepIntent())
-        sleepTimeMs = null
+        cleanUpSleepTimer()
     }
 
     val isRunning: Boolean
@@ -82,7 +97,7 @@ class SleepTimer @Inject constructor(@ApplicationContext private val context: Co
 
         val timeLeft = sleepTimeMs - System.currentTimeMillis()
         if (timeLeft < 0) {
-            SleepTimer.sleepTimeMs = null
+            cleanUpSleepTimer()
             return null
         }
         return (timeLeft / DateUtils.SECOND_IN_MILLIS).toInt()
@@ -95,5 +110,14 @@ class SleepTimer @Inject constructor(@ApplicationContext private val context: Co
 
     private fun getAlarmManager(): AlarmManager {
         return context.getSystemService(Context.ALARM_SERVICE) as AlarmManager
+    }
+
+    private fun cleanUpSleepTimer() {
+        sleepTimeMs = null
+        cancelAutomaticSleepTimerRestart()
+    }
+    private fun cancelAutomaticSleepTimerRestart() {
+        lastSleepAfterTime = 0
+        lastTimeHasFinished = 0
     }
 }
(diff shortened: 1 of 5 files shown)


CookieyedCodes:
I personally would find 5 mins a tad long, personally i think it should be between 90 seconds or a maximum of 3 mins but if it's a factor of accurate clocks then fair, I would potentially also have a toast because a user might be unaware of the functionality & that would be bad 🤔

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SleepTimer.kt
```diff
@@ -16,10 +18,19 @@ import javax.inject.Singleton
...
+class SleepTimer @Inject constructor(
+    private val analyticsTracker: AnalyticsTrackerWrapper,
+    @ApplicationContext private val context: Context,
+) {
 
     companion object {
         private var sleepTimeMs: Long? = null
+        private var lastSleepAfterTimeInMinutes: Int = 0
+        private var lastTimeHasFinishedInMillis: Long = 0
+        private var lastEpisodeUuidAutomaticEnded: String? = null
```
MiSikora, line 30: Can we move this properties out of `companion object`? For now it would be only a cosmetic change but it isn't good to keep state in global static objects. And `SleepTimer` is marked with `@Singleton` ensuring that Dagger will provide only single instance of it.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SleepTimer.kt
```diff
@@ -16,10 +18,19 @@ import javax.inject.Singleton
...
-class SleepTimer @Inject constructor(@ApplicationContext private val context: Context) {
+class SleepTimer @Inject constructor(
+    private val analyticsTracker: AnalyticsTrackerWrapper,
+    @ApplicationContext private val context: Context,
+) {
 
     companion object {
         private var sleepTimeMs: Long? = null
+        private var lastSleepAfterTimeInMinutes: Int = 0
+        private var lastTimeHasFinishedInMillis: Long = 0
```
MiSikora, line 29: Instead of juggling conversions between milliseconds and minutes which is very error prone could you change these properties to hold their values as `Duration` type?

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SleepTimer.kt
```diff
@@ -16,10 +18,19 @@ import javax.inject.Singleton
...
+        private var lastEpisodeUuidAutomaticEnded: String? = null
+        private const val MIN_TIME_TO_RESTART_SLEEP_TIMER_IN_MINUTES = 5
```
MiSikora, line 31: This also could be kept as a `Duration` type.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SleepTimer.kt
MiSikora, line 29: Having these values not-nullable is ambiguous. It's not obvious whether `0` means restart immediately or, like in the current implementation, never restart.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SleepTimer.kt
```diff
@@ -43,6 +56,23 @@ class SleepTimer @Inject constructor(@ApplicationContext private val context: Co
         }
         createAlarm(time.timeInMillis)
     }
+    suspend fun restartSleepTimerIfApplies(currentEpisodeUuid: String, isSleepTimerRunning: Boolean, onRestartSleepAfterTime: () -> Unit, onRestartSleepOnEpisodeEnd: suspend () -> Unit) {
```
MiSikora, line 59: Unless I'm missing something I don't think that this function and `onRestartSleepOnEpisodeEnd` parameter need to be suspending.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SleepTimer.kt
```diff
@@ -43,6 +56,23 @@ class SleepTimer @Inject constructor(@ApplicationContext private val context: Co
...
+
+        if (diffInMinutes < MIN_TIME_TO_RESTART_SLEEP_TIMER_IN_MINUTES && !lastEpisodeUuidAutomaticEnded.isNullOrEmpty() && currentEpisodeUuid != lastEpisodeUuidAutomaticEnded) {
+            onRestartSleepOnEpisodeEnd()
+            analyticsTracker.track(PLAYER_SLEEP_TIMER_RESTARTED, mapOf(TIME_KEY to END_OF_EPISODE_VALUE))
+        } else if (diffInMinutes < MIN_TIME_TO_RESTART_SLEEP_TIMER_IN_MINUTES && lastSleepAfterTimeInMinutes != 0 && !isSleepTimerRunning) {
+            analyticsTracker.track(PLAYER_SLEEP_TIMER_RESTARTED, mapOf(TIME_KEY to lastSleepAfterTimeInMinutes * 60)) // Convert to seconds
+            sleepAfter(lastSleepAfterTimeInMinutes, onRestartSleepAfterTime)
+        }
+    }
+    fun sleepTimerWasTriggeredForEndOfEpisode(uuid: String?) {
```
MiSikora, line 71: It's a bit hard for me to read the name of this function clearly. Could you rename it something simpler? Even `setEndOfEpisodeUuid()` would be enough I believe.

Also, this function shouldn't accent nullable uuid.

MiSikora (commented):
Nice work! Feature works as described. I've tried breaking it by kill app process and leaving only media notification but everything worked. I left some comments regarding design of the class that would make it easier to maintain in the long-term.

Also, a changelog entry is missing.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SleepTimer.kt
mebarbosa: No need to use `suspend` anymore. When I was creating this function, in one of the implementations options, I have to use `suspend`. Nice catch! 

I will remove it

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SleepTimer.kt
mebarbosa: I agree. [721153b](https://github.com/Automattic/pocket-casts-android/pull/2048/commits/721153bdc98495e648717e05f994416fcd0b44d3)

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-android/milestone/51">7.62</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

mebarbosa:
> Nice work! Feature works as described. I've tried breaking it by kill app process and leaving only media notification but everything worked. I left some comments regarding design of the class that would make it easier to maintain in the long-term.
> 
> Also, a changelog entry is missing.

Hey @MiSikora Thanks for your review! I've just addressed all changes

mebarbosa:
> I personally would find 5 mins a tad long, personally i think it should be between 90 seconds or a maximum of 3 mins but if it's a factor of accurate clocks then fair, I would potentially also have a toast with a button to cancel because a user might be unaware of the functionality & that would be quite bad in my opinion 🤔

@CookieyedCodes Thank you for share your feedback! I will share this with our team. I agree that we should at least let the user knows about it

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SleepTimer.kt
```diff
@@ -8,41 +8,78 @@ import android.os.Build
...
+        lastTimeSleepTimeHasFinished?.let { lastTimeHasFinished ->
+            val diffTimeInMillis = System.currentTimeMillis() - lastTimeHasFinished.inWholeMilliseconds
+            val diffInMinutes = diffTimeInMillis / (1000 * 60) // Convert to minutes
+
+            if (diffInMinutes < MIN_TIME_TO_RESTART_SLEEP_TIMER_IN_MINUTES && !lastEpisodeUuidAutomaticEnded.isNullOrEmpty() && currentEpisodeUuid != lastEpisodeUuidAutomaticEnded) {
+                onRestartSleepOnEpisodeEnd()
+                analyticsTracker.track(PLAYER_SLEEP_TIMER_RESTARTED, mapOf(TIME_KEY to END_OF_EPISODE_VALUE))
+            } else if (diffInMinutes < MIN_TIME_TO_RESTART_SLEEP_TIMER_IN_MINUTES && lastSleepAfterTime != null && !isSleepTimerRunning) {
+                lastSleepAfterTime?.let {
+                    analyticsTracker.track(PLAYER_SLEEP_TIMER_RESTARTED, mapOf(TIME_KEY to it * 60)) // Convert to seconds
```
MiSikora, line 72: AFTER converting `lastSleepAfterTime` to `Duration` this needs to be changed.

```kotlin
analyticsTracker.track(PLAYER_SLEEP_TIMER_RESTARTED, mapOf(TIME_KEY to it.inWholeSeconds))
```

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SleepTimer.kt
```diff
@@ -8,41 +8,78 @@ import android.os.Build
...
+                    sleepAfter(it.inWholeMinutes.toInt(), onRestartSleepAfterTime)
+                }
+            }
+        }
+    }
+    fun setEndOfEpisodeUuid(uuid: String) {
+        lastEpisodeUuidAutomaticEnded = uuid
+        lastTimeSleepTimeHasFinished = System.currentTimeMillis().toDuration(DurationUnit.MILLISECONDS)
```
MiSikora, line 80: [Kotlin provides](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.time/-duration/#companion-object-properties) useful extensions for these conversions.

```kotlin
1.days
10.seconds
378.milliseconds
```

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SleepTimer.kt
```diff
@@ -8,41 +8,78 @@ import android.os.Build
...
+                analyticsTracker.track(PLAYER_SLEEP_TIMER_RESTARTED, mapOf(TIME_KEY to END_OF_EPISODE_VALUE))
+            } else if (diffInMinutes < MIN_TIME_TO_RESTART_SLEEP_TIMER_IN_MINUTES && lastSleepAfterTime != null && !isSleepTimerRunning) {
+                lastSleepAfterTime?.let {
+                    analyticsTracker.track(PLAYER_SLEEP_TIMER_RESTARTED, mapOf(TIME_KEY to it * 60)) // Convert to seconds
+                    sleepAfter(it.inWholeMinutes.toInt(), onRestartSleepAfterTime)
+                }
+            }
+        }
+    }
+    fun setEndOfEpisodeUuid(uuid: String) {
```
MiSikora, line 78: A new line would be helpful here.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SleepTimer.kt
```diff
@@ -8,41 +8,78 @@ import android.os.Build
...
             return
         }
         val time = Calendar.getInstance().apply {
             timeInMillis = currentTimeMs
-            add(Calendar.MINUTE, mins)
+            add(Calendar.MINUTE, minutes)
         }
         createAlarm(time.timeInMillis)
     }
+    fun restartSleepTimerIfApplies(currentEpisodeUuid: String, isSleepTimerRunning: Boolean, onRestartSleepAfterTime: () -> Unit, onRestartSleepOnEpisodeEnd: () -> Unit) {
```
MiSikora, line 62: Also a new line to separate this function for readability would be nice.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SleepTimer.kt
```diff
@@ -96,4 +133,16 @@ class SleepTimer @Inject constructor(@ApplicationContext private val context: Co
...
+
+    private fun cleanUpSleepTimer() {
+        sleepTimeMs = null
+        cancelAutomaticSleepAfterTimeRestart()
+        cancelAutomaticSleepOnEpisodeEndRestart()
+    }
+    private fun cancelAutomaticSleepAfterTimeRestart() {
+        lastSleepAfterTime = null
+    }
+    private fun cancelAutomaticSleepOnEpisodeEndRestart() {
```
MiSikora, line 145: Some empty space formatting is missing here as well.

MiSikora (approved):
LGTM! Left one formatting nitpick.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SleepTimer.kt
mebarbosa: ✅

I thought `./gradlew spotlessApply` would handle cases like this.

MiSikora: Yeah, I'm surprised it doesn't complain about it. Maybe something to configure.

PR #2065, opened 2024-04-16
Make widget texts non scaling

MiSikora:
## Description

Using scaling units for fonts on widgets isn't very practical since we don't have a good control over UI. This PR introduces `NonScalingText` widget UI element. It delegates to XML layout that is then embedded inside Glance. This allows us to control typical properties of TextView like text size which can now be expressed in `dp` instead of `sp`. Because of this change I had to do some workaround to enable proper theming on our texts. However, this also enabled changing alpha.

### Text size

1. Place medium and large widgets on the home screen.
2. Change system font to large or tiny size.
3. Text size on widgets should stay the same regardless of the system settings.
4. Test this with different queue combinations. Single episode, multiple episodes, no episodes.

### Text color

1. Place medium and large widgets on the home screen.
2. Have something in the queue.
3. Test if texts use correct colors based on dark/light mode and based on `Use dynamic colors for widget` setting. Try different combinations of those.

### Small widget

Smoke test small widget and if it looks correctly. There's been a small fix for playback button computation.

## Checklist
 
- [x] with different themes


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-android/milestone/51">7.62 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

ashiagr:
Text sizes remain the same, but margins do not look right when the display size is increased: 

Small font | Large font
-----|-----
 | 

Can we do something about it?

MiSikora:
Hmmm… I'm not sure. I'll poke around and see.

MiSikora:
I think this is issue of widget not fitting described here: pdzxoO-oM-p2

I added one more fix though for playback controls clipping to the logo position when widget was scaled.

| Before | After |
| - | - |
| Screenshot_20240417-103110 | Screenshot_20240417-103358 |

ashiagr:
> I think this is issue of widget not fitting described here - pdzxoO-oM-p2 - and you should be able to adjust the size of the widget.

It does look like a limitation. I'll go ahead and approve the PR.

PR #2079, opened 2024-04-17
Sleep Timer: Play confirmation sound when restarting after shaking the phone

mebarbosa:
## Description
- This adds a confirmation sound for when we restart the sleep time after shaking phone and the device was in background
- Update track to match with [iOS](https://github.com/Automattic/pocket-casts-ios/pull/1627#discussion_r1569376086).

### Restarting timer after shaking phone in foreground
1. Run the app
2. Play something
3. Open the full player, tap the sleep timer icon (zZz), select X minutes.
4. Keep the app open
5. Shake your device
6. ✅ You should see the toast with `Sleep timer restarted after device shake` message

### Restarting timer after shaking phone in background
1. Run the app
2. Play something
3. Open the full player, tap the sleep timer icon (zZz), select X minutes
4. Put your device in background
5. Shake your device
6. ✅ You should hear a confirmation sound and the sleep timer restarted

## Checklist
 
- [ ] with different themes


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SleepTimerRestartWhenShakingDevice.kt
```diff
@@ -56,23 +62,37 @@ class SleepTimerRestartWhenShakingDevice @Inject constructor(
...
+    private fun trackSleepTimeRestart(time: Duration) {
+        val currentTime = System.currentTimeMillis().milliseconds
+        val elapsedTime: Duration? = lastTrackTime?.let { currentTime - it }
+        if (elapsedTime == null || elapsedTime >= 3.seconds) { // Make sure we don't send the same report in 3 seconds
+            analyticsTracker.track(
+                PLAYER_SLEEP_TIMER_RESTARTED,
+                mapOf(TIME_KEY to time.inWholeSeconds, REASON_KEY to DEVICE_SHAKE_VALUE),
+            )
+            lastTrackTime = currentTime
+        }
```
mebarbosa, line 90: I implemented this because I noticed we were reporting 10ish times per restart because of the device shaken detection

MiSikora (approved):

PR #2082, opened 2024-04-18
Split episodes update into smaller batches

MiSikora:
## Description

When app requests to refresh podcasts and a user has a lot of them a call to `https://refresh.pocketcasts.com/user/update` can time out because there is so much to process on the server side. There are two potential solutions:

1. Extend timeout for this particular call from our default 1 minute to something else.
2. Split update into small calls and combine them on the client side.

I decided to go with the second solution because it's hard to determine a correct timeout as number of subscribed podcasts does not have a limit. I split into batches of 200 per request. During my tests such a split over 4000 podcasts took about 150 seconds on average.

I do not limit concurrency of the updates on my own. I delegate everything to OkHttp which has it's own parallelism limit.

Update fails and cancels all calls if any of the calls fail. I don't do partial updates. Mainly due to regression concerns. Previous behavior allowed only for complete updates and I want to keep that behavior.

See p1713111918959379-slack-C02A333D8LQ for more context.

## Testing Instructions

1. Have a clean app installation.
2. Sign in with an account that has A LOT of podcasts. You can use account 11930 from the Secret Store. It is subscribed to 4000 podcasts. 
3. Wait until app is synchronized and all 4000 podcasts are added. It will take some time (like 10+ minutes). You can observe podcast count on the Profile page.
4. Go to Podcasts page.
5. Pull to refresh. If you see in the logs `Not refreshing as internet not connected` close the app, reopen it, and pull again. There is a separate bug that network connection might not be correctly detected.
6. Observe in the logs that there are multiple requests issued to `https://refresh.pocketcasts.com/user/update`. After some time (couple of minutes) it should succeed and you should be able to access episodes. I had about `750_000` of them in the database after refreshing.

## Checklist
 
- [ ] with different themes


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

MiSikora:
I added `do not merge` label to wait for the feedback from server folks. I don't think we could have a different solution for now but maybe I'm missing something.

MiSikora:
@Automattic/pocket-casts-server Is this an acceptable solution or should we look somewhere different?

ashiagr (commented):
I understand the PR does not address the initial sync, but it took a lot of time and the connection was reset too. I have not been able to sync the account fully.  

I'll try it again and share my findings.

ashiagr (approved):

MiSikora:
Discussed here: p1713783561299969-slack-C029BMW150R

Issue #2085, opened 2024-04-18
Podcast Artwork flashes


joashrajin:
### Description

On certain podcasts, the artwork flashes every 3-5 seconds

### Step-by-step reproduction instructions

1. Go to an episode like this: https://pca.st/episode/90659d04-2a3b-4b12-b641-beb75dc0326a
2. Enable 'Use Episode Artwork' in Settings > Appearance

### Screenshots or screen recording

https://github.com/Automattic/pocket-casts-android/assets/16253818/c6d5fd07-1ce0-41b7-b264-057e4cfac45f

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

Fold 4, Android 14, Pocket Casts v7.62-rc-3

8059216-zd-a8c

PR #2086, opened 2024-04-18
Check episode ID instead of episode before loading artwork

MiSikora:
## Description

We were checking episode instead of episode ID before loading a new artwork. I'm not sure why it failed only when `Use episode artwork` was enabled. I added a similar check for loading chapters as well. 

Closes: #2085

## Testing Instructions

1. Enable `Use episode artwork` setting.
2. Play an episode like [this one](https://pca.st/episode/90659d04-2a3b-4b12-b641-beb75dc0326a).
3. Open the player view.
4. Artwork should not flash.
5. Minimize the player.
6. Artwork on the mini player should not flash.

---

1. Smoke test that chapters artwork is displayed with podcast like [this one](https://pca.st/upgrade).

## Checklist
 
- [ ] with different themes


ashiagr (approved):
LGTM

PR #2097, opened 2024-04-22
Sleep Timer: Sleep time by episodes

mebarbosa:
## Description
- Implements the ability to set more than one end of episode for sleep time
- Deletes `player_sleep_when_episode_ends` string that we are not using anymore
- Updates the existing `player_sleep_timer_enabled`, `player_sleep_timer_extended ` and `player_sleep_timer_restarted ` track events to include the property `number_of_episodes` that will have the number of episodes set for sleep time end of episode

### End of Episode + Automatic restart
1. Run the app
2. Add 5/6 episodes to Up next
3. Open the full player, tap the sleep timer icon (zZz)
4. Click on plus button to set 2 episodes
5. ✅ Ensure 🔵 Tracked: `player_sleep_timer_enabled, Properties: {"time":"end_of_episode","number_of_episodes":2` is tracked
6. Tap the sleep timer icon (zZz) again
7.  ✅  You should see `Sleeping in 2 episodes` message and the cancel button
8. Scroll the episode to almost the end
9. Wait for the episode to finish
10. ✅ The next episode should start
11. Tap the sleep timer icon (zZz) again
12. ✅  You should see `Sleeping in 1 episode` message and the cancel button
13. Scroll the episode to almost the end
14. Wait for the episode to finish
15.  ✅ When the episode finishes, it should be stoped
16. Tap on `Play` button within 5 minutes
17. ✅ Automatic sleep time should be triggered with 2 episodes set
18. ✅  Ensure 🔵 Tracked: `player_sleep_timer_restarted, Properties: {"time":"end_of_episode","number_of_episodes":2` is tracked

### Smoke test
You can smoke test the sleep time feature. Make sure you the cancel button still works, you can also play with sleep after x minutes, etc

## Screenshots or Screencast 
https://github.com/Automattic/pocket-casts-android/assets/42220351/db9a66b7-0b24-411f-b5a1-a84ae24db1a9

## Checklist
 
- [ ] with different themes


CookieyedCodes:
@mebarbosa when I was doing the mockup when this was a personal project I kept the end of episode as an option because 

1. Not to disrupt existing muscle memory 
2. To keep a quick option for end of episode if the user wanted it, mainly because I don't want to constantly be adjusting 
3. To distinguished the timer spefically being an after episode timer mainly to make sure their was no confusion
4. To make it easier for the preset manager 
Ignore this if you want I don't mind

but these other things I did was extending the timer on the cancel timer screen was to add odd & even episodes, I also added an option so the number the user inputs into the timer would be passed through as an option these would be nice enhancements.
(Tho the latter is slightly less required because of the repeat timer functionality I still wouldn't mind the option tbh)

https://github.com/Automattic/pocket-casts-android/discussions/1452#discussioncomment-7422816

modules/services/localization/src/main/res/values/strings.xml
```diff
@@ -308,10 +308,13 @@
     <string name="player_sleep_custom_minus">Custom minus</string>
     <string name="player_sleep_custom_plus">Custom plus</string>
     <string name="player_sleep_end_of_episode">End of episode</string>
+    <string name="player_sleep_end_of_episode_singular">End of 1 episode</string>
+    <string name="player_sleep_end_of_episode_plural">End of %d episodes</string>
     <string name="player_sleep_timer_title">Sleep Timer</string>
     <string name="player_sleep_timer">Sleep timer</string>
     <string name="player_sleep_timer_start_failed">Unable to start timer. Please allow \'Alarms &amp; reminders\' permission.</string>
-    <string name="player_sleep_when_episode_ends">Sleeping when episode ends</string>
```
ashiagr, line 314: Nitpick: Just removing a string from this main strings file is sufficient. It will get automatically removed from other translations when they're updated.

ashiagr:
@CookieyedCodes 👋 

We went with this design for the end-of-episode timer as it is more flexible for updating episode count without adding too many elements on the screen. Tapping the `End of 1 episode` should retain existing behavior. 

There are some nice suggestions in your mockups. We'll collect more feedback and continue to improve the functionality. Thanks so much for the mockups!  

/cc @david-gonzalez-a8c

ashiagr (approved):
Works as described! I just left a comment regarding translations.

modules/services/localization/src/main/res/values/strings.xml
mebarbosa: hummm... Got it! Thanks. I did not know that

PR #2111, opened 2024-04-25
Store episode artwork granular settings

MiSikora:
## Description

This PR is only code changes. It doesn't affect the user. It migrates boolean flag for episode artwork to a setting that allows disabling episode artwork in particular contexts:

* Filters
* UpNext
* Downloads
* FIles
* Starred
* Bookmarks
* Listening History

## Testing Instructions

1. Install the app from 4a41c33bd1db36a823a2ea2605dfd3fc7aca6c51.
2. Enable the `Use episode artwork setting`.
3. Install the app from this PR.
4. Check that the `Use episode artwork setting` is still enabled.
5. Smoke test the episode artwork shows on different screens.
6. Disable the `Use episode artwork setting` setting.
7. Smoke test that the episode artwork does not show on different screens.

## Checklist
 
- [ ] with different themes


ashiagr (approved):
LGTM

PR #2140, opened 2024-04-30
Select min dimension for small widget

MiSikora:
## Description

Some launchers allow to resize widgets regardless of the configuration. This can lead to unexpected results for the small widget.

I also started using size for fetching artwork to account for big widget sizes.

## Testing Instructions

1. Install Nova Launcher.
2. Add the small widget to the Nova's home screen and play something.
3. Resize the widget and check that it looks correctly.

## Screenshots or Screencast 

| Before | After |
| - | - |
| 1-before | 1-after |

| Before | After |
| - | - |
| 2-before | 2-after |

| Before | After |
| - | - |
| 3-before | 3-after |

## Checklist

- [ ] with different themes

### DIFF
--- a/modules/features/widgets/src/main/kotlin/au/com/shiftyjelly/pocketcasts/widget/ui/EpisodeImage.kt
+++ b/modules/features/widgets/src/main/kotlin/au/com/shiftyjelly/pocketcasts/widget/ui/EpisodeImage.kt
@@ -33,7 +33,7 @@ internal fun EpisodeImage(
     backgroundColor: ((WidgetTheme) -> ColorProvider)? = null,
     onClick: Action? = null,
 ) {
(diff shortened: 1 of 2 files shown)


CookieyedCodes:
🤔🤔, could we get the play pause icon to be in the corner of the 1x1 icon I kind of want to see the artwork 😅

MiSikora:
I'll pass your feedback.

CookieyedCodes:
Yes I'm aware, still would prefer a smaller indicator

ashiagr:
I can see that the artwork is resized properly. If we make the play/ pause grey background more transparent, we can see artwork through it.

ashiagr (approved):
Got you! 

Limiting button size serves the purpose. Thanks.

PR #2164, opened 2024-05-06
Dispatch refresh state when refresh task is throttled

### LINKED ISSUE #2153: The podcasts are taking too long to load when refreshing the Podcasts tab
### Description

We're experiencing infinite loading when refreshing podcasts in the Podcasts tab. I was only able to reproduce this issue after adding cloud files (videos) to my account, and I had around 100 podcast subscriptions.

I can't reproduce this consistently every time.

1. Have an account with around 100 podcast subscriptions and some cloud videos uploaded to the account
2. Run the app
3. Go to Podcasts tab
4. Refresh the podcast list
5. The first refresh should work as expected
6. After finishes the refresh, try to refresh it again
7. ❌ You will notice the refresh gets stuck or takes loo long to finish

https://github.com/Automattic/pocket-casts-android/assets/42220351/493cc360-c69e-4331-9574-0d4c976a8876

Pocket Casts `7.63-rc-3`

MiSikora:
## Description

This PR addresses an issue with our manual refresh task, which implements a 15-second throttle in pull-to-refresh scenarios. Previously, we did not dispatch a "refreshed" state upon task completion due to this throttling mechanism. Additionally, throttling is disabled in debug builds, rendering this bug non-reproducible in those environments.

I have prepared a patch that enables throttling in debug builds to facilitate testing of this fix.
Closes #2153

## Testing Instructions

1. Apply [this patch](https://github.com/Automattic/pocket-casts-android/files/15219568/refresh.patch).
2. Install the app.
3. Go to the podcasts tab.
4. Pull to refresh and wait for the sync to finish.
5. Pull to refresh again.
6. Spinner should disappear.

## Checklist
 
- [x] ~with different themes~

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/refresh/RefreshPodcastsTask.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/refresh/RefreshPodcastsTask.kt
@@ -103,9 +103,6 @@ class RefreshPodcastsTask @AssistedInject constructor(
                     applicationScope = applicationScope,
                     runNow = true,
                 )
-                if (!refreshThread.isAllowedToRun(runNow = true)) {
-                    return@withContext
-                }
                 refreshJob = async { refreshThread.run() }
             }
 
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/refresh/RefreshPodcastsThread.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/refresh/RefreshPodcastsThread.kt
@@ -275,7 +275,7 @@ class RefreshPodcastsThread(
 
     private fun dispatchCurrentRefreshedState() {
         val settings = getEntryPoint().settings()
-        (settings.getLastSuccessRefreshState() ?: RefreshState.Never).let(settings::setRefreshState)
+        settings.setRefreshState(settings.getLastSuccessRefreshState() ?: RefreshState.Never)
     }
 
     private fun updatePodcasts(result: RefreshResponse?): List<String> {
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/refresh/RefreshPodcastsThread.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/refresh/RefreshPodcastsThread.kt
@@ -55,6 +55,8 @@ import dagger.hilt.EntryPoints
 import dagger.hilt.InstallIn
 import dagger.hilt.components.SingletonComponent
 import java.util.Date
+import kotlin.time.Duration.Companion.minutes
+import kotlin.time.Duration.Companion.seconds
 import kotlinx.coroutines.CoroutineScope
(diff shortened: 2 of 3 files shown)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/refresh/RefreshPodcastsTask.kt
```diff
@@ -103,9 +103,6 @@ class RefreshPodcastsTask @AssistedInject constructor(
                     applicationScope = applicationScope,
                     runNow = true,
                 )
-                if (!refreshThread.isAllowedToRun(runNow = true)) {
-                    return@withContext
-                }
```
MiSikora, line 108: This isn't needed here as `RefreshPodcastsThread` already handles this scenario.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/refresh/RefreshPodcastsThread.kt
```diff
@@ -270,6 +273,11 @@ class RefreshPodcastsThread(
         getEntryPoint().settings().setRefreshState(RefreshState.Failed(message))
     }
 
+    private fun dispatchCurrentRefreshedState() {
+        val settings = getEntryPoint().settings()
+        (settings.getLastSuccessRefreshState() ?: RefreshState.Never).let(settings::setRefreshState)
```
MiSikora, line 278: There is an edge case where we lose failure state due to this but I don't see any way around it without rewriting a lot of code. I think this is an acceptable compromise.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/refresh/RefreshPodcastsThread.kt
```diff
@@ -364,8 +372,8 @@ class RefreshPodcastsThread(
 
         private const val GROUP_NEW_EPISODES = "group_new_episodes"
 
-        private val THROTTLE_RUN_NOW_MS: Long = if (BuildConfig.DEBUG) 0 else 15000 // 15 seconds
-        private val THROTTLE_PERIODIC_MS: Long = if (BuildConfig.DEBUG) 0 else 5L * 60L * 1000L // 5 minutes
+        private val THROTTLE_RUN_NOW_MS: Long = if (BuildConfig.DEBUG) 0 else 15.seconds.inWholeMilliseconds
+        private val THROTTLE_PERIODIC_MS: Long = if (BuildConfig.DEBUG) 0 else 5.minutes.inWholeMilliseconds
```
geekygecko, line 376: I will try to do this in the future. It's so much cleaner.

geekygecko (approved):
Great, that worked well in testing.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/refresh/RefreshPodcastsThread.kt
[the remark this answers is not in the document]
MiSikora: No, not really. I like your version better. 👍

PR #2179, opened 2024-05-07
Introduce performance monitoring

wzieba:
## Description

This PR brings Performance Monitoring integration to the app.

## Testing Instructions
1. Apply below PATCH or manually change values
<details>
<summary>
PATCH
</summary>

```PATCH
Subject: [PATCH] Verify network state when enqueuing encrypted logs

If event is fatal (meaning: crash), do not send encrypted logs immediately because it won't likely finish. Encrypted logs will be sent on the next app start
---
Index: modules/services/crashlogging/src/main/java/au/com/shiftyjelly/pocketcasts/crashlogging/PocketCastsCrashLoggingDataProvider.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/modules/services/crashlogging/src/main/java/au/com/shiftyjelly/pocketcasts/crashlogging/PocketCastsCrashLoggingDataProvider.kt b/modules/services/crashlogging/src/main/java/au/com/shiftyjelly/pocketcasts/crashlogging/PocketCastsCrashLoggingDataProvider.kt
--- a/modules/services/crashlogging/src/main/java/au/com/shiftyjelly/pocketcasts/crashlogging/PocketCastsCrashLoggingDataProvider.kt	(revision 36a2297abb02bc9f23b2d72a079a2d633b5e9615)
+++ b/modules/services/crashlogging/src/main/java/au/com/shiftyjelly/pocketcasts/crashlogging/PocketCastsCrashLoggingDataProvider.kt	(date 1715159293195)
@@ -32,8 +32,8 @@
     override val locale: Locale = Locale.getDefault()
 
     override val performanceMonitoringConfig = PerformanceMonitoringConfig.Enabled(
-        sampleRate = 0.01,
-        profilesSampleRate = 0.01,
+        sampleRate = 1.0,
+        profilesSampleRate = 1.0,
     )
 
     override val releaseName = ReleaseName.SetByTracksLibrary
```
</details>

2. Run the app (can be in debug)
3. Assert there's a new performance transaction and profile for MainActivity in Sentry dashboard

## Screenshots or Screencast 

image

## Checklist
 
#### ~I have tested any UI changes...~

- [ ] with different themes

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/di/RepositoryProviderModule.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/di/RepositoryProviderModule.kt
@@ -35,7 +35,7 @@ class RepositoryProviderModule {
         }
         return OkHttpClient.Builder()
             .dispatcher(dispatcher)
-            .apply { networkInterceptors().addAll(networkInterceptors) }
+            .apply { networkInterceptors.forEach(::addNetworkInterceptor) }
             .connectTimeout(30, TimeUnit.SECONDS)
             .writeTimeout(30, TimeUnit.SECONDS)
             .readTimeout(30, TimeUnit.SECONDS)
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/di/RepositoryProviderModule.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/di/RepositoryProviderModule.kt
@@ -4,6 +4,7 @@ import androidx.lifecycle.LifecycleOwner
 import androidx.lifecycle.ProcessLifecycleOwner
 import au.com.shiftyjelly.pocketcasts.crashlogging.di.ProvideApplicationScope
 import au.com.shiftyjelly.pocketcasts.repositories.sync.SyncAccountManager
+import au.com.shiftyjelly.pocketcasts.servers.di.CrashLoggingInterceptor
 import au.com.shiftyjelly.pocketcasts.servers.sync.TokenHandler
 import dagger.Module
 import dagger.Provides
@@ -29,13 +30,13 @@ class RepositoryProviderModule {
     @Provides
     @Singleton
     @DownloadOkHttpClient
-    fun downloadOkHttpClient(networkInterceptors: Set<@JvmSuppressWildcards Interceptor>): OkHttpClient {
+    fun downloadOkHttpClient(@CrashLoggingInterceptor crashLoggingInterceptor: Interceptor): OkHttpClient {
         val dispatcher = Dispatcher().apply {
             maxRequestsPerHost = 5
         }
         return OkHttpClient.Builder()
             .dispatcher(dispatcher)
-            .apply { networkInterceptors.forEach(::addNetworkInterceptor) }
+            .addInterceptor(crashLoggingInterceptor)
             .connectTimeout(30, TimeUnit.SECONDS)
             .writeTimeout(30, TimeUnit.SECONDS)
             .readTimeout(30, TimeUnit.SECONDS)
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/di/RepositoryProviderModule.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/di/RepositoryProviderModule.kt
@@ -15,6 +15,7 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
 import okhttp3.Dispatcher
+import okhttp3.Interceptor
 import okhttp3.OkHttpClient
 
 @Module
@@ -28,12 +29,13 @@ class RepositoryProviderModule {
     @Provides
     @Singleton
     @DownloadOkHttpClient
-    fun downloadOkHttpClient(): OkHttpClient {
+    fun downloadOkHttpClient(networkInterceptors: Set<@JvmSuppressWildcards Interceptor>): OkHttpClient {
         val dispatcher = Dispatcher().apply {
             maxRequestsPerHost = 5
         }
         return OkHttpClient.Builder()
             .dispatcher(dispatcher)
+            .apply { networkInterceptors().addAll(networkInterceptors) }
             .connectTimeout(30, TimeUnit.SECONDS)
             .writeTimeout(30, TimeUnit.SECONDS)
             .readTimeout(30, TimeUnit.SECONDS)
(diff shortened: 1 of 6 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/di/RepositoryProviderModule.kt
```diff
@@ -28,12 +29,13 @@ class RepositoryProviderModule {
...
     @Singleton
     @DownloadOkHttpClient
-    fun downloadOkHttpClient(): OkHttpClient {
+    fun downloadOkHttpClient(networkInterceptors: Set<@JvmSuppressWildcards Interceptor>): OkHttpClient {
         val dispatcher = Dispatcher().apply {
             maxRequestsPerHost = 5
         }
         return OkHttpClient.Builder()
             .dispatcher(dispatcher)
+            .apply { networkInterceptors().addAll(networkInterceptors) }
```
geekygecko, line 38: Sorry, I know this is a draft PR but I wonder if we shouldn't bother with monitoring the network performance of episode downloads as these come from podcast producer servers and not our own. iOS does this with Firebase Performance and it adds a lot of noise to the metrics.

Screenshot 2024-05-07 at 10 24 18 PM

wzieba: That's a good point! WooCommerce Android has a similar concern when store owners connect to the stores hosted on their own hosting providers.

I agree, that it's not useful data in isolation. It is useful though in context of performance transactions. I mean instead of answering questions "which network queries are the slowest" we'll be able to answer questions "why loading podcasts is slow" - is it because of e.g. performance bottleneck of our app, or because slow http request to a 3rd party podcast server? If we don't introduce the http interceptor, we'll have "Missing instrumentation" in places when http happens and we won't be able to identify what was happening during this time.

geekygecko: I'm happy to take your advice on including them if you recommend it.

wzieba: Gotcha, thanks. To visualize my point above, having [a sample transaction](https://a8c.sentry.io/performance/trace/5b82af1a0b2547a79026fbae92ae8a2c/?node=trace-root&statsPeriod=1h&timestamp=1715158865):
Screenshot 2024-05-08 at 11 05 24

MiSikora (approved):

modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/di/ServersModule.kt
```diff
@@ -154,10 +154,15 @@ class ServersModule {
...
+        @SyncServerCache cache: Cache,
+        networkInterceptors: Set<@JvmSuppressWildcards Interceptor>,
+    ): OkHttpClient.Builder {
         var builder = OkHttpClient.Builder()
-            .addNetworkInterceptor(INTERCEPTOR_CACHE_MODIFIER)
-            .addNetworkInterceptor(INTERCEPTOR_USER_AGENT)
+            .apply {
+                (networkInterceptors + INTERCEPTOR_CACHE_MODIFIER + INTERCEPTOR_USER_AGENT)
+                    .forEach(::addNetworkInterceptor)
+            }
```
MiSikora, line 165: This seems a bit cleaner in terms of code formatting and it doesn't need to append lists.

```suggestion
            .apply { networkInterceptors.forEach(::addNetworkInterceptor) }
            .addNetworkInterceptor(INTERCEPTOR_CACHE_MODIFIER)
            .addNetworkInterceptor(INTERCEPTOR_USER_AGENT)
```

modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/di/ServersModule.kt
```diff
@@ -224,19 +229,26 @@ class ServersModule {
...
-    internal fun provideOkHttpClientNoCacheBuilder(): OkHttpClient.Builder {
+    internal fun provideOkHttpClientNoCacheBuilder(
+        networkInterceptors: Set<@JvmSuppressWildcards Interceptor>,
+    ): OkHttpClient.Builder {
         val dispatcher = Dispatcher()
         dispatcher.maxRequestsPerHost = 5
         var builder = OkHttpClient.Builder()
             .dispatcher(dispatcher)
-            .addNetworkInterceptor(INTERCEPTOR_USER_AGENT)
+            .apply { (networkInterceptors + INTERCEPTOR_USER_AGENT).forEach(::addNetworkInterceptor) }
```
MiSikora, line 251:
```suggestion
            .apply { networkInterceptors.forEach(::addNetworkInterceptor) }
            .addNetworkInterceptor(INTERCEPTOR_USER_AGENT)
```

modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/di/ServersModule.kt
wzieba: Thanks! Order is something I didn't consider - I don't know if it matters (interceptors are already applied) but it's maybe better to add Sentry's one as the last. Addressed in 29f12cc857a9090f93fbaa96a85f0e8812579848

wzieba: Or it _does_ matter, as interceptors are executed in order of applying?

wzieba: Ok I answered myself, it makes sense 😅 

> OkHttp uses lists to track interceptors, and interceptors are called in order. (https://square.github.io/okhttp/features/interceptors/)

So yeah, definitely it's better to add Sentry as the last one 👍

MiSikora: In that case wouldn't it be better to not include Sentry interceptor in set and inject it explicitly with some named annotation? Otherwise if we include other interceptors in the set in the future Sentry one might not be added as the last one.

wzieba: Sure 👍 the injection via set wasn't the best idea here, especially as it's the only interceptor. Addressed in 6e4023aa53198c8797bb37034c20df0b14900f1c

PR #2184, opened 2024-05-08
Do not send background restricions errors to Sentry

MiSikora:
## Description

We do not want to send `ForegroundServiceStartNotAllowedException` to Sentry due to quota limits. See: p1715090579752019-slack-C028JAG44VD

## Testing Instructions

> [!note]
> Test with SDK 31 or higher

1. Open the app.
2. Play something.
3. Pause it.
4. Kill the app and remove all Pocket Cast notifications.
5. Wait a for a minute or so.
6. Execute this ADB command `adb shell am broadcast -a au.com.shiftyjelly.pocketcasts.action.PLAY au.com.shiftyjelly.pocketcasts.debug`.
7. Playback will start for a couple of seconds and the it might stop due to OS restrictions.
8. You should see following in the logs.

```
Playback: attempted startForeground for state: 3, but that threw an exception we caught: android.app.ForegroundServiceStartNotAllowedException: Service.startForeground() not allowed due to mAllowStartForeground false: service au.com.shiftyjelly.pocketcasts.debug/au.com.shiftyjelly.pocketcasts.repositories.playback.PlaybackService
🔵 Tracked: playback_foreground_service_error
```

## Checklist
 
- [x] ~with different themes~

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackService.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackService.kt
@@ -18,7 +18,6 @@ import android.support.v4.media.session.PlaybackStateCompat
 import androidx.media.MediaBrowserServiceCompat
 import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsEvent
 import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsTrackerWrapper
-import au.com.shiftyjelly.pocketcasts.analytics.FirebaseAnalyticsTracker
 import au.com.shiftyjelly.pocketcasts.localization.BuildConfig
 import au.com.shiftyjelly.pocketcasts.models.entity.BaseEpisode
 import au.com.shiftyjelly.pocketcasts.models.entity.Podcast
@@ -278,7 +277,6 @@ open class PlaybackService : MediaBrowserServiceCompat(), CoroutineScope {
                             if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S && e is ForegroundServiceStartNotAllowedException) {
                                 addBatteryWarnings()
                                 analyticsTrackerWrapper.track(AnalyticsEvent.PLAYBACK_FOREGROUND_SERVICE_ERROR)
-                                FirebaseAnalyticsTracker.foregroundServiceStartNotAllowedException()
                             }
                         }
                     } else {
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackService.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackService.kt
@@ -16,6 +16,8 @@ import android.support.v4.media.session.MediaControllerCompat
 import android.support.v4.media.session.MediaSessionCompat
 import android.support.v4.media.session.PlaybackStateCompat
 import androidx.media.MediaBrowserServiceCompat
+import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsEvent
+import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsTrackerWrapper
 import au.com.shiftyjelly.pocketcasts.analytics.FirebaseAnalyticsTracker
 import au.com.shiftyjelly.pocketcasts.localization.BuildConfig
 import au.com.shiftyjelly.pocketcasts.models.entity.BaseEpisode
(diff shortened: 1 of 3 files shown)


modules/services/analytics/src/main/java/au/com/shiftyjelly/pocketcasts/analytics/TracksAnalyticsTracker.kt
```diff
@@ -91,6 +92,9 @@ class TracksAnalyticsTracker @Inject constructor(
                 AppPlatform.Phone -> "phone"
                 AppPlatform.WearOs -> "watch"
             },
+            PredefinedEventProperty.APP_VERSION_NAME to settings.getVersion(),
+            PredefinedEventProperty.APP_VERSION_CODE to settings.getVersionCode(),
```
wzieba, line 96: The Tracks library should attach this data already to events. It uses `PackageInfo` for this, and `settings.getVersionCode` uses `BuildConfig` but I think the value should be the same.

https://github.com/Automattic/Automattic-Tracks-Android/blob/trunk/AutomatticTracks/src/main/java/com/automattic/android/tracks/DeviceInformation.java#L81-L82

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackService.kt
```diff
@@ -272,11 +275,9 @@ open class PlaybackService : MediaBrowserServiceCompat(), CoroutineScope {
...
                         } catch (e: Exception) {
                             LogBuffer.e(LogBuffer.TAG_PLAYBACK, "attempted startForeground for state: $state, but that threw an exception we caught: $e")
-                            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S &&
-                                e is ForegroundServiceStartNotAllowedException
-                            ) {
+                            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S && e is ForegroundServiceStartNotAllowedException) {
                                 addBatteryWarnings()
-                                SentryHelper.recordException(e)
+                                analyticsTrackerWrapper.track(AnalyticsEvent.PLAYBACK_FOREGROUND_SERVICE_ERROR)
                                 FirebaseAnalyticsTracker.foregroundServiceStartNotAllowedException()
```
geekygecko, line 281: Should we remove the Firebase event as we now use Tracks?

modules/services/analytics/src/main/java/au/com/shiftyjelly/pocketcasts/analytics/TracksAnalyticsTracker.kt
MiSikora: Ahh, thanks! I'll drop the commit. I couldn't find it in tracks previously but it is all under `device_info` section.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackService.kt
MiSikora: I wasn't sure if we want it or not. I'll remove it.

modules/services/analytics/src/main/java/au/com/shiftyjelly/pocketcasts/analytics/TracksAnalyticsTracker.kt
wzieba: Thanks! Yeah, it's confusing - it's not really a "device info" 😕

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackService.kt
```diff
@@ -272,12 +274,9 @@ open class PlaybackService : MediaBrowserServiceCompat(), CoroutineScope {
                             LogBuffer.i(LogBuffer.TAG_PLAYBACK, "startForeground state: $state")
                         } catch (e: Exception) {
                             LogBuffer.e(LogBuffer.TAG_PLAYBACK, "attempted startForeground for state: $state, but that threw an exception we caught: $e")
-                            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S &&
-                                e is ForegroundServiceStartNotAllowedException
-                            ) {
+                            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S && e is ForegroundServiceStartNotAllowedException) {
```
mebarbosa, line 277: Should we log any other exception besides `ForegroundServiceStartNotAllowedException` ?

MiSikora: Do you think we should?

mebarbosa: I was wondering if we could add just because playback is a critical area and we have some users reporting playback issues.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackService.kt
geekygecko: Thanks

Issue #2081, opened 2024-04-18
Show Up Next in Tab Bar


ashiagr:
Project Thread: pdeCcb-4tX-p2
Designs: G60UqxPhTynbalo0Vfr1NC-fi-1060_3209

<img width="829" alt="Screenshot 2024-04-18 at 4 52 14 PM" src="https://github.com/Automattic/pocket-casts-android/assets/1405144/13a9fa1e-c2ad-4080-b4f1-3a5f95a85793">

```[tasklist]
### Tab bar changes
- [ ] https://github.com/Automattic/pocket-casts-android/pull/2077
- [ ] https://github.com/Automattic/pocket-casts-android/pull/2101
- [ ] https://github.com/Automattic/pocket-casts-android/pull/2102
- [ ] https://github.com/Automattic/pocket-casts-android/pull/2104
- [ ] https://github.com/Automattic/pocket-casts-android/pull/2113
- [ ] https://github.com/Automattic/pocket-casts-android/pull/2143
- [ ] https://github.com/Automattic/pocket-casts-android/pull/2161 
```

```[tasklist]
### Other design changes
- [ ] https://github.com/Automattic/pocket-casts-android/pull/2165
- [ ] https://github.com/Automattic/pocket-casts-android/pull/2197
- [ ] https://github.com/Automattic/pocket-casts-android/pull/2192
```

CookieyedCodes:
Can you share a screenshot for what this is going to look like, I generally like the double swipe up but maby I can like this if I know what it looks like :)

PR #2192, opened 2024-05-09
Up Next design updates

ashiagr:
Part of: #2081 

## Description

Figma Design: G60UqxPhTynbalo0Vfr1NC-fi-1584_17364

This applies following updates to the Up Next screen:

1. Add divider between queue items
2. Update queue item background color to match existing screen background **
3. Add image shadow so that (white) thumbnails are shown properly on certain themes like `Light Contrast`
4. Update header title to include episode count
5. Update Clear Queue dialog theme
6. Update Clear Queue text color to `primary_interactive_01`

** Existing screen background is still `primary_ui_04` instead of `primary_ui_01`. I have to apply an appropriate theme to the Now Playing item before I switch to the new background color.

*** Font sizes and styles for texts shown in Figma are not yet applied as they need to be used throughout the app. They're kept outside the scope of this PR.

### FF enabled

1. Open the app
2. Notice that the above changes are applied 
3. Smoke test theme changes on Up Next screen for tab and Up Next screen on the Player screen

### FF disabled

1. Disable feature flag UPNEXT_IN_TAB_BAR 
2. Kill and restart the app
4. Notice that the original design is shown

## Screenshots or Screencast 

Up Next Tab | Up Next - Clear Queue Light | Up Next - BottomSheet Dark | Up Next - Clear Queue Dark
---|---|---|---
 |    |  |

## Checklist
 
- [x] with different themes

### DIFF
--- a/modules/services/localization/src/main/res/values/strings.xml
+++ b/modules/services/localization/src/main/res/values/strings.xml
@@ -343,7 +343,7 @@ up    <string name="player_sleep_in_one_chapter">Sleeping in 1 chapter</string>
     <string name="player_up_next_move_to_top">Move to top</string>
     <string name="player_up_next_select">Select</string>
     <string name="player_up_next_time_remaining">Total time remaining %s</string>
-    <string name="player_up_next_time_left">%s left</string>
+    <string name="player_up_next_time_left">@string/time_left</string>
     <plurals name="player_up_next_header_title">
         <item quantity="one" formatted="true">%1$d episode - %2$s left</item>
         <item quantity="other" formatted="true">%1$d episodes - %2$s left</item>
--- a/modules/services/localization/src/main/res/values/strings.xml
+++ b/modules/services/localization/src/main/res/values/strings.xml
@@ -343,6 +343,11 @@ up    <string name="player_sleep_in_one_chapter">Sleeping in 1 chapter</string>
     <string name="player_up_next_move_to_top">Move to top</string>
     <string name="player_up_next_select">Select</string>
     <string name="player_up_next_time_remaining">Total time remaining %s</string>
+    <string name="player_up_next_time_left">%s left</string>
+    <plurals name="player_up_next_header_title">
+        <item quantity="one" formatted="true">One episode - %2$s left</item>
+        <item quantity="other" formatted="true">%1$d episodes - %2$s left</item>
+    </plurals>
     <string name="player_open_url_failed">Could not find a browser to open %s.</string>
     <string name="player_archive_title" translatable="false">@string/archive</string>
     <string name="player_archive_summary">Are you sure you want to archive this episode?</string>
--- a/modules/services/localization/src/main/res/values/strings.xml
+++ b/modules/services/localization/src/main/res/values/strings.xml
@@ -345,7 +345,7 @@ up    <string name="player_sleep_in_one_chapter">Sleeping in 1 chapter</string>
     <string name="player_up_next_time_remaining">Total time remaining %s</string>
     <string name="player_up_next_time_left">%s left</string>
     <plurals name="player_up_next_header_title">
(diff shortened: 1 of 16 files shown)


modules/services/localization/src/main/res/values/strings.xml
```diff
@@ -343,6 +343,11 @@ up    <string name="player_sleep_in_one_chapter">Sleeping in 1 chapter</string>
     <string name="player_up_next_move_to_top">Move to top</string>
     <string name="player_up_next_select">Select</string>
     <string name="player_up_next_time_remaining">Total time remaining %s</string>
+    <string name="player_up_next_time_left">%s left</string>
+    <plurals name="player_up_next_header_title">
+        <item quantity="one" formatted="true">One episode - %2$s left</item>
```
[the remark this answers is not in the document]
ashiagr: Updated in 40613eb

modules/services/localization/src/main/res/values/strings.xml
```diff
@@ -343,6 +343,11 @@ up    <string name="player_sleep_in_one_chapter">Sleeping in 1 chapter</string>
     <string name="player_up_next_move_to_top">Move to top</string>
     <string name="player_up_next_select">Select</string>
     <string name="player_up_next_time_remaining">Total time remaining %s</string>
+    <string name="player_up_next_time_left">%s left</string>
```
geekygecko, line 346: We already have `time_left` should we use this instead.
```
    <string name="time_left">%s left</string>
```
or we could link back to it
```
   <string name="player_up_next_time_left" translatable="false">@string/time_left</string>
```

modules/services/localization/src/main/res/values/strings.xml
```diff
@@ -343,6 +343,11 @@ up    <string name="player_sleep_in_one_chapter">Sleeping in 1 chapter</string>
...
+    <plurals name="player_up_next_header_title">
+        <item quantity="one" formatted="true">%1$d episode - %2$s left</item>
+        <item quantity="other" formatted="true">%1$d episodes - %2$s left</item>
```
geekygecko, line 349: We have these separately `episodes_singular`, `episodes_plural`, and `time_left `. Would it be better to use those rather than creating a new string to combine them?

geekygecko (approved):

modules/services/localization/src/main/res/values/strings.xml
MiSikora: Regular strings divided into `singular` and `plural` do not handle plurality rules correctly. English has only two forms `one` and `zero`. But other languages might have different forms. For example in Polish you'd say.

```
0 episodes - 0 epizodów
1 episode - 1 epizod
2/3/4 episodes - 2/3/4 epizody
5 episodes - 5 episodów
```

modules/services/localization/src/main/res/values/strings.xml
ashiagr: Good point. Linked back in [9e8d5ad](https://github.com/Automattic/pocket-casts-android/pull/2192/commits/9e8d5add81afacc7dd6084c2ff8c7c8019d75703)

modules/services/localization/src/main/res/values/strings.xml
ashiagr: Oh nice! Thanks for explaining @MiSikora! I'll leave it as is then.

geekygecko: Changing `episodes_singular` and `episodes_plural` to use Android plurals would good. I more meant that it might be a good idea to not combine episode and left in the same string.

MiSikora: > I more meant that it might be a good idea to not combine episode and left in the same string.

But the position of these might depend entirely on language. In some languages "time left" might make more sense to be displayed at the beginning.

PR #2194, opened 2024-05-09
Fix auto archiving episodes

MiSikora:
## Description

Episodes were being archived after playing even when global settings was set to never and podcasts wasn't technically overriding the value. Some users were affected by it during settings sync test.

Internal ref: p1713427004164729-slack-C02A333D8LQ

### Bug fix

1. Open the app.
2. Go to global auto archive settings and set `Archive played episodes` to `Never`.
3. Run `UPDATE podcasts SET auto_archive_played_after = 1, override_global_archive = 0` in the `Database Inspector`.
4. Play an episode from any podcast.
5. Skip to the end.
6. Podcast should not be archived.

### Regressions

Test that episode playback respects all custom auto archive settings. Both global and podcast specific ones. You might want to clear you DB after the previous test before conducting this test.

## Checklist
 
- [x] ~with different themes~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 *   Bug Fixes
     *   Fix pull to refresh icon sometimes being stuck in a loading state
         ([#2164](https://github.com/Automattic/pocket-casts-android/pull/2164))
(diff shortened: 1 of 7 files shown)


ashiagr (approved):
Tested. Works as described!

Issue #2228, opened 2024-05-14
Folders not rendering correctly on Landscape Mode


joashrajin:
### Description

When I use my Fold 4 in Landscape Mode, the folders aren't rendering correctly anymore when on v7.64 rc-1

### Step-by-step reproduction instructions

1. Switch to landscape mode on a smaller tablet or foldable device
2. Notice how the folders render

### Screenshots or screen recording

![Screenshot_20240514_195757_Pocket Casts](https://github.com/Automattic/pocket-casts-android/assets/16253818/500acd9e-54c9-4212-945b-137eb0b8f500)

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

Samsung Fold 4, Android 14, Pocket Casts v7.64 rc-1

PR #2230, opened 2024-05-15
Fix folder size on large screens

geekygecko:
## Description

This fixes the folder size issue on the Samsung Fold in landscape. 

The issue was that the logic was still assuming a navigation drawer was displayed on larger devices, which hasn't been the case for a while. I have removed this logic. I realised that it's probably best to try just filling the available space instead of using an exact size with the folders. This worked, which makes me wonder why we were setting it in the past. Hopefully, I haven't broken anything in the process. Please let me know if you can think of anything. 

The change has also fixed the slight folder size issue on a regular sized phone.

Fixes https://github.com/Automattic/pocket-casts-android/issues/2228

## Testing Instructions
1. Create an emulator with the virtual configuration of a Pixel Fold (if you haven't already)
2. Put the emulator in landscape 
3. Subscribe to some podcasts
4. Add a folder to the home screen with podcasts in it
5. ✅ Verify the folder size matches podcasts in the podcast grid
6. Rotate the emulator
7. ✅ Verify the folder size again
8. Change the layout to the small grid and list view
9. ✅ Verify the folder size again
10. Try the same on a regular sized phone device

## Screenshots 
| Before | After |
| --- | --- |
| Screenshot_20240515_105503 | Screenshot_20240515_105418 |
| Screenshot_20240515_110454 | Screenshot_20240515_110421 |

### DIFF
--- a/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/podcasts/FolderAdapter.kt
+++ b/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/podcasts/FolderAdapter.kt
@@ -37,7 +37,6 @@ import au.com.shiftyjelly.pocketcasts.ui.extensions.getThemeColor
 import au.com.shiftyjelly.pocketcasts.ui.extensions.themed
 import au.com.shiftyjelly.pocketcasts.ui.theme.Theme
 import au.com.shiftyjelly.pocketcasts.utils.extensions.dpToPx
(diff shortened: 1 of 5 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

ashiagr (approved):
> I realised that it's probably best to try just filling the available space instead of using an exact size with the folders.

You did the right thing. I haven't found issues in my testing, all good 👍

PR #2235, opened 2024-05-15
Fix: Android doesn't handle timestamp parameter in share links properly

### LINKED ISSUE #21: Links with position in them do not start from the shared position
Test steps:

1. Share a link with position from the app
2. Close the app and wipe the app's data
3. Open the link
4. Observe that the web player has the podcast loaded to the proper position
5. Tap the button to open the podcast in the app
6. Begin playing the podcast within the app and observe that it does not start playing from the proper position

### LINKED ISSUE #2227: Android doesn't handle timestamp parameter in share links properly
### Description

When sharing links like `https://pca.st/sb2ksvqv?t=28,60 ` or `https://pca.st/y0ebrkcc?t=10m15s` we should parse two timestamps instead of expecting it to be only start. The general format is `t=<start>,<end>` where timestamp can be expressed in different patterns. See https://www.w3.org/TR/media-frags/#naming-time for specification.

We don't need to handle end time in the player yet but we need to parse timestamps correctly.

Internal ref: p1715694841931039-slack-C02A333D8LQ

1. Use `https://pca.st/y0ebrkcc?t=10m15s` to open Pocket Casts.
2. It should start playback of [this episode](https://pca.st/y0ebrkcc) at `10:15`.

_No response_

_No response_

mebarbosa:
## Description
- This PR fixes the timestamp parsing when opening Pocket Casts from shared URL. It will support:

```
https://pca.st/sb2ksvqv?t=28,60

Either one or both parameters may be omitted, with the begin time defaulting to 0 seconds and the end time defaulting to the duration of the source media

t=10,20   # => results in the time interval [10,20)
t=,20     # => results in the time interval [0,20)
t=10      # => results in the time interval [10,end)

```

```
https://pca.st/y0ebrkcc?t=1:02:03

hh:mm:ss.fraction - Either be specified as seconds, with an optional fractional part to indicate miliseconds, or as colon-separated hours, minutes and seconds (again with an optional fraction). Minutes and seconds must be specified as exactly two digits, hours and fractional seconds can be any number of digits

```

```
https://pca.st/y0ebrkcc?t=10m15s

[hour]h[minute]m[second]s

```

Closes #2227
Closes #21

## Testing Instructions
1. Use https://pca.st/4adhfj4d?t=7m15s to open Pocket Casts.
2. ✅ Ensure the app start playback of the episode at 07:15
3. Use https://pca.st/sb2ksvqv?t=28,60 to open Pocket Casts
4. ✅ Ensure the app start playback of the episode at 28 seconds
5. Use https://pca.st/sb2ksvqv?t=,60 to open Pocket Casts
6. ✅ Ensure the app start playback of the episode at 0 seconds
7. Use https://pca.st/p9ipuzdr?t=10s to open Pocket Casts
8. ✅ Ensure the app start playback of the episode at 10 seconds
9. Use https://pca.st/p9ipuzdr?t=0:02:00 to open Pocket Casts
10. ✅ Ensure the app start playback of the episode at 2 minutes

## Screenshots or Screencast 
[Screen_recording_20240515_141048.webm](https://github.com/Automattic/pocket-casts-android/assets/42220351/d1b6fa5d-14e1-4fd5-a1a7-9807c5b71287)

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/SharingUrlTimestampParser.kt
+++ b/modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/SharingUrlTimestampParser.kt
@@ -22,8 +22,8 @@ class SharingUrlTimestampParser {
             val hours = match.groups[1]?.value?.toLong() ?: 0
             val minutes = match.groups[2]?.value?.toLong() ?: 0
             val seconds = match.groups[3]?.value?.toLong() ?: 0
-            val fraction = match.groups[4]?.value?.toDouble() ?: 0.0
-            val totalSeconds = (hours * 3600) + (minutes * 60) + seconds + (fraction / 1000.0)
+            val fractionInMilliseconds = match.groups[4]?.value?.take(3)?.toDouble() ?: 0.0 // Extract first three digits
+            val totalSeconds = (hours * 3600) + (minutes * 60) + seconds + (fractionInMilliseconds / 1000.0)
             return Pair(totalSeconds.seconds, null)
         }
 
--- a/modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/SharingUrlTimestampParser.kt
+++ b/modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/SharingUrlTimestampParser.kt
@@ -1,0 +1,18 @@
+package au.com.shiftyjelly.pocketcasts.utils
+
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.seconds
+
+class SharingUrlTimestampParser {
+    fun parseTimestamp(t: String): Pair<Duration?, Duration?> {
+        val intervalPattern = Regex("""(\d*\.?\d*)?,?(\d*\.?\d*)?""") // t=10,20 or t=,20 or t=10
+
+        intervalPattern.find(t)?.let { match ->
+            val startTime = match.groups[1]?.value?.takeIf { it.isNotEmpty() }?.toDouble()?.toInt() ?: 0
+            val endTime = match.groups[2]?.value?.takeIf { it.isNotEmpty() }?.toDouble()?.toInt() ?: 0
+            return Pair(startTime.seconds, endTime.seconds)
+        }
+
+        return Pair(null, null)
+    }
+}
--- a/modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/SharingUrlTimestampParser.kt
+++ b/modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/SharingUrlTimestampParser.kt
@@ -18,7 +18,7 @@ class SharingUrlTimestampParser {
             return Pair(extractTimeFrom(startParsedTime)?.seconds, extractTimeFrom(endParsedTime)?.seconds)
         }
         val result = parseTime(timestamp)
-        return Pair((result.first ?: 0.0).seconds, (result.second ?: 0.0).seconds)
+        return Pair(result.first?.seconds, result.second?.seconds)
     }
 
     private fun parseTime(t: String): Pair<Double?, Double?> {
@@ -27,9 +27,6 @@ class SharingUrlTimestampParser {
             val minutes = match.groups[2]?.value?.toInt() ?: 0
             val seconds = match.groups[3]?.value?.toDouble() ?: 0.0
             val totalSeconds = ((hours * 3600) + (minutes * 60) + seconds)
-            if (totalSeconds == 0.0) {
-                return Pair(0.0, null)
-            }
             return Pair(totalSeconds, null)
         }
 
@@ -39,15 +36,12 @@ class SharingUrlTimestampParser {
             val seconds = match.groups[3]?.value?.toLong() ?: 0
             val fractionInMilliseconds = match.groups[4]?.value?.take(3)?.toDouble() ?: 0.0 // Extract first three digits
             val totalSeconds = (hours * 3600) + (minutes * 60) + seconds + (fractionInMilliseconds / 1000.0)
-            if (totalSeconds == 0.0) {
-                return Pair(0.0, null)
-            }
             return Pair(totalSeconds, null)
         }
 
         intervalPattern.find(t)?.let { match ->
-            val startTime = match.groups[1]?.value?.takeIf { it.isNotEmpty() }?.toDouble()?.toInt()
-            val endTime = match.groups[2]?.value?.takeIf { it.isNotEmpty() }?.toDouble()?.toInt()
+            val startTime = match.groups[1]?.value?.takeIf { it.isNotEmpty() }?.toDouble()?.toInt()?.takeIf { it != 0 }
+            val endTime = match.groups[2]?.value?.takeIf { it.isNotEmpty() }?.toDouble()?.toInt()?.takeIf { it != 0 }
             return Pair(startTime?.toDouble(), endTime?.toDouble())
         }
 
@@ -55,7 +49,7 @@ class SharingUrlTimestampParser {
     }
 
     private fun extractTimeFrom(startParsedTime: Pair<Double?, Double?>) = if (startParsedTime.first == null && startParsedTime.second == null) {
-        0.0
+        null
     } else if (startParsedTime.first != null) {
         startParsedTime.first
     } else {
--- a/modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/SharingUrlTimestampParser.kt
+++ b/modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/SharingUrlTimestampParser.kt
@@ -7,7 +7,7 @@ class SharingUrlTimestampParser {
     companion object {
         val intervalPattern = Regex("""(\d*\.?\d*)?,?(\d*\.?\d*)?""")
         val hmsPattern = Regex("""(?:(\d+)h)?(?:(0?[0-5]?\d)m)?(?:(0?[0-5]?\d)s)?""")
-        val hhmmssFractionPattern = Regex("""(\d+):(0[0-9]|[1-5][0-9]):(0[0-9]|[1-5][0-9])(?:\.(\d+))?""")
+        val hhmmssFractionPattern = Regex("""(?:(\d+):)?(0[0-9]|[1-5][0-9]):(0[0-9]|[1-5][0-9])(?:\.(\d+))?""")
     }
 
     fun parseTimestamp(timestamp: String): Pair<Duration?, Duration?> {
--- a/modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/SharingUrlTimestampParser.kt
+++ b/modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/SharingUrlTimestampParser.kt
@@ -4,11 +4,12 @@ import kotlin.time.Duration
 import kotlin.time.Duration.Companion.seconds
 
 class SharingUrlTimestampParser {
-    fun parseTimestamp(t: String): Pair<Duration?, Duration?> {
+    companion object {
         val intervalPattern = Regex("""(\d*\.?\d*)?,?(\d*\.?\d*)?""") // t=10,20 or t=,20 or t=10
         val hmsPattern = Regex("""(?:(\d+)h)?(?:(\d+)m)?(?:(\d+(?:\.\d+)?)s)?""") // t=10h20m30s or t=10m30s or or 10m or t=10s
         val hhmmssFractionPattern = Regex("""(\d+):(\d{2}):(\d{2})(?:\.(\d+))?""") // t=10:20:30.456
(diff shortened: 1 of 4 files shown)


modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/SharingUrlTimestampParser.kt
```diff
@@ -0,0 +1,37 @@
+package au.com.shiftyjelly.pocketcasts.utils
+
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.seconds
+
+class SharingUrlTimestampParser {
+    fun parseTimestamp(t: String): Pair<Duration?, Duration?> {
+        val intervalPattern = Regex("""(\d*\.?\d*)?,?(\d*\.?\d*)?""") // t=10,20 or t=,20 or t=10
+        val hmsPattern = Regex("""(?:(\d+)h)?(?:(\d+)m)?(?:(\d+(?:\.\d+)?)s)?""") // t=10h20m30s or t=10m30s or or 10m or t=10s
+        val hhmmssFractionPattern = Regex("""(\d+):(\d{2}):(\d{2})(?:\.(\d+))?""") // t=10:20:30.456
```
MiSikora, line 10: Can we move this to companion object? Creating regex is a bit heavy operation because it needs to create the whole internal state machine.

modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/SharingUrlTimestampParser.kt
```diff
@@ -0,0 +1,37 @@
+package au.com.shiftyjelly.pocketcasts.utils
+
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.seconds
+
+class SharingUrlTimestampParser {
+    fun parseTimestamp(t: String): Pair<Duration?, Duration?> {
+        val intervalPattern = Regex("""(\d*\.?\d*)?,?(\d*\.?\d*)?""") // t=10,20 or t=,20 or t=10
```
MiSikora, line 8: This needs to handle also decimal points as well. Number after the decimal point is considered to be milliseconds.

modules/services/utils/src/test/java/au/com/shiftyjelly/pocketcasts/utils/SharingUrlTimestampParserTest.kt
```diff
@@ -0,0 +1,33 @@
...
+    @Test
+    fun parseTimestampInterval() {
+        assertEquals(Pair(28.seconds, 60.seconds), parser.parseTimestamp("28,60"))
+        assertEquals(Pair(0.seconds, 20.seconds), parser.parseTimestamp(",20"))
+        assertEquals(Pair(10.seconds, 0.seconds), parser.parseTimestamp("10"))
+    }
+
+    @Test
+    fun testParseTimestampHoursMinutesSeconds() {
+        assertEquals(Pair(454070.seconds, null), parser.parseTimestamp("125h67m50s"))
```
MiSikora, line 20: This isn't correct. Something like `67m` shouldn't have passed. Minutes and seconds must be between `0` and `59`.

modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/SharingUrlTimestampParser.kt
MiSikora, line 10: This should handle mixed patterns. For example `120,0:02:01.5` is valid. Adding test for these would be great.

modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/SharingUrlTimestampParser.kt
```diff
@@ -0,0 +1,37 @@
...
+        val hmsPattern = Regex("""(?:(\d+)h)?(?:(\d+)m)?(?:(\d+(?:\.\d+)?)s)?""") // t=10h20m30s or t=10m30s or or 10m or t=10s
+        val hhmmssFractionPattern = Regex("""(\d+):(\d{2}):(\d{2})(?:\.(\d+))?""") // t=10:20:30.456
+
+        hmsPattern.matchEntire(t)?.let { match ->
+            val hours = match.groups[1]?.value?.toInt() ?: 0
+            val minutes = match.groups[2]?.value?.toInt() ?: 0
+            val seconds = match.groups[3]?.value?.toDouble() ?: 0.0
```
MiSikora, line 15: Can you add a test for handling decimal part for this pattern?

modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/SharingUrlTimestampParser.kt
```diff
@@ -0,0 +1,37 @@
...
+            val totalSeconds = ((hours * 3600) + (minutes * 60) + seconds)
+            return Pair(totalSeconds.seconds, null)
+        }
+
+        hhmmssFractionPattern.matchEntire(t)?.let { match ->
+            val hours = match.groups[1]?.value?.toLong() ?: 0
+            val minutes = match.groups[2]?.value?.toLong() ?: 0
+            val seconds = match.groups[3]?.value?.toLong() ?: 0
+            val fraction = match.groups[4]?.value?.toDouble() ?: 0.0
+            val totalSeconds = (hours * 3600) + (minutes * 60) + seconds + (fraction / 1000.0)
```
MiSikora, line 25: The `fraction / 1000.0` part is faulty. For example this test fails and evaluates timestamp to `5d 18h 53m 21s`.

```kotlin
assertEquals(Pair(1500.milliseconds, null), parser.parseTimestamp("00:00:01.500000000"))
```

modules/services/utils/src/test/java/au/com/shiftyjelly/pocketcasts/utils/SharingUrlTimestampParserTest.kt
```diff
@@ -0,0 +1,33 @@
...
+        assertEquals(Pair(3600.seconds, null), parser.parseTimestamp("1h"))
+        assertEquals(Pair(60.seconds, null), parser.parseTimestamp("1m"))
+        assertEquals(Pair(1.seconds, null), parser.parseTimestamp("1s"))
+        assertEquals(Pair(3661.seconds, null), parser.parseTimestamp("1h1m1s"))
+    }
+
+    @Test
+    fun testParseTimestampHoursMinutesSecondsFraction() {
+        assertEquals(Pair(3723.seconds, null), parser.parseTimestamp("1:02:03"))
+        assertEquals(Pair(3728.seconds, null), parser.parseTimestamp("1:02:03.5000"))
```
MiSikora, line 30: This isn't correct. This is correct evaluation. Fractional part doesn't mean 5000 milliseconds here. It means 500_000 microseconds. Basically we only care about first three digits and convert them to milliseconds.

```kotlin
assertEquals(Pair(1.hours + 2.minutes + 3.seconds + 500.milliseconds, null), parser.parseTimestamp("1:02:03.5000"))
```

MiSikora (changes requested):
I know I left a lot of comments but it is because I want us to get this right. In general I really like this implementation.

modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/SharingUrlTimestampParser.kt
mebarbosa: Done: [5c8e07d](https://github.com/Automattic/pocket-casts-android/pull/2235/commits/5c8e07dcb433aa4d3aafc07d7a38a30e0b6d7ef5)

modules/services/utils/src/test/java/au/com/shiftyjelly/pocketcasts/utils/SharingUrlTimestampParserTest.kt
mebarbosa: [ad8b83a](https://github.com/Automattic/pocket-casts-android/pull/2235/commits/ad8b83a6190469c06444789215e405526dffcbca)

modules/services/utils/src/test/java/au/com/shiftyjelly/pocketcasts/utils/SharingUrlTimestampParserTest.kt
mebarbosa: Got it! I've updated here [46e81a4](https://github.com/Automattic/pocket-casts-android/pull/2235/commits/46e81a4f70a2870526be23c79cfe7df0ffc6fa13)

modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/SharingUrlTimestampParser.kt
mebarbosa: I am struggling to implement the mixed patterns with my limited regex knowledge. I will give another try tomorrow

MiSikora: Right, that's why I'm suggesting to have patterns only for single parts. More or less something like this.

```kotlin
fun parseTimestamp(t: String): Pair<Duration, Duration>? {
    val parts = t.split(",")
    if (parts.isEmpty() || parts.size > 2) {
        return null
    }
    val start = parts[0]
    val end = part.getOrNull(1)
    TODO("Parse start and end separately")
}
```

modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/SharingUrlTimestampParser.kt
```diff
@@ -0,0 +1,38 @@
+package au.com.shiftyjelly.pocketcasts.utils
+
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.seconds
+
+class SharingUrlTimestampParser {
+    companion object {
+        val intervalPattern = Regex("""(\d*\.?\d*)?,?(\d*\.?\d*)?""") // t=10,20 or t=,20 or t=10
+        val hmsPattern = Regex("""(?:(\d+)h)?(?:(0?[0-5]?\d)m)?(?:(0?[0-5]?\d(?:\.\d+)?)s)?""") // t=10h20m30s or t=10m30s or or 10m or t=10s
+        val hhmmssFractionPattern = Regex("""(\d+):(\d{2}):(\d{2})(?:\.(\d+))?""") // t=10:20:30.456
```
MiSikora, line 10: All of these, not only `hmsPattern`, need to handle minutes and seconds above `59`. Can you add tests for these cases as well?

modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/SharingUrlTimestampParser.kt
mebarbosa: Yeah. I was trying to do this, but I faced with some issues. Anyways, please, check here: [5bc9733](https://github.com/Automattic/pocket-casts-android/pull/2235/commits/5bc9733fb4358d2e589df3d9d3643e3510209ba3)

modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/SharingUrlTimestampParser.kt
mebarbosa: [a18bb80](https://github.com/Automattic/pocket-casts-android/pull/2235/commits/a18bb8044bbe92f1a0354d65aa2fb29ecb46d6f9)

modules/services/utils/src/test/java/au/com/shiftyjelly/pocketcasts/utils/SharingUrlTimestampParserTest.kt
```diff
@@ -0,0 +1,53 @@
...
+        assertEquals(Pair(3723.seconds, 0.seconds), parser.parseTimestamp("1:02:03"))
+        assertEquals(Pair(120.seconds, 0.seconds), parser.parseTimestamp("0:02:00"))
+        assertEquals(Pair(1500.milliseconds, 0.seconds), parser.parseTimestamp("00:00:01.500000000"))
+        assertEquals(Pair(1.hours + 2.minutes + 3.seconds + 500.milliseconds, 0.seconds), parser.parseTimestamp("1:02:03.5000"))
+    }
+
+    @Test
+    fun testParseTimestampMixedPatterns() {
+        assertEquals(Pair(120.seconds, 2.minutes + 1.seconds + 5.milliseconds), parser.parseTimestamp("120,0:02:01.5"))
+    }
```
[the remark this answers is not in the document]
mebarbosa: [5587dba](https://github.com/Automattic/pocket-casts-android/pull/2235/commits/5587dbad59af94d13b48478ced83eb15e91c7b09)

MiSikora (approved):

Issue #2237, opened 2024-05-16
Podcast artwork doesn't display with Tesla


geekygecko:
### Description

When playing an episode through Bluetooth the Tesla media screen does not display the artwork. This is working in the other apps I tested so it must be possible. The other apps were Google Podcasts, Audible, Spotify, and AntennaPod.

Internal reference: p1713956545483609-slack-C02A333D8LQ

### Step-by-step reproduction instructions

1. Connect your Bluetooth to a Tesla 🚗
2. Play an episode
3. See the artwork is not displayed

### Screenshots or screen recording

![20240516_101548](https://github.com/Automattic/pocket-casts-android/assets/308331/37aab9d6-6a40-446b-bb02-16e4395db996)

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

Samsung Fold5, Pocket Casts 7.54-rc-2

P-Burk:
Just wanted to chime in because I've also experienced this. 

Also, maybe I'm not remembering correctly because I don't pay a lot of attention to it, but I could have sworn that sometimes it works and sometimes it doesn't. It might be podcast specific. But like I said, I can't make a definitive comment on it as I don't really pay it much attention.

PR #2254, opened 2024-05-20
Fix Tesla artwork issue

geekygecko:
## Description

The podcast or episode artwork doesn't show in Tesla cars when only the artwork URL is set in the media session metadata. If the artwork is set in a bitmap it's displayed. 

Fixes https://github.com/Automattic/pocket-casts-android/issues/2237

## Testing Instructions
This is a hard one to test without a car but it would be good to check that using the bitmap doesn't break other places where we use the artwork. 

1. Tap play on an episode
2. ✅ Verify the notification shade artwork is correct and not low-resolution
3. Start an Android Auto head unit
```
cd /Applications/android-sdk/extras/google/auto/ (or wherever your Android SDK folder is located)
# first time add the execute permission
chmod +x ./desktop-head-unit
./desktop-head-unit --usb
```
4. ✅ Verify the playing screen has the correct artwork

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/MediaSessionManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/MediaSessionManager.kt
@@ -305,7 +305,7 @@ class MediaSessionManager(
                     val podcastUuid = if (episode is PodcastEpisode) episode.podcastUuid else null
                     val podcast = podcastUuid?.let { podcastManager.findPodcastByUuid(it) }
                     val podcastTitle = episode.displaySubtitle(podcast)
-                    val localUri = AutoConverter.getBitmapUriForPodcast(podcast, episode, context, settings.artworkConfiguration.value.useEpisodeArtwork)
+                    val localUri = AutoConverter.getPodcastArtworkUri(podcast, episode, context, settings.artworkConfiguration.value.useEpisodeArtwork)
                     val description = MediaDescriptionCompat.Builder()
                         .setDescription(episode.episodeDescription)
                         .setTitle(episode.title)
@@ -400,19 +400,22 @@ class MediaSessionManager(
             nowPlayingBuilder.putString(MediaMetadataCompat.METADATA_KEY_ALBUM, podcast.author)
         }
 
-        val nowPlaying = nowPlayingBuilder.build()
         Timber.i("MediaSession metadata. Episode: ${episode.uuid} ${episode.title} Duration: ${episode.durationMs.toLong()}")
-        mediaSession.setMetadata(nowPlaying)
 
         if (settings.showArtworkOnLockScreen.value) {
-            val bitmapUri = AutoConverter.getBitmapUriForPodcast(podcast, episode, context, useEpisodeArtwork)?.toString()
+            val bitmapUri = AutoConverter.getPodcastArtworkUri(podcast, episode, context, useEpisodeArtwork)?.toString()
             nowPlayingBuilder = nowPlayingBuilder.putString(MediaMetadataCompat.METADATA_KEY_ALBUM_ART_URI, bitmapUri)
             if (Util.isAutomotive(context)) nowPlayingBuilder = nowPlayingBuilder.putString(MediaMetadataCompat.METADATA_KEY_DISPLAY_ICON_URI, bitmapUri)
 
-            val nowPlayingWithArtwork = nowPlayingBuilder.build()
+            // Send the bitmap, as some devices don't support URLs, such as a Tesla.
+            AutoConverter.getPodcastArtworkBitmap(episode, context, useEpisodeArtwork)?.let { bitmap ->
+                nowPlayingBuilder = nowPlayingBuilder.putBitmap(MediaMetadataCompat.METADATA_KEY_ALBUM_ART, bitmap)
+            }
             Timber.i("MediaSession metadata. With artwork.")
-            mediaSession.setMetadata(nowPlayingWithArtwork)
         }
+
+        val nowPlaying = nowPlayingBuilder.build()
+        mediaSession.setMetadata(nowPlaying)
     }
 
     private fun addCustomActions(stateBuilder: PlaybackStateCompat.Builder, currentEpisode: BaseEpisode, playbackState: PlaybackState) {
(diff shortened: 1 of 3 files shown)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/auto/AutoConverter.kt
```diff
@@ -125,7 +131,7 @@ object AutoConverter {
         return MediaBrowserCompat.MediaItem(mediaDescription, MediaBrowserCompat.MediaItem.FLAG_BROWSABLE)
     }
 
-    fun getBitmapUriForPodcast(podcast: Podcast?, episode: BaseEpisode?, context: Context, showRssArtwork: Boolean): Uri? {
+    fun getPodcastArtworkUri(podcast: Podcast?, episode: BaseEpisode?, context: Context, showRssArtwork: Boolean): Uri? {
```
geekygecko, line 134: I renamed this method as it didn't make sense with bitmap in the name when it doesn't return a bitmap.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/auto/AutoConverter.kt
```diff
@@ -194,8 +217,14 @@ object AutoConverter {
...
-        val drawableName = context.resources.getResourceEntryName(drawable)
-        return Uri.parse("android.resource://" + context.packageName + "/drawable/" + drawableName)
+        val resources = context.resources
+        // This is an example of the URI android.resource://au.com.shiftyjelly.pocketcasts/drawable/auto_folder_01
+        return Uri.Builder()
+            .scheme(ContentResolver.SCHEME_ANDROID_RESOURCE) // android.resource
+            .authority(resources.getResourcePackageName(drawable)) // au.com.shiftyjelly.pocketcasts
+            .appendPath(resources.getResourceTypeName(drawable)) // drawable
+            .appendPath(resources.getResourceEntryName(drawable)) // auto_folder_01
+            .build()
```
geekygecko, line 227: This isn't really related to the change, but I thought it was a better way to do it, rather than building up a string.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/MediaSessionManager.kt
```diff
@@ -400,19 +400,25 @@ class MediaSessionManager(
             nowPlayingBuilder.putString(MediaMetadataCompat.METADATA_KEY_ALBUM, podcast.author)
         }
 
-        val nowPlaying = nowPlayingBuilder.build()
         Timber.i("MediaSession metadata. Episode: ${episode.uuid} ${episode.title} Duration: ${episode.durationMs.toLong()}")
-        mediaSession.setMetadata(nowPlaying)
```
geekygecko, line 405: Setting here and below caused the artwork to be displayed and then removed.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/auto/AutoConverter.kt
```diff
@@ -143,6 +149,23 @@ object AutoConverter {
         return getArtworkUriForContentProvider(podcastArtUri, context)
     }
 
+    fun getPodcastArtworkBitmap(episode: BaseEpisode, context: Context, useEpisodeArtwork: Boolean): Bitmap? {
+        val imageRequestFactory = PocketCastsImageRequestFactory(
+            context,
+            isDarkTheme = true,
+            size = 480,
+            placeholderType = PocketCastsImageRequestFactory.PlaceholderType.Small,
+        )
```
geekygecko, line 158: I wasn't sure what size to send over. It can be displayed larger, but I didn't want to send too much data, so I picked the middle image size.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/MediaSessionManager.kt
```diff
@@ -400,19 +400,25 @@ class MediaSessionManager(
...
             if (Util.isAutomotive(context)) nowPlayingBuilder = nowPlayingBuilder.putString(MediaMetadataCompat.METADATA_KEY_DISPLAY_ICON_URI, bitmapUri)
 
-            val nowPlayingWithArtwork = nowPlayingBuilder.build()
+            // Send the bitmap, as some devices don't support URLs, such as a Tesla.
+            // Don't do this for Wear OS or Automotive to reduce the amount memory used.
+            if (!Util.isWearOs(context) && !Util.isAutomotive(context)) {
+                AutoConverter.getPodcastArtworkBitmap(episode, context, useEpisodeArtwork)?.let { bitmap ->
+                    nowPlayingBuilder = nowPlayingBuilder.putBitmap(MediaMetadataCompat.METADATA_KEY_ALBUM_ART, bitmap)
+                }
+            }
```
MiSikora, line 416: Is it correct to do this only when `showArtworkOnLockScreen` is `true`?

MiSikora (commented):
I smoke tested the app with different scenarios - embedded artwork, chapters, and so on. Everything works as before. I left one question though I'd like to clarify before approving.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/auto/AutoConverter.kt
MiSikora: Yeah, this is a nice change. I wanted to do it while working on unifying artwork loading but forgot. Thanks!

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/auto/AutoConverter.kt
jaredkotoff: Can you compare with a few other podcast apps? I noticed that Pocket Casts uses an image that's quite low quality in many places compared to other apps.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/MediaSessionManager.kt
geekygecko: We only send the URL artwork when the user has chosen to show it on the lock screen, so maybe this should be the same.

MiSikora: I think it might be a little bit confusing if artwork on Tesla reacts to the lock screen setting. Personally, I would expect it to behave similarly to the media notification center. But I'm not sure if there is a good way to do that. I think we can proceed with this fix since having something is for sure better than nothing. 😅

MiSikora (approved):

PR #2261, opened 2024-05-21
Open clip links in a web view

MiSikora:
## Description

Since our app doesn't have capabilities to play clips of episodes we need to redirect to a web view. And web view allows to play the full episode in the app.

Internal refs: pdeCcb-3RS-p2#comment-4181, pdeCcb-3RS-p2#comment-4169 

## Testing Instructions

> [!note]
> Before starting the test make sure in the settings that the app is able to intercept the links.

Screenshot_20240521-092556

1. Install the app and sign in.
2. Close the app.
3. Send intent via ADB with start and end timestamps. For example: `adb shell am start -a android.intent.action.VIEW -d https://pca.st/y0ebrkcc\?t=20,60`.
4. App should intercept the link and open it in a web view.
5. You should be able to play the clip.
6. Tap on the `Full episode in Pocket Casts` button.
7. The episode should start playing in the app from the start timestamp.

## Screenshots or Screencast 

https://github.com/Automattic/pocket-casts-android/assets/30936061/24cd1ee9-ab6f-430e-b5fe-cd9276e7b4ac

## Checklist
 
- [x] ~with different themes~

### DIFF
--- a/modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/activity/WebViewActivity.kt
+++ b/modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/activity/WebViewActivity.kt
@@ -82,7 +82,7 @@ class WebViewActivity : AppCompatActivity(), CoroutineScope {
                     return false
                 }
                 val parsedUri = Uri.parse(url)
-                return if (parsedUri != null && !uriIsInternal(parsedUri)) {
+                return if (parsedUri != null) {
                     val intent = Intent(Intent.ACTION_VIEW)
                     intent.data = Uri.parse(url)
                     try {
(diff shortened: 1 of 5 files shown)


modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/activity/WebViewActivity.kt
```diff
@@ -115,11 +115,6 @@ class WebViewActivity : AppCompatActivity(), CoroutineScope {
         }
     }
 
-    private fun uriIsInternal(url: Uri): Boolean {
-        val host = url.host ?: return false
-        return INTERNAL_HOSTS.count { host.endsWith(it) } > 0
-    }
```
MiSikora, line 121: For more context see [this PR](https://github.com/Automattic/pocket-casts-android-private/issues/2558). This is most likely no longer necessary since we don't support that feature.

geekygecko (approved):
Looks good.

Should we make the toolbar match the theme? Here I'm using the light theme and the toolbar is dark.

Screenshot_20240522_211121

MiSikora:
> Should we make the toolbar match the theme? Here I'm using the light theme and the toolbar is dark.

We could but I think it would be a better fit for a separate task since it is about theming the web view activity and it might affect other parts of the app that use it.

PR #2267, opened 2024-05-22
Get Exoplayer cache size from remote config

ashiagr:
## Description
This gets the Exoplayer cache size from remote config (increased from 50 to 100 MB).

## Testing Instructions

1. Search for `setMinimumFetchIntervalInSeconds`(...) in the code and set to 5s
2. Add a LogBuffer.i(...) just before line 29 in `ExoPlayerCacheUtil` to print value for `FeatureFlag.isEnabled(Feature.CACHE_PLAYING_EPISODE)` and `firebaseRemoteConfig
                .getLong(FirebaseConfig.EXOPLAYER_CACHE_SIZE_IN_MB)`
4. Install release build
5. Open the app
6. Play an episode
7. Open `Profile` -> `Settings` -> `Help & feedback` -> `Logs`
8. ✅ Notice that correct values are printed
9. Uninstall the build

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/config/FirebaseConfig.kt
+++ b/modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/config/FirebaseConfig.kt
@@ -12,6 +12,7 @@ object FirebaseConfig {
     const val SLUMBER_STUDIOS_YEARLY_PROMO_CODE = "slumber_studios_yearly_promo_code"
     const val SLEEP_TIMER_DEVICE_SHAKE_THRESHOLD = "sleep_timer_device_shake_threshold"
     const val REFRESH_PODCASTS_BATCH_SIZE = "refresh_podcasts_batch_size"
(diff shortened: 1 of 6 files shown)


modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/config/FirebaseConfig.kt
```diff
@@ -20,6 +22,7 @@ object FirebaseConfig {
         CLOUD_STORAGE_LIMIT to 10L,
         SLEEP_TIMER_DEVICE_SHAKE_THRESHOLD to 30L,
         REFRESH_PODCASTS_BATCH_SIZE to 200L,
+        EXOPLAYER_CACHE_SIZE_IN_MB to if (BuildConfig.DEBUG) 100L else 0L,
```
ashiagr, line 25: Not sure if it is needed.

I explicitly added a default value for debug build here as the feature may be enabled even when Firebase config could not be fetched, in this case, we do not want the cache size to be 0.

PR #2283, opened 2024-05-29
Configure Lint to fail builds

MiSikora:
## Description

Our Lint did effectively nothing. This changes configuration to treat warnings as errors and fail builds on CI. I also added baseline profiles for our current Lint issues so we don't have to deal with them right now.

## Testing Instructions

Review only 55440d1ca0df6d5307b3d45d4586e3f05f338371 commit.

a8d0383ee70e0925a98d2f0a7f76aea3f4220249 Should fail Lint CI check.

## Checklist
- [x] ~Ensure the linter passes (`./gradlew spotlessApply` to automatically apply formatting/linting)~
 
- [x] ~with different themes~


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

ashiagr (approved):
LGTM. Nice job!

automotive/lint-baseline.xml
```diff
@@ -0,0 +1,289 @@
...
+        <location
+            file="$GRADLE_USER_HOME/caches/transforms-3/e02c82fc04458c4a3fac73d2cfdcf943/transformed/runtime-release/jars/lint.jar"/>
+    </issue>
+
+    <issue
+        id="InvalidPackage"
+        message="Invalid package reference in library; not included in Android: `java.awt`. Referenced from `com.sun.jna.Native.AWT`.">
+        <location
+            file="$GRADLE_USER_HOME/caches/transforms-3/61f2ace0377fec944d0761aa7634924b/transformed/jna-5.13.0/jars/classes.jar"/>
+    </issue>
```
wzieba, line 16: I think we have to fix this - the baseline shouldn't reference a specific hash key. For me, after updating a dependency, the started to fail.

MiSikora: Do you know where does the hash come from? If it is constant for a dependency than it is correct that it failed for you. The baseline did its job - it ignored the rule as long as it matched the conditions. For me it is the same thing as baseline failing after moving failing code to another file. Baseline will correctly no longer detect it. At which point we need to either decide whether we want to update the baseline or fix the issue that breaks the rule.

wzieba: It's not constant for a dependency - I haven't updated jna and the lint failed on this rule. I don't think we should rely on cache keys as it's internal Gradle implementation detail, that might include many factors like cache validity period.

I've tried `lint.absolutePaths false` and bumping AGP to 8.4.1 (to use newer lint). So far nothing helps.

PR #2315, opened 2024-06-06
Move chapter selection to chapter manager

MiSikora:
## Description

This PR moves chapter selection from `PlaybackManager` to `ChapterManager` and make it transactional. I'm doing it in a preparation for adding chapters to episode details. 

## Testing Instructions

Test chapter selection. Deselect and select different chapters, check syncing, check skipping chapters in the player, etc.

## Checklist
 
- [x] ~with different themes~

### DIFF
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/ChapterDao.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/ChapterDao.kt
@@ -5,6 +5,12 @@ import androidx.room.Insert
 import androidx.room.OnConflictStrategy
 import androidx.room.Query
 import androidx.room.Transaction
+import androidx.room.Update
+import au.com.shiftyjelly.pocketcasts.models.entity.BaseEpisode
+import au.com.shiftyjelly.pocketcasts.models.entity.ChapterIndices
+import au.com.shiftyjelly.pocketcasts.models.entity.PodcastEpisode
+import au.com.shiftyjelly.pocketcasts.models.entity.UserEpisode
+import java.util.Date
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.distinctUntilChanged
 import au.com.shiftyjelly.pocketcasts.models.to.DbChapter as Chapter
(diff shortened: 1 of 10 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/ChapterDao.kt
```diff
@@ -46,4 +52,34 @@ abstract class ChapterDao {
...
+        val episode = findPodcastEpisode(episodeUuid) ?: findUserEpisode(episodeUuid) ?: return
+        val deselectedChapters = episode.deselectedChapters
+        episode.deselectedChapters = when {
+            select && chapterIndex in deselectedChapters -> ChapterIndices((deselectedChapters - chapterIndex).distinct())
+            !select && chapterIndex !in deselectedChapters -> ChapterIndices((deselectedChapters + chapterIndex).distinct())
+            else -> return
+        }
+        episode.deselectedChaptersModified = modifiedAt
+        update(episode)
+    }
```
geekygecko, line 67: Sorry, I might be misunderstanding this. Why does this need to be in a transaction if it's only updating one episode? Would it be safer to update only the episodes `deselectedChapters` and `deselectedChaptersModified` columns, rather than every column?

geekygecko (approved):

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/ChapterDao.kt
MiSikora: > Why does this need to be in a transaction if it's only updating one episode?

It's a series of operations. First read then write.

> Would it be safer to update only the episodes `deselectedChapters` and `deselectedChaptersModified` columns, rather than every column?

What risk do you see?

geekygecko: > What risk do you see?
If the playback progress changed while this method was happening would it be possible to overwrite the playback progress to a different value?

MiSikora: > If the playback progress changed while this method was happening would it be possible to overwrite the playback progress to a different value?

No, because it is in transaction.

geekygecko: No need to change it. Thanks for explaining.

PR #2378, opened 2024-06-19
Cache playing episode feature flag + advanced setting

ashiagr:
## Description

This is a preparatory PR  that

1. Adds feature flag for caching the entire streaming episode to minimize streaming while playing (this is different from on-the-fly caching introduced earlier).
2. Adds advanced setting if the above feature flag is enabled (it is disabled by default, we can revisit the default state, and summary text later).
3. Extracts data source factory to helper singleton so that they can be reused later to cache entire episode.

## Testing Instructions
1. Install the app
2. Go to `Profile` -> `Settings`-> `Advanced`
3. ✅ Notice there is an advanced setting to enable caching the playing episode 
4. Toggle the setting
5. ✅ Notice that `settings_advanced_cache_entire_playing_episode` is tracked
6. Go to `Profile` -> `Settings`->  `Beta features` 
7. ✅ Disable "Cache entire playing episode" feature
8. Go to `Profile` -> `Settings`-> `Advanced`
9. ✅ Notice there is no advanced setting to enable caching the playing episode 

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/SettingsImpl.kt
+++ b/modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/SettingsImpl.kt
@@ -194,6 +194,12 @@ class SettingsImpl @Inject constructor(
         sharedPrefs = sharedPreferences,
     )
 
+    override val cacheEntirePlayingEpisode = UserSetting.BoolPref(
+        sharedPrefKey = "cacheEntirePlayingEpisode",
+        defaultValue = false,
+        sharedPrefs = sharedPreferences,
+    )
+
     override fun setSelectPodcastsSortType(sortType: PodcastsSortType) {
         sharedPreferences.edit().apply {
             putString(Settings.PREFERENCE_SELECT_PODCAST_LIBRARY_SORT, sortType.clientId.toString())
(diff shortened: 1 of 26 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please add tests for class <code>ExoPlayerHelper</code> (or add <code>unit-tests-exemption</code> label to ignore this).</td>
    </tr>
  </tbody>
</table>

</p>

modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/SettingsImpl.kt
```diff
@@ -194,6 +194,12 @@ class SettingsImpl @Inject constructor(
         sharedPrefs = sharedPreferences,
     )
 
+    override val cacheEntirePlayingEpisode = UserSetting.BoolPref(
+        sharedPrefKey = "cacheEntirePlayingEpisode",
+        defaultValue = false,
+        sharedPrefs = sharedPreferences,
+    )
```
MiSikora, line 201: It will entirely depend on a PR that depends on this setting but I think it would be safer to have a custom adapter for this setting that always returns `false` if the feature flag is disabled. Otherwise we might get into a situation where users enable the flag, we disable the feature, and they can't come back to the default state because option is not available in the settings.

MiSikora (approved):

modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/SettingsImpl.kt
ashiagr: That's a good point. I updated it in [bc9450c](https://github.com/Automattic/pocket-casts-android/pull/2378/commits/bc9450c2fb38cef4cbcc9f305e6cc0dcdbe5c1f6).

Issue #2149, opened 2024-05-01
Remove flickering when double-tapping an item in the tab bar


david-gonzalez-a8c:
### Description

The content strangely flickers when repeatedly tapping any item in the tab bar. I wish the items were unactionable if active (this seems to be the most typical behaviour amongst Android apps).

### Step-by-step reproduction instructions

1. Repeatedly tap any item in the tab bar

### Screenshots or screen recording

https://github.com/Automattic/pocket-casts-android/assets/92024441/cc8c1054-fc14-4abc-a884-40564ea020ce

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

Google Pixel 7a, Pocket Casts 7.63-rc 2

joashrajin:
8261507-zd-a8c

OutshineIssue:
I noticed in #2356 that the fix was rolled back. The flickering is a real bug and hard to ignore, it should take priority over some nice-to-have features that disappeared because of the PR. @mebarbosa could you reconsider it? And later on bring those missing features back.

PR #2381, opened 2024-06-19
Track `up_next_shown ` when tapping on navigation bar

### LINKED ISSUE #2363: [Up Next] - `up_next_shown` track event not fired when tap on up next bottom navigation
### Description

Currently we have this event

up_next_shown | When the up next view is displayed to the user | Android- source: [String] [mini_player, now_playing, player, up_next_shortcut] - The name of the view that is displaying the up next view
-- | -- | --

But this one is not being fired when taps on Up Next from bottom navigation

**Expected behavior**: We need to track this event only once like we are doing for Discover tab
- See: p1718725092247959-slack-C06EHRAPW12
- See: https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?gid=0#gid=0&range=A246

Tap on Up Next from bottom navigation bar and see that `up_next_shown` does not fire this event

_No response_

_No response_

mebarbosa:
## Description
- This PR fires `up_next_shown` event with `tab_bar` source when tapping on navigation bar 
- NOTE: Both discover and up next navigation bar button should trigger these `up_next_shown` and `discover_shown` events once, but due the opened issue we have https://github.com/Automattic/pocket-casts-android/issues/2149 we are recreating the fragment every time we tap on navigation bar so the event will be fired every time the navigation is tapped

Fixes #2363

## Testing Instructions
1. Open the app
2. Tap on Up Next bottom navigation bar button
3. ✅ Ensure `🔵 Tracked: up_next_shown, Properties: {"source":"tab_bar"` is tracked

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/UpNextFragment.kt
+++ b/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/UpNextFragment.kt
@@ -190,6 +190,9 @@ class UpNextFragment : BaseFragment(), UpNextListener, UpNextTouchCallback.ItemT
         val binding = FragmentUpnextBinding.inflate(themedInflater, container, false).also {
             realBinding = it
         }
+        if (upNextSource == UpNextSource.UP_NEXT_TAB) {
+            analyticsTracker.track(AnalyticsEvent.UP_NEXT_SHOWN, mapOf("source" to "tab_bar"))
+        }
         return binding.root
     }
 


modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/UpNextFragment.kt
```diff
@@ -190,6 +190,9 @@ class UpNextFragment : BaseFragment(), UpNextListener, UpNextTouchCallback.ItemT
         val binding = FragmentUpnextBinding.inflate(themedInflater, container, false).also {
             realBinding = it
         }
+        if (upNextSource == UpNextSource.UP_NEXT_TAB) {
```
mebarbosa, line 193: I added this check because I noticed that when we open the up next from another place that is not navigation bar this event is also tracked.

ashiagr (approved):

Issue #2405, opened 2024-06-27
App doesn't handle direct episode links with timestamps


MiSikora:
### Description

Sharing something like `https://pca.st/episode/c510a66f-f9c0-45c8-bca2-8c0cc1d5f055?t=1327` with the app doesn't recognize the `1327` timestamp. It is because the link is treated as a regular episode link due to 2 path segments here:

https://github.com/Automattic/pocket-casts-android/blob/9cf0a90a2910d455900db0347047725fb6454d36/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt#L1548-L1564

The reason why something like `https://pca.st/9mjo6ijv?t=1327` works is because it is not a direct and the `/episode/` path is not there.

### Step-by-step reproduction instructions

1. Run `adb shell am start -a android.intent.action.VIEW -d https://pca.st/episode/c510a66f-f9c0-45c8-bca2-8c0cc1d5f055\?t=1327`
2. Notice that episode details show but the playback from `1327` timestamp doesn't start.

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

Google Pixel 6, Android 14, 9cf0a90a2910d455900db0347047725fb6454d36

PR #2406, opened 2024-06-27
Add clip selector

MiSikora:
## Description

This PR adds clip selection to clip sharing. After several iterations, I ultimately used `LazyRow` as the UI element for the timeline. Initially, I experimented with `SubcomposeLayout`, but I soon realized that I would be implementing something very similar to `LazyRow`, which would likely be not as optimized. Although it might be possible to create a more efficient custom layout, I am satisfied with the current design.

The UI is divided into two main components: `ClipTimeline` and `ClipBox`. `ClipTimeline` is responsible for displaying ticks, while `ClipBox` handles the clipping window. They communicate via `ClipSelectorState`, which is remembered in the `ShareClipPage`, a higher-level component. This setup preserves the state of zoom, scroll offset, etc., during configuration changes. The state also manages the translation between pixels and duration, and vice versa.

Double-tapping is implemented in `ClipTimeline` instead of `ClipBox` for a key reason. Adding it to `ClipBox` makes `ClipTimeline` unresponsive to scroll events. While this could be addressed by adding another `.pointerInput()` that delegates to scroll state, my testing found it a bit janky. For example, it wasn't always 100% responsive, and simple delegation lost nice features like flings or bounce animations.

The zoom gesture is somewhat problematic, but I don't know if we can improve it. I'm currently delegating to `.transformable()`, which handles zooming. The issue seems to stem from `LazyRow` also supporting scrolling, which interferes with zooming.

Designs: [Figma](https://www.figma.com/design/HR2vcQrWcS0scsolKZBitg/Sharing?node-id=2203-3599&t=YWEw1W0PZCsN0nLE-4)

Next steps:
* Handle playback progress.
* Manage the initial clip range depending on where the clip was started.
* Implement analytics and logging.
* Accessibility.

## Testing Instructions

1. Start clip sharing.
2. Verify that the clip window follows the timeline scroll position.
3. Ensure you can drag handles on the timeline.
4. Check that double-tapping on the timeline moves handles to the tapped position, moving the nearest handle.
5. Confirm that you can zoom in and out on the timeline. The clip box should stay locked to the same time marks when zooming. If an episode is very short, like a couple of seconds, zooming out is intentionally impossible to prevent very narrow timelines displayed.
6. Share the clip and verify that the timestamps are correct. Note that we support 1-second resolution, even though the UI might suggest otherwise. This can be adjusted in the future. Be aware that you won't be able to open this link in the app due to issue #2405. You can test them in the web.
7. Check if the box and timeline state is preserved during configuration changes, such as device rotation.

### Demo

https://github.com/Automattic/pocket-casts-android/assets/30936061/c19689fb-e786-4cc6-92f5-02bc462ff806

### Scrolling

scroll-ezgif com-crop

### Dragging

drag-ezgif com-crop

### Double tapping

tap-ezgif com-crop

### Zooming

zoom-ezgif com-crop

## Checklist
 
- [x] ~with different themes~


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please add tests for class <code>ClipSelectorState</code> (or add <code>unit-tests-exemption</code> label to ignore this).</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

modules/features/clip/src/main/kotlin/au/com/shiftyjelly/pocketcasts/clip/ClipSelector.kt
```diff
@@ -57,24 +89,289 @@ internal fun ClipSelector(
...
+    val localDensity = LocalDensity.current
+    val largeTickHeight = maxHeight / 3
+    val mediumTickHeight = maxHeight / 6
+    val smallTickHeight = maxHeight / 12
+
+    val tickWidth = 1.5.dp
+    val spaceWidth = 4.dp * state.scale
+    val itemWidth = tickWidth + spaceWidth
+    state.itemWidth = with(localDensity) { itemWidth.toPx() }
+    LaunchedEffect(state.scale) { state.scaleBoxOffsets(clipRange) }
```
MiSikora, line 163: I'm not super happy about this part but I think it is a chicken and egg problem. I need `itemWidth` to correctly translate duration into pixels but an actual value is not available before the view is laid out. So I just update it whenever possible and let Compose handle the rest by propagating the state.

modules/features/clip/src/main/kotlin/au/com/shiftyjelly/pocketcasts/clip/ClipSelector.kt
```diff
@@ -57,24 +89,289 @@ internal fun ClipSelector(
...
+                Box(
+                    modifier = Modifier
+                        .width(2.dp)
+                        .height(this@ClipBox.maxHeight / 2)
+                        .clip(RoundedCornerShape(1.dp))
+                        .background(clipColors.selectorHandleColor),
+                )
+            }
+        }
+        // Outer box to increast touch area of the handle
```
MiSikora, line 308:
```suggestion
        // Outer box to increas the touch area of the handle
```

modules/features/clip/src/main/kotlin/au/com/shiftyjelly/pocketcasts/clip/ClipSelector.kt
```diff
@@ -57,24 +89,289 @@ internal fun ClipSelector(
...
+                GenericShape { size, _ ->
+                    val baseRect = size.toRect()
+                    val adjustedRect = baseRect.copy(left = baseRect.left - handleWidthPx)
+                    addRect(adjustedRect)
+                },
+            )
+            .offset { IntOffset(-scrollOffset.roundToInt(), 0) }
+            .fillMaxSize(),
+    ) {
+        // Outer box to increast touch area of the handle
```
MiSikora, line 272:
```suggestion
        // Outer box to increase the touch area of the handle
```

modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/FlowUtil.kt
```diff
@@ -0,0 +1,24 @@
+package au.com.shiftyjelly.pocketcasts.utils
+
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.combine as kotlinCombine
+
+public fun <T1, T2, T3, T4, T5, T6, R> combine(
```
ashiagr, line 6: Can we reuse existing flow extension: `fun <T1, T2, T3, T4, T5, T6, R> combine6` instead of creating a new one?

modules/features/clip/src/main/kotlin/au/com/shiftyjelly/pocketcasts/clip/ClipSelector.kt
```diff
@@ -57,24 +89,290 @@ internal fun ClipSelector(
...
+    val spaceWidth = 4.dp * state.scale
+    val itemWidth = tickWidth + spaceWidth
+    state.itemWidth = with(localDensity) { itemWidth.toPx() }
+    LaunchedEffect(state.scale) { state.scaleBoxOffsets(clipRange) }
+
+    val maxTickResolution by remember {
+        val secondsCount = episodeDuration.inWholeSeconds.toInt()
+        var resolution = 1
+        val minItemWidth = 1.5.dp + 4.dp
+        while (resolution < 125) {
```
ashiagr, line 169: Can we add an explanatory comment here?

ashiagr (approved):
This is so awesome!

I tested the following and enjoyed every part of the interaction 🌟 

- Scrolling ✅ 
- Dragging clip box edges ✅ 
- Double tap gesture ✅ 
- Colors ✅ 
- Landscape layout ✅ 
- Remember the last state on orientation changes ✅ 
- Large display ✅ 

Just a couple of findings: 

1. Grey box at the beginning:

When the clip box edge is moved to the start of the timeline, and  timeline view is scrolled, I see a small grey box:

https://github.com/Automattic/pocket-casts-android/assets/1405144/316ad219-94c5-4f21-adc2-c7afa85d4f04

2. In Figma, I can see the current position in the timeline view, do we plan to add it in a later PR?

current position

modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/FlowUtil.kt
MiSikora: Oh didn't notice we have one. Would you mind if I removed `combine6()` and keep this one instead? I'd prefer to follow Kotlin's naming.

modules/features/clip/src/main/kotlin/au/com/shiftyjelly/pocketcasts/clip/ClipSelector.kt
MiSikora: Sure, I'll add it.

modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/FlowUtil.kt
ashiagr: yep, you can remove combine6

MiSikora:
> When the clip box edge is moved to the start of the timeline, and timeline view is scrolled, I see a small grey box:

Yes, this is expected. The start handle is offset so that its end position aligns with a time mark. When you scroll to the left, the handle moves out of the screen, but the connecting frame remains visible. If the frame were clipped to the timeline edge, there would be an empty space between the handle and the frame when you start scrolling.

> In Figma, I can see the current position in the timeline view, do we plan to add it in a later PR?

Yes, I plan to address it in one of the next PRs.

modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/FlowUtil.kt
MiSikora: Fixed in 21a3f2eb08c0162bdfb455b78e667877864d7f6e

modules/features/clip/src/main/kotlin/au/com/shiftyjelly/pocketcasts/clip/ClipSelector.kt
MiSikora: Fixed in 0eb6eddbc0c05c773eeeabed7bb101aed2eb9fd5

Issue #2427, opened 2024-07-02
Playback Speed: Add 1.0x to the options when clicking on the speed


joashrajin:
### Description

If you click on the playback speed (without clicking on the '+' or '-' icons), 1.0x is missing, and 1.5x is the lowest it goes. You need to manually use the '-' icon to get it down to 1.0x

### Step-by-step reproduction instructions

1. Play an episode and attempt to adjust the playback speed
2. Click on the speed itself to have it adjusted by 0.5 increments
3. Notice that once you go past 4.0x, it reverts to the lowest being 1.5x

### Screenshots or screen recording

https://github.com/Automattic/pocket-casts-android/assets/16253818/726a2a68-569d-407a-afca-6f5d8e506368

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

Samsung Fold 4, Android 14, Pocket Casts v7.67-rc-2

joashrajin:
https://forums.pocketcasts.com/forums/topic/please-reincorporate-the-middle-click-playback-speed-1-0-function

joashrajin:
8405492-zd-a8c

joashrajin:
8404063-zd-a8c

azollafox:
8416866-zen

azollafox:
8420154-zen

mdrockwell:
8407169-zd-a8c

PR #2439, opened 2024-07-03
Revert playback speed label taps

geekygecko:
## Description

The recent change to add 5x playback speed changed the behaviour of tapping the speed label. We have received lots of negative user feedback, so until we figure out a better way, this will restore the original functionality. 

Fixes https://github.com/Automattic/pocket-casts-android/issues/2427

## Testing Instructions
1. Play an episode
3. Tap on the effects icon
4. Tap the + button to increase the speed to 1.2
5. Tap the speed label
6. ✅ Verify the speed returns to 1x
7. Tap the speed label
8. ✅ Verify the speed changes to 1.5x
9. Tap the speed label
8. ✅ Verify the speed changes to 2x 
Tap the speed label
8. ✅ Verify the speed changes to 1x 

## Screencast 

https://github.com/Automattic/pocket-casts-android/assets/308331/c52697a0-88a8-4965-b6a4-301eef0cd5c2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 *   Updates:
     *   Update Give rating view to include the average rating
         ([#2421](https://github.com/Automattic/pocket-casts-android/pull/2421))
+*   Bug Fixes
+    *   Fix playback speed label tap behavior
+        ([#2439](https://github.com/Automattic/pocket-casts-android/pull/2439))
 
 7.67
 -----
--- a/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/EffectsFragment.kt
+++ b/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/EffectsFragment.kt
@@ -232,13 +232,8 @@ class EffectsFragment : BaseDialogFragment(), CompoundButton.OnCheckedChangeList
             binding.btnClear.id -> viewModel.clearPodcastEffects(podcast)
             binding.lblSpeed.id -> {
                 when (effects.playbackSpeed) {
-                    in 0.0..<1.05 -> changePlaybackSpeed(effects, podcast, 1.5)
-                    in 1.05..<1.55 -> changePlaybackSpeed(effects, podcast, 2.0)
-                    in 1.55..<2.05 -> changePlaybackSpeed(effects, podcast, 2.5)
-                    in 2.05..<2.55 -> changePlaybackSpeed(effects, podcast, 3.0)
-                    in 2.55..<3.05 -> changePlaybackSpeed(effects, podcast, 3.5)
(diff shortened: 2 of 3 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

mebarbosa (changes requested):
Thanks for putting this PR! 

I've noticed that we still has support for `5x` speed from notification media center. Is it something we want to address on this PR? 

image

b-aaron:
One suggestion I have for this is being able to long press the speed label to get back to 1x. This would fix the amount of presses needed to get back quickly.

CookieyedCodes:
Ya predicted this would likely be an issue,
Which is why I designed a dial ui & preset manager https://github.com/Automattic/pocket-casts-android/discussions/1452#discussioncomment-7422777

But agreed long press going back to 1x or global default is a good fix that would help while the UI is a bit more work in to work on I get that 🤔

mebarbosa (approved):
Looks good to me! Thanks for the change

geekygecko:
> One suggestion I have for this is being able to long press the speed label to get back to 1x.

Thanks for the suggestion, I do wonder if a long press is too hidden.

> Which is why I designed a dial ui & preset manager

I like that it's easy to switch between the playback speeds! I was wondering about having quick adjust buttons, but they might not work on narrow phones. 
-1, -0.5, -0.1, (reset, show current speed), +0.1, +0.5, +1

CookieyedCodes:
No bothers :)

Hidden yes but tbh better then nothing and it's not like your app dosen't have hidden - functionality -

Oh that's not a bad idea 🤔🤔, still wouldn't mind a speed preset manager tbh but that also wouldn't be toto bad as an in-between 

                     -- -.1 [1.0x] +.1 + ++

Hmmmmm if it's on a new line instead of numbers below a certain screen size you could do smaller symbols 🤔🤔,
I do agree beyond 3x+ should only be button tho until a potential diall UI could be done 🤔🤔 some analytics should be assed for your smallest android devices and I suppose watch os need to be considered too 🤔 

Got to love the the complexity of even buttons 😅

CookieyedCodes:
Another suggestion that might be easier actually 

Is 
-- - [1.0x] + ++
& Make the -- & ++ user customisable 🤔🤔

MtndaleRedneck:
An idea is a setting to be able to set custom speed change defaults, meaning there'll be like 5 buttons below the speed adjustment that you can assign various speeds to those buttons. I have a few speeds that I tend to use for podcasts and have no need for all the speeds in between. Also long pressing the speed button should revert it to 1x.

PR #2468, opened 2024-07-08
Transcripts - Style cues into paragraphs

ashiagr:
## Description
This styles cues into paragraphs.

** I also refactored the code so the transcript is parsed only when the `Transcript` tab is selected. 

## Testing Instructions
1. Play an episode from "Cautionary Tales" (https://pcast.pocketcasts.net/8zpmjiru)
2. Open full screen player
3. Go to the `Transcript` tab
4. ✅ Notice that the transcript is styled as paragraphs

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/transcripts/TranscriptViewModel.kt
+++ b/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/transcripts/TranscriptViewModel.kt
@@ -9,57 +9,77 @@ import androidx.media3.extractor.text.CuesWithTiming
 import androidx.media3.extractor.text.SubtitleParser
 import au.com.shiftyjelly.pocketcasts.models.entity.Podcast
 import au.com.shiftyjelly.pocketcasts.models.to.Transcript
+import au.com.shiftyjelly.pocketcasts.repositories.di.IoDispatcher
 import au.com.shiftyjelly.pocketcasts.repositories.playback.PlaybackManager
 import au.com.shiftyjelly.pocketcasts.repositories.podcast.TranscriptsManager
 import au.com.shiftyjelly.pocketcasts.utils.UrlUtil
 import com.google.common.collect.ImmutableList
 import dagger.hilt.android.lifecycle.HiltViewModel
 import javax.inject.Inject
+import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.ExperimentalCoroutinesApi
-import kotlinx.coroutines.flow.SharingStarted
+import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.flatMapLatest
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.stateIn
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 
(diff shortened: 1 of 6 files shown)


modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/transcripts/TranscriptViewModel.kt
```diff
@@ -1,65 +1,92 @@
...
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.flatMapLatest
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.stateIn
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 
+// TODO: [Transcript] Modify SpeakerRegex for non english languages
+private val SpeakerRegex = Pattern.compile("Speaker \\d*: ")
+private val NewLineRegex = Pattern.compile("\\.$")
```
MiSikora, line 35: No change needed but using `"""<PATTERN>""".toRegex()` allows you to avoid escaping. So something like `Pattern.compile("Speaker \\d*: ")` becomes `"""Speaker d*: """.toRegex()`.

MiSikora, line 35: Should this also account for `?` and `!` characters?

And I think that similarly to the comment about speakers in non English languages this should also account for other languages like Japanese period `。`. This can be addressed in a separate PR but I wanted to note it.

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/transcripts/TranscriptViewModel.kt
```diff
@@ -1,65 +1,92 @@
...
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.flatMapLatest
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.stateIn
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 
+// TODO: [Transcript] Modify SpeakerRegex for non english languages
+private val SpeakerRegex = Pattern.compile("Speaker \\d*: ")
```
MiSikora, line 34: `\\d*` should probably be `\\d+` to indicate at least one digit. Though I think that in practice it won't make any difference.

MiSikora (approved):

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/transcripts/TranscriptViewModel.kt
ashiagr: Good point. Included ? and ! characters also for new line regex in [eb83ed0](https://github.com/Automattic/pocket-casts-android/pull/2468/commits/eb83ed00789b93a32439868ae600bb274679691f)

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/transcripts/TranscriptViewModel.kt
ashiagr: Applied it in [eb83ed0](https://github.com/Automattic/pocket-casts-android/pull/2468/commits/eb83ed00789b93a32439868ae600bb274679691f) even though it doesn't matter in a real world.

PR #2475, opened 2024-07-08
[Give Rating] Add rate podcast

mebarbosa:
## Description
- This PR adds the ability to rate a podcast calling the API
- See: p1720438636666779-slack-C077XU4GF9D

## Testing Instructions
1. Run the app in `debug`
2. Open a podcast
3. Listen to at least 2 episodes from this podcast (scroll to almost the end of the episode)
4. Tap to rate
5. If you have a success response you will see something like `{
  "podcastUuid": "3c5900ab-dc52-4af2-bed8-f488f3e565ce",
  "modifiedAt": "2024-07-08T23:12:13.658Z",
  "podcastRating": 3
}` as response. 
6. Ensure you see the toast with `Thank you for rating!` message
7. Otherwise, you should see a toast with `Could not rate this podcast` message

## Screenshots or Screencast 
[Screen_recording_20240708_201733.webm](https://github.com/Automattic/pocket-casts-android/assets/42220351/e272d3f5-b73a-46e5-b26f-43cef13be026)

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/components/ratings/GiveRatingFragment.kt
+++ b/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/components/ratings/GiveRatingFragment.kt
@@ -73,13 +73,19 @@ class GiveRatingFragment : BaseDialogFragment() {
                         onDismiss = ::dismiss,
                         setRating = viewModel::setRating,
                         submitRating = {
-                            viewModel.submitRating(
-                                context = context,
-                                onSuccess = {
-                                    Toast.makeText(context, "TODO: Submit rating", Toast.LENGTH_LONG).show()
-                                    dismiss()
-                                },
-                            )
+                            coroutineScope.launch {
+                                viewModel.submitRating(
+                                    podcastUuid = podcastUuid,
+                                    context = context,
+                                    onSuccess = {
+                                        Toast.makeText(context, getString(LR.string.thank_you_for_rating), Toast.LENGTH_LONG).show()
+                                        dismiss()
+                                    },
+                                    onError = {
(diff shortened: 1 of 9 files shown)


geekygecko:
It's not related to a change in this PR but when you go to rate a podcast for the first time it has 1 star selected as that is the minimum allow. I wonder if it would be nicer to have 0 stars selected and the button disabled until you select your rating. It feels like we are suggesting a 1 star rating.

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/components/ratings/GiveRatingFragment.kt
```diff
@@ -73,18 +73,26 @@ class GiveRatingFragment : BaseDialogFragment() {
...
-                                    Toast.makeText(context, "TODO: Submit rating", Toast.LENGTH_LONG).show()
-                                    dismiss()
-                                },
-                            )
+                            coroutineScope.launch {
+                                viewModel.submitRating(
+                                    podcastUuid = podcastUuid,
+                                    context = context,
+                                    onSuccess = {
+                                        Toast.makeText(context, getString(LR.string.thank_you_for_rating), Toast.LENGTH_LONG).show()
```
geekygecko, line 81: This is just an opinion but I think this would be nicer at a Snackbar instead of a Toast message. I don't think they need an action but having it in the app rather than float above the app is nicer. 

> Alternatives to using toasts
> If your app is in the foreground, consider using a [snackbar](https://material.io/components/snackbars) instead of using a toast. Snackbars include user-actionable options, which can provide a better app experience.

https://developer.android.com/guide/topics/ui/notifiers/toasts#:~:text=If%20your%20app%20is%20in,action%2C%20use%20a%20notification%20instead.

geekygecko (approved):

mebarbosa:
@geekygecko Thank you for your suggestions! I will address these in the next PR

PR #2485, opened 2024-07-11
Task/upload files crash fix

### LINKED ISSUE #2285: Files: Crashing when uploading files
### Description

We have had a few reports from users mentioning that uploading files to Pocket Casts causes the app to crash, or revert back to the previous screen, without adding the file

Internal Ref: p1716983632473029-slack-C02A333D8LQ

We have not been able to reproduce just yet, but we've had one user mention it was happening with a 40mb mp3 file

N/A

Samsung Galaxy S21 5G, App version: 7.64, Android 14 (SDK 34)

CalebKL:
## Description

This PR fixes file uploading crash. I have tried to reproduce this crash but I was unsuccessful after numerous attempts. Although after going through the logs, the crash was a permission error specifically for `WRITE_EXTERNAL_STORAGE` and `INTERNET` permissions. The `WRITE_EXTERNAL_STORAGE` was causing the `FileNotFoundException` and `INTERNET` permissions caused a connection error leading to a `StorageFileLoadException`. 
The PR adds checks to confirm that the permissions have been enabled and request permissions if they are not enabled. 

Fixes #2285 
## Screenshots or Screencast 

1. Sign in 
2. Go to profile 
3. Click files 
4. Add a new file

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/cloud/AddFileActivity.kt
+++ b/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/cloud/AddFileActivity.kt
@@ -1,7 +1,6 @@
 package au.com.shiftyjelly.pocketcasts.profile.cloud
 
-import android.Manifest.permission.INTERNET
-import android.Manifest.permission.WRITE_EXTERNAL_STORAGE
+import android.Manifest.permission.READ_EXTERNAL_STORAGE
 import android.app.Activity
 import android.content.ActivityNotFoundException
 import android.content.Context
@@ -90,8 +89,7 @@ private const val EXTRA_EXISTING_EPISODE_UUID = "fileUUID"
 private const val EXTRA_FILE_CHOOSER = "filechooser"
 private const val STATE_LAUNCHED_FILE_CHOOSER = "LAUNCHED_FILE_CHOOSER"
 private const val STATE_DATAURI = "DATAURI"
-private const val WRITE_STORAGE_PERMISSION_REQUEST_CODE = 50
-private const val INTERNET_PERMISSION_REQUEST_CODE = 51
+private const val READ_STORAGE_PERMISSION_REQUEST_CODE = 50
 
 @AndroidEntryPoint
 class AddFileActivity :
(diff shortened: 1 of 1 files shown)


MiSikora:
I'm sorry but the reasoning doesn't explain the #2285 issue. The stack trace shows that the crash happens for an input stream which means reading a file and not writing to one.

`WRITE_EXTERNAL_STORAGE` is needed for writing to an external shared storage. And since SDK 33 this permission is deprecated and instead file interactions should be handled via [createWriteRequest()](https://developer.android.com/reference/android/provider/MediaStore?utm_source=lint&utm_medium=lint&utm_campaign=lint#createWriteRequest(android.content.ContentResolver,%20java.util.Collection%3Candroid.net.Uri%3E)). This is something we should probably look into.

`INTERNET` permission doesn't make sense because it is a [normal permission](https://developer.android.com/guide/topics/permissions/overview#normal) granted at the installation time.

CalebKL:
> I'm sorry but the reasoning doesn't explain the #2285 issue. The stack trace shows that the crash happens for an input stream which means reading a file and not writing to one.
> 
> `WRITE_EXTERNAL_STORAGE` is needed for writing to an external shared storage. And since SDK 33 this permission is deprecated and instead file interactions should be handled via [createWriteRequest()](https://developer.android.com/reference/android/provider/MediaStore?utm_source=lint&utm_medium=lint&utm_campaign=lint#createWriteRequest(android.content.ContentResolver,%20java.util.Collection%3Candroid.net.Uri%3E)). This is something we should probably look into as a separate issue.
> 
> `INTERNET` permission doesn't make sense because it is a [normal permission](https://developer.android.com/guide/topics/permissions/overview#normal) granted at the installation time.

Thanks for the feedback, so we should check permissions for `READ_EXTERNAL_STORAGE` 
I will go ahead and remove the `INTERNET` permissions check. The reasoning to that was since we also had a connection error, we should have a check for the permission, but since its a normal permission we don't need that.

mukul-newsuk (approved):

CalebKL:
@MiSikora any update on this?

MiSikora:
The solution isn't correct. You're invoking the permission request after the file request has been submitted. Also, this permission is deprecated in SDK 33+ and doesn't work there. And most importantly this permission shouldn't be needed since we're not reading internal files when accessing them through the `android.intent.action.GET_CONTENT` Intent, which we do.

If you want to fix this issue I'm going to have to ask you to provide a reproducible scenario that is fixed with your changes.

MiSikora:
I'm closing this PR as there is no reproducible scenario that we could test this with and the proposed solution has issues that I mentioned above.

PR #2487, opened 2024-07-12
[Dependency Update] Update `a8c-ci-toolkit` to 3.4.2

ParaskP7:
## Description

This PR updates `a8c-ci-toolkit` to [3.4.2](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/releases/tag/3.4.2). To be more specific, this PR actually also migrates from the old `bash-cache` to the new `a8c-ci-toolkit` naming convention ([context](https://github.com/Automattic/pocket-casts-android/pull/2433#discussion_r1664015761)).

FYI: The breaking change in `3.0.0` is due to removing the `nvm_install` command, but this repo doesn't use that anywhere anyway. As such, it is safe to apply this update.

## Testing Instructions

- Make sure CI is green (🟢).

#### I have tested any UI changes... (`N/A`)

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,12 +1,6 @@
 # yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
 ---
 
-# Nodes with values to reuse in the pipeline.
-common_params:
-  # Common plugin settings to use with the `plugins` key.
-  - &common_plugins
-    - automattic/a8c-ci-toolkit#3.4.2
-
 agents:
   queue: "android"
 
@@ -14,7 +8,7 @@ steps:
   - label: "Gradle Wrapper Validation"
     command: |
       validate_gradle_wrapper
-    plugins: *common_plugins
+    plugins: [$CI_TOOLKIT]
 
   # Wait for Gradle Wrapper to be validated before running any other jobs
   - wait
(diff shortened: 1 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

AliSoftware (commented):
While this looks good, I'd love to use this opportunity to adopt `shared-pipeline-vars` in this repo.

It seems that it [the Buildkite pipeline already uses the default steps](https://github.com/Automattic/buildkite-ci/blob/trunk/src/buildkite/pipelines/pocket-casts-android.tf) and thus will already do the `source .buildkite/shared-pipeline-vars` if that file exists.
So it'd just be a matter of creating that file in the repo to define an env var for `CI_TOOLKIT` and use it in all the `.yml`—so you can then remove the `common_params` anchors and update the plugin version in a single place.

ParaskP7:
> While this looks good, I'd love to use this opportunity to adopt shared-pipeline-vars in this repo.

On it @AliSoftware , I was thinking about that too, but wasn't 100% sure if I should do it, thanks for the suggestion! 👍

ParaskP7:
FYI @AliSoftware that this is now done: 4a09e07

AliSoftware (approved):

PR #2500, opened 2024-07-23
Add missing latest episode info to the podcast

geekygecko:
## Description

When testing the podcast refresh we found that Android wasn't setting the latest episode uuid and publish date on the podcast when first signing in. This would mean that the latest episode uuid would be sent to the server as null. If a new episode was never released, then every time the server would return episodes. This will speed up the refresh for lots of users. Do you think I should add it to the change log?

Parameters before. I have replaced `%2C` with `,` to make it easier to read.

```
podcasts=bb793090-0424-012e-f9a0-00163e1b201c,2743d720-0edf-0133-2204-059c869cc4eb,04364440-031e-012e-f99b-00163e1b201c,4eb5b260-c933-0134-10da-25324e2a541d,b1090430-da7a-0135-9e62-5bb073f92b78
last_episodes=null,null,null,null,null
```

Parameters after

```
podcasts=bb793090-0424-012e-f9a0-00163e1b201c,2743d720-0edf-0133-2204-059c869cc4eb,04364440-031e-012e-f99b-00163e1b201c,4eb5b260-c933-0134-10da-25324e2a541d,b1090430-da7a-0135-9e62-5bb073f92b78
last_episodes=1fb6e853-aabf-4104-b0f7-a98612628234,68cf7e90-303b-4806-bcca-9c54e823b5a3,209753f1-1426-4bd7-952e-d436d8108ad5,1cf06916-3eb6-4f20-b5a6-2116db33007b,0e4601ee-f35b-4196-8932-8900401756e2
```

## Testing Instructions

1. Sign up for a new Pocket Casts account
2. Subscribe to some podcasts
3. Sign out and clear the storage
4. Sign in with the account
5. Filter logcat for last_episodes
6. Go to the Profile tab and tap Refresh
7. ✅ Verify that the episode uuids are not null

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 *   Updates
     *   Episodes shared from the player are updated with the playback status.
         ([#2499](https://github.com/Automattic/pocket-casts-android/pull/2499))
+    *   Improved the podcast refresh performance.
+        ([#2500](https://github.com/Automattic/pocket-casts-android/pull/2500))
 *   New Features
     *   Add an option in your account settings to modify your avatar with Gravatar
         ([#2263](https://github.com/Automattic/pocket-casts-android/pull/2263))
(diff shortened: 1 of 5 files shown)


MiSikora (approved):
Nice! I've ran into it when working on podcast refresh batching and forgot about it.

> Do you think I should add it to the change log?

I think it is a good idea..

PR #2539, opened 2024-07-30
Bump media3 lib

ashiagr:
## Description
This bumps `media3` lib from `1.2.1` to the latest stable release `1.4.0`.

Release notes:
[1.4.0](https://github.com/androidx/media/releases/tag/1.4.0)
[1.3.1](https://github.com/androidx/media/releases/tag/1.3.1)
[1.3.0](https://github.com/androidx/media/releases/tag/1.3.0)

I'm mainly interested in getting this fix from [1.3.1](https://github.com/androidx/media/releases/tag/1.3.1#:~:text=Text%3A,from%20WebvttParser.parse) but it should be good to update to the latest version:
> Text:
> WebVTT: Prevent directly consecutive cues from creating spurious additional CuesWithTiming instances from WebvttParser.parse.

I could reproduce the bug on the default `WebVTT` format transcript for any episode from `The Audacity to Podcast`.

## Testing Instructions
- CI should be green
- Smoke test playback

** No need to test transcript yet, I'll create a separate PR with fixes for different formats. 

## Checklist
 
- [ ] with different themes


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

MiSikora (changes requested):
Trim silence stopped working for me. I tested with this podcast  https://open.spotify.com/show/7KImr4BaJgg2KAOtgke48p

ashiagr:
I suppressed missing super call lint warning in 0c7c6cf to fix it. Let me know if you can suggest a better way to fix it.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/ShiftyTrimSilenceProcessor.kt
```diff
@@ -76,7 +77,8 @@ class ShiftyTrimSilenceProcessor(
         return if (enabled) inputAudioFormat else AudioFormat.NOT_SET
     }
 
-    override fun isActive(): Boolean = super.isActive() && enabled
+    @SuppressLint("MissingSuperCall")
```
MiSikora, line 80: Could you add a comment with a reason for suppressing it?

MiSikora (approved):

MiSikora:
> Let me know if you can suggest a better way to fix it.

I'm not sure. From looking at the API it looks like we're not setting `pendingOutputAudioFormat`. Which normally is set in `configure()` but we don't call `super()` in the `ShiftyTrimSilenceProcessor`. It probably has something to do with it.

```java
@Override
@CanIgnoreReturnValue
public final AudioFormat configure(AudioFormat inputAudioFormat) throws UnhandledAudioFormatException {
  pendingInputAudioFormat = inputAudioFormat;
  pendingOutputAudioFormat = onConfigure(inputAudioFormat);
  return isActive() ? pendingOutputAudioFormat : AudioFormat.NOT_SET;
}
```

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/ShiftyTrimSilenceProcessor.kt
ashiagr: Good point. Added in [f7aae59](https://github.com/Automattic/pocket-casts-android/pull/2539/commits/f7aae59bb347cf521c4a903eb660566ea4711461).

MiSikora:
Ohhh, I think I know what it is. Because we return `AudtioFormat` from `onCofigure()` only when `enabled` is set to `true` we get into a chicken and an egg problem. We do not set `pendingOutputAudioFormat` because `enabled` is initially `false` and `isActive()` returns `false` because `pendingOutputAudioFormat` is not set. I think it should be ok to keep the logic as with with suppression.

PR #2554, opened 2024-08-01
Transcripts - Add select text capability

ashiagr:
## Description

This adds select text capability.

Default text selection  in Compose Text view only allows `Copy` option: https://issuetracker.google.com/issues/235259768

I added a custom menu `Select All` to select all text and `Share` for older devices (API level < 33). In newer devices,  `Copy` option allows you to share copied text.

EDIT: `Select All` is added to [Compose 1.7.0-alpha08](https://developer.android.com/jetpack/androidx/releases/compose-foundation#1.7.0-beta06:~:text=Added%20%22Select%20all%22%20to%20all%20text%20contextual%20menus%20in%20SelectionContainer.). We will not need this custom action once 1.7.0 gets stable and we switch to it.

## Testing Instructions
1. Play an episode with transcript
2. Open full-screen player
3. Tap on the transcript bottom shelf icon
4. Long-press on the transcript text
5. ✅ Notice that text is selected with a text toolbar to copy, and select all for newer devices (API level >=33 and copy, share, and select all for older devices where copy does not allow you to share copied text.
 
## Screenshots or Screencast 

API 33

https://github.com/user-attachments/assets/4fde2850-b12b-4800-b418-e85405d6519a

API 32

https://github.com/user-attachments/assets/ea84a74d-09f7-4057-bf3c-a572a48721a4

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/transcripts/TranscriptPage.kt
+++ b/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/transcripts/TranscriptPage.kt
@@ -17,11 +17,14 @@ import androidx.compose.foundation.verticalScroll
 import androidx.compose.material.MaterialTheme
 import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.platform.LocalConfiguration
+import androidx.compose.ui.platform.LocalTextToolbar
+import androidx.compose.ui.platform.LocalView
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.ParagraphStyle
 import androidx.compose.ui.text.SpanStyle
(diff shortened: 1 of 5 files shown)


modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/transcripts/TranscriptPage.kt
```diff
@@ -145,19 +153,37 @@ private fun TranscriptContent(
...
-                modifier = Modifier
-                    .padding(horizontal = 16.dp)
-                    .padding(bottom = bottomPadding)
-                    .verticalScroll(rememberScrollState()),
-            )
+            val customMenu = buildList {
+                if (Build.VERSION.SDK_INT < Build.VERSION_CODES.TIRAMISU) {
+                    add(CustomMenuItemOption.Share)
+                }
+                add(CustomMenuItemOption.SelectAll)
```
ashiagr, line 167: I have not added `Share` custom menu for newer devices (API level >33) as copy option provides sharing once the text is selected. 

Note that transcripts with HTML mime-type use Android view to render HTML content and the text selection toolbar displays a share option even on newer devices. But I plan to replace `Android View` with `Compose View` once Compose `1.7.0` is stable that can parse HTML ([ref](https://developer.android.com/jetpack/androidx/releases/compose-foundation#1.7.0-alpha06:~:text=Added%20parseAsHtml%20method%20for%20styled%20strings)).

mebarbosa (commented):
Nice work!! I have one question

Should the `Select All` label select all texts from transcripts ?

[Screen_recording_20240801_145431.webm](https://github.com/user-attachments/assets/e335ff95-b282-424f-8895-a7f10d05fbe5)

ashiagr:
> Should the Select All label select all texts from transcripts ?

It is tricky to select all text and may need more customization as the text selection part is added to another internal Compose class.  I can remove `Select All` for now and include it when it is available in 1.7.0. Wdyt?

mebarbosa (approved):
LGTM

PR #2556, opened 2024-08-01
Transcripts - Show scrollbar

ashiagr:
## Description
This displays a scrollbar on the transcript view.

I added a basic implementation without any drag capabilities (p1722425522642369-slack-C078746V6UX). Note that scrollbars are in the [Compose roadmap](https://developer.android.com/jetpack/androidx/compose-roadmap?authuser=1#:~:text=headers%20for%20lists-,Scrollbars,-Flow%20layouts%20new), so hopefully we'll have a fully functional scrollbar soon. 

## Testing Instructions
1. Open transcript view with scrollable text (e.g. episode from podcast: "Cautionary Tales" (https://pcast.pocketcasts.net/8zpmjiru))
2. Try to scroll the view
3. Notice that scrollbars are visible.
** Test for a transcript with HTML mime-type (eg: episode from podcast `If books could kill` https://pca.st/tqib657g also, as those are displayed on Android TextView.

## Screenshots or Screencast 

https://github.com/user-attachments/assets/db4b766f-d5f8-4466-8edf-36253a8b4b27

## Checklist
 
- [ ] with different themes


pachlava:
@ashiagr 👋 

I had a look into the implementation, and noticed only one minor issue. When the transcript is very long, the scrollbar is compressed so much that it's hardly seen. Is it possible to set a minimum size for a scrollbar?

Another point is that the very top and very bottom of the bounding box for the transcript content is overlapped by the X button panel (at the top) and player (at the bottom), so when the scrollbar is too small, it's fully invisible.

Other than that, it looks good! Thanks for working on this! 🙂 

Test podcast: https://pcast.pocketcasts.net/c8nzu6xo

https://github.com/user-attachments/assets/a519315c-3017-4e58-8dc4-8579702f7fc6

MiSikora (approved):
LGTM!

modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/extensions/Modifier.kt
```diff
@@ -86,3 +94,37 @@ enum class FadeDirection {
...
+        label = "",
+    )
+
+    drawWithContent {
+        drawContent()
+        if (scrollState.maxValue == 0) {
+            return@drawWithContent
+        }
+        val contentHeight = size.height
+        val scrollBarHeight = viewPortHeight * viewPortHeight / contentHeight
```
MiSikora, line 119:
```suggestion
    val miScrollBarHeight = LocalDensity.current.run { 16.dp.toPx() }

    drawWithContent {
        drawContent()
        if (scrollState.maxValue == 0) {
            return@drawWithContent
        }
        val contentHeight = size.height
        val scrollBarHeight = (viewPortHeight * viewPortHeight / contentHeight).coerceAtLeast(miScrollBarHeight)
```

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/transcripts/TranscriptPage.kt
```diff
@@ -146,6 +153,21 @@ private fun TranscriptContent(
...
+            .padding(bottom = bottomPadding)
+            .verticalScroll(scrollState)
+            .verticalScrollBar(
+                thumbColor = colors.textColor(),
+                scrollState = scrollState,
+                viewPortHeight = viewportHeight,
+            )
+            .onGloballyPositioned {
+                viewportHeight = it.boundsInParent().height
+            }
```
MiSikora, line 169: It would nice to avoid `onGloballyPositioned` here and make it part of `verticalScrollBar()`. I think something like this should work.

```kotlin
fun Modifier.verticalScrollBar() = composed {
    var viewPortHeight by remember { mutableFloatStateOf(0f) }
    drawWithContent {

    }.onGloballyPositioned { viewPortHeight = it.boundsInParent().height }
}
```

ashiagr: Moved in [2e93c16](https://github.com/Automattic/pocket-casts-android/pull/2556/commits/2e93c1646151a1c49401f9669adda2f9067cad62)

modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/extensions/Modifier.kt
ashiagr: Applied in [55fd90d](https://github.com/Automattic/pocket-casts-android/pull/2556/commits/55fd90da04772cd004d327687ea0b289c59ea893)

ashiagr:
@pachlava 👋 
Thanks for testing! I've added content padding and set a minimum height for the scroll bar. PR is ready for another look.

pachlava:
@ashiagr 👋 The updated scrollbar looks good, thanks!

PR #2560, opened 2024-08-01
[Give Rating] Adding whats new for Give Rating

### LINKED ISSUE #2543: [Give Rating] Add a What's new announcement

mebarbosa:
## Description
- This PR adds the give rating for whats new with animation. 
- The initial state should show no stars and then play the animation.
- See: p1722454541296259/1722365320.671129-slack-C077XU4GF9D

Fixes #2543 

## Testing Instructions
1. Make give rating feature flag is enabled
2. Apply this [enable_whats_new.patch](https://github.com/user-attachments/files/16470652/enable_whats_new.patch)
3. Run the App
4. Check the whats new modal for give rating

## Screenshots or Screencast 
screen-20240805-150228-ezgif com-crop

https://github.com/user-attachments/assets/bb5d2caa-067d-46bd-9da1-ee994f85dc87

## Checklist
 
- [x] with different themes


mebarbosa:
I am tagging you all since this should be available in Beta next week @Automattic/pocket-casts-android

modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/Settings.kt
```diff
@@ -62,7 +62,7 @@ interface Settings {
 
         const val CHROME_CAST_APP_ID = "2FA4D21B"
 
-        const val WHATS_NEW_VERSION_CODE = 9121
+        const val WHATS_NEW_VERSION_CODE = 9256
```
mebarbosa, line 65: Please, double check if this is the expected version code 🙏

MiSikora: This isn't related to app version code. It is a separate code. It should be incremented normally

MiSikora:
I haven't checked the code but the animation feels rough when compared to the spec.

| Specification | Implementation |
| - | - |
| spec | impl |

1. The stars do not appear and disappear one after another in a smooth way.
2. There is almost no bounciness to the vertical motion.
3. The animation starts with the stars visible while in specification we start with an empty state.
4. Animation loops only 4 times. I've checked this thread (p1722527470060899/1722365320.671129-slack-C077XU4GF9D) and I don't understand the problem. Looping forever shouldn't affect the performance. If this happens for longer loops it means that there is something wrong with how the animation is set up.

MiSikora:
I haven't checked the code but the animation feels rigid when compared to the spec. Ignore choppiness of the implementation. It is only due to lower GIF frame rate.

| Specification | Implementation |
| - | - |
| spec | impl |

1. There is no vertical motion when stars disappear.
2. The animation gets out of sync. Stars start to appear one after another but then they bounce in a random order.
3. Fade and vertical movement are not played together. First stars fade in then they move vertically.
4. Animation loops only 4 times. I've checked this thread (p1722527470060899/1722365320.671129-slack-C077XU4GF9D) and I don't understand the problem. Looping forever shouldn't affect the performance. If this happens for longer loops it means that there is something wrong with how the animation is set up.

mebarbosa:
Hey @ashiagr @geekygecko , could any of you review this PR since Michal also pushed code here?

PR #2565, opened 2024-08-02
Transcripts - Handle no content, no network

ashiagr:
## Description
This handles transcripts with no content and displays no network error if transcript cannot be downloaded from network.

** I also added previews in e6187b4d4092172630d9a6dfb62983fc387ddea5. Note that the error states design is not final (p1722354522092579-slack-C078746V6UX).

** Caching will handled as a separate task

#### Test.1 No Network

1. Play an episode with transcript
2. Open full screen player
3. Disable internet on phone
4. Tap the transcript shelf icon
5. ✅ Notice that no network available error is shown
6. Go back 
7. Enable internet on phone
8. Re-tap the transcript shelf icon
9. ✅ Notice that transcript is loaded

#### Test.2 Transcript with no content

1. Play an episode having transcript with no content (e.g. [ChaosRadio](https://pca.st/r2ct46up))
2. Open full-screen player
3. Tap the transcript shelf icon
4. ✅ Notice that `Transcript is empty` message is shown

## Screenshots or Screencast 

https://github.com/user-attachments/assets/39cf654b-8965-43ed-9d13-41142532d4e9

## Checklist
 
- [ ] with different themes


MiSikora (approved):
Would it possible to hide transcript icon if transcript content is empty or is the information not really accessible prior to opening the transcripts page?

ashiagr:
> Would it possible to hide transcript icon if transcript content is empty or is the information not really accessible prior to opening the transcripts page?

We get transcript URL content only after tapping the icon, so this information is unavailable beforehand. 
It will be rare that we get into this situation. I see that I need to support [PodcastIndex.org transcript json format](https://github.com/Podcastindex-org/podcast-namespace/blob/main/transcripts/transcripts.md#json) also for this content to load. I've added it to Project task.

MiSikora:
I think it might be worth considering in the future to cache transcript content like we do with chapters. This way we could support offline mode and show the button only when there's content.

ashiagr:
That's the plan. I added it to PR description:

** Caching will handled as a separate task

PR #2581, opened 2024-08-06
[Kids Profile] Send Anonymous feedback

mebarbosa:
## Description
- This sends the anonymous feedback using the `/anonymous/feedback` endpoint. I will add another PR to handle an non anonymous feedback, witch is for the case the user is logged in.
- This PR does not handle UI feedback after sending the request. I will work on the snack bar feedback in a later PR
- See: p1722564959720889/1722564947.563029-slack-C07CGGZR59U

## Testing Instructions
1. Have the Kids Feature flag enabled
2. Go to Profile -> Request Early access -> Send Feedback
3. Type something to enable the Send Feedback button
4. Send feedback
5. ✅ Ensure the `/anonymous/feedback` is called (check on logcat) 
6. If you want to see the feedback sent to the email you can follow these steps: p1722955409841369/1722954751.866389-slack-C07CGGZR59U

## Screenshots or Screencast 
[Screen_recording_20240806_123157.webm](https://github.com/user-attachments/assets/fd32d6c1-6e58-4044-8a41-99bec891970e)

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/services/protobuf/src/main/proto/sync_api.proto
+++ b/modules/services/protobuf/src/main/proto/sync_api.proto
@@ -225,7 +225,9 @@ message PodcastRatingShowRequest {
 }
 
 message SupportFeedbackRequest {
-     string subject = 1;
-      string inbox = 2;
-      string message = 3;
+    string message = 1;
+    string email = 2;
+    string subject = 3;
+    string debug = 4;
+    string inbox = 5;
 }
--- a/modules/services/protobuf/src/main/proto/sync_api.proto
+++ b/modules/services/protobuf/src/main/proto/sync_api.proto
@@ -223,3 +223,10 @@ message PodcastRatingAddRequest {
 message PodcastRatingShowRequest {
     string podcast_uuid = 1;
 }
+
+message SupportFeedbackRequest {
+     string subject = 1;
+      string inbox = 2;
+      string message = 3;
+      string debug = 4;
+}
--- a/modules/services/protobuf/src/main/proto/sync_api.proto
+++ b/modules/services/protobuf/src/main/proto/sync_api.proto
@@ -228,5 +228,4 @@ message SupportFeedbackRequest {
      string subject = 1;
       string inbox = 2;
       string message = 3;
-      string debug = 4;
 }
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/SyncManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/SyncManager.kt
@@ -99,5 +99,5 @@ interface SyncManager : NamedSettingsCaller {
     suspend fun loadStats(): StatsBundle
     fun upNextSync(request: UpNextSyncRequest): Single<UpNextSyncResponse>
     suspend fun getBookmarks(): List<Bookmark>
-    suspend fun sendAnonymousFeedback(subject: String, inbox: String, message: String): Single<Response<Void>>
+    suspend fun sendAnonymousFeedback(subject: String, inbox: String, message: String): Response<Void>
 }
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/SyncManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/SyncManager.kt
@@ -99,5 +99,5 @@ interface SyncManager : NamedSettingsCaller {
     suspend fun loadStats(): StatsBundle
     fun upNextSync(request: UpNextSyncRequest): Single<UpNextSyncResponse>
     suspend fun getBookmarks(): List<Bookmark>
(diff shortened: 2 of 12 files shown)


modules/services/protobuf/src/main/proto/sync_api.proto
```diff
@@ -223,3 +223,9 @@ message PodcastRatingAddRequest {
 message PodcastRatingShowRequest {
     string podcast_uuid = 1;
 }
+
+message SupportFeedbackRequest {
+     string subject = 1;
+      string inbox = 2;
+      string message = 3;
```
MiSikora, line 230:
```suggestion
     string subject = 1;
     string inbox = 2;
     string message = 3;
```

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/SyncManager.kt
```diff
@@ -99,4 +99,5 @@ interface SyncManager : NamedSettingsCaller {
     suspend fun loadStats(): StatsBundle
     fun upNextSync(request: UpNextSyncRequest): Single<UpNextSyncResponse>
     suspend fun getBookmarks(): List<Bookmark>
+    suspend fun sendAnonymousFeedback(subject: String, inbox: String, message: String): Single<Response<Void>>
```
MiSikora, line 102: `suspend` function shouldn't return `RxJava` types.

MiSikora (changes requested):
LGTM! But sync code should be changed to not mix coroutines and `RxJava`.

There also a backend issue I think that I reported here: p1723009570417369-slack-C07CGGZR59U

modules/services/protobuf/src/main/proto/sync_api.proto
```diff
@@ -223,3 +223,9 @@ message PodcastRatingAddRequest {
...
+}
```
geekygecko, line 231: Can you change the message to the following:
```
message SupportFeedbackRequest {
    string message = 1;
    string email = 2;
    string subject = 3;
    string debug = 4;
    string inbox = 5;
}
```
This is from the following location https://github.com/Automattic/pocket-casts-sync-api/blob/develop/api/modules/protobuf/src/main/proto/api.proto#L105

mebarbosa: I've just updated the code with this! [232b852](https://github.com/Automattic/pocket-casts-android/pull/2581/commits/232b8524ef3a7ed77f02b8bebf9cb089e9a87269)

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/SyncManager.kt
mebarbosa: Done: [838ab4b](https://github.com/Automattic/pocket-casts-android/pull/2581/commits/838ab4b15fd70ce91c443b5e14b596ae532feee2)

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/SyncManager.kt
```diff
@@ -99,4 +100,5 @@ interface SyncManager : NamedSettingsCaller {
     suspend fun loadStats(): StatsBundle
     fun upNextSync(request: UpNextSyncRequest): Single<UpNextSyncResponse>
     suspend fun getBookmarks(): List<Bookmark>
+    fun sendAnonymousFeedback(subject: String, inbox: String, message: String): Call<Void>
```
MiSikora, line 103: There are still some issues with the current implementation. Let me explain why.

The original code was:

```kotlin
suspend fun sendAnonymousFeedback(@Body request: SupportFeedbackRequest): Single<Response<Void>>
```

`suspend fun foo(): Single<T>: This approach mixes coroutines and RxJava, which are two different asynchronous programming models. It's like combining two different callback mechanisms, which complicates the code and makes it harder to manage.

The proposed change is:

```kotlin
fun sendAnonymousFeedback(subject: String, inbox: String, message: String): Call<T>
```

This isn't necessarily wrong on its own. However, it doesn't fit the app at all. `fun foo(): Call<Void>`: This requires you to handle the Call object manually by either calling `.execute()` (which is blocking and not suitable for asynchronous operations) or `.enqueue()` (which requires handling callbacks). This approach doesn't integrate well with coroutines, making the code harder to maintain and less consistent. You can see that this approach "forces" you to call `.execute()` in the `FeedbackManager`.

The actual source of this function is a Retrofit interface, which already supports suspending functions. To properly integrate with coroutines, the function should be a suspend function that returns the result directly, without using future-like types such as `Single` or `Call`.

mebarbosa: Got it. I was following the same approach as `fun uploadFile(@Url url: String, @Body requestBody: RequestBody): Call<Void>`. I've updated the code again

MiSikora (approved):

PR #2594, opened 2024-08-08
Share clips to all platforms

MiSikora:
## Description

This PR introduces clip sharing in all formats across all platforms.

Initially, I thought I could share directly from the `rememberCoroutineScope()`, but this approach has a couple of drawbacks:

* It cancels sharing during configuration changes.
* It is not very testable to put the logic for creating clip requests inside the Compose UI.

To address these issues, I moved the logic to the `ShareClipViewModel`. This change resolves both problems and allows the view model to handle all snackbar-related messages.

Clip creation animation might be temporary for now as we don't have exact specification for it: pdeCcb-50w-p2#comment-4934

The landscape layout will be added in a future PR.

### Sharing

I can't provide exact instructions that would be comprehensive. Test sharing clips to different platforms in different formats. Instagram has a limit of 20 seconds so even if you try to share a longer video it will trim it.

### Snackbar

1. Start clip sharing for an episode that is not downloaded.
2. Disable internet.
3. Play the clip.
4. You should see an error message.
5. Try to share a video or an audio clip.
6. You should see an error message.

## Screenshots or Screencast 

screen-20240808-121840-ezgif com-optimize

## Checklist
 
- [x] with different themes


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

modules/features/sharing/src/main/kotlin/au/com/shiftyjelly/pocketcasts/sharing/clip/ShareClipFragment.kt
```diff
@@ -81,22 +89,38 @@ class ShareClipFragment : BaseDialogFragment() {
...
                 listener = listener,
+                snackbarHostState = snackbarHostState,
             )
+
+            LaunchedEffect(Unit) {
+                viewModel.snackbarMessages.collect { message ->
+                    val text = when (message) {
+                        is SnackbarMessage.SharingResponse -> message.message
+                        is SnackbarMessage.PlayerIssue -> getString(LR.string.podcast_episode_playback_error)
+                        is SnackbarMessage.GenericIssue -> getString(LR.string.error)
```
geekygecko, line 119: Maybe we should have a nicer error message here, something like "Unable to create clip. Please try again later."

Screenshot_20240809_135748_Pocket Debug

modules/features/sharing/src/main/kotlin/au/com/shiftyjelly/pocketcasts/sharing/clip/ShareClipFragment.kt
```diff
@@ -81,22 +89,38 @@ class ShareClipFragment : BaseDialogFragment() {
...
                 assetController = assetController,
                 listener = listener,
+                snackbarHostState = snackbarHostState,
             )
+
+            LaunchedEffect(Unit) {
+                viewModel.snackbarMessages.collect { message ->
+                    val text = when (message) {
+                        is SnackbarMessage.SharingResponse -> message.message
+                        is SnackbarMessage.PlayerIssue -> getString(LR.string.podcast_episode_playback_error)
```
geekygecko, line 118: This error looks good, but I have a minor comment on the UI behaviour. I'm not sure if it's possible, but it would be nice if the play button changed to a pause button as soon as it's tapped. At the moment, there is a delay, and it can feel like the button hasn't been pressed.

modules/features/sharing/src/main/kotlin/au/com/shiftyjelly/pocketcasts/sharing/clip/ShareClipViewModel.kt
```diff
@@ -98,6 +124,109 @@ class ShareClipViewModel @AssistedInject constructor(
...
+            return
+        }
+        isSharingState.value = true
+        viewModelScope.launch {
+            val animation = async {
+                // Make sure that animation lasts for at least 1 seconds when sharing
+                // something other than a link
+                if (platform != SocialPlatform.PocketCasts) {
+                    delay(1.seconds)
+                }
```
geekygecko, line 146: Why do we need to wait a second? Does it look strange if it flashes on the screen too quickly?

geekygecko (approved):

modules/features/sharing/src/main/kotlin/au/com/shiftyjelly/pocketcasts/sharing/clip/ShareClipViewModel.kt
MiSikora: Yeah, it can look ugly if it just flashes between two states.

modules/features/sharing/src/main/kotlin/au/com/shiftyjelly/pocketcasts/sharing/clip/ShareClipFragment.kt
MiSikora: Sure, makes sense 👍

modules/features/sharing/src/main/kotlin/au/com/shiftyjelly/pocketcasts/sharing/clip/ShareClipFragment.kt
MiSikora: I agree with you. I just didn't want to risk our managed state get out of sync with the ExoPlayer. But it probably is possible to use something like 'isLoading' for one of ExoPlayer's callbacks. I'll look into it in a separate PR because it might put me on a journey of its own.

modules/features/sharing/src/main/kotlin/au/com/shiftyjelly/pocketcasts/sharing/clip/ShareClipFragment.kt
MiSikora: Fixed in b41a92b58da8bbdb961abce002874a7d411d38b8

PR #2603, opened 2024-08-09
Use parent fragment manager instead of the child's one when creating Kids dialog

MiSikora:
## Description

Attempt to fix [the crash from Sentry](https://a8c.sentry.io/issues/5690494505/?project=6711064&query=release%3Aau.com.shiftyjelly.pocketcasts.debug%407.70-rc-2%2B9257&referrer=release-issue-stream) in beta.

From the stack trace it looks like the lambda for requesting early access is getting called. ~I'm not entirely sure how is that possible because the feature is disabled. Maybe someone is experimenting with it and enabling it on its own with memory or code manipulation.~ Ok, now I know. It is from one of our devices because the crash happened in the `.debug` version. It's still better to fix in beta.

In any case, I'm not able to reproduce the bug. But I think I found the root cause. We're using `childFragmentManager` when instantiating `KidsBottomSheetDialog`. This is wrong because then the view model of this dialog is connected to the parent fragment, which is `ProfileFragment` which might get detached. I'm fixing it by using `parentFragmentManager`.

## Testing Instructions

Not sure since I can't reproduce it. Verify the code.

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/ProfileFragment.kt
+++ b/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/ProfileFragment.kt
@@ -291,7 +291,7 @@ class ProfileFragment : BaseFragment() {
                     KidsProfileCard(
                         onDismiss = { viewModel.dismissKidsBanner() },
                         onRequestEarlyAccess = {
-                            KidsBottomSheetDialog().show(childFragmentManager, "KidsBottomSheetDialog")
+                            KidsBottomSheetDialog().show(parentFragmentManager, "KidsBottomSheetDialog")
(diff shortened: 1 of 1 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-android/milestone/61">7.70 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

geekygecko (approved):

PR #2613, opened 2024-08-12
Update the episode star value in the Media Session

geekygecko:
## Description

Car manufacturers using Pocket Casts have reported that when they change the episode star it doesn't update. I believe this is because we aren't setting the rating in the media session. We don't say it's an action we support and instead use custom actions but as they are adding the action and we do support setting the value it seems like a good change to make. 

## Testing Instructions

1. As this isn't possible to test in the Automotive emulator as a Timber below the `putRating` call.
2. Deploy to the Automotive emulator
3. Play an episode
4. Open the full screen player
5. Tap the overflow dots
6. Star the episode
7. ✅ Verify the media session is updated
8. Unstar the episode
9. ✅ Verify the media session is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 *   New Features
     *   Adds the kids banner in profile
         ([#2591](https://github.com/Automattic/pocket-casts-android/pull/2591))
+*   Bug Fixes
+    *   Update the episode star value in the Media Session to fix places where it's shown outside the app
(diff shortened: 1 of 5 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

ashiagr (approved):

PR #2615, opened 2024-08-12
Constraint max sharing card size on all screens

MiSikora:
## Description

Allowing cards to occupy the whole available space is wrong because when it is shared as an image or a video we scale it down to `1080x1920` bitmap. This would make fonts extremely small. This PR address this issue by constraining card sizes.

## Testing Instructions

1. Preview UI in landscape layout for clip and podcast sharing on a tablet or a foldable device.
2. Preview UI in portrait mode for podcast, episode, and episode timestamp sharing on a tablet or a foldable device.

## Screenshots or Screencast 

Example of a constrained square card

| Before | After |
| - | - |
|  Screenshot 2024-08-12 at 09 17 29 | Screenshot 2024-08-12 at 09 17 11 |

## Checklist
 
- [x] with different themes


ashiagr:
To address this problem for large screens, we displayed a sharable asset screen in a pop-up in EoY. 



Do you think we can do something similar? We can take David's input.

MiSikora:
> Do you think we can do something similar? We can take David's input.

Yes, this is a good idea but the effort to do this isn't trivial. I'll bring it up to the designers but I don't think we will prioritize it.

ashiagr (approved):

PR #2626, opened 2024-08-13
Transcipts - Display text using LazyColumn

ashiagr:
## Description

This improves loading large transcripts. 

Loading an entire annotated text in the `Compose` text view was inefficient.  During search implementation, it became more apparent due to styling for highlighting search results. To address this, this PR adopts a `LazyColumn` for text rendering.

I'll submit search implementation PRs on top of this PR.

## Testing Instructions
1. Load large transcript (E.g. for episode https://pca.st/1iofzvym with ~4 hrs duration)
2. ✅ Notice that UI does not block while setting large text
3. Scroll the text
4. ✅ Notice that the scrollbar works properly 

## Screenshots or Screencast 

After

https://github.com/user-attachments/assets/f2e653d8-dc63-49b3-b465-994552497309

Before (notice lag at timestamp 0:06)

https://github.com/user-attachments/assets/89cb437b-a1e7-479f-9323-e6b5ffae27cd

## Checklist
- [x] ~~If this is a user-facing change, I have added an entry in CHANGELOG.md~~
- [x] ~~I have considered whether it makes sense to add tests for my changes~~
- [x] ~~All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml`~~
- [x] ~~Any jetpack compose components I added or changed are covered by compose previews~~
- [x] ~~I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.~~
 
- [x] ~~with different themes~~
- [x] ~~with a landscape orientation~~
- [x] ~~with the device set to have a large display and font size~~
- [x] ~~for accessibility with TalkBack~~

### DIFF
--- a/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/transcripts/TranscriptPage.kt
+++ b/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/transcripts/TranscriptPage.kt
@@ -12,6 +12,9 @@ import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.lazy.LazyColumn
+import androidx.compose.foundation.lazy.itemsIndexed
+import androidx.compose.foundation.lazy.rememberLazyListState
 import androidx.compose.foundation.rememberScrollState
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.foundation.text.selection.SelectionContainer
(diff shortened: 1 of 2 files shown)


MiSikora:
It's worth noting that splitting text into rows makes it not possible to highlight and select multiple separate paragraphs. But this is definitely better than loading the whole string and rendering because it can potentially lead even into OOM crashes.

MiSikora (approved):

ashiagr:
> It's worth noting that splitting text into rows makes it not possible to highlight and select multiple separate paragraphs.

It was a tricky decision to switch to `LazyColumn`. Thankfully, we can highlight and select multiple paragraphs even with `LazyColumn`



And also highlight search term with some workaround:


I'll start submitting search PRs tomorrow.

MiSikora:
> Thankfully, we can highlight and select multiple paragraphs even with `LazyColumn`

How do you do this? I can't select it like that. I don't even know how this could work in principle since each paragraph is a separate row in the `LazyColumn` so they don't share cursor and selection. Even in principle it would be hard because there might be other composables between one text block and another text block. I'm not sure about implementation details but maybe it depends on the format of transcripts and sometimes all of paragraphs are split into separate strings and sometimes they are not?

https://github.com/user-attachments/assets/115a932c-1340-4143-bb87-c7ce178c0893

ashiagr:
Sorry, I should have been more clear. I was just wrapping up my day. You can't select it in this PR. I moved `SelectionContainer` outside `LazyColumn` to see if it works. I'll include it in the next PR.

PR #2686, opened 2024-08-22
Transcripts - Misc fixes

ashiagr:
## Description
This fixes following two issues:

1. Flicker when selecting another episode with transcript
2. 301 redirect cache issue

#### Flicker
1. Play an episode with a transcript
2. Open transcript view
3. Close transcript view
4. Go to the podcast 
5. Select another episode from the podcast
6. ✅ Notice there's no flicker

#### HTTPS 301 redirect 
1. Play this episode https://pcast.pocketcasts.net/c8nzu6xo
2. Open transcript view
3. Notice that the transcript is cached for HTTP URL as well as the 301 redirect HTTPS URL 
4. Go to the podcast 
5. Play another episode and open transcript view
6. Go to the podcast 
7. Disable internet
8. Play previous episode and open transcript view
9.  ✅ Notice that the transcript is loaded from HTTPS URL cache (earlier it displayed network error)

#### Flicker 

Before

https://github.com/user-attachments/assets/b5340acc-d02a-4334-a86b-161b5d18508e

After

https://github.com/user-attachments/assets/22e3bf75-2c49-4731-baf6-ae15559dee78

#### HTTPS 301 redirect 

https://github.com/user-attachments/assets/bd4664aa-8615-4a54-bc0d-c0c2504dbedd

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/di/ServersModule.kt
+++ b/modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/di/ServersModule.kt
@@ -81,9 +81,12 @@ class ServersModule {
             val originalResponse = chain.proceed(request)
             var responseBuilder = originalResponse.newBuilder()
             if (request.cacheControl.noCache) {
-                responseBuilder = responseBuilder.header(HEADER_CACHE_CONTROL, "public, max-age=$CACHE_FIVE_MINUTES")
+                responseBuilder = responseBuilder
+                    .header(HEADER_CACHE_CONTROL, "public, max-age=$CACHE_FIVE_MINUTES")
+                    .removeHeader("Pragma")
             } else if (request.cacheControl.onlyIfCached) {
                 responseBuilder.header(HEADER_CACHE_CONTROL, "public, only-if-cached, max-stale=${request.cacheControl.maxStaleSeconds}")
+                    .removeHeader("Pragma")
             }
             responseBuilder.build()
         }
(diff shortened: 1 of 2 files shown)


modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/di/ServersModule.kt
```diff
@@ -81,9 +81,12 @@ class ServersModule {
             val originalResponse = chain.proceed(request)
             var responseBuilder = originalResponse.newBuilder()
             if (request.cacheControl.noCache) {
-                responseBuilder = responseBuilder.header(HEADER_CACHE_CONTROL, "public, max-age=$CACHE_FIVE_MINUTES")
+                responseBuilder = responseBuilder
+                    .header(HEADER_CACHE_CONTROL, "public, max-age=$CACHE_FIVE_MINUTES")
+                    .removeHeader("Pragma")
             } else if (request.cacheControl.onlyIfCached) {
                 responseBuilder.header(HEADER_CACHE_CONTROL, "public, only-if-cached, max-stale=${request.cacheControl.maxStaleSeconds}")
+                    .removeHeader("Pragma")
```
MiSikora, line 89: I'm not sure what effects it might have on other requests. Shouldn't we do this only for the transcript requests?

ashiagr: `TranscriptsManagerImpl` uses `TranscriptCacheServer` to load only transcript URLs. They do not have a specific format.

MiSikora: But if you don't think we need I'm happy to approve this PR.

ashiagr: Thanks for the snippet! ❤️ I like how you added a custom header to identify such requests.
We only use the transcript interceptor for transcript URLs, so we don't need it. I added a comment in 61f0978 to make it more explicit.

MiSikora (approved):

PR #2729, opened 2024-08-27
Clean up Moshi configuration

MiSikora:
## Description

This PR cleans up Moshi configuration and makes sure that we use a single instance of it that is injected.

I also discovered a potential issue with handling refresh token on Wear OS. We used a `value class` for it, which isn't supported by Moshi.

## Testing Instructions

1. Smoke test the phone app.
2. Test signing in on the Wear OS app.

## Checklist
- [x] ~All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml``
 
- [ ] with different themes

### DIFF
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/di/ModelModule.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/di/ModelModule.kt
@@ -1,6 +1,9 @@
 package au.com.shiftyjelly.pocketcasts.models.di
 
 import android.app.Application
+import androidx.room.Room
+import androidx.room.RoomDatabase
+import au.com.shiftyjelly.pocketcasts.model.BuildConfig
 import au.com.shiftyjelly.pocketcasts.models.db.AppDatabase
 import au.com.shiftyjelly.pocketcasts.models.db.dao.ChapterDao
 import au.com.shiftyjelly.pocketcasts.models.db.dao.EpisodeDao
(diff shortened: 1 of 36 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please add tests for class <code>WatchSyncAuthDataSerializer</code> (or add <code>unit-tests-exemption</code> label to ignore this).</td>
    </tr>
  </tbody>
</table>

</p>

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/di/ModelModule.kt
```diff
@@ -39,3 +65,11 @@ object ModelModule {
     @Provides
     fun provideExternalDataDao(database: AppDatabase): ExternalDataDao = database.externalDataDao()
 }
+
+@Qualifier
+@Retention(AnnotationRetention.BINARY)
+annotation class RoomConverters
+
+fun <T : RoomDatabase> RoomDatabase.Builder<T>.addTypeConverters(converters: List<Any>) = converters.fold(this) { builder, converter ->
```
geekygecko, line 73: It took me a while to get my head around the `fold` command. Is this a common pattern if you want to return `this` as the result?

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/di/ModelModule.kt
```diff
@@ -1,24 +1,50 @@
...
+import javax.inject.Singleton
 
 @Module
 @InstallIn(SingletonComponent::class)
 object ModelModule {
     @Provides
-    fun providesAppDatabase(application: Application): AppDatabase {
-        return AppDatabase.getInstance(application)
+    @RoomConverters
+    fun provideRoomConveretrs(moshi: Moshi): List<Any> {
```
geekygecko, line 28:
```suggestion
    fun provideRoomConverters(moshi: Moshi): List<Any> {
```
A minor typo

geekygecko (approved):
LGTM

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/di/ModelModule.kt
MiSikora: Always! 😠

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/di/ModelModule.kt
MiSikora: It's actually not returning `this`. It's returning whatever is the result of the lambda in `fold()`. So

```kotlin
val list = listOf(1, 2, 3, 4, 5)
val result = list.fold("") { str, int -> "$str$int" }
println(result) // "12345" 
```

It doesn't matter much in this instance because `Builder` is most likely mutable. I haven't checked but I assume so. So these two are equivalent.

```kotlin
converters.forEach { converter -> 
  builder.add(converter)
}
return builder

return converters.fold(builder) { builder, converter -> builder.add(converter) } 
```

But it does matter when object which we fold upon is immutable.

Whether it is common it depends on who you ask I guess. It's built into the standard library. It's super common and natural for me. In functional programming it is a standard way.

geekygecko: Thanks for taking the time to explain that to me.

PR #2735, opened 2024-08-27
Clean up OkHttp configuration

MiSikora:
## Description

Our current `OkHttp` client dependency injection configuration needs some improvement. We don't share the dispatcher between multiple clients, we have unused clients, and navigating the `ServersModule` has become challenging due to the extensive DI configuration. To address these issues, I've made the following changes:

- **Moved interceptor configuration**: All interceptor configurations have been relocated to a separate `InterceptorsModule`.
- **Removed unused clients**: Cleaned up the code by removing clients that are no longer in use.
- **Simplified qualifier annotations**: Consolidated the annotations by replacing `@TranscriptsCache`, `@TranscriptsClientBuilder`, and `@TranscriptsClient` with a single `@Transcripts` annotation.
- **Shared connection pool**: Most clients now share the same connection pool, with the exception of the client used for downloads, which is kept separate to allow it do its own thing.

I would like to provide a custom client for `Coil`, but since it resides in the `:ui` module, I've decided not to address it in this PR. This might be revisited in the future.

## Testing Instructions

Test networking of the app. Using APIs, downloads, caching transcripts, etc.

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/di/InterceptorModule.kt
+++ b/modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/di/InterceptorModule.kt
@@ -1,5 +1,12 @@
 package au.com.shiftyjelly.pocketcasts.servers.di
 
+import au.com.shiftyjelly.pocketcasts.preferences.AccessToken
+import au.com.shiftyjelly.pocketcasts.preferences.Settings
+import au.com.shiftyjelly.pocketcasts.servers.BuildConfig
+import au.com.shiftyjelly.pocketcasts.servers.OkHttpInterceptor
+import au.com.shiftyjelly.pocketcasts.servers.sync.TokenHandler
+import au.com.shiftyjelly.pocketcasts.servers.toClientInterceptor
+import au.com.shiftyjelly.pocketcasts.servers.toNetworkInterceptor
 import com.automattic.android.tracks.crashlogging.CrashLoggingOkHttpInterceptorProvider
 import com.automattic.android.tracks.crashlogging.FormattedUrl
 import com.automattic.android.tracks.crashlogging.RequestFormatter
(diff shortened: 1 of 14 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/di/InterceptorModule.kt
```diff
@@ -1,24 +1,188 @@
...
 
 @InstallIn(SingletonComponent::class)
 @Module
-class InterceptorModule {
-    @Provides
-    @CrashLoggingInterceptor
-    fun provideMonitoringInterceptor(): Interceptor = CrashLoggingOkHttpInterceptorProvider
+object InterceptorModule {
+    private val sixMonths = 180.days.inWholeSeconds
+    private val fiveMinutes = 180.minutes.inWholeSeconds
```
ashiagr, line 29:
```suggestion
    private val fiveMinutes = 5.minutes.inWholeSeconds
```

modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/di/InterceptorModule.kt
```diff
@@ -1,24 +1,188 @@
...
+            responseBuilder = responseBuilder.header(cacheControlHeader, "public, max-age=$fiveMinutes")
+        }
+        responseBuilder.build()
+    }
+
+    private val cacheControlNotesInterceptor = Interceptor { chain ->
+        val request = chain.request()
+        val originalResponse = chain.proceed(request)
+        originalResponse.newBuilder().header(cacheControlHeader, "public, max-age=$sixMonths").build()
+    }
```
ashiagr, line 59: `cacheControlNotesInterceptor` doesn't seem to be used anymore. Can we remove it?

ashiagr (approved):
Thanks for cleaning it up! I left two minor comments, aside from that, everything looks good. 🌟

modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/di/InterceptorModule.kt
MiSikora: Yes, thanks. It was used in an used client. I'll double check to make sure of it.

PR #2756, opened 2024-08-29
Include feature flags state in reports

MiSikora:
## Description

This PR include feature flags state in Sentry reports and in user logs.

## Testing Instructions

1. Go to our developer settings and long press "Report a crash".
2. Set a custom message.
3. Tap a couple of time on "Report a crash" to send a report to Sentry. You need to tap a couple of times due to sampling.
4. Go to Sentry and find your crash.
5. You should see `feature_` tags.
6. Go back to the app.
7. Go to "Help & feedback".
8. Open logs.
9. Notice feature flag entries in the logs.

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/support/Support.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/support/Support.kt
@@ -30,6 +30,8 @@ import au.com.shiftyjelly.pocketcasts.utils.FileUtil
 import au.com.shiftyjelly.pocketcasts.utils.Network
 import au.com.shiftyjelly.pocketcasts.utils.SystemBatteryRestrictions
 import au.com.shiftyjelly.pocketcasts.utils.Util
+import au.com.shiftyjelly.pocketcasts.utils.featureflag.Feature
+import au.com.shiftyjelly.pocketcasts.utils.featureflag.FeatureFlag
 import au.com.shiftyjelly.pocketcasts.utils.log.LogBuffer
 import com.jaredrummler.android.device.DeviceName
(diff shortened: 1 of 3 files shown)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/support/Support.kt
```diff
@@ -270,6 +272,13 @@ class Support @Inject constructor(
                 }
             }
 
+            runCatching {
+                val features = Feature.entries.map { "${it.key}: ${FeatureFlag.isEnabled(it)}" }
+                output.append("Feature flags").append(eol)
+                features.forEach { output.append(it).append(eol) }
+                output.append(eol)
```
geekygecko, line 279: The last eol isn't needed as there is one before the next section, so it adds two lines.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/support/Support.kt
```diff
@@ -270,6 +272,13 @@ class Support @Inject constructor(
                 }
             }
 
+            runCatching {
```
geekygecko, line 275: Do we need to catch the exception here? There is a try catch around the whole block.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/support/Support.kt
MiSikora: Ah, haven't noticed. I'll make a follow-up PR.

PR #2791, opened 2024-09-03
Clean up Gradle depenednecies

### LINKED ISSUE #2583: Tech Debt: move dependencies into modules and remove unused dependencies
Our dependency monitoring tool states we have 2947 unused dependencies. 

[build-health-report.txt](https://github.com/user-attachments/files/16518768/build-health-report-2024.txt)

MiSikora:
## Description

**This PR addresses the issue of dependency configuration in our project.**

Screenshot 2024-09-04 at 09:31:20

Previously, most dependencies were applied globally to all projects. This has now been adjusted so that dependencies are correctly applied only when necessary.

Here's a summary of the changes in this PR:

1. **Analysis Plugin Update and Configuration:**
   * Enabled the `ignoreKtx()` option to ignore unused KTX dependencies where a regular one suffices. This reduces noise, and R8 should strip out unused methods anyway.
   * Disabled the transitive dependencies check. Maintaining this check would require us to explicitly include all transitive dependencies, which I believe is unnecessary and tedious.
   * Disabled the check requiring `kotlinx-coroutines-android` to be used only as a runtime dependency. While the reason for this check is unclear, I want to explicitly include it in app binaries to ensure access to the main thread dispatcher.
   * Disabled the suggestion to configure `kotlin-stdlib` as `implementation()` in app modules. This is because the Kotlin plugin applies it as `api()`, and to my knowledge, we have no control over this behavior.

2. **Version Catalogs Adjustment:**
   * Removed bundles from Version Catalogs. While they are convenient, using them often introduces unwanted dependencies, making them impractical for our use case.

3. **Plugin Cleanup:**
   * Removed unnecessary plugins, such as KSP, Hilt, or Parcelize, from modules where they are not needed.

4. **Kotlin Version Enforcement:**
   * Ensured that our app uses the specific Kotlin version we define. The `aboutlibraries` dependency had introduced Kotlin 2.0 into our `7.72` builds. To maintain control over our Kotlin version, I've enforced that dependencies cannot upgrade it.

5. **Leak Canary Configuration:**
   * Leak Canary can now be applied by setting the `au.com.shiftyjelly.pocketcasts.leakcanary=true` project property.

5. **Typesafe Project Accessors:**
   * Projects can now be included using `api(projects.some.inner.project)` syntax instead of `api(project(":some:inner:project"))`

Fixes #2583

## Testing Instructions

I'm not sure what's the best way to test it. Smoke testing our apps and different features should be ok.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

geekygecko (approved):
Thanks for making this change! I tested with mobile, Wear OS, and Automotive. Everything seems to be working fine.

wzieba:
Great work, thank you @MiSikora for working on this! 

> Disabled the transitive dependencies check. Maintaining this check would require us to explicitly include all transitive dependencies, which I believe is unnecessary and tedious.

I don't know if it's a mental shortcut, but just in case, I'd like to clarify: it didn't require us to explicitly include **all** transitive dependencies, but only these which **we actually use** in our codebase.

>  This is because the Kotlin plugin applies it as api(), 

I didn't know Kotlin plugin applies stdlib as api(), I always thought it was implementation() configuration. Do you have some materials from where you found out it's api()?

> and to my knowledge, we have no control over this behavior.

In case you referred to generally applying stdlib (and not about configuration it uses), the only control we have, AFAIK is to disable adding stdlib ([source](https://kotlinlang.org/docs/gradle-configure-project.html#dependency-on-the-standard-library)).

MiSikora:
> I don't know if it's a mental shortcut, but just in case, I'd like to clarify: it didn't require us to explicitly include all transitive dependencies, but only these which we actually use in our codebase.

Yes, this was a mental shortcut.

> Do you have some materials from where you found out it's api()?

https://github.com/JetBrains/kotlin/blob/66602bad39e3e5edb856e1769d310f11d4e97290/libraries/tools/kotlin-gradle-plugin/src/common/kotlin/org/jetbrains/kotlin/gradle/internal/stdlibDependencyManagement.kt#L106-L112

```kotlin
val scope = if (compilation.isTest() ||
    (this is KotlinAndroidTarget && kotlinSourceSet.isRelatedToAndroidTestSourceSet())
) {
    KotlinDependencyScope.IMPLEMENTATION_SCOPE
} else {
    KotlinDependencyScope.API_SCOPE
}
```

It wouldn't be error-prone to apply it as `implementation` because it is super easy to leak Kotlin APIs and then your compilation fails if you try to use them in, for example, Java project that uses a project with Kotlin.

> In case you referred to generally applying stdlib (and not about configuration it uses), the only control we have, AFAIK is to disable adding stdlib 

No, I was referring to how is it applied. I think that applying it through plugin is good.

wzieba:
Right, thanks for sharing, TIL!

PR #2794, opened 2024-09-04
Listening history - Add local search

ashiagr:
## Description
This adds local search to listening history

## Testing Instructions
1. Login with an account with sufficient listening history
2. Go to `Profile` tab -> `Listening History`
3. Search for an episode that exists in listening history
4. ✅ Notice that the episode appears in the results if the search term exists in episode title or podcast title
5. Search for an episode that does not exists in listening history
6. ✅ Notice that empty results screen is displayed

## Screenshots or Screencast 

https://github.com/user-attachments/assets/c33c94ee-3ad6-4d56-b12d-80e44350d465

## Checklist
 
- [ ] with different themes


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/EpisodeDaoTest.kt
```diff
@@ -160,4 +163,40 @@ class EpisodeDaoTest {
...
+    @Test
+    fun getOrderedFilteredPlaybackHistoryResults() = runTest {
+        val query = "%test%"
+        val episode1 = PodcastEpisode(uuid = "1", title = "Test Episode 1", publishedDate = Date(), lastPlaybackInteraction = 1000)
+        val episode2 = PodcastEpisode(uuid = "2", title = "Test Episode 2", publishedDate = Date(), lastPlaybackInteraction = 2000)
+        val episodes = listOf(episode2, episode1)
+        episodeDao.insertAll(episodes)
+
+        val result = episodeDao.filteredPlaybackHistoryFlow(query).first()
+        assertEquals(episodes, result)
```
mebarbosa, line 200: @ashiagr Do you think it would make sense to add a test for no matched results?

mebarbosa (approved):
I've tested with different themes, devices, landscape and also checked the accessibility and all worked as expected!

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/EpisodeDao.kt
```diff
@@ -277,6 +277,20 @@ abstract class EpisodeDao {
...
 
+    @Query(
+        """
+        SELECT podcast_episodes.*
+        FROM podcast_episodes
+        LEFT JOIN podcasts ON podcast_episodes.podcast_id = podcasts.uuid
+        WHERE last_playback_interaction_date IS NOT NULL
+          AND last_playback_interaction_date > 0
+          AND (UPPER(podcast_episodes.title) LIKE UPPER(:query)
+               OR UPPER(podcasts.title) LIKE UPPER(:query))
```
MiSikora, line 288: It's probably not extremely important but when doing user input searches we should escape some special characters.

```sql
-- this example doesn't handle mixed case like yours with UPPER
WHERE column LIKE '%' || :query || '%'  ESCAPE '\'
```

```kotlin
val results = dao.search(term.escapeLike('\\'))

fun String.escapeLike(escapeChar: Char) = replace("$escapeChar", "$escapeChar$escapeChar")
    .replace("%", "$escapeChar%")
    .replace("_", "${escapeChar}_")
```

Otherwise you won't be able to search using `%` and `_` characters. Which is probably not needed given that we search in episode titles. But I think it's worth mentioning.

You can test with this podcast: https://anchor.fm/s/f68090c0/podcast/rss

ashiagr: Nice! Added in 18b5d28, tested in a657423

app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/EpisodeDaoTest.kt
ashiagr: Added in a657423

PR #2815, opened 2024-09-09
Add dependency analysis step to default CI pipeline

mokagio:
## Description

@MiSikora suggested running the dependency analysis check on every CI build, internal ref pdnsEh-1QL-p2, and there seemed to be appetite for it.

This draft PR aims to get the work started...

One argument _against_ adopting this, and which I guess was the rationale for running the check on a schedule, is that it takes a good 10+ minutes to run. If this were to become a requirement for the PR to be merged, it might slow down the dev cycle.

image

Having said that, it's something that would be easy to try to find out directly, and revert if necessary. 

### Implemented

- Add the check to the default pipeline, that is, run the check on every standard CI build

### To Do

- Update the configuration to be as strict as possible – I don't actually know if that's required but the wording of the request made me think so. Plugin docs https://github.com/autonomousapps/dependency-analysis-gradle-plugin/wiki/Customizing-plugin-behavior
- Moving the check to run on every build makes the existing [schedule](https://github.com/Automattic/pocket-casts-android/blob/924c3df9c4fe6875d96d431c6a33515334b9d3f0/.buildkite/schedules/dependency-analysis.yml) redundant, so I would suggest removing it. This is a two step task:
  - Remove from this repo
  - Remove the schedule configuration on the Buildkite end
- If/when this PR lands on `main`, update the GitHub configuration to require the "Dependency Analysis" step check

## Testing Instructions
See CI build: TBD

## Checklist
 
- [x] with different themes

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -57,6 +57,17 @@ steps:
     artifact_paths:
       - "**/build/instrumented-tests/**/*"
 
+  - label: "Dependency Analysis"
+    command: |
+      echo "--- 📊 Analyzing"
+      ./gradlew buildHealth
+    plugins: [$CI_TOOLKIT]
+    artifact_paths:
+      - "build/reports/dependency-analysis/build-health-report.*"
+    notify:
+      - slack: "#android-core-notifs"
+        if: build.state == "failed"
+
   - label: "Assemble release APK"
     command: |
       echo "--- :rubygems: Setting up Gems"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -65,6 +65,8 @@ steps:
     artifact_paths:
       - "build/reports/dependency-analysis/build-health-report.*"
     notify:
+      - github_commit_status:
+        context: Dependency Analysis
       - slack: "#android-core-notifs"
         if: build.state == "failed"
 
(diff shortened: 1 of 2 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

MiSikora:
Just a thought: wouldn't it be better to run it inside GH Actions? This way it won't occupy any of our Buildkite workers.

wzieba:
> Just a thought: wouldn't it be better to run it inside GH Actions? This way it won't occupy any of our Buildkite workers. Or do we want to use Builtkite for mostly everything? I'm not really sure when do we use GH Actions and when do we use Buildkite.

Our approach is to use Buildkite whenever we can as Buildkite workers are more performant and we have working configuration for collecting metrics from Buildkite.

We use GH Actions only when the benefits of using GH are high: e.g. for uploading Gradle dependencies to GitHub there is a first-class Gradle GitHub Action that handles both Gradle and GitHub API part for us. It's also not vital for performance, as resolving dependency graph is not that resource-consuming.

For dependency analysis I'd suggest we keep using Buildkite because we both need strong performance and we collect data from these builds.

ParaskP7:
After discussing this, we are not going to move forward with enabling this one every PR, not just yet. It is being too expensive for little benefit (ie. few PRs have the potential for such dependency related changes). But, we'll keep iterating on what's the best way to further integrate dependency analysis into our workflows, potentially blocking a PR from merging when a new advise presents itself (pdnsEh-1QL-p2#comment-4158).

PR #2828, opened 2024-09-11
Upload Android lint results to GitHub code scanning registry

wzieba:
## Description

This PR proposes a set of changes related to Android Lint configuration:

#### What?
Send lint report files to GitHub Code Scanning registry

#### Why?
1. To bring more attention to lint warnings. Right now, if a warning (not error) is reported, it's only reported in one of 37 lint reports, available only on Buildkite artifacts. This makes it basically hidden from PR author.
2. To have a common source of all Lint warnings existing in project, in case someone would want to tackle them ([link](https://github.com/Automattic/pocket-casts-android/security/code-scanning?query=pr%3A2828+is%3Aopen)).

#### What?
Instead of running lint for every module, run lint only for application modules with `checkDependencies` option

#### Why?
1. To generate a single (per app) report file, allowing us to upload results to GitHub Code Scanning registry. If we send reports for each module, they would override each other (fake-closing, example).
2. To reduce number of reports we have (I think it's easier to browse 3 reports instead of 37)

#### What?
Regenerate lint baselines to only contain errors.

#### Why?
I don't think there's need to baseline warnings: they don't crash build and if we baseline them, we hide them, making them more difficult to address when there's opportunity on a daily basis.

### Drawbacks
Because now we generate baseline per-application (app, wear, automotive) instead per-module, the newly generated baselines contains duplication. Basically, these 3 modules are treated as separate projects.

I think this is an acceptable drawback, counting the benefits I explained above.

## Testing Instructions
Verify code, green checks on CI are just fine.

## Screenshots or Screencast 

image

image

## Checklist
 
- [ ] with different themes


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

wzieba:
@MiSikora @ParaskP7 can I have a review from you on this? I remember that last time we worked on Lint integration, you were discussing it. Maybe you have some opinions to share here? WDYT about the proposed changes?

build.gradle.kts
```diff
@@ -184,14 +184,19 @@ subprojects {
         val isLocalizationsProject = project.name == "localization"
         lintConfig = rootProject.file(if (isLocalizationsProject) "lint-no-op.xml" else "lint.xml")
         xmlReport = true
+        sarifReport = true
```
MiSikora, line 187: Can you change this to generate SARIF reports only on CI?

build.gradle.kts
```diff
@@ -184,14 +184,19 @@ subprojects {
         val isLocalizationsProject = project.name == "localization"
         lintConfig = rootProject.file(if (isLocalizationsProject) "lint-no-op.xml" else "lint.xml")
```
MiSikora, line 185: This part requires further consideration. During the release process, Lint will block us if there are any missing translations or other issues, which would prevent us from building release builds. I don't think it’s feasible to expect release managers to handle this during release process.

We should explore a way to exclude Lint issues related to the `:localization` project if we decide to aggregate the results.

MiSikora (commented):
> Because now we generate baseline per-application (app, wear, automotive) instead per-module, the newly generated baselines contains duplication. Basically, these 3 modules are treated as separate projects.
> 
> I think this is an acceptable drawback, counting the benefits I explained above.

The reason I configured Lint to run on each module individually was to prevent multiple checks on the same module. Lint is known for being slow, and I wanted to minimize redundant iterations. It's possible that Lint is smart enough to avoid this duplication. Do you know if Lint can handle this efficiently?

> To reduce number of reports we have (I think it's easier to browse 3 reports instead of 37)

I’m not sure I fully agree with that. In my experience, it’s easier to browse reports specific to particular modules when you want to actually fix things and not just have an overview.

That said, I understand why aggregation is necessary for this feature to function as intended.

build.gradle.kts
```diff
@@ -184,14 +184,19 @@ subprojects {
...
         xmlReport = true
+        sarifReport = true
+        abortOnError = false
```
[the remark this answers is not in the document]
wzieba: for the context: I've added `abortOnError = false` because of this: https://buildkite.com/automattic/pocket-casts-android/builds/8293#0191e554-9473-4617-914b-60b0f2cc2f85/179-271 : after failed lint task, the `sarif` file for `app` was not generated (because `automotive:lint` execution stopped the task)

Maybe we should run `:app:lint` as the first task then, and only then `:automotive` and `:wear`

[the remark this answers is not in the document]
MiSikora: Maybe we could 2 separate jobs? One to run `:app:lint` for the GH integration. And another one for the other apps.

wzieba:
> The reason I configured Lint to run on each module individually was to prevent multiple checks on the same module. Lint is known for being slow, and I wanted to minimize redundant iterations. It's possible that Lint is smart enough to avoid this duplication. Do you know if Lint can handle this efficiently?

Because of tasks-depenencies mechanism, lint won't be executed multiple times on the same module. To verify it's true, you can run `./gradlew :app:lint :wear:lint :automotive:lint --dry-run` and browse which tasks will be indeed executed.

Since AGP 7.0 (or close to this) Lint also runs the checks for library modules when `checkDependencies=true` in parallel, so I also don't expect the performance drawback.

build.gradle.kts
wzieba: Thank you for additional context, I've missed this.

Would something like this: eafe896 work for you? I think disabling lint in a project is project-specific configuration, so I put it directly there.

build.gradle.kts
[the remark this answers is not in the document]
wzieba: Probably addressed in: 6de6674

build.gradle.kts
MiSikora: Yup, this works. Thanks!

MiSikora (approved):

PR #2857, opened 2024-09-13
Comment with dependency diff

wzieba:
## Description

Similarly to other projects, this PR integrate a new tool to make it easier to understand changes that a PR might introduce to the dependency graph.

## Testing Instructions
See the comment below.

After approval, I'll revert 9cec647 and merge.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -16,8 +16,8 @@
 #    horologist = Horologist Version
 
 [versions]
-android-gradle-plugin = "8.5.1"
 aboutlibraries = "11.2.3"
+android-gradle-plugin = "8.6.1"
 billing = "7.0.0"
 coil = "2.7.0"
 compose = "2024.02.00" # https://developer.android.com/jetpack/compose/bom/bom-mapping


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

.buildkite/commands/dependency-diff.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash -u
+
+echo "--- :memo: Commenting on the PR with the dependency diff"
+comment_with_dependency_diff 'app' 'releaseRuntimeClasspath'
```
ParaskP7, line 4: Question (❓): Comparing that to the other projects, why did you decide to create a separate command for dependency diff, that is, instead of using `comment_with_dependency_diff 'app' 'releaseRuntimeClasspath'` directly within the `pipeline.yml` step? 🤔

.buildkite/pipeline.yml
```diff
@@ -37,6 +37,13 @@ steps:
       ./gradlew testDebugUnitTest
     plugins: [$CI_TOOLKIT]
 
+  - label: 'Dependency diff'
+    if: build.pull_request.id != null
```
ParaskP7, line 41: Question (❓): This is a very interesting addition, I like that, should we maybe apply the same on all our projects, wdyt? 🤔

ParaskP7 (approved):
👋 @wzieba !

I have reviewed this PR and everything LGTM, great job on this too! 🌟 🌟 🌟

> After approval, I'll revert https://github.com/Automattic/pocket-casts-android/commit/9cec647bd56477ece9a4f35414b611f4d0cc92b8 and merge.

Let's not forget to revert this 6d571705f2e089606a7d3c07fd25188763eda99c too. 😊

.buildkite/pipeline.yml
wzieba: But I think this is applied in other projects! Did I miss it somewhere?

ParaskP7: Ah, interesting, I just checked all JP/WPAndroid, WCAndroid and DOAndroid and indeed this is included in them all, I might be dreaming, apologies for the false alarm, I just probably noticed that just now, just because it was an actual change/addition... 🤷 😊

.buildkite/commands/dependency-diff.sh
ParaskP7: Awesome, thanks! 👍

PR #2868, opened 2024-09-16
Run lint in release variant

wzieba:
## Description

To catch any issues that could block a release. This was discussed internally: p1716988994467289/1716983402.292699-slack-CC7L49W13

Recently, we had a scenario in which `debug` lint was successful, while the `release` one was failing. This caused a potential release blocker.

I think we should run lint in release variant on PRs to catch issues like this as early as possible.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

.buildkite/commands/lint.sh
```diff
@@ -1,11 +1,17 @@
 #!/bin/bash -u
 
+echo "--- :rubygems: Setting up Gems"
+install_gems
+
+echo "--- :closed_lock_with_key: Installing Secrets"
+bundle exec fastlane run configure_apply
```
MiSikora, line 7: Out of my own curiosity - is it needed because of the release flavor and missing secrets, etc?

MiSikora (approved):

.buildkite/commands/lint.sh
wzieba: Yes, precisle this 👍

PR #2923, opened 2024-09-23
Fix excessive scrolling behavior in discovery search

### LINKED ISSUE #2557: Search podcast results scroll back to the start after subscribing
### Description

When subscribing to a podcast in the search results the scroll position shouldn't be reset to the start. 

1. Tap the "Discover" tab
2. Tap the search text box at the top of the screen
3. Type a search such as "Daily News"
4. Scroll the podcast results
5. Tap subscribe

The podcast should be subscribed to and the scroll position shouldn't be reset to the start.

https://github.com/user-attachments/assets/45109f52-c852-4247-96d0-db6d0389c189

Samsung Fold5 Android 14, Pocket Casts 7.69-rc-2

semanticer:
## Description
Essentially, from what I've gathered, this piece of code: 

```kotlin
val podcasts by remember { mutableStateOf(state.podcasts) }
val podcastsRowState = rememberLazyListState()
val episodesRowState = rememberLazyListState()
LaunchedEffect(key1 = state.podcasts) {
    // Only reset the scroll state if we have podcasts and they are different. This prevents resetting the
    // scroll state when navigating back to the search results after tapping on one of the results.
    if (state.podcasts.isNotEmpty() && state.podcasts != podcasts) {
        podcastsRowState.scrollToItem(0)
    }
}
```

was supposed to make sure that:

1. When new podcasts are fetched (due to updated search term) it should scroll podcasts row back to the beginning
2. When the user subscribes, the podcasts row should stay where it is (#2557)
3. When selecting a podcast and going back, the podcasts row should stay where it is.

However, the original code above was only handling the first case correctly. See my video below where I demonstrate these three cases, to confirm that all three cases are now working as expected. 

Fixes #2557

## Testing Instructions
Check the original issue and my video below

## Screenshots or Screencast 

https://github.com/user-attachments/assets/0e513cf3-507f-4f7a-b3ff-0a8b4ccf01cc

## Checklist


modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchInlineResultsPage.kt
```diff
@@ -130,13 +130,13 @@ private fun SearchResultsView(
...
         }
     }
-    val podcasts by remember { mutableStateOf(state.podcasts) }
     val podcastsRowState = rememberLazyListState()
     val episodesRowState = rememberLazyListState()
-    LaunchedEffect(key1 = state.podcasts) {
-        // Only reset the scroll state if we have podcasts and they are different. This prevents resetting the
-        // scroll state when navigating back to the search results after tapping on one of the results.
-        if (state.podcasts.isNotEmpty() && state.podcasts != podcasts) {
+    val initialPodcasts by remember(state.podcasts.isEmpty()) { mutableStateOf(state.podcasts) }
```
MiSikora, line 135: The expression `state.podcasts.isEmpty()` is not a reliable key. For instance, if I type "redwood," the app will store a list of podcasts. However, when I delete characters to leave only "red," the original list of podcasts is still retained.

A better approach would be to use `state.podcasts.map { it.uuid }` as a `val`. It is already created during every recomposition for the `LaunchedEffect` anyway. Then it can be used inside the `LaunchedEffect` for comparison to control scrolling.

MiSikora (changes requested):

MiSikora:
Oh, and it'd be great if you could update the changelog with a fix. 🙏

semanticer:
> The expression state.podcasts.isEmpty() is not a reliable key. For instance, if I type "redwood," the app will store a list of podcasts. However, when I delete characters to leave only "red," the original list of podcasts is still retained.

This is an intentional behavior. My idea was to store the first non-empty list in the `initialPodcasts` to support this behavior mentioned above:
> 1. When new podcasts are fetched (due to updated search term) it should scroll podcasts row back to the beginning

Storing the first non-empty list and not updating it will ensure that the condition `state.podcasts != initialPodcasts` will be satisfied as soon as the `state.podcasts` changes (via search query change), and the scroll will be performed. 

If I understand your suggestion correctly, you suggested something like this, right?

```kotlin
val podcasts = state.podcasts.map { it.uuid }
val initialPodcasts by remember(podcasts) { mutableStateOf(state.podcasts) }
LaunchedEffect(key1 = podcasts) {
    if (state.podcasts.isNotEmpty() && state.podcasts != initialPodcasts) {
        podcastsRowState.scrollToItem(0)
    }
}
```

This will never trigger the `podcastsRowState.scrollToItem(0)` because `state.podcasts` will always be the same as `initialPodcasts`.

MiSikora:
No, I suggest this:

```kotlin
LaunchedEffect(key1 = state.podcasts.map { it.uuid }) {
    if (state.podcasts.isNotEmpty()) {
        podcastsRowState.scrollToItem(0)
    }
}
```

There's no need to keep track of `podcasts`. The coroutine inside the effect is guaranteed to be called only when the key changes.

semanticer:
But with that code, the list will scroll back to the start when returning from podcast detail:

https://github.com/user-attachments/assets/58b6ec8b-2a4c-437e-a343-8b6fcd8707a9

MiSikora:
Oh, you're right. But the proposed solution exchanges the current bug for another bug.

1. Search for "red" and let the results come in.
2. Search for "redwood" and let the results come int.
3. Scroll the list to some position.
4. Search again for "red" and let the results come in.
5. The list won't scroll to the start.

I'll admit that this is much more obscure bug. But it implies that the solution isn't correct.

MiSikora:
Thanks, I think it is the most stable key here. I've moved your changelog entry to `7.75` because we've just cut the `7.74` release. I'll run the CI checks now.

MiSikora (approved):

PR #2979, opened 2024-10-07
Speed up listening history search by limiting the number of results

geekygecko:
## Description

We had a report that searching over a large listening history was very slow. The cause of this was that it could return a large number of results. I have added a limit to the search results query to speed up this. 

## Testing Instructions
1. Subscribe to lots of podcasts
2. Run the following query in the Android Studio window "App Inspection" `UPDATE podcast_episodes SET last_playback_interaction_date = 1700000000000`
3. In the app open the Profile tab
4. Tap on Listening History
4. Search using a simple term so lots of results are return

✅ Verify the search results load quickly.

## Video

https://github.com/user-attachments/assets/a4f8fe4f-7e19-4143-9b37-712cd56aefea

## Checklist
 
- [x] with different themes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
     *   Add an advanced setting to display artwork in episode listing
         ([#2958](https://github.com/Automattic/pocket-casts-android/pull/2958))
 *   Bug Fixes
+    *   Speed up listening history search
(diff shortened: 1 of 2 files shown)


ashiagr (approved):
Thanks for fixing it!

PR #2985, opened 2024-10-08
Sync user ratings

MiSikora:
## Description

This adds syncing of user ratings. Initially I wanted to add `user_rating` column to the `podcasts` table but this made it inefficient when it came to synchronization. A separate table that consists only of ratings is simpler to manage.

### DB Migration

1. Install the app from 85d7c1f0e3e6b77fd8d286b23a26b79e9fdad6ce
2. Sign in.
3. Install the app from this branch.
4. The app shouldn't crash and you should not lose any data.

### Syncing

1. Rate a podcast on the device A.
2. Sync the device B.
3. Verify that rating synced. `SELECT p.title, rating FROM user_podcast_ratings JOIN podcasts AS p ON podcast_uuid = p.uuid ORDER BY title ASC`

### Local rating

1. Rate a podcast.
2. Verify that the rating is stored locally. `SELECT p.title, rating FROM user_podcast_ratings JOIN podcasts AS p ON podcast_uuid = p.uuid ORDER BY title ASC`

## Checklist

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/PodcastSyncProcess.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/PodcastSyncProcess.kt
@@ -12,6 +12,7 @@ import au.com.shiftyjelly.pocketcasts.models.entity.Folder
 import au.com.shiftyjelly.pocketcasts.models.entity.Playlist
 import au.com.shiftyjelly.pocketcasts.models.entity.Podcast
 import au.com.shiftyjelly.pocketcasts.models.entity.PodcastEpisode
+import au.com.shiftyjelly.pocketcasts.models.entity.UserPodcastRating
 import au.com.shiftyjelly.pocketcasts.models.to.StatsBundle
 import au.com.shiftyjelly.pocketcasts.models.to.SubscriptionStatus
 import au.com.shiftyjelly.pocketcasts.models.type.EpisodePlayingStatus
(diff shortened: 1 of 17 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

ashiagr (approved):
LGTM

mebarbosa:
> I wanted to add user_rating column to the podcasts table but this made it inefficient when it came to synchronization

Hey @MiSikora I am just curious. How did you find out this? Did you do any specific test for it?

MiSikora:
It is because call to refresh podcast doesn't return user given rating so it would be hard to fit this update for partial podcast updates. I could potentially keep ratings in the podcast entity but doing it in a separate one is better for the db normalization.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/PodcastSyncProcess.kt
```diff
@@ -321,6 +325,21 @@ class PodcastSyncProcess(
...
+            syncManager.getPodcastRatings()?.podcastRatingsList
+                ?.mapNotNull { rating ->
+                    val modifiedAt = rating.modifiedAt.toDate() ?: return@mapNotNull null
+                    UserPodcastRating(
+                        podcastUuid = rating.podcastUuid,
+                        rating = rating.podcastRating,
+                        modifiedAt = modifiedAt,
+                    )
+                }
+                ?.let { ratingsManager.updateUserRatings(it) }
```
geekygecko, line 339: Should we try to add this to the regular sync or include the greatest modifiedAt in the call so we don't download all the ratings with every refresh? Another option could be downloading the ratings before the end of year stats if that's the only time we use the table. I could be overthinking this and maybe it doesn't matter. 🤔

MiSikora: > Should we try to add this to the regular sync or include the greatest modifiedAt in the call so we don't download all the ratings with every refresh?

Maybe? I don't know. It doesn't seem like a lot of data. Especially considering it is a proto call. But I don't think that using the greatest `modifiedAt` is correct. If I change a rating on a different device it might never update locally.

> Another option could be downloading the ratings before the end of year stats if that's the only time we use the table.

I do that using `EndOfYearSync` sync class. I think overall syncing and syncing for EoY are separate things and shouldn't be conflated. Syncing here allows us to migrate our rating process to not check a rating with a network call when opening the screen.

PR #2986, opened 2024-10-08
[Auto Downloading] Update UI for Podcast Auto Downloading

### LINKED ISSUE #2971: [Project] Auto Downloading Episodes - Implement podcast settings
This should add the new settings to the UI and persist the value in database

mebarbosa:
## Description
- This adds the auto download settings to podcast setting
- Persist the download limit in database d0b94ac7cd763b924628274e217ce209d44221ca
- Figma: [VvERCFIopLjFL8Fnczsf8r-fi-1_756](https://href.li/?https://www.figma.com/file/VvERCFIopLjFL8Fnczsf8r/?node-id=1-756)

**Note:** Default behavior: Two latest episodes
**Note2**: I will align and address in another PR with designs if we should reset the auto download limit to two latest download limit after toggle `Auto Download` off
**Note 3**: I will address in another PR the database update after the user subscribes to a podcast

Fixes #2971 

### AUTO_DOWNLOAD feature flag OFF
1. Have the FF disabled
2. Go to some podcast -> Subscribe to this podcast -> Tap on Settings
3. Toggle the `Auto Download` ON
4. ✅ Ensure you don't see the limit download section

### AUTO_DOWNLOAD feature flag ON
1. Have the FF enabled
2. Go to some podcast -> Subscribe to this podcast -> Tap on Settings
3. Toggle the `Auto Download` ON
4. ✅ Ensure you see the limit download section
5. Tap on limit downloads label and select one option
6. ✅ Ensure your option is selected
8. ✅ Ensure 🔵 Tracked: podcast_settings_auto_download_limit_toggled, Properties: {"value":... is tracked

## Checklist
 
- [x] with different themes

### DIFF
--- a/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/podcast/PodcastSettingsFragment.kt
+++ b/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/podcast/PodcastSettingsFragment.kt
@@ -477,6 +477,11 @@ class PodcastSettingsFragment : BasePreferenceFragment(), FilterSelectFragment.L
                 ?.let { AutoDownloadLimitSetting.fromPreferenceString(it) }
                 ?: AutoDownloadLimitSetting.TWO_LATEST_EPISODE
 
+            analyticsTracker.track(
+                AnalyticsEvent.PODCAST_SETTINGS_AUTO_DOWNLOAD_LIMIT_TOGGLED,
+                mapOf("value" to autoDownloadLimitSetting.analyticsString),
+            )
+
             viewModel.setAutoDownloadLimit(autoDownloadLimitSetting)
 
             true
--- a/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/podcast/PodcastSettingsFragment.kt
+++ b/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/podcast/PodcastSettingsFragment.kt
@@ -475,15 +475,15 @@ class PodcastSettingsFragment : BasePreferenceFragment(), FilterSelectFragment.L
         autoDownloadPodcastsLimit?.setOnPreferenceChangeListener { _, newValue ->
             val autoDownloadLimitSetting = (newValue as? String)
                 ?.let { AutoDownloadLimitSetting.fromPreferenceString(it) }
-                ?: AutoDownloadLimitSetting.TWO_LATEST_EPISODE
 
-            analyticsTracker.track(
-                AnalyticsEvent.PODCAST_SETTINGS_AUTO_DOWNLOAD_LIMIT_TOGGLED,
-                mapOf("value" to autoDownloadLimitSetting.analyticsString),
-            )
+            autoDownloadLimitSetting?.let {
+                analyticsTracker.track(
+                    AnalyticsEvent.PODCAST_SETTINGS_AUTO_DOWNLOAD_LIMIT_TOGGLED,
+                    mapOf("value" to it.analyticsString),
+                )
 
-            viewModel.setAutoDownloadLimit(autoDownloadLimitSetting)
-
+                viewModel.setAutoDownloadLimit(it)
+            }
             true
         }
 
--- a/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/podcast/PodcastSettingsFragment.kt
+++ b/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/podcast/PodcastSettingsFragment.kt
@@ -23,6 +23,7 @@ import au.com.shiftyjelly.pocketcasts.podcasts.R
 import au.com.shiftyjelly.pocketcasts.podcasts.viewmodel.PodcastSettingsViewModel
 import au.com.shiftyjelly.pocketcasts.preferences.Settings
 import au.com.shiftyjelly.pocketcasts.preferences.model.AutoAddUpNextLimitBehaviour
+import au.com.shiftyjelly.pocketcasts.preferences.model.AutoDownloadLimitSetting
 import au.com.shiftyjelly.pocketcasts.repositories.podcast.PodcastManager
 import au.com.shiftyjelly.pocketcasts.settings.AutoAddSettingsFragment
 import au.com.shiftyjelly.pocketcasts.ui.extensions.getThemeColor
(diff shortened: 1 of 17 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please add tests for class <code>PodcastAutoDownloadLimitConverter</code> (or add <code>unit-tests-exemption</code> label to ignore this).</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

CookieyedCodes:
Just checking with this auto download limit will it be completely customisable?, be able to be based on the sort order? and will it not delete downloads/archive episodes from feeds that are already have more episides downloaded?

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/AppDatabase.kt
```diff
@@ -841,6 +843,10 @@ abstract class AppDatabase : RoomDatabase() {
             }
         }
 
+        val MIGRATION_100_101 = addMigration(100, 101) { database ->
+            database.execSQL("ALTER TABLE episode ADD COLUMN auto_download_limit INTEGER DEFAULT 2")
+        }
```
MiSikora, line 848: This migration is incorrect.

1. It doesn't update `auto_download_limit` for existing podcasts.
2. The definition doesn't match `Podcast` entity. The entity doesn't define default value on the SQL level but on the Kotlin level so there no `DEFAULT 2` when the table is created without migration.
3. The migration crashes the app due to a typo. It should be `TABLE podcasts` .instead of `TABLE episode`.

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/Podcast.kt
```diff
@@ -59,6 +60,7 @@ data class Podcast(
     @ColumnInfo(name = "show_notifications") var isShowNotifications: Boolean = false,
     @ColumnInfo(name = "show_notifications_modified") var showNotificationsModified: Date? = null,
     @ColumnInfo(name = "auto_download_status") var autoDownloadStatus: Int = 0,
+    @ColumnInfo(name = "auto_download_limit") var autoDownloadLimit: AutoDownloadLimitSetting = AutoDownloadLimitSetting.TWO_LATEST_EPISODE,
```
MiSikora, line 63: This default value leads to some significant issues. When I sign in and to a clean sync all of the podcasts have `auto_download_limit` set to `TWO_LATEST_EPISODES`. Moreover, all podcasts that I see in discover have this column set to `TWO_LATEST_EPISODES`.

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/type/AutoDownloadLimitSetting.kt
```diff
@@ -50,10 +50,12 @@ enum class AutoDownloadLimitSetting(
         fun fromPreferenceString(stringValue: String): AutoDownloadLimitSetting {
             return try {
                 val intValue = stringValue.toInt()
-                entries.firstOrNull { it.preferenceInt == intValue } ?: TWO_LATEST_EPISODE
+                entries.firstOrNull { it.id == intValue } ?: TWO_LATEST_EPISODE
```
MiSikora, line 53: I think it would be better to change this method to return `AutoDownloadLimitSetting?` instead. The current default value, `TWO_LATEST_EPISODES`, mixes different use cases and could lead to potential errors when we need different behaviors.

I suggest removing the default value and determining the appropriate setting at the call sites instead.

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/type/AutoDownloadLimitSetting.kt
```diff
@@ -50,10 +50,12 @@ enum class AutoDownloadLimitSetting(
...
             } catch (e: Exception) {
                 TWO_LATEST_EPISODE
             }
         }
+
+        fun fromInt(id: Int) = (AutoDownloadLimitSetting.entries.firstOrNull { it.id == id } ?: TWO_LATEST_EPISODE)
```
MiSikora, line 59: Similarly to `fromPreferenceString` I think I'd be better to use any default value in case of missing entries.

MiSikora (changes requested):

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/Podcast.kt
mebarbosa: I am going to update it to default OFF and change to TWO_LATEST_EPISODE after the user subscribes to the podcast

mebarbosa:
@MiSikora I addressed all changes. 

After implementing them I tried to run the migration and I had to do it twice. When I checked to `main` and build the app, I realized the code from the previous branch was installed. I cleaned build, deleted manually the local cache and the old code was still there. So I run again with `./gradlew :app:installDebug --rerun-tasks` command and finally was able to get the `main` code installed

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/podcast/PodcastSettingsFragment.kt
```diff
@@ -490,6 +517,18 @@ class PodcastSettingsFragment : BasePreferenceFragment(), FilterSelectFragment.L
         preferenceFilters?.isVisible = availableFilters.isNotEmpty()
     }
 
+    private fun updateAutoDownloadLimit(podcast: Podcast) {
+        val options = AutoDownloadLimitSetting.entries
+        val podcastAutoDownloadLimit = podcast.autoDownloadLimit ?: AutoDownloadLimitSetting.TWO_LATEST_EPISODE
```
MiSikora, line 522: Nothing to change here right now but it will be important to remember to use `TWO_LATEST_EPISODE` as the default value when doing actual downloading.

MiSikora (approved):
I think there is a UX issue. It doesn't make sense to me to have an `Off` setting when there is a toggle to disable auto downloads. What does it mean to enable the setting and have the limit set to `Off`?

share_8631750866837509173

mebarbosa:
> I think there is a UX issue. It doesn't make sense to me to have an `Off` setting when there is a toggle to disable auto downloads. What does it mean to enable the setting and have the limit set to `Off`? If it means that there is no limit there is a separate value for it.
> 
> Off	All
> share_8631750866837509173	share_2201720753321164508

@MiSikora I agree. It does not make sense to me to have this toggle. I reported this here p1728473112997409/1728311298.538509-slack-C07QMC2GB1B and I am waiting for the response. In the next PR I can address it

mebarbosa:
> Just checking with this auto download limit will it be completely customisable?, be able to be based on the sort order? and will it not delete downloads/archive episodes from feeds that are already have more episides downloaded?

@CookieyedCodes We want to implement something to download X latest episodes for now. 
For the deletion we are building some rules for this to exclude some cases

CookieyedCodes:
Hmmmmm while I could see that as useful I think a per podcast auto download that is tied to the list sort order & ignoreing archived episodes would be more helpful imo because the vast majority of podcasts are listened to in backlog imo & once caught up on it would behave like the current auto download functionality and as the default is newest to oldest it would automatically have that functionality, even a toggle between both options latest and list would be a massive improvement 🤔

PR #2989, opened 2024-10-09
Sync end of year data in a custom job

MiSikora:
## Description

This pull request introduces `CachedAction` class that memoizes result of a job and prevents calls of the same lambda if the result is already memoized or the lambda is being executed. This class is then used to sync data necessary for EoY.

## Testing Instructions

Right now there isn't much to test since there is no preview of EoY. Code review and CI checks should be enough.

## Checklist

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/endofyear/EndOfYearSync.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/endofyear/EndOfYearSync.kt
@@ -1,0 +1,119 @@
+package au.com.shiftyjelly.pocketcasts.repositories.endofyear
+
+import au.com.shiftyjelly.pocketcasts.models.entity.UserPodcastRating
+import au.com.shiftyjelly.pocketcasts.preferences.Settings
+import au.com.shiftyjelly.pocketcasts.repositories.podcast.EpisodeManager
+import au.com.shiftyjelly.pocketcasts.repositories.podcast.HistoryManager
+import au.com.shiftyjelly.pocketcasts.repositories.ratings.RatingsManager
+import au.com.shiftyjelly.pocketcasts.repositories.sync.SyncManager
+import au.com.shiftyjelly.pocketcasts.servers.extensions.toDate
+import au.com.shiftyjelly.pocketcasts.utils.coroutines.SyncedAction
+import au.com.shiftyjelly.pocketcasts.utils.coroutines.run
+import au.com.shiftyjelly.pocketcasts.utils.featureflag.Feature
+import au.com.shiftyjelly.pocketcasts.utils.featureflag.FeatureFlag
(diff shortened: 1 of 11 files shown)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/endofyear/EndOfYearSync.kt
```diff
@@ -0,0 +1,118 @@
...
+            )
+        }
+        ratingsManager.updateUserRatings(userRatings)
+    }
+
+    private suspend fun syncHistoryForYear(year: Int) {
+        // check for an episode interacted with before this year and assume they have the full listening history if they exist
+        if (anyEpisodeInteractionBeforeYear(year)) {
+            return
+        }
```
ashiagr, line 87: I remember a case where a user had a listening history for 2022. However, they did not access the app on their secondary Android device between Jan and Nov 2023. They continued to use the app on their primary device during that period. Because of this, they noticed a mismatch in their listening time comparison from last year between different devices.

The issue was only resolved when the app was reinstalled on the secondary device.

I think that removing this condition will fix the problem.

MiSikora: Yes, I've been wondering about it as well and if should keep this check when migrating syncing to this process. On the one hand it's quite clever but on the other it fails for scenarios that you mentioned.

I'm wondering if there is any check that we could do to avoid network requests. Maybe I could store the eoy sync date in the preferences and check that. What do you think?

ashiagr: > Maybe I could store the eoy sync date in the preferences and check that. What do you think?

That makes sense to me. How long would you wait to allow another network request?

MiSikora: I think something in the range of 6 to 12 hours is ok. This would only affect users listening on multiple devices so it shouldn't be that big of a deal to have this accuracy window.

ashiagr (approved):

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/endofyear/EndOfYearSync.kt
ashiagr: I've approved the PR. Feel free to tackle it in another PR.

MiSikora: I'll add it here to avoid changing this file in the future PRs.

PR #2955, opened 2024-09-30
Use playback speed and skipped chapters for remaining playback time

MiSikora:
## Description

This PR accounts for playback speed and skipped chapters when displaying remaining playback time. Affected UI is remaining time under the player's seek bar and remaining time under next chapter button.

Closes #2249

## Testing Instructions

1. Play an episode with chapters.
2. Open full screen player.
3. Notice that the remaining duration of the whole episode and of a chapter changes with the playback speed.
4. Notice that the remaining duration changes with the chapters selection.
5. Smoke test seeking behavior for audio and video episodes.

## Checklist
- [x] If this is a user-facing change, I have added an entry in CHANGELOG.md
- [x] Ensure the linter passes (`./gradlew spotlessApply` to automatically apply formatting/linting)
- [x] I have considered whether it makes sense to add tests for my changes
- [x] ~All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml`~
- [x] ~Any jetpack compose components I added or changed are covered by compose previews~
- [x] ~I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.~

PR #3044, opened 2024-10-18
Add a setting for adjusting remaining time

MiSikora:
## Description

We got a feedback [here](https://github.com/Automattic/pocket-casts-android/pull/2955#issuecomment-2421956935) that this should be a setting.

I'm targeting beta since I suspect that a lot of users might be confused about this change when enabled by default.

## Testing Instructions

1. Play something with chapters.
2. Change playback speed.
3. Deselect some chapters.
4. Check that the player doesn't display remaining time as adjusted time.
5. Go to `Profile -> Settings -> General`.
6. Enable `Adjust remaining time` setting.
7. Check that the player displays adjusted time as the remaining time.
8. Disable `Adjust remaining time` setting.
9. Check that the player no longer displays adjusted time as the remaining time.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-android/milestone/66">7.75 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

geekygecko:
I wonder if we should wait to add this setting until we see if people complain about it. Without the change, the time remaining is wrong. I'm a bit cautious about adding more settings. But I am happy to put the setting in if you think we need to.

> I would potentially suggest as an alternative instead of adding a setting, that spefically while scrubbing it would toggle the time temporarily 🤔🤔

This seems too hidden, and if you are scrubbing wouldn't you lose your playback position?

MiSikora:
> I wonder if we should wait to add this setting until we see if people complain about it.

Personally I really wouldn't want to use this feature. The main problem that I see is that without this setting people who don't like it have to wait for the next release because it won't be a hot fix. And I don't want our support to suffer if we get backlash. On the other hand there's been only one public voice against it so far. So I'm fine with either. How do you want to proceed?

geekygecko:
> How do you want to proceed?

Let's add it to the settings, as a few people already want to turn it off.

geekygecko (approved):

jamesdeluk:
`Adjust remaining time` is not in `Profile -> Settings -> General`. Has this been removed? If so, why?

PR #3073, opened 2024-10-22
[Project] Up Next Shuffle - Implement Shuffle Action

### LINKED ISSUE #3064: [Project] Up Next Shuffle - Implement Shuffle Action
Implement the shuffle action when the user taps on shuffle button

mebarbosa:
## Description
- This PR add the action to shuffle when tapping on up next shuffle button

> [!NOTE]  
> In another PR I will handle the shuffle state sync

Fixes #3064

## Testing Instructions
Have the shuffle feature flag enabled

### Shuffle Feature

1. Add some episodes to Up Next
2. Play an episode
3. Have the shuffle disabled
4. Seek the current playing episode to almost the end
5. Wait for the episode ends
6. ✅ Ensure the first episode from up next starts playing
7. Enable shuffle
8. Seek the current playing episode to almost the end
9. Wait for the episode ends
10. ✅ Ensure a random episode from up next starts playing _(Since we are using a random way to get the next episode, we can also pick the first one from up next, so I recommend to test this a couple of times to ensure that is taking random episodes from up next)_
11. ✅ Ensure the random episode that started playing was removed from the up next queue

#### Play Next / Play Last
1. Have some episodes in Up next and shuffle enabled
2. Open an podcast
3. Swipe to add episode to Up next
4. ✅ Ensure you can add episodes to both end of the queue and top of the queue

image

#### Play Up Next episode on Tap
1. Have some episodes in Up next and shuffle enabled
2. Go to Profile -> Settings -> General -> Enable `Play Up Next episode on Tap`
3. Go to Up next
4. Tap on some episode from Up next on queue
5. ✅ Ensure the episode you tapped starts playing
6. ✅ Ensure the up next kept the episode sort (only changed the episode you tapped and the current playing episode)

#### Sync Up next Queue
1. Have two devices logged with the same account
2. Have some populated Up Next
3. Tap on Refresh now in both of them to ensure they are synced
4. Ensure the two devices have the same up next queue
5. On device A, enable shuffle, wait for the current episode finishes
6. ✅ Ensure a random episode was picked
7. Observe the current Up Next
8. Go to Profile and Refresh Now
9. On device B, go to Profile and Refresh Now
10. Check the Up Next
11. ✅ Ensure the Up Next is the same you checked on step 7

## Screenshots or Screencast 
[Screen_recording_20241022_161358.webm](https://github.com/user-attachments/assets/5e93b8af-c9f0-4cf9-a6d3-ed2c792ed091)

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/UpNextAdapter.kt
+++ b/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/UpNextAdapter.kt
@@ -173,30 +173,43 @@ class UpNextAdapter(
                 }
 
                 shuffle.isVisible = FeatureFlag.isEnabled(Feature.UP_NEXT_SHUFFLE)
-                shuffle.setShuffleDisabledButton()
+                shuffle.updateShuffleButton()
 
                 shuffle.setOnClickListener {
-                    shuffle.setShuffleEnabledButton()
+                    settings.upNextShuffle.set(!settings.upNextShuffle.value, updateModifiedAt = false)
+                    shuffle.updateShuffleButton()
                 }
 
                 root.layoutParams.height = ViewGroup.LayoutParams.WRAP_CONTENT
             }
         }
 
-        private fun ImageButton.setShuffleDisabledButton() {
-            this.setImageDrawable(ContextCompat.getDrawable(context, IR.drawable.shuffle))
-            this.contentDescription = context.getString(LR.string.up_next_shuffle_button_content_description)
-            this.setImageTintList(ColorStateList.valueOf(ThemeColor.primaryIcon02(theme)))
+        private fun ImageButton.updateShuffleButton() {
+            if (!FeatureFlag.isEnabled(Feature.UP_NEXT_SHUFFLE)) return
 
-            TooltipCompat.setTooltipText(this, context.getString(LR.string.up_next_shuffle_button_content_description))
-        }
+            val isEnabled = settings.upNextShuffle.value
 
-        private fun ImageButton.setShuffleEnabledButton() {
-            this.setImageDrawable(ContextCompat.getDrawable(context, IR.drawable.shuffle_enabled))
-            this.contentDescription = context.getString(LR.string.up_next_shuffle_disable_button_content_description)
-            this.setImageTintList(ColorStateList.valueOf(ThemeColor.primaryIcon01(theme)))
+            this.setImageDrawable(
+                ContextCompat.getDrawable(
+                    context,
+                    if (isEnabled) IR.drawable.shuffle_enabled else IR.drawable.shuffle,
+                ),
+            )
 
-            TooltipCompat.setTooltipText(this, context.getString(LR.string.up_next_shuffle_button_content_description))
+            this.contentDescription = context.getString(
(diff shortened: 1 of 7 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please add tests for class <code>RemoveAndShuffle</code> (or add <code>unit-tests-exemption</code> label to ignore this).</td>
    </tr>
  </tbody>
</table>

</p>

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/UpNextQueueImpl.kt
```diff
@@ -232,9 +238,13 @@ class UpNextQueueImpl @Inject constructor(
         episodes.forEach { playLastNow(it, downloadManager, null) }
     }
 
-    override suspend fun removeEpisode(episode: BaseEpisode) {
+    override suspend fun removeEpisode(episode: BaseEpisode, shouldShuffleUpNext: Boolean) {
```
MiSikora, line 241: There's no need to pass `shouldShuffleUpNext` to this function. You already have access to settings that hold the property you can read. This also simplifies a lot of things because there will be no need to for `RemoveAndShuffle` type since everything can be determined by reading the user setting.

MiSikora (changes requested):
Overall works well but the code can be simplified by not passing the `shuffle` state between different layers.

There's also an issue when an episode is archived or marked as played while listening to it. If that happens the next episode isn't chosen at random.

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/UpNextAdapter.kt
```diff
@@ -173,30 +173,43 @@ class UpNextAdapter(
...
-            TooltipCompat.setTooltipText(this, context.getString(LR.string.up_next_shuffle_button_content_description))
-        }
+            val isEnabled = settings.upNextShuffle.value
+
+            this.setImageDrawable(
+                ContextCompat.getDrawable(
+                    context,
+                    if (isEnabled) IR.drawable.shuffle_enabled else IR.drawable.shuffle,
+                ),
+            )
```
geekygecko, line 197:
```suggestion
            this.setImageResource(if (isEnabled) IR.drawable.shuffle_enabled else IR.drawable.shuffle)
```

Can you use `setImageResource` instead here?

mebarbosa:
Hey @MiSikora . Thank you for testing and catching the played / archived scenario. I've addressed them already and updated the PR description with more scenarios to test. When you have time, could you please test them again?

MiSikora (approved):
I found one minor thing but I'm not sure if it is even an issue. When I archive or mark an episode not from the player but, for example, from a podcast's episode list the Up Next doesn't use shuffle. But I think it is a very obscure scenario that we might not even want to support.

mebarbosa:
> I found one minor thing but I'm not sure if it is even an issue. When I archive or mark an episode not from the player but, for example, from a podcast's episode list the Up Next doesn't use shuffle. But I think it is a very obscure scenario that we might not even want to support.

I will check this scenario as well and address in another PR if we decided to handle. Thank you again!

PR #3094, opened 2024-10-25
Keep the playing service in the foreground when casting

geekygecko:
## Description

Chromecasting from Pocket Casts has an issue that the Pocket Casts PlaybackService doesn’t run in the foreground, so the system kills off the service while it's still needed. This means that the episode progress would not be tracked, and when the episode ends, the next episode in the Up Next would not start. For the service to run in the foreground, we need to create a notification. We didn't create a notification as the Chromecast library creates a separate playback notification causing  two playback notifications but it seems recent versions of Android only show one notification. 

## Testing Instructions

1. Play an episode
2. ✅ Verify there is a playback notification
3. Cast the episode 
4. ✅ Verify there is only a casting playback notification
5. ✅ Verify logcat doesn't contain the following message "can't startForeground as the notification is null"

## Checklist
 
- [x] with different themes

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -565,6 +565,11 @@
             </intent-filter>
         </receiver>
 
+        <receiver
+            android:name="androidx.mediarouter.media.MediaTransferReceiver"
+            android:exported="true">
+        </receiver>
+
         <service
                 android:name="au.com.shiftyjelly.pocketcasts.repositories.playback.PlaybackService"
                 android:exported="true"
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackService.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackService.kt
@@ -243,12 +243,6 @@ open class PlaybackService : MediaBrowserServiceCompat(), CoroutineScope {
             val isForegroundService = isForegroundService()
             val state = playbackState.state
 
-            // If we have switched to casting we need to remove the notification
-            if (isForegroundService && notification == null && playbackManager.isPlaybackRemote()) {
-                stopForeground(STOP_FOREGROUND_REMOVE)
-                LogBuffer.i(LogBuffer.TAG_PLAYBACK, "stopForeground as player is remote")
-            }
-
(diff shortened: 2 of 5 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

app/src/main/AndroidManifest.xml
```diff
@@ -565,6 +565,11 @@
             </intent-filter>
         </receiver>
 
+        <receiver
+            android:name="androidx.mediarouter.media.MediaTransferReceiver"
+            android:exported="true">
+        </receiver>
```
geekygecko, line 571: Adding this "enables media transfer among devices with system UI" which means that on the casting notification there is the option to switch where you are casting to.

https://developers.google.com/cast/docs/android_sender/output_switcher#enable_output_switcher_in_androidmanifestxml

| Before | After |
| --- | --- |
| Screenshot_20241025_134511 | Screenshot_20241025_134548 |

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackService.kt
```diff
@@ -303,8 +297,11 @@ open class PlaybackService : MediaBrowserServiceCompat(), CoroutineScope {
                             LogBuffer.i(LogBuffer.TAG_PLAYBACK, "stopForeground state: $state removing notification: $removeNotification")
                         }
 
-                        @Suppress("DEPRECATION")
-                        stopForeground(removeNotification)
+                        // When paused keep the notification otherwise remove it
+                        stopForeground(if (removeNotification) STOP_FOREGROUND_REMOVE else STOP_FOREGROUND_DETACH)
+                        if (removeNotification) {
+                            notificationManager.cancel(Settings.NotificationId.PLAYING.value)
```
geekygecko, line 303: If you long press on the mini player and choose "Close and clear Up Next" before this change there would be a blank notification without any controls left in the notification drawer.

ashiagr (approved):

mkevins:
This may resolve the issue reported here: 8987110-zd-a8c

PR #3126, opened 2024-10-29
Change launchMode on MainActivity from "singleTop" to "singleTask"

hamorillo:
## Description

Looking at the [Android documentation](https://developer.android.com/guide/components/activities/tasks-and-back-stack#ManifestForTasks), it seems like the `singleTask` `launchMode` is more correct to the Gravatar's `MainActivity`.

> "singleTask"
> The system creates the activity at the root of a new task or locates the activity on an existing task with the same affinity. If an instance of the activity already exists, the system routes the intent to the existing instance through a call to its [onNewIntent()](https://developer.android.com/reference/android/app/Activity#onNewIntent(android.content.Intent)) method, rather than creating a new instance. Meanwhile all of the other activities on top of it are destroyed.

vs 

> "singleTop"
> If an instance of the activity already exists at the top of the current task, the system routes the intent to that instance through a call to its [onNewIntent()](https://developer.android.com/reference/android/app/Activity#onNewIntent(android.content.Intent)) method, rather than creating a new instance of the activity. The activity is instantiated multiple times, each instance can belong to different tasks, and one task can have multiple instances (but only if the activity at the top of the back stack is not an existing instance of the activity).

Despite being just a one-line change, please smoke test everything that comes to your mind related to how the activity is launched. I've done the same, but I have limited knowledge of the code base.

cc: @MiSikora 

## Testing Instructions
- Smoke test the app. Test deep links, close the app, and do everything you can about how the activity is launched.

## Checklist
 ~~- [ ] If this is a user-facing change, I have added an entry in CHANGELOG.md~~ 
 ~~- [] Ensure the linter passes (`./gradlew spotlessApply` to automatically apply formatting/linting)~~ 
 ~~- [ ] I have considered whether it makes sense to add tests for my changes~~ 
 ~~- [ ] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml`~~ 
 ~~- [ ] Any jetpack compose components I added or changed are covered by compose previews~~ 
 ~~- [ ] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.~~ 
 
~~- [ ] with different themes~~ 
~~- [ ] with a landscape orientation~~ 
~~- [ ] with the device set to have a large display and font size~~ 
~~- [ ] for accessibility with TalkBack~~

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -53,7 +53,7 @@
         <activity
                 android:name=".ui.MainActivity"
                 android:configChanges="screenSize|smallestScreenSize|screenLayout"
-                android:launchMode="singleTop"
+                android:launchMode="singleTask"
                 android:colorMode="default"
                 android:exported="true"
                 android:windowSoftInputMode="stateAlwaysHidden|adjustPan">


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

MiSikora (changes requested):
Unfortunately, this breaks the app. When a different activity is on top minimizing the app and then opening the app again collapses the activity stack to the `MainActivity`. In the attached video you can see that. The Upsell screen disappears after I reopen the active application either by tapping on the app's icon, on a widget, or on a media notification. This happens with other activities as well.

https://github.com/user-attachments/assets/2889bba0-4579-4d8f-949d-2fa96f4adfd9

hamorillo:
Thanks for the testing @MiSikora! 

I've been reading, and investigating about that, and it seems to be the expected behavior when using the `singletrack` in the `Main` activity. There are some threads on Stackoverflow like [this](https://stackoverflow.com/a/3002890) where they mentioned that, despite not being clearly explained in the documentation that's how it works.

Some people suggest some workarounds that consist of not using the `singleTask` mode for the Main Activity.

Said that, I think we shouldn't move forward with this change. PocketCast is working well and create a new activity with the suggested workarounds just to change the launch mode of the `MainActivity` doesn't make sense. If you agree, I'll close this PR.

hamorillo:
Anyway, thanks for you help @MiSikora!

PR #3183, opened 2024-11-06
Playback: Attempt to fix 416 source error

ashiagr:
## Description
Internal Ref: p1730809737477079-slack-C02A333D8LQ

This attempts to fix `416` playback source error by resetting episode cache if [416 `responseCode`](https://github.com/androidx/media/blob/c35a9d62baec57118ea898e271ac66819399649b/libraries/datasource_okhttp/src/main/java/androidx/media3/datasource/okhttp/OkHttpDataSource.java#L300) is encountered. 

Relevant discussions:
https://github.com/androidx/media/issues/1032#issuecomment-1921375048
https://github.com/google/ExoPlayer/issues/10577

To be safe, changes are guarded behind a feature flag. 

## Testing Instructions
I was not able to reproduce the issue. Code review should be sufficient.

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/CacheWorker.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/CacheWorker.kt
@@ -11,6 +11,7 @@ import androidx.media3.datasource.DataSpec
 import androidx.media3.datasource.cache.CacheWriter
 import androidx.work.Constraints
 import androidx.work.Data
+import androidx.work.ExistingWorkPolicy
 import androidx.work.NetworkType
 import androidx.work.OneTimeWorkRequest
 import androidx.work.WorkInfo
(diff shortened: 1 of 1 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

geekygecko (approved):
Do you think this could accidentally queue up the same episode multiple times for the CacheWorker?

We don’t have any examples of using enqueueUniqueWork with episode uuids as the tag but maybe that would work. 🤔

ashiagr:
> Do you think this could accidentally queue up the same episode multiple times for the CacheWorker?

The current playing episode caching worker has a unique tag that is not dependent on the episode uuid. Before starting a new worker, I explicitly canceled the previous one here:

https://github.com/Automattic/pocket-casts-android/blob/56f8e72da242f0312cd4d5f02cce392aa84dac48/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/CacheWorker.kt#L96

As the same behavior can be enforced using `enqueueUniqueWork`, I've updated it in [36354e8](https://github.com/Automattic/pocket-casts-android/pull/3183/commits/36354e88be90e7bf3e4c5934d3bd9ec82d8459d6).

PR #3194, opened 2024-11-08
[Paywall Experiment] - Use the same UUID for both Tracks and Explat

### LINKED ISSUE #3193: [Paywall Experiment] - Use the same UUID for both Tracks and Explat
### Description

- Currently we are not initializing the Explat with the same UUID that we use for Tracks. This was causing an issue that we are not getting data from the AA experiment
- See: p1730936149575659/1729729858.793479-slack-C046HLX37K2

mebarbosa:
## Description
- We encountered an issue where data from the AA experiment was not being collected. After investigation, the data team identified the problem: we were using different UUIDs for Tracks and Explat when we should have been using the same one. So this PR make sure we use the same uuid for both cases.
- See: p1730936149575659/1729729858.793479-slack-C046HLX37K2

Fixes #3193

## Testing Instructions
1. Apply [log-track.patch](https://github.com/user-attachments/files/17670975/log-track.patch) to log the uuid every time an event is tracked
2. Fresh install the app
3. Look for `ExperimentsProvider` in logcat
4. See what uuid the explat was initialized: ExperimentsProvider: Initializing experiments with uuid: {uuid}
5. Navigate to some screen to track event
6. ✅ Ensure the event tracked has the same uuid you saw in step 4
7. Log in with a new account
8.  ✅ See the Experiment was refreshed with another uuid
9. Track some events
10. ✅ Ensure the event tracked has the same uuid you saw in step 8
11. Log out
12. ✅ See the Experiment was refreshed with another uuid
13. Track some events
14. ✅ Ensure the event tracked has the same uuid you saw in step 12

## Screenshots or Screencast 

https://github.com/user-attachments/assets/cb0b8838-3c51-43e2-bdbc-7aeff9fdb6d4

## Checklist

### DIFF
--- a/modules/services/analytics/src/main/java/au/com/shiftyjelly/pocketcasts/analytics/IdentifyingTracker.kt
+++ b/modules/services/analytics/src/main/java/au/com/shiftyjelly/pocketcasts/analytics/IdentifyingTracker.kt
@@ -29,7 +29,7 @@ abstract class IdentifyingTracker(
         }
     }
 
-    protected val anonID: String?
+    val anonID: String?
         get() {
             if (anonymousID == null) {
                 anonymousID = preferences.getString(anonIdPrefKey, null)
@@ -37,7 +37,7 @@ abstract class IdentifyingTracker(
             return anonymousID
         }
 
-    protected fun generateNewAnonID(): String {
+    internal fun generateNewAnonID(): String {
         val uuid = UUID.randomUUID().toString().replace("-", "")
         Timber.d("\uD83D\uDD35 New anonID generated in " + this.javaClass.simpleName + ": " + uuid)
         preferences.edit {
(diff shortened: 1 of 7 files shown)


modules/services/analytics/src/main/java/au/com/shiftyjelly/pocketcasts/analytics/IdentifyingTracker.kt
```diff
@@ -29,15 +29,15 @@ abstract class IdentifyingTracker(
         }
     }
 
-    protected val anonID: String?
+    val anonID: String?
```
mebarbosa, line 32: I updated this to be public because I need to access this from outside of the class, but it's safe because is `val` and the variable that is modified is actually `anonymousID`. We also do the same for uuid in this classe

ashiagr (approved):

PR #3218, opened 2024-11-12
Inject OkHttpClient to episode details task and to status checker

MiSikora:
## Description

`UpdateEpisodeDetailsTask` and `ServiceStatusChecker` didn't use DI for `OkHttpClient` which can lead to performance degradation and generally is harder to maintain as we update our network layer.

## Testing Instructions

1. Test if status check still works.
2. Play an episode and check if you can see the episode details.

## Checklist

### DIFF
--- a/modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/status/ServiceStatusChecker.kt
+++ b/modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/status/ServiceStatusChecker.kt
@@ -3,6 +3,7 @@ package au.com.shiftyjelly.pocketcasts.settings.status
 import android.content.Context
 import android.net.ConnectivityManager
 import android.net.NetworkCapabilities
+import au.com.shiftyjelly.pocketcasts.servers.di.NoCache
 import au.com.shiftyjelly.pocketcasts.utils.extensions.await
 import dagger.hilt.android.qualifiers.ApplicationContext
 import java.util.concurrent.TimeUnit
@@ -11,7 +12,10 @@ import okhttp3.OkHttpClient
 import okhttp3.Request
 import okhttp3.Response
 
-class ServiceStatusChecker @Inject constructor(@ApplicationContext val context: Context) {
+class ServiceStatusChecker @Inject constructor(
+    @ApplicationContext private val context: Context,
+    @NoCache private val httpClient: OkHttpClient,
+) {
 
     suspend fun check(check: Check): ServiceStatus {
         return when (check) {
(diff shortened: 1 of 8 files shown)


modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/status/ServiceStatusChecker.kt
```diff
@@ -32,7 +36,8 @@ class ServiceStatusChecker @Inject constructor(@ApplicationContext val context:
     private suspend fun checkUrl(url: String): ServiceStatus {
         val log = StringBuilder()
 
-        val okHttpClient = OkHttpClient.Builder()
+        val okHttpClient = httpClient
+            .newBuilder()
```
MiSikora, line 40: I would like to avoid doing this as well but I don't know how we could access `chain.connection()` other than through an interceptor. And this way seems much simpler than synchronizing an interceptor and logs across multiple threads/coroutines.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/download/UpdateEpisodeDetailsTask.kt
```diff
@@ -91,7 +89,6 @@ class UpdateEpisodeDetailsTask @AssistedInject constructor(
                 val downloadUrl = episode.downloadUrl?.toHttpUrlOrNull() ?: continue
                 val request = Request.Builder()
                     .url(downloadUrl)
-                    .addHeader("User-Agent", "Pocket Casts")
```
geekygecko, line 94: It would be good to check this user agent stays the same. I could be wrong but I have a vague memory that we made the user agent stay the same across all the apps so it was easier for podcasters to track the Pocket Casts platform.

MiSikora: This particular user agent was different from the one that we use in all other places. Now it is aligned to `Pocket Casts/Android/<VERSION>`.

MiSikora: I changed it in https://github.com/Automattic/pocket-casts-android/pull/3218/commits/69ee862e26105ba8ca3950f29732852629ec9ec6. Now communication with our servers uses the full user-agent and with external ones* `Pocket Casts` one.

*this is mostly true because we still fetch things like chapters through `PodcastCacheService`. But downloads, streaming, and episode details use the simple version.

geekygecko: Thanks for doing that. I can see iOS uses "Pocket Casts" [here](https://github.com/Automattic/pocket-casts-ios/blob/trunk/Modules/Server/Sources/PocketCastsServer/Public/Sharing/Structs/ServerConstants.swift#L96).

geekygecko (approved):

PR #3226, opened 2024-11-13
[Tooling] Automate creation of backmerge PRs

AliSoftware:
## Description

This PR automates 🤖  the creation of the backmerge PR after having triggered a new beta build (post-code-freeze or post-beta) and after having finalized the final release.

It also removes the `complete_code_freeze` lane after having moved the action that it did (triggering a new build) directly in the `code_freeze` lane. That way, Release Managers don't have to run `code_freeze` then `complete_code_freeze` lanes back-to-back every time, and will instead now just have a single `code_freeze` lane to run instead of two. 🎉 

> [!NOTE]
> This PR has a corresponding diff on the ReleasesV2 scenario (see D166226-code) to adjust the tasks in the checklist accordingly.

## Testing Instructions

The whole scenario is not easily testable without going through the real release steps.

But if there are new commits on the current `release/*` branch that have not been merged yet, we can at least test the `create_backmerge_pr` lane in isolation:
 - Run `bundle exec fastlane create_backmerge_pr`
 - Validate that it creates an intermediate branch, push it, and [creates the PR in GitHub](https://github.com/Automattic/pocket-casts-android/pull/3225) with the appropriate commits.

## Extras

This PR also updates the `release-toolkit` to [its latest version `12.3.1`](https://github.com/wordpress-mobile/release-toolkit/releases/tag/12.3.1)—which amongst other things will [fix the issue reported recently](https://github.com/wordpress-mobile/release-toolkit/pull/610) by @MiSikora about links not working on GitHub Releases (internal ref: p1731397370353649/1729610708.303749-slack-C028JAG44VD)

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -541,19 +541,6 @@ platform :android do
       labels: ['Releases'],
       milestone_title: release_version_next
     )
-  rescue StandardError => e
-    error_message = <<-MESSAGE
-      Error creating backmerge pull request:
-
-      #{e.message}
-
-      If this is not the first time you are running the release task, the backmerge PR for the version `#{version}` might have already been previously created.
-      Please close any previous backmerge PR for `#{version}`, delete the previous merge branch, then run the release task again.
-    MESSAGE
-
-    buildkite_annotate(style: 'error', context: 'error-creating-backmerge', message: error_message) if is_ci
-
-    UI.user_error!(error_message)
   end
 
   #####################################################################################
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -129,25 +129,14 @@ platform :android do
     )
     push_to_git_remote(tags: false)
 
+    trigger_release_build(branch_to_build: "release/#{new_version}")
+    create_backmerge_pr
+
     # We cannot use the `copy_branch_protection` action here as it would create a branch protection rule with the same
-    # restrictions we have in `trunk`, and the release managers wouldn't be able to push due to permissions.
+    # restrictions we have in `main`, and the release managers wouldn't be able to push due to permissions.
     # This should be changed only when we have PC Android releases done on CI, when the CI bot is the one running `git push`.
     set_branch_protection(repository: GH_REPOSITORY, branch: "release/#{new_version}")
     set_milestone_frozen_marker(repository: GH_REPOSITORY, milestone: new_version)
-  end
-
-  lane :complete_code_freeze do |options|
-    ensure_git_branch(branch: '^release/') # Match branch names that begin with `release/`
-    ensure_git_status_clean
-
-    new_version = release_version_current
-
-    UI.important("Completing code freeze for: #{new_version}")
-    unless options[:skip_confirm] || UI.confirm('Do you want to continue?')
-      UI.user_error!("Terminating as requested. Don't forget to run the remainder of this automation manually.")
-    end
-
-    trigger_release_build(branch_to_build: "release/#{new_version}")
   end
 
   # @option [String] base_version The "x.y" version number to create a beta from. Defaults to the current version as defined in the version file of `main` branch
@@ -197,6 +186,7 @@ platform :android do
     push_to_git_remote(tags: false)
 
     trigger_release_build(branch_to_build: "release/#{release_version_current}")
+    create_backmerge_pr
   end
 
   # Sets the stage to start working on a hotfix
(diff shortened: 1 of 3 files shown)


geekygecko (approved):

geekygecko:
> Release Managers don't have to run code_freeze then complete_code_freeze lanes back-to-back every time

Thanks for making this change. This week is the first time I'm doing a release and I noticed this. Where do I suggest other changes to the process. For example:

### Idea 1.

> Move [all open issues and PRs in the 7.77 milestone](https://github.com/Automattic/pocket-casts-android/milestones/7.77) to next version

This isn't a valid URL.

### Idea 2.

After a release there is a step "Post a message in #pocket-casts-dev-android on Slack, notifying the team that you've released the release builds to 10% of users." the team also do this for a beta so it would be good to add a reminder for that.

fastlane/Fastfile
```diff
@@ -539,6 +533,29 @@ platform :android do
...
+
+      #{e.message}
+
+      If this is not the first time you are running the release task, the backmerge PR for the version `#{version}` might have already been previously created.
+      Please close any previous backmerge PR for `#{version}`, delete the previous merge branch, then run the release task again.
+    MESSAGE
+
+    buildkite_annotate(style: 'error', context: 'error-creating-backmerge', message: error_message) if is_ci
+
+    UI.user_error!(error_message)
```
AliSoftware, line 556: I just realized that I copy-pasted this error handling from the Fastfile of another repo… which already had the lanes being executed by CI instead of developers' machines.

Moving PCAndroid to have the fastlane lanes being run on CI is the next step in my work on improving Release Automation, but it's not implemented yet. Which means that in the meantime, this `rescue` block in case an error occurs creating the PR: (1) not only provides a slightly misleading error message / recovery suggestion in the meantime and (2) will likely fail when it'll try to call `buildkite_annotate` to annotate the CI build with the error message… since for the time being that lane is executed on devs' Macs not on CI.

Also, [now that our `create_release_backmerge_pull_request` action from `release-toolkit` has been improved to auto-delete past intermediate branch of past backmerge PRs before creating a more recent one](https://github.com/wordpress-mobile/release-toolkit/pull/601/files), thus making this error use case less likely if not impossible… I'm not sure the error handling and recovery suggestion are necessary to be added in our `Fastfiles` anymore 🤔 (cc @iangmaia @spencertransier thoughts on that last part?)

AliSoftware:
@geekygecko Thanks for the feedback & ideas!

> Where do I suggest other changes to the process. For example:

I think maybe the best approach would be to collect your notes while you go through the tasks during this sprint, and at the end of your release rotation post those notes in P2

You can post those either in a dedicated P2 post (like some other teams have done, posting in their team p2 and X-Post to appsinfrap2), or directly as a comment on the Project Thread dedicated to my work on improving PocketCasts Release Automation (ref: paaHJt-78B-p2)

> > Move [all open issues and PRs in the 7.77 milestone](https://github.com/Automattic/pocket-casts-android/milestones/7.77) to next version
> 
> This isn't a valid URL.

Ah good catch, I'll take note to fix this. Though I also have in my todo-list for later to actually automate that task via fastlane as well (we have a dedicated action for that all other repos already use, so shouldn't be too much work), so maybe I'll instead directly implement the automation for that at some point this week or the next which would allow to actually remove the manual task from the ReleaseV2 scenario entirely 😛

> After a release there is a step "Post a message in #pocket-casts-dev-android on Slack, notifying the team that you've released the release builds to 10% of users." the team also do this for a beta so it would be good to add a reminder for that. 

Taking note, thanks for the idea 👍

fastlane/Fastfile
AliSoftware: Update: removed the `rescue` block in 71668bd31

iangmaia: Ah that's a good point 👍 
I mean, there can still be errors in the process but I guess we're fine just letting them propagate -- this specific handling was indeed for the "backmerge already existing" case.

fastlane/Fastfile
```diff
@@ -539,6 +533,16 @@ platform :android do
     )
   end
 
+  lane :create_backmerge_pr do |version: release_version_current|
+    create_release_backmerge_pull_request(
+      repository: GH_REPOSITORY,
+      source_branch: "release/#{version}",
+      default_branch: DEFAULT_BRANCH,
```
iangmaia, line 540: Ah yes, this is required here as the default branch for PCAndroid is `main` 👍

fastlane/Fastfile
```diff
@@ -129,27 +129,16 @@ platform :android do
...
+
     # We cannot use the `copy_branch_protection` action here as it would create a branch protection rule with the same
-    # restrictions we have in `trunk`, and the release managers wouldn't be able to push due to permissions.
+    # restrictions we have in `main`, and the release managers wouldn't be able to push due to permissions.
     # This should be changed only when we have PC Android releases done on CI, when the CI bot is the one running `git push`.
     set_branch_protection(repository: GH_REPOSITORY, branch: "release/#{new_version}")
     set_milestone_frozen_marker(repository: GH_REPOSITORY, milestone: new_version)
   end
 
-  lane :complete_code_freeze do |options|
```
iangmaia, line 139: Nice 🧹

iangmaia (approved):
👍

fastlane/Fastfile
AliSoftware: Yep, didn't have it at first but noticed it was necessary during testing 😅 

Note that I found out the other day that the `Git` ruby gem has a [`repository_default_branch` method](https://github.com/ruby-git/ruby-git/blob/master/lib/git/base.rb#L28) that maybe we could start using in our actions of `release-toolkit` to avoid the need for those kind of parameters in the future 🤔

PR #3253, opened 2024-11-20
Referrals - Attach screenshot while sharing guest pass

ashiagr:
## Description
This attaches screenshot while sharing referral guest pass.

** I'm not sending an image from landscape phone layout as the card isn't displayed in that layout. I think it should be fine as statistically, Profile screen is opened mostly in portrait mode.

## Testing Instructions

1. Tap referral icon on Profile tab
2. Tap Share Guest Pass
3. Share using email
4. ✅ Notice that guest pass image is attached to the email p1732059551376689-slack-C07HDV91FBQ

## Checklist
 
- [ ] with different themes

### DIFF
--- a/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/sharing/SharingClientTest.kt
+++ b/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/sharing/SharingClientTest.kt
@@ -677,6 +677,7 @@ class SharingClientTest {
 
     @Test
     fun shareReferralLink() = runTest {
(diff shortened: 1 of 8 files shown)


MiSikora (commented):
I'm not sure we can do this. By setting content type to `image/*` a lot of apps will prioritize it and not pick up the text as the don't support it. For example sharing to Instagram, Tumblr, or Slack ignores the message.

ashiagr:
You're right. It seems it is not just the content type but setting `EXTRA_STREAM` causes it to ignore `EXTRA_TEXT` for sharing to certain apps.

I'll close this PR as the change isn't straightforward and it isn't a priority for now.

PR #3283, opened 2024-11-25
Add support for system navigation to the low storage dialog

### LINKED ISSUE #3232: Add support for system navigation to the low storage dialog
### Description

The following issue was found during the beta test. The bottom sheet that appears in the “Storage & data use” and the “Auto download” screens does not respect system navigation and does not react to it.

Internal reference pdeCcb-7BG-p2

1. To easily test, edit the file `DeviceStorageUtil.kt` and return true from the function `isDeviceRunningOnLowStorage`.
2. Go to Profile -> Settings -> Storage & data use
3. Swipe the side of the screen to go back (or tap the navigation back button)

Notice the page goes back but the dialog remains open. The dialog should close first.

https://github.com/user-attachments/assets/d881609c-ec8e-4736-8a5b-df1e5d229cfb

Google Pixel 9, Android 14, Pocket Casts 7.77-rc-1

mebarbosa:
## Description
- This PR adds the support of back navigation gesture for low storage modal.

Fixes #3232

## Testing Instructions
1. To easily test, edit the file `DeviceStorageUtil.kt` and return true from the function `isDeviceRunningOnLowStorage`
2. Have at least one downloaded episode

### Storage screen
1. Go to Profile -> Settings -> Storage & data use
2. See the Low storage modal opened
3. Swipe the side of the screen to go back (or tap the navigation back button)
4. The low storage modal should be closed ✅
5. Swipe the side of the screen again to go back
6. The Storage screen should be closed ✅
7. Repeat the test by pressing back button instead of swipe. The expected behavior should be the same ✅

### Auto Download
1. Go to Profile -> Settings -> Auto Download
2. Turn New Episodes toggle on
3. See the Low storage modal opened
4. Swipe the side of the screen to go back (or tap the navigation back button)
5. The low storage modal should be closed ✅
6. Swipe the side of the screen again to go back
7. The Auto Download screen should be closed ✅
8. Repeat the test by pressing back button instead of swipe. The expected behavior should be the same ✅

## Screenshots or Screencast 
[Uploading Screen_recording_20241125_151510.webm…]()

## Checklist
 
- [ ] with different themes


mebarbosa:
I will update the CHANGELOG later once the 7.79 is available

MiSikora (commented):
It works, but the solution feels a bit hacky. Could you explain why this sheet is managed using a custom interface and a variable in the activity? It seems like a more natural solution would be to use something like a `BottomSheetDialogFragment`, which could be managed through the back stack.

mebarbosa:
> It works, but the solution feels a bit hacky. Could you explain why this sheet is managed using a custom interface and a variable in the activity? It seems like a more natural solution would be to use something like a `BottomSheetDialogFragment`, which could be managed through the back stack.

@MiSikora I wanted to use the compose bottom sheet dialog similar what we have for end of the year, so that's why I did not use a `BottomSheetDialogFragment `, but the hard part was to make this compose dialog works with the back stack. I thought that it could have something wrong with the `BackHandler`, but it was working as expected, after some investigation and attempts to solve this issue I found the `HasBackstack` interface and saw that some fragments do a similar approach. But I can migrate this dialog to `BottomSheetDialogFragment`

MiSikora:
IMO the end of year doesn't use a correct approach. I didn't touch it while working on Playback 2024 because the solution was already in place for Playback 2023 and I wanted to focus the PB24 itself. Imagine we want to add a couple more of different bottom sheet dialogs—it wouldn't be manageable to track this state with a boolean for each new dialog. A more scalable approach is something like we use here:

https://github.com/Automattic/pocket-casts-android/blob/e8edc8819b06c651d73bdd3d216fd92ac8c24f6e/modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/PlaybackSettingsFragment.kt#L684-L700

But this doesn't use Compose. So if you want to keep your Compose UI I suggest writing a custom fragment that wraps your UI inside `BottomSheetDialogFragment`.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please add tests for class <code>LowStorageDialog</code> (or add <code>unit-tests-exemption</code> label to ignore this).</td>
    </tr>
  </tbody>
</table>

</p>

mebarbosa:
> IMO the end of year doesn't use a correct approach. I didn't touch it while working on Playback 2024 because the solution was already in place for Playback 2023 and I wanted to focus on the PB24 itself. Imagine we want to add a couple more of different bottom sheet dialogs—it wouldn't be manageable to track this state with a boolean for each new dialog. A more scalable approach is something like we use here:
> 
> https://github.com/Automattic/pocket-casts-android/blob/e8edc8819b06c651d73bdd3d216fd92ac8c24f6e/modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/PlaybackSettingsFragment.kt#L684-L700
> 
> But this doesn't use Compose. So if you want to keep your Compose UI I suggest writing a custom fragment that wraps your UI inside `BottomSheetDialogFragment`.

Hey @MiSikora I ended up using the `ConfirmationDialog ` approach after trying different other ones. The main issue I faced with the other solutions was the need to open another fragment from this dialog. I also tried using the Compose Navigation Controller, but it didn’t work as I expected.

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/LowStorageDialog.kt
```diff
@@ -0,0 +1,58 @@
...
+import au.com.shiftyjelly.pocketcasts.utils.Util
+import au.com.shiftyjelly.pocketcasts.utils.featureflag.Feature
+import au.com.shiftyjelly.pocketcasts.utils.featureflag.FeatureFlag
+import au.com.shiftyjelly.pocketcasts.utils.isDeviceRunningOnLowStorage
+import au.com.shiftyjelly.pocketcasts.views.dialog.ConfirmationDialog
+import au.com.shiftyjelly.pocketcasts.images.R as IR
+import au.com.shiftyjelly.pocketcasts.localization.R as LR
+import au.com.shiftyjelly.pocketcasts.ui.R as UR
+
+class LowStorageDialog(
```
MiSikora, line 17: Can you rename this class. It isn't a dialog. It creates and shows dialog.

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/LowStorageDialog.kt
```diff
@@ -0,0 +1,58 @@
...
+    val analyticsTracker: AnalyticsTracker,
+    val settings: Settings,
+) {
+
+    suspend fun shouldShow(downloadedFiles: Long): Boolean = isDeviceRunningOnLowStorage() &&
+        downloadedFiles != 0L &&
+        settings.shouldShowLowStorageModalAfterSnooze() &&
+        FeatureFlag.isEnabled(Feature.MANAGE_DOWNLOADED_EPISODES)
+
+    fun show(
```
MiSikora, line 28: This function name is a bit awkward. In order to show a dialog you need to do `dialog.show().show()`.

MiSikora (approved):

PR #3295, opened 2024-11-27
Migrate account details to Compose UI

MiSikora:
## Description

This migrates the account detail sections to Compose UI. There is one UI difference—the switch is styled according to the theme and is vertically centered in the item. Theming is applied because handling it properly was cumbersome. Vertical centering is done because the base `Switch` component includes significant padding, and refactoring would be required to align it to the top.

## Testing Instructions

Verify these areas:
* Tapping on sections should use correct actions—including signing out, and deleting an account
* Theming
* Automotive

## Screenshots or Screencast 

Screenshot 2024-11-27 at 17 10 51

## Checklist
 
- [x] with different themes


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

ashiagr:
Looks good!

Could you check this one thing: 
I noticed that after changing the email, the new email is not reflected on the Account details screen:

https://github.com/user-attachments/assets/f6a05fad-fe38-4bf3-bc36-061c84569641

It was updated in the production version.

MiSikora:
>  I noticed that after changing the email, the new email is not reflected on the Account details screen:

Fixed in https://github.com/Automattic/pocket-casts-android/pull/3295/commits/b230d7d7dd88e51ec6a146063a2916f5f393af46.

The issue was that email changes were not being observed, only the logged-in state. Previously, this worked because the collections in the profile and account detail pages were observed with the lifecycle. This caused the subscription to restart whenever a fragment was restarted. While I could have tried to follow the same approach here, but it would be harder in case of hot flows from view models. Also, it would not address the root cause and would rely somewhat on chance when observing the sign in state in the future.

There is, however, a potential risk that propagating email changes might cause unintended side effects. This is because `getSignInState()` is used in multiple places, and current observers might perform side-effect actions when it is triggered. That said, email changes are not a frequent action and are not triggered from many places, so I believe this approach is safe.

ashiagr (approved):
Works well now. 

> There is, however, a potential risk that propagating email changes might cause unintended side effects. This is because getSignInState() is used in multiple places, and current observers might perform side-effect actions when it is triggered. That said, email changes are not a frequent action and are not triggered from many places, so I believe this approach is safe.

I agree, it is not a frequent action so the solution looks safe to me too.

PR #3309, opened 2024-11-29
Create NumberStepper component in Compose UI and replace it for Sleep Timer

mebarbosa:
## Description
- This migrates the stepper from sleep timer to Compose UI and creates it as a component

## Testing Instructions
1. Play an episode
2. Open Sleep timer
3. Make sure you can update the `plus` and `minor` buttons ✅

## Screenshots or Screencast 
| Before | After |
|--------|--------|
| before | after | 

## Checklist
- [ ] ~I have updated (or requested that someone edit) [the spreadsheet]~(https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.
 
- [x] with different themes


modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/SleepFragment.kt
```diff
@@ -114,6 +112,51 @@ class SleepFragment : BaseDialogFragment() {
             }
         }
 
+        binding.customStepperComposeView.setContentWithViewCompositionStrategy {
+            AppTheme(theme.activeTheme) {
```
geekygecko, line 116:
```suggestion
            AppTheme(Theme.ThemeType.DARK) {
```

The player is always dark so if the active theme is used it might have the wrong colours such as for the press state.


https://github.com/user-attachments/assets/26f95043-0a2f-4c75-88a9-be0070a65983

It's not a big deal, but the pressed state is being clipped, and it might be able to be fixed by attaching to something different in the constraint layout.

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/SleepFragment.kt
```diff
@@ -114,6 +112,51 @@ class SleepFragment : BaseDialogFragment() {
...
+                NumberStepper(
+                    onMinusClick = {
+                        minusButtonClicked()
+                    },
+                    onPlusClick = {
+                        plusButtonClicked()
+                    },
+                    minusContentDescription = LR.string.player_sleep_custom_minus,
+                    plusContentDescription = LR.string.player_sleep_custom_plus,
```
MiSikora, line 125: I think all of these content descriptions could be more descriptive. Something like `Increase custom duration`, `Increase chapter count`, etc. Otherwise it's doesn't help to know what the buttons do when seeing/hearing `minus` and `plus`.

MiSikora (approved):
Overall looks good. I left one accessibility issue that I found apart form what Phil mentioned.

mebarbosa:
Thank you @geekygecko and @MiSikora for these suggestions! I already addressed all of them

PR #3324, opened 2024-12-03
Hack Week Project: Add options to sort Up Next

mebarbosa:
## Description
- This is part of the Hack Week project, which adds 3 options to sort Up next
- ⚠️ If this gets approved, we will need to review the design proposed here
- ⚠️ To merge this PR, we need to do some regression tests first
- See: pdjfGa-6h-p2

## Testing Instructions
- Add some episodes to up next and see the sort icon at the top of the screen
- Try all options to sort up next

## Screenshots or Screencast 

[Screen_recording_20241203_144705.webm](https://github.com/user-attachments/assets/54e7b2d4-7f3b-4e11-a908-95344f47577b)

| Up Next | Modal |
|--------|--------|
| Screenshot_20241203_144630 | Screenshot_20241203_144642
 |

### DIFF
--- a/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/UpNextFragment.kt
+++ b/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/UpNextFragment.kt
@@ -31,6 +31,7 @@ import au.com.shiftyjelly.pocketcasts.player.viewmodel.PlayerViewModel
 import au.com.shiftyjelly.pocketcasts.player.viewmodel.UpNextViewModel
 import au.com.shiftyjelly.pocketcasts.preferences.Settings
 import au.com.shiftyjelly.pocketcasts.repositories.playback.PlaybackManager
+import au.com.shiftyjelly.pocketcasts.repositories.playback.UpNextSortType
 import au.com.shiftyjelly.pocketcasts.repositories.playback.UpNextSource
 import au.com.shiftyjelly.pocketcasts.repositories.podcast.EpisodeManager
 import au.com.shiftyjelly.pocketcasts.ui.helper.FragmentHostListener
@@ -38,6 +39,7 @@ import au.com.shiftyjelly.pocketcasts.ui.helper.StatusBarColor
 import au.com.shiftyjelly.pocketcasts.ui.theme.Theme
 import au.com.shiftyjelly.pocketcasts.ui.theme.ThemeColor
 import au.com.shiftyjelly.pocketcasts.utils.extensions.hideShadow
+import au.com.shiftyjelly.pocketcasts.views.dialog.OptionsDialog
 import au.com.shiftyjelly.pocketcasts.views.extensions.tintIcons
 import au.com.shiftyjelly.pocketcasts.views.fragments.BaseFragment
 import au.com.shiftyjelly.pocketcasts.views.fragments.BaseFragmentToolbar.ChromeCastButton
@@ -294,7 +296,7 @@ class UpNextFragment : BaseFragment(), UpNextListener, UpNextTouchCallback.ItemT
                     true
                 }
                 R.id.sort_up_next -> {
-                    onSortQueue()
+                    onSortUpNext()
                     true
                 }
                 else -> false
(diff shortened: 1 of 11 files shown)


modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/UpNextDao.kt
```diff
@@ -33,6 +34,7 @@ abstract class UpNextDao {
         // update the existing episode positions
         val newPosition = updatePositionsBlocking(upNextEpisodes, position, replaceOneEpisode)
         upNextEpisode.position = newPosition
+        upNextEpisode.addedToQueueAt = System.currentTimeMillis()
```
mebarbosa, line 37: I added this new column so we can track when the episode was added to Up next. With this info we can sort back to the original sort

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Errors
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/UpNextFragment.kt
```diff
@@ -517,6 +524,36 @@ class UpNextFragment : BaseFragment(), UpNextListener, UpNextTouchCallback.ItemT
...
+            .addTextOption(
+                titleId = LR.string.up_next_sort_oldest_to_newest,
+                click = {
+                    viewLifecycleOwner.lifecycleScope.launch {
+                        playbackManager.upNextQueue.sortUpNextBy(UpNextSortType.OLDEST_TO_NEWEST)
+                    }
+                },
+            )
+        dialog.show(parentFragmentManager, "up_next_sort_options")
+    }
```
mebarbosa, line 556: We can also refine this dialog here. I copied the same one that we use in Podcast episodes view

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/UpNextDao.kt
geekygecko: I could be wrong but if you add an episode to the top or bottom of your Up Next using row swipe this won't restore the previous to queue order. I think it would be fine to remove the "Added to Up next" option. 

https://github.com/user-attachments/assets/d6f0a1ed-3626-4d33-8819-0ac589b6c7f3

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/UpNextFragment.kt
```diff
@@ -517,6 +524,36 @@ class UpNextFragment : BaseFragment(), UpNextListener, UpNextTouchCallback.ItemT
...
+
+    override fun onSortUpNext() {
+        val dialog = OptionsDialog()
+            .setTitle(getString(LR.string.sort_by))
+            .addTextOption(
+                titleId = LR.string.up_next_sort_added_to_up_next,
+                click = {
+                    viewLifecycleOwner.lifecycleScope.launch {
+                        playbackManager.upNextQueue.sortUpNextBy(UpNextSortType.ADDED_TO_UP_NEXT)
+                    }
```
geekygecko, line 536: This is only a suggestion, I wonder if it would be better to push this into the view model.

```suggestion
                    upNextViewModel.sort(UpNextSortType.ADDED_TO_UP_NEXT)
```

and have the following in the view model

```kotlin
    fun sort(addedToUpNext: UpNextSortType) {
        viewModelScope.launch {
            playbackManager.upNextQueue.sortUpNextBy(addedToUpNext)
        }
    }
```

CookieyedCodes: Ummmmm phil in my opinion "add to upnext" should still be an option so you could hop between lists that's super powerful imo, 
I would probably just call it "custom default" or something like that
but yes the swipe or changing it should overwrite the existing 'custom default" & go to "custom default" option (add to up next would probably switch it too, which if you consider below I wouldn't be against but sort A-Z/Z-A should stick to those)

I would also love an option to sort into grouped feeds/subscribed so if you have multiple episodes from a feed it will group then together based off the existing feed order 🤔🤔 (as in the first episode the other episodes clings to that & then if you move then one of the episodes to the bottom it would move back to "custom default"

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/UpNextDao.kt
msurdi-a8c: @mebarbosa @geekygecko  is this value something that we could somehow sync and have available in the `/up_next/list` endpoint?

Otherwise I think it will be impossible to mirror this feature in the Web Player.

mebarbosa:
Closing this PR until we get it reviewed by our designs and the task gets prioritized

PR #3351, opened 2024-12-11
Do not use local files for ParcelFileDescriptor

### LINKED ISSUE #3346: Error Playing Uploaded Files on Android 15
### Description

> It is downloaded and played locally. This happens with almost every file. Maybe one or two where it works. [...] Sometimes it stays at that stage and sometimes the whole system crashes and restarts.

9140979-zd-a8c

Another similar report: 9156237-zd-a8c

We were not able to reproduce this issue in our end

See the video provided here: p1733831046713619/1733830216.036979-slack-C02A333D8LQ

Android 15

MiSikora:
## Description

Our current implementation of providing artwork to media notifications crashes the OS on Android 15. The issue arises because we always try to convert image paths to URLs, which is not applicable for embedded artwork. This throws an `IllegalArgumentException` when attempting to convert a URI's path to a URL. This wasn't an issue before because the OS handled it gracefully, but Android 15 apparently changed the implementation and shuts down when an exception is thrown from a `ContentProvider` while getting a file descriptor.

This PR addresses the issue by adding more safeguards and checking whether the data can be processed, preventing exceptions from being thrown Additionally, I wrapped everything in a `runCatching` block to ensure that even if something fails, we do not crash but simply return `null`. If a crash occurs, I log the issue to our `LogBuffer`.

Closes #3346

Not updating the changelog because it will be upstreamed to `7.78` and `7.79`..

## Testing Instructions

> [!note]
> You need a real device with OS 15. It's even possible that not all devices crash like this. I tested mine with Pixel 6, Android 15.

### Fix

1. Have a file with an embedded cover. You can use [this file](https://drive.google.com/file/d/1iqHw9Q9AhPICzGlx68oyt1KRVBXe5koq/view?usp=sharing).
2. Add this file to Pocket Casts.
   1. If you're unable to do it form the Pocket Casts app go to a file manager app.
   2. Share the file.
   3. Choose Pocket Casts as the destination.
3. Play the file.
4. It should not crash the OS or the app.
5. You should see the file's cover in the media notification.

### Artwork smoke tests

Test if playing episodes and artwork still work with Automotive and the Media Notification.

## Checklist

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/auto/AlbumArtContentProvider.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/auto/AlbumArtContentProvider.kt
@@ -7,15 +7,20 @@ import android.database.Cursor
 import android.net.Uri
 import android.os.ParcelFileDescriptor
 import au.com.shiftyjelly.pocketcasts.servers.di.Downloads
+import au.com.shiftyjelly.pocketcasts.utils.log.LogBuffer
 import dagger.hilt.EntryPoint
 import dagger.hilt.InstallIn
 import dagger.hilt.android.EntryPointAccessors
 import dagger.hilt.components.SingletonComponent
 import java.io.File
-import java.io.IOException
+import okhttp3.HttpUrl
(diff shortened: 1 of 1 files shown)


mebarbosa (approved):
Since I am unable to reproduce this issue, I did a smoke test. Everything continued working as expected.

PR #3364, opened 2024-12-12
Fix: Sleep timer stops podcast  after given time

### LINKED ISSUE #2219: Sleep timer gets randomly cancelled
### Description

We have two users reporting that got sleep timer randomly cancelled

> Sleep timer rarely works. I set it to 25 minutes but it’s still playing after that time. I can’t rely on it.
8163546-zd-a8c

> i see it at least once per day maby more, it mostly seems random but & the thing is I have noticed it count down randomly throughout the episode but also few seconds before the timer ends (this happens more, the former happend randomly) and once the timer becomes bad the next autoplay timer also repeats the same problem so it's repeatable but also not trackable [...] Most likely in the background but I saw the visual countdown of the episodes happen in the foreground especially with it happening on an repeat timer (Clock timer is fine btw, tho less then 5 mins can be a tad inaccurate but I don't think much can be done with that) [...] Hmmmm I generally set the timer to 4 or 5 & it's good for the first 3+ times but then it gets caught & will randomly choose to count down in the last few seconds of the podcast I noticed (autoplay will repeat the behaviour hence why I noticed)-I think I noticed it happened more with episodes that have skip last but some is happening no problems -
It is at least twice a day+ tho what's interesting is tho is once the timer is bugged & cancled it will display and hold the set number but if I go into it and then cancel it to actually fix(because it's technically still bugged) the timer it actually goes back to 1 so maby that's playing into it
[Beta user](https://pocketcastsbeta.slack.com/archives/C013F9S9C2W/p1715336380969659?thread_ts=1714544495.220039&cid=C013F9S9C2W)

I can't reproduce this issue. 

To be registered what I tried:
- I tried with 25 minutes custom time with the app in background and device locked and it worked as expected ✅
- Fixed time, remove the app from the list of opened apps and lock the device. It also worked as expected  ✅
- Tried with skip last and skip first episodes set and it worked as expected  ✅

_No response_

Report was from `7.62.2`, but maybe we had this issue before

### LINKED ISSUE #3355: Sleep timer stops podcast **after** given time
### Description

From the user report:

> Hello,
the sleep timer is not working for me anymore correctly. 
The sleep timer turns the volume after the defined time down but doesn't Pause the Podcast. 
The Podcast only pauses after double the defined time. 
Example:
> - Sleep Timer is defined for 5 minutes
> - the volumes gets muted after 5 minutes but the episode does not Pause
> - the episode pauses after 10 minutes

I can confirm this behaviour happen with 1 minute interval.

1. Open a podcast
2. Set sleep timer to 1 minute
3. See that after 1 minute, podcast volume is 0, but it's still in a state of `play` (progress + pause button visible)
4. In my case, the `play` state moved to `paused` after 1m 47s

https://github.com/user-attachments/assets/19c8c6a4-92d8-426c-b35f-389214d8187f

Google Pixel 8a, Android 15, Pocket Casts 7.78

mebarbosa:
## Description
- We got some reports related to Sleep Timer not working as expected. Specially not respecting the time set
- The problem is that we were using the method `setAndAllowWhileIdle`, which might sets an inexact alarm. This method also might be affected by battery optimizations. See here for more details: https://developer.android.com/about/versions/14/changes/schedule-exact-alarms#use-cases
- But Android also provides another option to set exact alarm, which is `setExactAndAllowWhileIdle`. This one requires user's permission and the permission was introduced in Android 12 (`SCHEDULE_EXACT_ALARM`). See: https://developer.android.com/about/versions/14/changes/schedule-exact-alarms
- For devices with Android < 12 this permission request will not be necessary
- Also, users who already have granted the `ACTION_REQUEST_IGNORE_BATTERY_OPTIMIZATIONS ` permission our app won't need to request the `SCHEDULE_EXACT_ALARM` permission, which might cover a lot of our users since we request this for the battery restriction.

Fixes #3355
Fixes #2219

## Testing Instructions
- I have two devices but I was able to reproduce this issue https://github.com/Automattic/pocket-casts-android/issues/3355 only in one device. Something to note is that the device that I was able to reproduce this issue had low remaining battery.

### Android > 12
1. Have a fresh install
2. Play an episode
3. Do not grant the battery permission
4. Tap on sleep timer
5. You should see the new permission request for alarm
6. Grant PC permission
7. Tap on sleep timer again
8. Set 1 minute
9. Ensure you have a sleep timer more accurate ✅

### Android < 12
1. Have a fresh install
2. Play an episode
3. Do not grant the battery permission
4. Tap on sleep timer
5. Set 1 minute
6. Ensure you have a sleep timer more accurate ✅

### Android > 12 and migration
1. Instal the app from release/7.79
2. Play an episode
3. Grant the battery permission
4. Install the app from this branch
5. Play an episode
6. Tap on sleep timer
7. Set 1 minute
8. Ensure you have a sleep timer more accurate ✅

## Checklist

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SleepTimer.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SleepTimer.kt
@@ -4,7 +4,10 @@ import android.app.AlarmManager
 import android.app.PendingIntent
 import android.content.Context
 import android.content.Intent
+import android.os.Build
+import android.provider.Settings
 import android.text.format.DateUtils
+import androidx.annotation.RequiresApi
 import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsEvent.PLAYER_SLEEP_TIMER_RESTARTED
 import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsTracker
 import au.com.shiftyjelly.pocketcasts.utils.log.LogBuffer
@@ -38,9 +41,9 @@ class SleepTimer @Inject constructor(
     private var lastEpisodeUuidAutomaticEnded: String? = null
 
     fun sleepAfter(duration: Duration, onSuccess: () -> Unit) {
-        val sleepAt = System.currentTimeMillis().milliseconds + duration
+        val sleepAt = System.currentTimeMillis() + duration.inWholeMilliseconds
 
-        if (createAlarm(sleepAt.inWholeMilliseconds)) {
+        if (createAlarm(sleepAt)) {
             lastSleepAfterTime = duration
             cancelAutomaticSleepOnEpisodeEndRestart()
             cancelAutomaticSleepOnChapterEndRestart()
@@ -126,13 +129,28 @@ class SleepTimer @Inject constructor(
 
     private fun shouldRestartSleepEndOfChapter(diffTime: Duration, isSleepEndOfChapterRunning: Boolean) = diffTime < MIN_TIME_TO_RESTART_SLEEP_TIMER_IN_MINUTES && !isSleepEndOfChapterRunning && lastSleepAfterEndOfChapterTime != null
 
+    @RequiresApi(Build.VERSION_CODES.S)
+    private fun requestExactAlarmPermissionIfNeeded(context: Context) {
+        val intent = Intent(Settings.ACTION_REQUEST_SCHEDULE_EXACT_ALARM)
+        intent.flags = Intent.FLAG_ACTIVITY_NEW_TASK
+        context.startActivity(intent)
+    }
+
     private fun createAlarm(timeMs: Long): Boolean {
         val sleepIntent = getSleepIntent()
         val alarmManager = getAlarmManager()
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S && !alarmManager.canScheduleExactAlarms()) {
+            LogBuffer.e(TAG, "Cannot schedule exact alarms. Permission not granted.")
+            requestExactAlarmPermissionIfNeeded(context)
+            return false
+        }
+
         alarmManager.cancel(sleepIntent)
+
         return try {
             LogBuffer.i(TAG, "Starting...")
-            alarmManager.setAndAllowWhileIdle(AlarmManager.RTC_WAKEUP, timeMs, sleepIntent)
+            alarmManager.setExactAndAllowWhileIdle(AlarmManager.RTC_WAKEUP, timeMs, sleepIntent)
             sleepTimeMs = timeMs
             lastTimeSleepTimeHasFinished = timeMs.milliseconds
             true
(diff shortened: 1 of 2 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-android/milestone/70">7.79 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SleepTimer.kt
```diff
@@ -126,13 +129,28 @@ class SleepTimer @Inject constructor(
...
+            requestExactAlarmPermissionIfNeeded(context)
+            return false
+        }
+
         alarmManager.cancel(sleepIntent)
+
         return try {
             LogBuffer.i(TAG, "Starting...")
-            alarmManager.setAndAllowWhileIdle(AlarmManager.RTC_WAKEUP, timeMs, sleepIntent)
+            alarmManager.setExactAndAllowWhileIdle(AlarmManager.RTC_WAKEUP, timeMs, sleepIntent)
```
geekygecko, line 153: I have been wondering if there is a way to do this without using the alarm permission. When playback happens, we have the playback service running in the foreground. This service saves the playback progress to the database. It could also check if the sleep timer should have been triggered and pause the podcast. I could be wrong about this. Is there a reason it could work like this?

MiSikora (commented):
I'm experiencing the same issue in both `release/7.79` and this branch, but it doesn't seem related to the changes. As a result, I can't verify whether the fix works. Here's the issue I encountered:

1. Play something.  
2. Set a sleep timer.  
3. Wait for the sleep timer to finish.  
4. You'll hear the audio fading out.  
5. The audio fades to 0, but playback continues and progresses.  
6. After about 30 seconds, playback stops, with a notification stating it was paused by the sleep timer.  

Sometimes, I experienced a variation where the audio didn't fade to 0.

---

Regarding the PR itself, I'm not confident about using `SCHEDULE_EXACT_ALARM` and ignoring the lint warning. This permission is intended for system apps, which we are not. Ashita did a brief analysis here: [p1730446379362099-slack-C07J5LNP4SF](#).

---

From my current experience with the sleep timer, I feel it needs a rewrite. It's a small enough component that we could manage this, and I really like @geekygecko's idea of running it as part of the service. We could implement a simple timer scheduler that resides within the service.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SleepTimer.kt
mebarbosa: hummm... I don't know why we implemented this using system alarm, but I think we can try this approach. I am going to do some test so this how this could work

mebarbosa:
I am closing this PR to open another one with the approach we discussed

PR #3414, opened 2025-01-08
Edge-to-edge dialog support

geekygecko:
## Description

This pull request aims to improve the dialog styling. Switching to Android 15 caused issues with the padding and colours, especially regarding the navigation bar. 

The latest version might look worse on the older SDK 30 and 24 but it look a lot better on the higher SDKs when using the dialog. The content that slides behind the navigation bar handle isn't a solid navigation bar colour. For example before the change:

https://github.com/user-attachments/assets/1349c93b-7ce2-469d-8e2b-fd850f098f4f

After:

https://github.com/user-attachments/assets/acc11114-2468-404e-9d66-0a85b61d13ba

This change took a while, so it doesn't take even longer I haven't focused on the landscape version. I will do that in a future PR if we have time to spend on it. 

## Testing Instructions

Try different bottom sheet dialogs in the app. And verify the content is displayed correctly with the navigation bar.

- Podcast grid options: go to the Podcast tab, three dots in the toolbar.
- Filter sort by: go to the Filters tab, select a filter, three dots in the toolbar, Sort By.
- Clear Up Next: add some episodes to your Up Next, go to the Up Next tab, clear Up Next button in the toolbar.
- File options: go to the Profile tab, tap Files, tap a user file.
- Champion: login as a lifetime user, go to the Profile tab, tap account, tap Pocket Casts Champion.
- Full screen player: open the full screen player, tap effects, tap sleep timer, tap the shelf overflow button.

### Filter sort by dialog

After 

Screenshot 2025-01-10 at 2 31 20 pm
Screenshot 2025-01-10 at 2 32 10 pm

Before (main)

Screenshot 2025-01-10 at 3 08 03 pm

### Podcast grid options dialog

After

Screenshot 2025-01-10 at 2 33 43 pm
Screenshot 2025-01-09 at 9 10 49 am

Before (main)

Screenshot 2025-01-10 at 3 08 16 pm

### Clear Up Next dialog

After

Screenshot 2025-01-10 at 2 56 21 pm
Screenshot 2025-01-10 at 2 55 35 pm

Before (main)

main-clear-up-next

### User file dialog

After
Screenshot 2025-01-10 at 2 57 57 pm
Screenshot 2025-01-10 at 2 58 59 pm

Before (main)
Screenshot 2025-01-10 at 3 08 54 pm

### Pocket Casts champion

After

Screenshot 2025-01-09 at 9 07 00 am
Screenshot 2025-01-09 at 9 08 25 am

### Player - Long press next episode

Screenshot 2025-01-10 at 2 59 59 pm

Before (main)

Screenshot 2025-01-10 at 3 07 30 pm

### Player - Playback effects

The extra bottom padding will go once we remove the old custom playback speed settings.

Screenshot 2025-01-10 at 3 01 00 pm

Before (main)

Screenshot 2025-01-10 at 3 07 05 pm

### Player - Sleep timers

Screenshot 2025-01-10 at 3 01 37 pm

Before (main)

Screenshot 2025-01-10 at 3 07 16 pm

## Checklist
 
- [x] with different themes


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

modules/services/ui/src/main/res/values/themes.xml
```diff
@@ -197,6 +197,7 @@
         <item name="isDark">false</item>
         <item name="isImageTintEnabled">false</item>
         <item name="imageTintThemeCacheTag">light</item>
+        <item name="enableEdgeToEdge">true</item>
```
geekygecko, line 200: The `BottomSheetDialog` material component uses this to know it should draw to the edge which fixes the navigation bar being a different colour to the background.

MiSikora (approved):
I didn't find any major issues with dialogs. The only thing I found is that the Rearrange Actions bottom sheet is clipped at the bottom unlike other dialogs.

clipped

I also noticed that the sharing screen now experiences a bit of a parallax effect when scrolling content down. I don't think it is an issue and I even like the effect but I wanted to mention it here.

There are also some other issues that I found:

Profile tab and account details top content can be obscured by the  status bar. I think we should follow other screens and exclude top bar from the scrolling content. That's also what iOS does.

| Profile | Account |
| - | - |
| profile | acc_details |

Another issue that I found is with importing files. It seems it doesn't account for the edge-to-edge mode.

add_file

geekygecko:
Thanks @MiSikora for the feedback. I will address these in some separate PRs.

PR #3446, opened 2025-01-16
Add Winback view model

MiSikora:
## Description

This PR adds Winback VM with logic for fetching available plans and matching user's subscription with them. It makes a lot of assumptions about the subscriptions:
* Only a single acknowledged purchase is allowed.
* Purchase must have an order ID.
* Purchase cannot have multiple products.

This matches iOS behavior and is done this way because there's no UI to handle different cases. Also, in practice, it should be the only viable option as Google Play prevents users from having multiple subscriptions and we don't have single-purchase products.

## Testing Instructions

There's nothing to test here manually.

## Checklist

### DIFF
--- a/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/winback/WinbackViewModel.kt
+++ b/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/winback/WinbackViewModel.kt
@@ -6,9 +6,11 @@ import au.com.shiftyjelly.pocketcasts.models.to.SignInState
 import au.com.shiftyjelly.pocketcasts.models.to.SubscriptionStatus
 import au.com.shiftyjelly.pocketcasts.models.type.Subscription
 import au.com.shiftyjelly.pocketcasts.models.type.SubscriptionMapper
+import au.com.shiftyjelly.pocketcasts.models.type.SubscriptionPricingPhase
 import au.com.shiftyjelly.pocketcasts.repositories.subscription.ProductDetailsState
 import au.com.shiftyjelly.pocketcasts.repositories.subscription.SubscriptionManager
 import au.com.shiftyjelly.pocketcasts.repositories.user.UserManager
+import com.android.billingclient.api.ProductDetails
 import dagger.hilt.android.lifecycle.HiltViewModel
 import javax.inject.Inject
 import kotlinx.coroutines.flow.SharingStarted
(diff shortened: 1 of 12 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/winback/WinbackViewModel.kt
```diff
@@ -0,0 +1,158 @@
...
+
+    private val _uiState = MutableStateFlow(UiState.Empty)
+    internal val uiState = _uiState.asStateFlow()
+
+    init {
+        loadInitialPlans()
+    }
+
+    private fun loadInitialPlans() {
+        _uiState.value = _uiState.value.copy(subscriptionPlansState = SubscriptionPlansState.Loading)
```
geekygecko, line 34: I could be wrong, but it seems like this line isn't needed, as the default is loading.

modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/winback/WinbackViewModel.kt
```diff
@@ -0,0 +1,158 @@
...
+        viewModelScope.launch {
+            val plansDeferred = async { loadPlans() }
+            val activePurchaseDeferred = async { loadActivePurchase() }
+
+            val plansResult = plansDeferred.await()
+            val activePurchaseResult = activePurchaseDeferred.await()
```
geekygecko, line 40: I like the way these are loading at the same time. Very nice!

modules/features/profile/src/test/kotlin/au/com/shiftyjelly/pocketcasts/profile/winback/WinbackViewModelTest.kt
```diff
@@ -0,0 +1,277 @@
...
+    on { this.isAcknowledged } doReturn isAcknowledged
+    on { this.orderId } doReturn orderId
+    on { this.products } doReturn productIds
+}
+
+private val BillingPeriod.value
+    get() = when (this) {
+        BillingPeriod.Monthly -> "P1M"
+        BillingPeriod.Yearly -> "P1Y"
+    }
```
geekygecko, line 277: Thanks for adding these great tests.

geekygecko (approved):
This looks great. 

I could be misunderstanding but I'm not sure this is true "Google Play prevents users from having multiple subscriptions". We have four subscription plans,  plus monthly, plus yearly, patron monthly, and patron yearly. As we have done it this way, it is up to us to make sure they only have one of these.

modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/winback/WinbackViewModel.kt
MiSikora: This would assume that the current state is always `Loading` which might not be the case for reloading when added.

MiSikora:
> I could be misunderstanding but I'm not sure this is true "Google Play prevents users from having multiple subscriptions".

You're right. I was going back and forth yesterday about it. I should've been more careful with what I'm saying and tested a bit more. I'll ask Dom what he suggest us to do as we can have multiple active subscriptions tied to a single Google Play account.

modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/winback/WinbackViewModel.kt
geekygecko: Oh I see, thanks.

MiSikora:
I asked what to do about it here: p1737090829465379-slack-C080ZKX4UCV

Regardless, I'll merge it since the feature is behind a feature flag and I want to proceed with the work.

PR #3451, opened 2025-01-17
[Gradle] Enable Build Scans on CI

ParaskP7:
## Description

This PR enables [build scans](https://gradle.com/develocity/product/build-scan/) on CI for this project. 🏗️

---

## Testing Instructions

- For `local` builds:
  - Run any `gradlew` task (ie. `assembleDebug`) and notice that, at the very end, there is no build scan publishing happening.
- For `ci` builds:
  - Per specific build job (below), notice that build scan is being published and a `gradle.com/s/xyz` link is available for that build job/task.
    - [Lint](https://buildkite.com/automattic/pocket-casts-android/builds/10139#01947449-a644-4ea7-a34e-38d82b2beebd/217-441) -> [Build Scan](https://scans.gradle.com/s/qnjjhihheuyoq)
    - [Unit tests](https://buildkite.com/automattic/pocket-casts-android/builds/10139#01947449-a647-4889-8b8a-9f042a1eb853/180-362) -> [Build Scan](https://scans.gradle.com/s/igtv3tdjknii2)
    - [Instrumented tests](https://buildkite.com/automattic/pocket-casts-android/builds/10139#01947449-a64a-4a55-a8d2-46bac2b624d2/223-3825) -> [Build Scan](https://scans.gradle.com/s/d4dhnat5w5taq)
    - [Assemble release APK](https://buildkite.com/automattic/pocket-casts-android/builds/10139#01947449-a64b-4798-9d2e-951ff2970acd/223-1012) -> [Build Scan](https://scans.gradle.com/s/gmljx2crb3l3a)
    - [Prototype Builds](https://buildkite.com/automattic/pocket-casts-android/builds/10139#01947449-a64c-45af-8f75-ec0ee0fe351f/218-1918) -> [Build Scan](https://scans.gradle.com/s/giljebeeaglei)

PS: Just because of this 7815d0591d8afe66c0798ab4bf423532d0801aa4 commit, which moves the `build cache` into the `gradle/config` folder, you would also want to verify that everything is still working as expected:

[Prototype Builds](https://buildkite.com/automattic/pocket-casts-android/builds/10139#01947449-a64c-45af-8f75-ec0ee0fe351f/218-1918) -> [Build Scan](https://scans.gradle.com/s/giljebeeaglei) -> [Performance [Build Cache]](https://scans.gradle.com/s/giljebeeaglei/performance/build-cache) -> Check `Remote cache -> Configuration -> URL` & Any Other Configurations

---

#### I have tested any UI changes... `N/A`


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

wzieba (approved):

settings.gradle.kts
```diff
@@ -14,9 +14,10 @@ pluginManagement {
 }
 
 plugins {
-    id("com.gradle.enterprise").version("3.16.2")
+    id("com.gradle.develocity").version("3.19")
 }
-apply(from = File("gradle_build_cache.gradle.kts"))
+apply(from = File("./config/gradle/gradle_build_cache.gradle.kts"))
+apply(from = File("./config/gradle/gradle_build_scan.gradle"))
```
geekygecko, line 20: Is it possible to get this as a `.gradle.kts` as I believe the rest of our build scripts are in Kotlin?

ParaskP7: 👋 @geekygecko and thanks so much for raising this question! 🌟 

1. Short story long, this is actually what I tried doing at first, only to realize/remember that this won't work, not at least in a straightforward way.
2. The problem is that when you use `apply(from = ...)` in `settings.gradle.kts`, or elsewhere for that matter, the applied script (`./config/gradle/gradle_build_scan.gradle.kts`) doesn't have access to the same context as the main script. This means that certain extensions and configurations available in the main script, such as the `develocity` block, are not accessible in the applied Kotlin script. Gradle Kotlin DSL scripts are compiled before execution and thus they run in a more restricted environment for better type safety and IDE support. Gradle generates accessor code for DSL elements, which can limit immediate access to some contexts and thus the compilation step introduces an additional layer between the script and the Gradle runtime.
3. Now, when using the Groovy syntax (`./config/gradle/gradle_build_scan.gradle`), the Groovy DSL script is executed directly in the Gradle runtime environment, which mean they have immediate access to the full Gradle API and project context. Thus, dynamic features on Groovy allow for more flexible runtime behavior, which is actually both an asset and a liability, but extensions and configuration are readily available without additional setup.
4. To solve this, one would need to use [convention plugins](https://docs.gradle.org/current/samples/sample_convention_plugins.html), which means, to create a custom plugin in the `buildSrc` or `build-logic` directory/module to encapsulate the build scan configuration logic. However, I didn't want to go that far, not a least now. Still, we do have it in our roadmap to migrate to convention plugins at some point (see [AIR#67](https://github.com/Automattic/apps-infra-plans/issues/67)).

Let me know if the above clarifies my decision to keep this script as a Groovy script, at least for now. 🙏 

PS: Note the fact that [gradle_build_cache.gradle.kts](https://github.com/Automattic/pocket-casts-android/blob/main/config/gradle/gradle_build_cache.gradle.kts) is indeed written in Gradle Kotlin DSL, and works fine. But this is solely because it doesn't require the usage of `develocity` block/api, it just using `buildCache` directly, which seems to be directly available. Btw, I noticed another such Groovy script in this project, the [install.gradle](https://github.com/Automattic/pocket-casts-android/blob/main/scripts/git-hooks/install.gradle), which is being using via the root `build.gradle.kts` script, same problem, some blocks/apis are not accessible.

geekygecko: Thanks for the great explanation. I should have known there was much more to it.

PR #3488, opened 2025-01-24
[Tooling] Allow `update_rollouts` to specify the track to update

AliSoftware:
## Why / Context

During his last run of the release scenario, @geekygecko noticed that, by the time the release manager is supposed to run `update_rollouts percent:1` to roll out the production build to 100% after it has been approved by Google, that command updates the rollout of the wrong build / track. [Internal ref: p1737528474331869-slack-CC7L49W13]

The reason this happens is because, so far, `update_rollouts` was always using the version stored in the `version.properties` file of the current branch to figure out which version name and code to update the rollout of.

But, by the time one wants to update the rollout of version `x.y` in production track to 100%, the `release/x.y` branch for that version has already been deleted (by the release manager running the `publish_release` lane earlier in the process, namely as soon as Google as approved the production release over the weekend and we started the rollout to production).

This means the release manager would not be able to run that `update_rollouts` lane from the `release/x.y` branch to update the rolllout of `x.y` in Production to 100%. And running that same lane from the `main` branch instead would end up updating the rollout of the latest beta that has been done after code-freeze of `x.y+1`, instead of updating the rollout of the production build for `x.y`

## How / The Fix

This PR updates the logic of `update_rollouts` so that one can now pass an explicit `track:production` to force the code to fetch the latest `versionCode` of the production track using the GooglePlay API, rather than relying on the value extracted from the `version.properties` of the current branch.

This should allow to update the rollout of the latest build in production track even if the `release/*` branch of that build version has been deleted since and the `version.properties` as already been bumped to a next beta version.

## Testing Instructions

 - Manually modify the `version.properties` files locally to set fake/invalid version name and code in it (save, but no need to commit)
 - Run `bundle exec fastlane update_rollouts percent:0.21 track:beta`
 - Observe from the logs that it fetches the `versionCode` of the beta track from the API (as opposed to relying on the value from `version.properties`) during the `Step: google_play_track_version_codes`
 - Validate that the latest Beta build in GooglePlay Console had its current rollout updated from 20% to 21%. (You can re-run the same lane with `percent:0.20` right after that test if you want to revert the rollout back to 20%)

Note: At the time this PR is created, the build currently in `production` track is already at 100%, so it won't be possible to run a similar test with `track:production`, since it's already at max rollout %.

## Internal References

171055-ghe-Automattic/wpcom
p1737528474331869-slack-CC7L49W13

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -303,7 +303,7 @@ platform :android do
   # @param track [String] The Google Play track for which to update the rollout of. Must be either `beta` or `production`.
   #
   lane :update_rollouts do |percent:, track:|
-    UI.user_error!('percent parameter must be between 0.0 and 1.0') if percent.to_f < 0 || percent.to_f > 1
+    UI.user_error!('percent parameter must be between 0.0 and 1.0') if percent.to_f.negative? || percent.to_f > 1
     UI.user_error!('track parameter must be either `beta` or `production`') unless %w[beta production].include?(track)
 
     is_beta = track == 'beta'
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -297,31 +297,23 @@ platform :android do
     end
   end
 
-  # Update the rollout of all 3 variants (app, automotive, wear) of the current version to the given value
+  # Update the rollout of all 3 variants/form-factors (app, automotive, wear) of the latest builds of the given Google Play track to the given % value
   #
   # @param percent [Float] The rollout percentage, between 0 and 1
-  # @param track [String]
-  #        If `beta`, will update the rollout of the latest build in all 3 Beta tracks.
-  #        If `production`, will update the rollout of the latest build in all 3 Production tracks.
-  #        If `nil`, will derive the track based on if the `version_name_current` contains `-rc-` or not.
+  # @param track [String] The Google Play track for which to update the rollout of. Must be either `beta` or `production`.
   #
-  lane :update_rollouts do |percent:, track: nil|
-    if track.nil?
-      # Derive values from current version in `version.properties`
-      is_beta = beta_version?(version_name_current)
-      base_version_code = build_code_current
-    else
-      # Use Google Play API to find the latest versionCode for the requested track
-      UI.user_error!('track parameter must be either `beta` or `production`') unless %w[beta production].include?(track)
+  lane :update_rollouts do |percent:, track:|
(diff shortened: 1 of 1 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

AliSoftware:
@MiSikora as the release manager for `7.81`, once this PR lands in `release/7.81` branch, you might want to remember to add this new `production_track:true` parameter to the calls to `update_rollouts` that would be done after the task that has called `publish_release` (which is the one that will be deleting the `release/7.81` branch), namely:

 - Call `bundle exec fastlane percent:0.5 production_track:true` on [this task of the "Release" milestone](https://mc.a8c.com/releases-v2/release.php?product=PCAndroid&version=7.81#task-78c71601-7134-48c5-921d-2348af8f3002-fragment)
 - Call `bundle exec fastlane percent:1 production_track:true` on [this task of the "Rollout to all" milestone](https://mc.a8c.com/releases-v2/release.php?product=PCAndroid&version=7.81#task-327852d7-043b-4586-9061-04bf7ae48fac-fragment)

_Since the scenario for `7.81` has already started I can't update the current scenario steps of it on the fly, so you'll have to remember about it for `7.81`; but I'll make sure to  update the scenario template for all the next versions after that one to include this addition_

fastlane/Fastfile
```diff
@@ -300,32 +300,49 @@ platform :android do
   # Update the rollout of all 3 variants (app, automotive, wear) of the current version to the given value
   #
   # @param percent [Float] The rollout percentage, between 0 and 1
+  # @param track [String]
+  #        If `beta`, will update the rollout of the latest build in all 3 Beta tracks.
+  #        If `production`, will update the rollout of the latest build in all 3 Production tracks.
+  #        If `nil`, will derive the track based on if the `version_name_current` contains `-rc-` or not.
```
MiSikora, line 306: I wonder if it wouldn't be safer to require `track` parameter. We don't update rollout often so it's not a big hindrance to type couple of letters more and it would prevent us from making a mistake either during manual deployment or when creating scripts. WDYT?

MiSikora (approved):
I tested it and it works as expected. I left one comment on handling no `track` parameter but feel free to ignore it or address at any later point.

PR #3494, opened 2025-01-27
Bookmarks Tweaks

### LINKED ISSUE #2147: UI tweaks to the bookmarks page
### Description

It would be good to make the following tweaks to the bookmarks page.

1. Round the corners of the artwork
2. Add drop shadow so white artwork doesn't blend into the background
3. Reduce the line height for the bookmark title
4. Consider removing the bookmark count and moving the overflow dots into the toolbar

1. Go to the profile tab
2. Tap on the Bookmarks section

| Design | Android |
| --- | --- |
|   |  | 

_No response_

mebarbosa:
## Description
- This fixes some smalls design issues for Bookmarks screen
   - Round the corners of the artwork
   - Reduce the line height for the bookmark title
   - Add Placeholder for Bookmark icon
- Figma: Io7PCz1H1hmOgEE9eljoYW-fi-2036_44518
- I will create another GH issue to address the more actions button to toolbar since I faced with some issues to move it here

Fixes #2147

## Testing Instructions
1. Create some bookmarks
2. Go to Profile -> Bookmarks
3. Check the updated screen

## Screenshots or Screencast 

| Before | After |
|--------|--------|
| before | after | 

## Checklist
 
- [x] with different themes

### DIFF
--- a/modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/components/TextStyles.kt
+++ b/modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/components/TextStyles.kt
@@ -143,12 +143,13 @@ fun TextH40(
     fontWeight: FontWeight = FontWeight.W500,
     disableAutoScale: Boolean = false,
     fontScale: Float = 1f,
+    lineHeight: TextUnit = 21.sp.scaled(disableAutoScale, fontScale),
 ) {
     Text(
         text = text,
         color = color,
         fontSize = 15.sp.scaled(disableAutoScale, fontScale),
-        lineHeight = 21.sp.scaled(disableAutoScale, fontScale),
+        lineHeight = lineHeight,
         fontWeight = fontWeight,
         textAlign = textAlign,
         maxLines = maxLines,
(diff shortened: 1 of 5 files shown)


aurarora:
One small feedback that I have for the Bookmarks page is being able to click on the body of each bookmark to open the Details page for its episode. This way I could immediately see all bookmarks for that episode.
To me it seems kind of counterintuitive for there to be no action at all when clicking a bookmark.

CookieyedCodes:
@aurarora 
I agree actually that would be nice
Personally wouldn't mind the bookmarks page to behave like a person podcasts outliner either in addition to this tho 😅

mebarbosa:
> One small feedback that I have for the Bookmarks page is being able to click on the body of each bookmark to open the Details page for its episode

@aurarora humm... This is a very good point. I will share this with our designers

MiSikora (approved):

modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/components/TextStyles.kt
```diff
@@ -143,12 +143,13 @@ fun TextH40(
...
     disableAutoScale: Boolean = false,
     fontScale: Float = 1f,
+    lineHeight: TextUnit = 21.sp.scaled(disableAutoScale, fontScale),
 ) {
     Text(
         text = text,
         color = color,
         fontSize = 15.sp.scaled(disableAutoScale, fontScale),
-        lineHeight = 21.sp.scaled(disableAutoScale, fontScale),
+        lineHeight = lineHeight,
```
MiSikora, line 152: This change breaks font scaling. Instead it should be `llineHeight: TextUnit = 21.sp` and `lineHeight = lineHeight.scaled(disableAutoScale, fontScale)` so that disabling and font scale is applied correctly.

PR #3514, opened 2025-01-29
[Tooling] Introduce pipelines to trigger release tasks on CI

AliSoftware:
## Description

As part of this Project Thread: paaHJt-78B-p2, this introduces `.buildkite/release-pipelines/*.yml` files that will take care of calling `fastlane code_freeze` and similar lanes called during the release process.

Those lanes are currently called by the Release Manager on their own Mac, so the point of introducing those pipelines is to be able to instead switch to make those kicker lanes be run on CI instead.

### Created pipelines

List of `.buildkite/release-pipelines/*.yml` files calling their associated `fastlane` lane during ReleaseV2 scenario:

 - [x] `code_freeze`
 - [x] `update_rollouts`
 - [x] `new_beta_release`
 - [x] `finalize_release`
 - [x] `publish_release`
 - [x] `new_hotfix_release`
 - [x] `finalize_hotfix_release`

## Scenario steps update in ReleasesV2

Once this lands, we will be able to also land the companion PR in wpcom to update the Release Scenario in MC (internal ref: 171698-ghe-Automattic/wpcom) to replace the currently-**_manual_** tasks like "Call `bundle exec fastlane code_freeze`" and the likes with automated tasks instead—i.e. tasks that will present a clickable button on ReleasesV2 which, when clicked, will trigger those lanes on CI—thus not requiring Release Managers to run those from their personal Macs anymore.

## Testing Instructions

 - [Dummy build testing that cutting a new `oha/test-release-on-ci` branch and pushing it to the remote will work thanks to `use-bot-for-git`](https://buildkite.com/automattic/pocket-casts-android/builds/10311#0194b3d6-c70c-4a98-a2a8-34a41e59e781)
 - [Dummy build validating that the steps pipeline-upload'ed from `release-build.yml` would properly ensure to `git checkout release/$RELEASE_VERSION` before starting the build, to account for the fact that when the steps of this pipeline are added to the current build that was started from `main` (as opposed to being triggered as a separate CI build directly from the `release/*` branch afterwards like was the case so far), the `release-build.yml` will ensure to switch to the right release branch first](https://buildkite.com/automattic/pocket-casts-android/builds/10313#0194b3e7-ab04-4602-919d-516a3fb9e616)

Otherwise I think the only way to test this is during the next release process that will adopt those changes (likely during `7.82`)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

fastlane/Fastfile
```diff
@@ -362,9 +361,7 @@ platform :android do
     version = release_version_current
 
     # Wrap up
-    remove_branch_protection(repository: GITHUB_REPO, branch: "release/#{version}")
```
AliSoftware, line 365: This is already done later during `publish_release`, which is a better moment to remove the release branch protection anyway (see https://github.com/Automattic/apps-infra-plans/issues/83)

.buildkite/release-pipelines/code-freeze.yml
```diff
@@ -0,0 +1,26 @@
...
+  - label: "🚂 Code Freeze"
+    command: |
+      echo '--- 🤖 Use bot for Git operations'
+      source use-bot-for-git
+
+      echo '--- :ruby: Setup Ruby Tools'
+      install_gems
+
+      echo '--- 🔐 Access Secrets'
+      bundle exec fastlane run configure_apply
```
mokagio, line 16: I don't think this is something we've ever done in the release pipelines, but how do you feel about extracting these common steps in a script to `source`?

Could be a followup / mini-project to log and pick up once someone has time to implement consistently. Just thought I'd toss the idea there for DRYing.

fastlane/Fastfile
```diff
@@ -413,12 +410,22 @@ platform :android do
...
+    if is_ci?
+      buildkite_pipeline_upload(**pipeline_args)
+    else
+      buildkite_trigger_build(
+        buildkite_organization: 'automattic',
+        buildkite_pipeline: 'pocket-casts-android',
+        branch: branch_to_build,
+        **pipeline_args
+      )
+    end
```
mokagio, line 428: Nice DRYing

mokagio (approved):
Looking good as far as I can see.

A similar feeling to what you wrote in regards to testing this:

> I think the only way to test this is during the next release process that will adopt those changes (likely during 7.82)

However, I'll say that the more of these we implement, the more confident I become because I can see the code being ported from one already real-world-tested implementation to a new one.

mokagio:
Side note: Glad you used consistent naming with `-` in the new pipeline files even though they don't match the lanes they call 1:1.

I feel like the confusing of seeing some files with dashes and some with underscores would be bigger than the feeling of inconsistency with the file being called `lane-name.yml` and the lane being `lane_name`.

.buildkite/release-pipelines/code-freeze.yml
AliSoftware: Yeah, as I copy/pasted the `release-pipelines/code-freeze.yml` multiple times to create the other `release-pipelines/` and seeing how similar they all were I thought about it too; a good idea to implement cross-repo in a subsequent iteration indeed!

.buildkite/release-pipelines/update-rollouts.yml
```diff
@@ -0,0 +1,28 @@
...
+      source use-bot-for-git
+
+      echo '--- :ruby: Setup Ruby Tools'
+      install_gems
+
+      echo '--- 🔐 Access Secrets'
+      bundle exec fastlane run configure_apply
+
+      echo '--- 🚀 Update Rollouts'
+      bundle exec fastlane update_rollouts track:$TRACK percent:$ROLLOUT_PERCENT
```
AliSoftware, line 21:
```suggestion
      bundle exec fastlane update_rollouts track:"$TRACK" percent:"$ROLLOUT_PERCENT"
```

.buildkite/release-pipelines/new-hotfix-release.yml
```diff
@@ -0,0 +1,28 @@
...
+      source use-bot-for-git
+
+      echo '--- :ruby: Setup Ruby Tools'
+      install_gems
+
+      echo '--- 🔐 Access Secrets'
+      bundle exec fastlane run configure_apply
+
+      echo '--- 🚀 New Hotfix Release'
+      bundle exec fastlane new_hotfix_release version_name:$RELEASE_VERSION version_code:$VERSION_CODE skip_confirm:true
```
AliSoftware, line 21:
```suggestion
      bundle exec fastlane new_hotfix_release version_name:"$RELEASE_VERSION" version_code:"$VERSION_CODE" skip_confirm:true
```

Jorge-Martin777 (approved):
250 changed

PR #3583, opened 2025-02-13
Improve OPML import logic

MiSikora:
## Description

Currently, the OPML import logic is messy and has several issues:

- We pre-process OPML files to escape invalid XML characters, but this is done only for files, not for URLs.  
- The escaping logic itself is incorrect and broken.  
  - The `&apos;` character is incorrectly escaped because we match `U+2019` (right quotation mark `’`) instead of `U+0027` (apostrophe `'`).  
  - Only some characters are escaped in specific XML attributes (`text`, `xmlUrl`, `type`), which can cause issues if these characters appear elsewhere.
- If certain characters remain unescaped, we resort to finding URLs using Regex patterns, adding unnecessary complexity.  
- The input streams are read multiple times, slowing down the process.
- We were creating a temporary file just to pre-process unescaped characters.
- The old implementation had test only for the preprocessing parts and it even had mistakes in it like escaping the `’` character.

This PR addresses these issues by simplifying the OPML import logic:

- Simple string manipulation is used to find URLs, eliminating the need for an XML parser and Regex patterns.
- The entire process is now handled in a single stream read, applying the same logic to both files and URLs.
- `OpmlUrlReader` is heavily covered with tests and tests our real logic of mapping OPML entries to URLs.

While using an XML parser could be beneficial, it is unnecessary in this case since we only extract URLs, and it's unlikely that additional data processing will be needed.

A small improvement was made in c20eeee3e134d699b44108fc3af0d43a050359a5. Since most of our Workers are `HiltWorkers`, there is no need to handle them manually in the `WorkerFactory`.

## Testing Instructions

1. Clear app's data.
2. Go to settings.
3. Tap on `Import & export OPML`.
4. Test importing logic with different OPML files. You can use ones these ones: p1739280028220679-slack-C06EHRAPW12. But it would be also good if you test it with other files not necessarily from our application.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

mebarbosa (approved):
I've tested with different OPML file and it also worked as expected 💯

PR #3586, opened 2025-02-13
Fix Suggested Folders Crash

mebarbosa:
## Description
- See: p1739448836612389-slack-C07J5LNP4SF

## Testing Instructions
1. Go to Podcasts tab
2. Ensure you don't see any crash

## Checklist

### DIFF
--- a/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/podcasts/PodcastsFragment.kt
+++ b/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/podcasts/PodcastsFragment.kt
@@ -13,7 +13,9 @@ import androidx.core.os.bundleOf
 import androidx.core.view.updatePadding
 import androidx.fragment.app.activityViewModels
 import androidx.fragment.app.viewModels
+import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.lifecycleScope
+import androidx.lifecycle.repeatOnLifecycle
 import androidx.recyclerview.widget.GridLayoutManager
 import androidx.recyclerview.widget.ItemTouchHelper
 import androidx.recyclerview.widget.LinearLayoutManager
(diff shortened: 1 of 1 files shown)


MiSikora (commented):
I found two issues. One is related to this change, and I'm not sure if the second one is actually an issue, but it seems strange.

### Folders suggestions constantly shows up

When I dismiss the folder suggestion, I wouldn’t expect it to reappear every time I return to the Podcasts page. Shouldn’t it show up at most once per some period of time? Right now, it feels almost like pestering. See the video:

https://github.com/user-attachments/assets/c0df64fa-67e4-4698-920c-5e5fc0cfb520

### Unsigned users are prompted to buy a subscription

1. Be signed out and have some followed podcasts.
2. Trigger the folders bottom sheet.
3. Tap on "Use these folders".
4. The subscription bottom sheet appears.

I don't think this is correct. We shouldn't offer a subscription to someone who doesn’t have an account. Instead, they should be shown the login flow first, and only then be nudged toward a subscription. Otherwise, they won’t be able to purchase a subscription because there won’t be an account to associate it with.

mebarbosa:
> When I dismiss the folder suggestion, I wouldn’t expect it to reappear every time I return to the Podcasts page. Shouldn’t it show up at most once per some period of time? Right now, it feels almost like pestering. See the video:

@MiSikora

> When I dismiss the folder suggestion, I wouldn’t expect it to reappear every time I return to the Podcasts page
You dismissed the paywall modal by dragging it down? Dom talked about hiding this when tapping on maybe later button only, but if you are talking about dragging it down to hide I agree with you  

> I don't think this is correct. We shouldn't offer a subscription to someone who doesn’t have an account. Instead, they should be shown the login flow first, and only then be nudged toward a subscription.

This is a great point. I will update this modal in another PR

MiSikora:
> You dismissed the paywall modal by dragging it down? Dom talked about hiding this when tapping on maybe later button only, but if you are talking about dragging it down to hide I agree with you

By tapping outside. But I think dragging it down should result in the same scenario.

> This is a great point. I will update this modal in another PR

Great!!

PR #3592, opened 2025-02-14
[Suggested Folders] Fetch data and persist it into database

mebarbosa:
## Description
- This PR fetches the suggested folder (mocked) endpoint and saves it into our database. This is necessary because fetching the real data would take about 10 seconds, so we are not going to show a loading progress to the user. Instead, we will fetch the data and display it once it is available

### Fetching data
1. Run the app in `debug`
2. Have a fresh install
3. Open the app using a free account or no account, and **do not follow** any podcasts
4. Go to Podcasts tab
5. Wait some time and ensure you **don't** see the paywall suggested folders modal ✅
6. Go to Discover and follow some podcasts
7. Go to Podcasts
8. Ensure you see the paywall suggested folders modal ✅

You can also add logs or put a breakpoint to ensure is fetching data:
https://github.com/Automattic/pocket-casts-android/blob/371d24814087efa6aa8ebeee25d2ebe6957249b3/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastsViewModel.kt#L306

### DB Migration
1.  Install the app from `main` branch
2. Interact with the app
3. Install the app from `task/fetch-suggested-folders` branch
4. Ensure the app does not crash ✅

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/SuggestedFoldersManager.kt
```diff
@@ -0,0 +1,43 @@
...
+            Timber.e(e, "Getting suggested folders failed")
+            return null
+        }
+    }
+
+    suspend fun refreshSuggestedFolders(podcastUuids: List<String>) {
+        try {
+            val request = SuggestedFoldersRequest(podcastUuids)
+            val folders: List<SuggestedFolder> = podcastCacheService.suggestedFolders(request)
+            suggestedFoldersDao.insertAll(folders)
```
MiSikora, line 38: This implementation results in these scenarios:
* When a different set of folders is returned from the server the old suggestions persist.
* When a different set of podcasts is returned for the safe folder the old podcasts are still present.

This can lead to us suggesting wrong folders or folders with wrong podcasts.

MiSikora (changes requested):

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/SuggestedFoldersManager.kt
mebarbosa: humm... Good point. So I think we should delete the existing one and add the new folders if the server returns something

mebarbosa: @MiSikora I've pushed the changes [d5b56f3](https://github.com/Automattic/pocket-casts-android/pull/3592/commits/d5b56f36e64cc4320e7f83fd1528b8295822ee81)

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/SuggestedFolder.kt
```diff
@@ -0,0 +1,16 @@
...
+import androidx.room.PrimaryKey
+
+@Entity(
+    tableName = "suggested_folders",
+    indices = [Index(value = ["folder_name", "podcastUuid"], unique = true)],
+)
+data class SuggestedFolder(
+    @PrimaryKey(autoGenerate = false) @ColumnInfo(name = "uuid") var uuid: String,
+    @ColumnInfo(name = "folder_name") var name: String,
+    @ColumnInfo(name = "podcastUuid") var podcastUuid: String,
```
MiSikora, line 15: Can you change column name to `podcast_uuid` to avoid mixing different case formatting?

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/SuggestedFolder.kt
```diff
@@ -0,0 +1,16 @@
+package au.com.shiftyjelly.pocketcasts.models.entity
+
+import androidx.room.ColumnInfo
+import androidx.room.Entity
+import androidx.room.Index
+import androidx.room.PrimaryKey
+
+@Entity(
+    tableName = "suggested_folders",
+    indices = [Index(value = ["folder_name", "podcastUuid"], unique = true)],
```
MiSikora, line 10: What's the purpose of this index?

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/SuggestedFolder.kt
```diff
@@ -0,0 +1,16 @@
...
+)
+data class SuggestedFolder(
+    @PrimaryKey(autoGenerate = false) @ColumnInfo(name = "uuid") var uuid: String,
```
MiSikora, line 13: What's the reason for having `uuid`? Can't the pair of `folder_name` and `podcastUuid` be used as a primary key?

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/SuggestedFoldersDao.kt
```diff
@@ -0,0 +1,34 @@
...
+    abstract suspend fun insert(folder: SuggestedFolder)
+
+    @Insert(onConflict = OnConflictStrategy.IGNORE)
+    abstract suspend fun insertAll(folders: List<SuggestedFolder>)
+
+    @Query("SELECT * FROM suggested_folders WHERE folder_name = :folderName")
+    abstract suspend fun findAllFolderPodcasts(folderName: String): List<SuggestedFolder>
+
+    @Query("SELECT * FROM suggested_folders")
+    abstract fun findAll(): Flow<List<SuggestedFolder>>
```
MiSikora, line 24: This method should be suspending.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/SuggestedFoldersManager.kt
```diff
@@ -0,0 +1,45 @@
...
+import kotlin.coroutines.CoroutineContext
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.Flow
+import timber.log.Timber
+
+class SuggestedFoldersManager @Inject constructor(
+    private val podcastCacheService: PodcastCacheServiceManager,
+    appDatabase: AppDatabase,
+) : CoroutineScope {
```
MiSikora, line 18: I think we should avoid extending `CoroutineScope` and inject one if it is needed. I know that we have this pattern in some parts of the code but it is not a good pattern. It hinders testing and doesn't offer much in return. More often then not we don't even need to have a scope and a simple suspending function is enough.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/SuggestedFoldersManager.kt
```diff
@@ -0,0 +1,45 @@
...
+    override val coroutineContext: CoroutineContext
+        get() = Dispatchers.Default
+
+    fun getSuggestedFolders(): Flow<List<SuggestedFolder>>? {
+        try {
+            return suggestedFoldersDao.findAll()
+        } catch (e: Exception) {
+            Timber.e(e, "Getting suggested folders failed")
+            return null
+        }
```
MiSikora, line 31: Can you explain why is there a need for `try/catch` here? Do you expect database to crash while reading from it?

modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/podcast/PodcastCacheServiceManagerImpl.kt
```diff
@@ -83,4 +85,26 @@ class PodcastCacheServiceManagerImpl @Inject constructor(
...
+                folderEntities.add(
+                    SuggestedFolder(
+                        uuid = UUID.randomUUID().toString(),
+                        name = folderName,
+                        podcastUuid = podcastUuid,
+                    ),
+                )
+            }
+        }
+        return folderEntities
```
MiSikora, line 108: This can be expressed as a regular `flatMap().map()` chaing without introducing local `val`s.

modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/podcast/PodcastCacheServiceManagerImpl.kt
```diff
@@ -83,4 +85,26 @@ class PodcastCacheServiceManagerImpl @Inject constructor(
...
+
+    override suspend fun suggestedFolders(request: SuggestedFoldersRequest): List<SuggestedFolder> {
+        return service.suggestedFolders(request).toSuggestedFolders()
+    }
+
+    private fun Map<String, List<String>>.toSuggestedFolders(): List<SuggestedFolder> {
+        val folderEntities = mutableListOf<SuggestedFolder>()
+
+        this.forEach { (folderName, podcastList) ->
+            // For each podcast in the list, create a SuggestedFolder entity
```
MiSikora, line 97: I'd like to understand the reasoning behind adding comments like this one or these:

https://github.com/Automattic/pocket-casts-android/blob/d5b56f36e64cc4320e7f83fd1528b8295822ee81/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/AppDatabase.kt#L944

https://github.com/Automattic/pocket-casts-android/blob/d5b56f36e64cc4320e7f83fd1528b8295822ee81/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/AppDatabase.kt#L956

Maybe I’m missing something, but do you find these comments helpful? They only describe what the code does, and the logic isn’t particularly complex. I understand that comments can be useful in certain cases, such as here:

https://github.com/Automattic/pocket-casts-android/blob/d5b56f36e64cc4320e7f83fd1528b8295822ee81/modules/features/reimagine/src/main/kotlin/au/com/shiftyjelly/pocketcasts/reimagine/ui/ClipSelectorState.kt#L69-L84

However, the comments in this PR feel somewhat redundant—akin to:

```kotlin
// Add two numbers, 2 and 5
val result = 2 + 5
```

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/AppDatabase.kt
```diff
@@ -936,6 +940,28 @@ abstract class AppDatabase : RoomDatabase() {
...
+                """.trimIndent(),
+            )
+
+            // Create the unique index for folder_name and podcastUuid if it doesn't already exist
+            database.execSQL(
+                """
+                    CREATE UNIQUE INDEX IF NOT EXISTS `index_suggested_folders_folder_name_podcastUuid`
+                    ON `suggested_folders` (`folder_name`, `podcastUuid`);
+                """.trimIndent(),
+            )
```
MiSikora, line 962: Multiple operations that happen during a migration should be wrapped in a transaction.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/SuggestedFoldersManager.kt
mebarbosa: I am being conservative here, but I don't expect a crash from the database

MiSikora: I'm asking because I don't think we do this in any other places. Maybe there are some single instances. I'm not saying that this is necessarily wrong. I'd even say that it is strictly correct to have `try/catch` when interacting with DB. But if that's something you think is worth doing I think we should discuss it in Slack or during one of our syncs instead of sometimes having it in the code.

mebarbosa: I don't have strong opinion about this. I will remove it in the next commits

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/SuggestedFolder.kt
mebarbosa: I wanted to ensure that I have only one Podcast A in an folder, but since I am cleaning the folders before adding new ones I don't need this anymore. I will remove it

MiSikora: Ah, makes sense. Didn't thought of it. But yeah, If you clear the DB and if you change the primary key to be defined like this index there shouldn't be a need for it.

mebarbosa:
Thank you for your review, @MiSikora ! I've addressed the changes

PR #3593, opened 2025-02-14
[Tooling] Fix `finalize_release` and the `release/*` branch checkout

iangmaia:
# Why?

During release process we need the lanes invoked from the `release-pipelines/*.yml` and that will require to do a `git push` to the release branch to actually `git fetch` + `git checkout` those release branches explicitly, otherwise the `git push` will fail (e.g. unknown `refspec`)

# How

 - We still need to checkout the `release/*` branch before the compilation in `release-build.sh`, because we want to ensure that we include the latest commit that was just pushed by the `new_beta_release` / `finalize_release` lanes from previous job, instead of building the commit that the CI build was initially created on
 - But we **also** need to checkout the `release/*` in `release-pipelines/*.yml` before calling any of the lanes that would do a version bump and push the commit, otherwise it won't try to do the commit on the branch but on the detached HEAD.

# Links

Internal ref: p1739547154659529/1739543498.075849-slack-CC7L49W13
iOS Counterpart: https://github.com/Automattic/pocket-casts-ios/pull/2754

### DIFF
--- a/.buildkite/commands/checkout-release-branch.sh
+++ b/.buildkite/commands/checkout-release-branch.sh
@@ -1,0 +1,19 @@
+#!/bin/bash -eu
+
+# Script to checkout a specific release branch
+# Usage: ./checkout-release-branch.sh <RELEASE_VERSION>
+
+# Buildkite, by default, checks out a specific commit, ending up in a detached HEAD state.
+# But in some cases, we need to ensure to be checked out on the `release/*` branch instead, namely:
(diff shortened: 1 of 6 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

AliSoftware (approved):

PR #3606, opened 2025-02-17
[Suggested Folders] Open suggested folders only if has pre cached them

mebarbosa:
## Description
- This opens the suggested folders screen when tapping on create folder button only if has pre cached for paid users
- Opens the suggested folders paywall for free users only if has pre cached them

## Testing Instructions
- Run the app in `debug`

#### Free user
1. Open the app
2. Do not follow any podcast
3. Go to Podcasts tab and ensure you don't follow any podcast
4. Ensure you don't see the suggested folders paywall ✅
5. Try to create a folder
6. Ensure you see the regular upgrade account screen ✅
7. Follow some podcasts
8. Go to podcasts tab again
9. Wait a few seconds
10. Ensure you see the suggested folders paywall ✅
11. Try to create a folder
12. Ensure you see the suggested folders paywall ✅

### Paid user
1. Open the app
2. Do not follow any podcast
3. Go to Podcasts tab and ensure you don't follow any podcast
4. Try to create a folder
5. Ensure you see the existing Create Folder screen ✅
6. Follow some podcasts
7. Go to podcasts tab again
8. Wait a few seconds
9. Try to create a folder
10. Ensure you see the suggested folder screen ✅

#### Paid User
[paid.webm](https://github.com/user-attachments/assets/56c7ba6c-ec95-41ec-b1ea-9dbdbd42828a)

#### Free User
[free.webm](https://github.com/user-attachments/assets/2db058d3-f3e7-493e-ab8e-337f4a0827e9)

## Checklist
- [ ] If this is a user-facing change, I have added an entry in CHANGELOG.md
- [x] Ensure the linter passes (`./gradlew spotlessApply` to automatically apply formatting/linting)
- [ ] I have considered whether it makes sense to add tests for my changes
- [ ] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml`
- [ ] Any jetpack compose components I added or changed are covered by compose previews
- [ ] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

PR #3615, opened 2025-02-18
[Suggested Folders] Filter podcasts to send to suggested folders API

mebarbosa:
## Description
- This PR sends only podcasts to the suggested folders API only if they are not already in a folder. This is to prevent that we have a podcast in a folder and create another folder with the same podcast from suggested folder API

> [!IMPORTANT]  
> There is an issue with the current podcast list. If you follow / unfollow a podcast or update its folder, it will not be automatically updated in the Podcast tab for suggested folders. To get the updated list updated sent to the suggested folders API, you will need to navigate away from the Podcast tab, go to another screen, and then return to the Podcast tab. (or tap 2x in Podcast tab to refresh the page) This was already addressed in https://github.com/Automattic/pocket-casts-android/pull/3606

## Testing Instructions
1. Run the app in `debug`
2. Log in with paid account
3. Follow some podcasts
4. Go to Podcasts tab
5. Manually create a folder and add some podcast (Tap on create custom folders button)
7. Tap on Podcasts tab again to refresh the folders due the issue I mentioned above
8. Check the breakpoint I shared above to see if the API returns only suggested folders for podcasts that are not currently in folder ✅

https://github.com/Automattic/pocket-casts-android/blob/ff738f51a3ff17376cbba419cba67023a4516cb0/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastsViewModel.kt#L303

## Checklist

### DIFF
--- a/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastsViewModel.kt
+++ b/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastsViewModel.kt
@@ -310,7 +310,7 @@ class PodcastsViewModel
     fun refreshSuggestedFolders() {
         viewModelScope.launch {
             if (FeatureFlag.isEnabled(Feature.SUGGESTED_FOLDERS)) {
-                val uuids = podcastManager.findSubscribedUuids()
+                val uuids = podcastManager.findFollowedPodcastsNotInFolderUuid()
                 suggestedFoldersManager.refreshSuggestedFolders(uuids)
             }
         }
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/PodcastDao.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/PodcastDao.kt
@@ -142,6 +142,9 @@ abstract class PodcastDao {
     @Query("SELECT * FROM podcasts WHERE folder_uuid IS NULL")
     abstract suspend fun findPodcastsNotInFolder(): List<Podcast>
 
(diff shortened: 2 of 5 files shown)


modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/PodcastDao.kt
```diff
@@ -142,6 +142,9 @@ abstract class PodcastDao {
     @Query("SELECT * FROM podcasts WHERE folder_uuid IS NULL")
     abstract suspend fun findPodcastsNotInFolder(): List<Podcast>
 
+    @Query("SELECT uuid FROM podcasts WHERE subscribed = 1 AND folder_uuid IS NULL")
+    abstract suspend fun findFollowedPodcastsNotInFolderUuid(): List<String>
```
geekygecko, line 146:
```suggestion
    abstract suspend fun findFollowedPodcastUuidsNotInFolder(): List<String>
```

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastsViewModel.kt
```diff
@@ -310,7 +310,7 @@ class PodcastsViewModel
     fun refreshSuggestedFolders() {
         viewModelScope.launch {
             if (FeatureFlag.isEnabled(Feature.SUGGESTED_FOLDERS)) {
-                val uuids = podcastManager.findSubscribedUuids()
+                val uuids = podcastManager.findFollowedPodcastsNotInFolderUuid()
```
geekygecko, line 313: It would be good to cache this so it doesn't call the server each time the page loads.

geekygecko (approved):

mebarbosa:
Closing this PR since we are going to continue sending all followed podcast to the API 

p1739971321715109-slack-C08BRS7N9UL

PR #3630, opened 2025-02-20
[Suggested Folders] Suggested Folders Tweaks

mebarbosa:
## Description
- Monitor list of subscribed podcasts to refresh suggested folders with the latest podcast list
- Pass the required argument for `SuggestedFolders` that I forgot to update on this screen

## Testing Instructions
1. Run in `debug`
2. Use free account without having followed any podcast
3. Follow a new podcast in discover
4. Go to Podcasts tab
5. Ensure you see this new podcast you just followed in the suggested folders paywall

## Screenshots or Screencast 
[Free.webm](https://github.com/user-attachments/assets/252edd4a-d1f4-4916-940b-cb87bfb3fe74)

## Checklist


modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastsViewModel.kt
```diff
@@ -314,8 +314,9 @@ class PodcastsViewModel
     fun refreshSuggestedFolders() {
         viewModelScope.launch {
             if (FeatureFlag.isEnabled(Feature.SUGGESTED_FOLDERS)) {
-                val uuids = podcastManager.findSubscribedUuids()
-                suggestedFoldersManager.refreshSuggestedFolders(uuids)
+                podcastManager.findSubscribedFlow().collect { podcasts ->
+                    suggestedFoldersManager.refreshSuggestedFolders(podcasts.map { it.uuid })
+                }
```
geekygecko, line 319: When signing into my account, I noticed that it keeps calling the server. 
Screenshot 2025-02-21 at 5 13 50 pm

mebarbosa: I will investigate how can I fix the issue of missing podcasts in suggested folders without needing to monitor the subscribed changes

mebarbosa: So, I changed the approach here [37a522](https://github.com/Automattic/pocket-casts-android/pull/3630/commits/d37a522d6c04f73fd0ef8bdc7d154dc0b47e94dd)

geekygecko: Thanks for changing that. Whenever I go back to the podcasts tab, it calls the endpoint even if I haven't added any podcasts or if I'm not tapping the suggested folder icon as a Plus user. I wonder if this is still too much? Especially if the server calls take a while to run.  

I initially thought we could store a setting for the latest podcast added date and only update it if the date changed, but that doesn't take into consideration when you unfollow a podcast. Is a hash of the subscribed UUIDs too much?

mebarbosa: I created this thread for us to talk about this p1740401875334509-slack-C08BRS7N9UL

mebarbosa: I've updated the code with the hash approach we talked about @geekygecko

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastsViewModel.kt
```diff
@@ -303,11 +305,18 @@ class PodcastsViewModel
...
         if (FeatureFlag.isEnabled(Feature.SUGGESTED_FOLDERS)) {
             _suggestedFoldersState.emit(SuggestedFoldersState.Loading)
-            suggestedFoldersManager.getSuggestedFolders().collect { folders ->
-                if (!folders.isEmpty()) {
-                    _suggestedFoldersState.emit(SuggestedFoldersState.Loaded(folders))
+            suggestedFoldersManager.getSuggestedFolders()
+                .collectLatest { folders ->
+                    // This is to fix a race condition where the database emits duplicate notifications,
+                    // which was causing the suggested folders paywall to incorrectly display after unfollowing all podcasts.
+                    delay(200)
```
MiSikora, line 312: I'm a bit confused. Where is the race condition? It sounds that there are frequent updates, which is normal. In which case you probably want to have more explicit intention with `debounce()` rather than using a raw `delay()`. 

On a side note, in general, if you suspect race conditions I wouldn't recommend solving them with delays because they are a walking minefields.

mebarbosa: The db is notifying twice when I remove an item.

Let's say I follow only one podcast and then and unfollow this one.

The DB notifies that I have 1 podcast and right after that notifies that I have 0 podcast. This first notification is leading to show the paywall when I don't want to.

I tried to solve this with:
- `debounce` ❌
- `distinctUntilChanged` ❌
- `collectLatest` ❌
- `conflate` ❌

But none of these approaches worked

MiSikora: This isn't a race condition but a normal DB behavior. You see 2 emissions because there are 2 changes. Can you show how do you apply `debouce()`? I don't understand why it wouldn't work. It should do exactly what you need. Roughly speaking the combination of  `combineLatest()` + `delay()` is a version of `debounce()`.

From [docs](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/debounce.html):

> Returns a flow that mirrors the original flow, but filters out values that are followed by the newer values within the given [timeout](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/debounce.html). The latest value is always emitted.

You can even verify it with this example.

```kotlin
suspend fun main() {
  val flow = flow {
    emit(1)
    delay(50)
    emit(2)
    delay(200)
    emit(3)
    delay(110)
    emit(4)
    delay(200)
    emit(5)
  }

  flow.debounce(200).collect { println("Debounce: $it") }
  flow.collectLatest { delay(200); println("Delay: $it") }
}
```

It prints the text below, which is expected.

```
Debounce: 2
Debounce: 4
Debounce: 5
Delay: 2
Delay: 4
Delay: 5
```

mebarbosa: I've pushed the update using `debounce` and I figured out what was the issue. Believe it or not, I dreamed about this last night, lol.

I had set the `debounce` in `PodcastsViewModel.loadSuggestedFolders()`, but this did not work like I said. 

So in `PodcastsFragment` I also changed from `viewModel.userSuggestedFoldersState.collect` to `viewModel.userSuggestedFoldersState.collectLatest` and it worked as I expect

MiSikora (approved):
Looks good to me. Left 1 comment about threading improvement.

PR #3711, opened 2025-03-05
[Tooling] Remove non-Ruby platforms from Gemfile.lock

AliSoftware:
Context: p1741171455910009/1741071779.090879-slack-C028JAG44VD

## Description

This PR:
 - Removes the non-ruby platforms from `Gemfile.lock`. I did this using the following command:
    ```
    bundle lock --remove-platform arm64-darwin-{21,22,23} x86_64-darwin-{20,21,22}
    ```
 - Add `BUNDLE_SPECIFIC_PLATFORM: "false"` to the `.bundle/config` file. This should not really be necessary in theory, because iinm, the already-existing `BUNDLE_FORCE_RUBY_PLATFORM: "true"` was supposed to be enough. But better safe than sorry I guess, and doesn't hurt to add it anyway.

Those changes are necessary to avoid issues with `GLIBC_2.28` not being available on some Linux distributions / CI agents we run the code on, while the _platform-specific, precompiled_ version (`.so` lib) of the `nokogiri` gem depends on it.

> [!NOTE]
> This PR targets the `release/7.84` branch so that it can unblock @ashiagr and allow her to do a new beta.

## Testing Instructions

Should be ok as long as CI is green, and that the next `new_beta_release` job we trigger after this is merged also passes without an issue at the `gem install` stage.

PR #3713, opened 2025-03-05
[Tooling] Attempt to fix RubyGems cache

AliSoftware:
This is a follow-up on #3711 and another attempt at fixing the RubyGems issue.

## What / How

The main idea is that I've made some dummy update to the `Gemfile.lock` so that its checksum has changed and it would trigger a cache miss, and then have [triggered a build with the dummy version of the `.buildkite/release-pipelines/new-beta-release.yml` pipeline on Buildkite](https://buildkite.com/automattic/pocket-casts-android/builds/10899/canvas?sid=0195660e-8a64-4d42-bcd9-e06efaff916d) so that the first time the gems cache is rebuilt for this new `Gemfile.lock` checksum it'd be build from the `tumblr-metal` agent (the one that does _not_ have the latest `GLIBC_2.28`.

The hope with this is that the new cache now does not contain the `.so` for the pre-compiled `nokogiri` that would require `GLIBC_2.28` anymore, but instead be a cache that can equally run on `tumblr-metal` and `android` CI agents used by this repo.

## Testing

 - [The cache has been rebuilt from the `tumblr-metal` agent](https://buildkite.com/automattic/pocket-casts-android/builds/10899/canvas?sid=0195660e-8a64-4d42-bcd9-e06efaff916d#0195660e-8a9b-40eb-bcbe-bf11eb1c8cc8/136-138)
 - And [the `bundle exec fastlane run configure_apply` command has run successfully on those agents](https://buildkite.com/automattic/pocket-casts-android/builds/10899/canvas?sid=0195660e-8a64-4d42-bcd9-e06efaff916d#0195660e-8a9b-40eb-bcbe-bf11eb1c8cc8/812-835)
 - So now all that's left is to validate that the CI jobs that will run on this PR and use the same rebuild gems cache but on other CI agents (especially `android` one) will still successfully install `nokogiri` from that cache without an incompatibility between platforms.

## What's Next

Note that this is just a temporary workaround to unblock the team and fix CI ASAP.

But this is not enough to prevent the issue from reappearing the next time we `bundle update` the gems via a PR and a new gems cache is once again regenerated but from an agent other than `tumblr-metal`.

To fix the issue long term, we'll probably have to adjust our implementation of `install_gems` in `a8c-ci-toolkit`, to:
 - Either exclude the `.so` files from the cache when creating the `.tar.gz` cache archive
 - Or deleting the `.so` file just after restoring the gems cache (but before that helper calling `bundle install`)
 - Or include the name of the Buildkite queue (`$BUILDKITE_AGENT_META_DATA_QUEUE`) in the hash we use to build the cache key, so that each different agent queue use a different gems cache.
 - Or just include the `glibc` version as part of the cache key (lighter solution than including the whole Buildkite queue name). This is actually already being considered in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/149

### DIFF
--- a/.buildkite/release-pipelines/new-beta-release.yml
+++ b/.buildkite/release-pipelines/new-beta-release.yml
@@ -4,24 +4,21 @@
 # This pipeline is meant to be triggered by the MC scenario in ReleasesV2
 
 steps:
-  - label: "🚂 DEBUG: New Beta Release / Ruby Setup"
+  - label: "🚂 New Beta Release"
     command: |
       echo '--- 🤖 Use bot for Git operations'
       source use-bot-for-git
 
-      # .buildkite/commands/checkout-release-branch.sh "${RELEASE_VERSION}"
+      .buildkite/commands/checkout-release-branch.sh "${RELEASE_VERSION}"
 
       echo '--- :ruby: Setup Ruby Tools'
       install_gems
 
-      echo '--- :ruby: Show Gemfile.lock'
-      cat Gemfile.lock
-
       echo '--- 🔐 Access Secrets'
       bundle exec fastlane run configure_apply
 
-      # echo '--- 🚀 New Beta Release'
-      # bundle exec fastlane new_beta_release skip_confirm:true
+      echo '--- 🚀 New Beta Release'
+      bundle exec fastlane new_beta_release skip_confirm:true
     plugins: [$CI_TOOLKIT]
     agents:
       queue: "tumblr-metal"


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

ashiagr (approved):

iangmaia (approved):

PR #3719, opened 2025-03-06
Use CI toolkit plugin v5.1.1 with fixes for nokogiri CI issue

mokagio:
## Description
Occasionally, we have seen CI issues due to steps running on `android` and `tumblr-metal` downloading the same Ruby gems cache. 

If the cache was generated on `android`, `tumblr-metal` was able to use it fine, but if it was the other way around it would fail. That was due to the `libc` version in the two queues, with the one on `android` being newer than the one on `tumblr-metal` and resulting in an incompatible `nokogiri` build on `tumblr-metal`.

See also https://github.com/Automattic/pocket-casts-android/pull/3711

The new CI toolkit plugin builds the Ruby gem cache in a way that accounts for the `libc` version of the machine building the cache, which should prevent this kind of clash in the future.

Internal refs:

- p1739222874153819-slack-C02KLTL3MKM
- p1741174663524979/1741071779.090879-slack-C028JAG44VD

## Testing Instructions
This is hard to test without artificially forcing `tumblr-metal` to build the cache before `android`. And that's even harder because we do not use the `tumblr-metal` queue in the default pipeline...

However, you can see that [CI is green](https://buildkite.com/automattic/pocket-casts-android/builds/10915/canvas?sid=019569a3-dfcd-4d95-8737-063e54cf4f40) and that it accounts for `libc` in the cache computation:

image

image

## Screenshots or Screencast 

N.A.

## Checklist
N.A.
 
N.A.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

MiSikora (approved):

AliSoftware:
> If the cache was generated on `android`, `tumblr-metal` was able to use it fine, but if it was the other way around it would fail.

For the record, in practice I think this part of the PR description is backwards, as `tumblr-metal` is the one having an older version of `libc` and thus unable to use a cache generated by `android` agent that are using a more recent version, so it's when the cache was generated by the `android` agent first (and thus depending on a more recent `libc` present in that agent) that `tumblr-metal` was unable to use that cache (only having an older `libc` version available) 🙃

mokagio:
Not sure why it didn't auto-merge. 

I resolved the merge conflict and was going to let it merge, but then I thought the resulting merge history and maybe commit message mentionin 5.1.1 when the value would have been 5.1.3 to be too confusing. 

Simply closing and doing without the splitting of version definition and plugin string definition.

PR #3762, opened 2025-03-13
Fix nav bar overlay issues

### LINKED ISSUE #3631: Bottom navigation hidden behind another view
### Description

A user has reported when they return to the app the bottom navigation is hidden. 

> With the current version of the app, when I switch back from another app, I often get the bottom menu blocked by a grey bar that looks like a broken version of the episode pop-up (with the add to up next top/bottom buttons etc), and I have to swipe to close the app to resolve the issue.

> Screenshot of behaviour - see grey area at the bottom of the screen, Pixel 6 with latest Android 15/February 2025, with Gemini installed if that's relevant. I can just see the top of 2 blue icons at the bottom matching actions in the episode menu.

I haven't been able to repeat this issue but the user says it's when they switch back to the app. 

Pixel 6, Android 15, Pocket Casts 7.83-rc-2

MiSikora:
## Description

This PR addresses issues related to edge-to-edge display. There are two fixes.

### Player Sheet

The player bottom sheet could reach a state where the navigation bar offset was incorrect. This occurred because the `BottomNavHideManager` class used an offset limit based on the navigation bar height. However, due to window insets application, it could get out of sync during configuration changes. To resolve this, I removed the class and replaced it with a simple offset based on the player's sheet offset, without any limits.

### Bottom Sheet

The `frameBottomSheet`'s `BottomSheetBehavior` did not have its state saved and restored because its parent `CoordinatorLayout` did not have an assigned ID. As a result, after configuration changes, the fragment remained displayed, but the sheet was collapsed. 

This is a more complex issue because, as far as I can tell, the Material library does not properly handle insets, causing the bottom sheet to be laid out on top of the content even when collapsed. See the image below:

Screenshot 2025-03-13 at 10 56 22

This is not an issue when no fragment is attached, as the view has no content and is not clickable. However, if a fragment is displayed, it blocks the bottom navigation bar. Ideally, we would offset the bottom sheet out of the screen, but I have not found a way to achieve this. Even if a solution exists, the current fix is orthogonal to that problem and directly addresses our primary concern: preventing the navigation bar from being blocked due to state loss.

Fixes #3631

There is a new bug but I think it is very minor and we can live with it for now. Someone can try to fix it in the future.

1. Open player.
2. Drag up Up Next queue.
3. Change configuration.
4. Up Next queue is gone and can't be dragged up. But it can be opened with a button.

It looks like somehow order of sheets isn't correct for a moment after recreation.

## Testing Instructions

1. Open the player.
2. Perform configuration changes couple of times.
3. At no point the navigation bar should be displayed on top of the player.
4. Drag the player up and down to see navigation bar behavior.

---

1. Open full screen Up Next queue with a button on the Mini Player.
2. Change configuration a couple of times.
3. At no point should the queue obstruct the navigation bar.

---

> [!note]
> In general do some monkey testing around the player sheet and the bottom sheet UI.

## Checklist
 
- [x] with different themes
- [ ] ~for accessibility with TalkBack~


geekygecko (approved):
Thanks for tackling this tricky issue!

PR #3781, opened 2025-03-17
Add padding to the podcast grid

geekygecko:
## Description

If the mini player wasn't open, the podcast grid didn't add the surrounding padding. 

## Testing Instructions

1. Fresh install the app.
2. Follow some podcasts.
3. Go to the Podcasts tab.
4. ✅ Verify the podcast grid has padding on the side so the artwork lines up with the Podcasts title.

## Screenshots 

| Before | After |
| --- | --- |
| Screenshot_20250317_171710 | Screenshot_20250317_171625 | 

## Checklist
 
- [x] with different themes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
         ([#3761](https://github.com/Automattic/pocket-casts-android/pull/3761))
     *   Fix issues with Classic widget not syncing its state in some scenarios.
         ([#3777](https://github.com/Automattic/pocket-casts-android/pull/3777))
+    *   Fix podcast grid padding
+        ([#3781](https://github.com/Automattic/pocket-casts-android/pull/3781))
 *   Updates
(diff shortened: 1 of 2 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

MiSikora (approved):
I left similar comment in Slack but for visibility leaving it out here as well. Can we fix it in a way where the outer padding doesn't depend on the mini player presence? Only bottom part should depend on it as this is needed for the mini player support.

We can go with this change to fix in in beta but it is an unpredictable logic to keep. And as a side note—with `onStart()` it seems more reasonable to change `_bottomInset` to `MutableStateFlow` with an initial value of `0`.

PR #3786, opened 2025-03-17
[Tooling] Update gems to latest version

AliSoftware:
Mostly to update `json` gem to the latest version to fix security issue reported by dependabot in https://github.com/Automattic/pocket-casts-android/security/dependabot/88


iangmaia (approved):
👍

PR #3802, opened 2025-03-21
Fix subscription mapper not using all offers

MiSikora:
## Description

This is a very ugly fix for the issue where users only see the yearly offer when a 1-month free trial is available.

The root cause of the issue is that the `SubscriptionMapper` logic is highly sensitive to any changes in our Play Console configuration. When the Winback offer was added, it wasn’t properly filtered out in most of the scenarios. I added a boolean flag to determine whether it should be filtered out or not. As mentioned, this is a temporary and inelegant workaround to restore expected behavior.

I believe our approach to retrieving subscription data needs a significant redesign and rewrite. We shouldn't be in a situation where adding an offer in the Play Console causes issues in unrelated parts of the app.

## Testing Instructions

> [!note]  
> You need to test with an account signed in to the Play Console that has never claimed a free trial offer. Otherwise, you won't be able to see it when purchasing a subscription. If the 1-month free offer doesn't appear, create a new Google account, preferably on a new device or emulator.

1. Sign in.
2. During the onboarding flow, when you see subscription offers, you should be able to switch between them.

I also smoke tested the Winback and Referral flows. You should do that as well. Winback should be tested using the production version of the app communicating with staging, due to the account duration requirement. You can apply [this patch](https://github.com/user-attachments/files/19389381/release.patch) and test with the `debug` variant.

## Screenshots or Screencast 

| Before | After |
| - | - |
| <video src ="https://github.com/user-attachments/assets/721fbae6-2951-45b4-a652-88abee935c10" /> | <video src="https://github.com/user-attachments/assets/feb05994-4f6e-4acc-b37d-616dbd95f1c2" /> |

### DIFF
--- a/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/winback/WinbackViewModel.kt
+++ b/modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/winback/WinbackViewModel.kt
@@ -261,7 +261,7 @@ class WinbackViewModel @Inject constructor(
             .await()
     }
 
-    private fun List<ProductDetails>.toSubscriptionPlans() = map(subscriptionMapper::mapFromProductDetails)
+    private fun List<ProductDetails>.toSubscriptionPlans() = map { productDetails -> subscriptionMapper.mapFromProductDetails(productDetails, removeWinbackOffer = false) }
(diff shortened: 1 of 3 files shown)


mebarbosa (approved):
It worked as expected in my tests!

PR #3650, opened 2025-02-24
Show redesign tooltip

MiSikora:
## Description

This PR adds a tooltip informing users about the header redesign. I removed the code for the previous refresh tooltip, as managing both would introduce unnecessary complexity, and users should have already seen the previous tooltip.

There is a small difference between the implementation and the designs: the overlay does not cover the navigation bar. I didn't want to introduce further changes to how we handle tooltips.

Designs: 3LuZimaaH5tGKxIkOpjYAZ-fi-1071_24625

## Testing Instructions

1. Install the app from the `task/podcast-tooltip` branch.
2. Go to the podcast view.
3. Ensure no tooltip is visible.
4. Uninstall the app.
5. Install the app from commit `4f34c35d7edf246631a02d92bee4dc049640dd2f`.
6. Open the app.
7. Update the app to a version from the `task/podcast-tooltip` branch.
8. Go to the podcast view.
9. You should see a tooltip.
10. Dismiss the tooltip.
11. Close the podcast view.
12. Open any podcast view.
13. You should not see the tooltip.
14. Close the app.
15. Open any podcast view.
16. You should not see the tooltip.

## Screenshots or Screencast 

Screenshot_20250224-155145

## Checklist
- [ ] ~If this is a user-facing change, I have added an entry in CHANGELOG.md~
- [x] Ensure the linter passes (`./gradlew spotlessApply` to automatically apply formatting/linting)
- [ ] ~I have considered whether it makes sense to add tests for my changes~
- [x] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml`
- [x] Any jetpack compose components I added or changed are covered by compose previews
- [ ] ~I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.~
 
#### I have tested any UI changes...

- [x] with different themes
- [x] with a landscape orientation
- [x] with the device set to have a large display and font size
- [x] for accessibility with TalkBack

PR #3815, opened 2025-03-25
[Empty State] Update for Filters

mebarbosa:
## Description
- Adds the tooltip for when have empty state in Filters screen
- There is a small difference between the implementation and the designs: the overlay does not cover the navigation bar. Same thing we did for #3650
- Designs: 2MA6Bz3Vi36BOhuR9pitCU-fi-50_7599

## Testing Instructions
1. Fresh install 
2. Go to Filters
3. You should see the tooltip
4. Tap outside of tooltip or close button
5. Tooltip should be closed
6. Close the app and open it again
7. Go to Filters
8. Ensure you don't see the tooltip anymore

## Checklist
 
- [x] with different themes

### DIFF
--- a/modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/filters/FiltersFragment.kt
+++ b/modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/filters/FiltersFragment.kt
@@ -6,6 +6,26 @@ import android.view.MenuItem
 import android.view.View
 import android.view.ViewGroup
 import androidx.appcompat.widget.Toolbar
+import androidx.compose.foundation.background
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.widthIn
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableFloatStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.layout.Layout
+import androidx.compose.ui.layout.MeasurePolicy
+import androidx.compose.ui.platform.LocalConfiguration
+import androidx.compose.ui.semantics.invisibleToUser
+import androidx.compose.ui.semantics.semantics
+import androidx.compose.ui.unit.dp
+import androidx.core.view.isVisible
 import androidx.core.view.updatePadding
 import androidx.fragment.app.viewModels
 import androidx.lifecycle.Lifecycle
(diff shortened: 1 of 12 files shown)


modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/filters/FiltersTooltip.kt
```diff
@@ -0,0 +1,136 @@
...
+                    painter = painterResource(IR.drawable.ic_close),
+                    contentDescription = stringResource(LR.string.close),
+                    tint = MaterialTheme.theme.colors.primaryText01,
+                )
+            }
+        }
+    }
+}
+
+private object UpwardTriangleShape : Shape {
```
geekygecko, line 102: The same logic here is already in the code https://github.com/Automattic/pocket-casts-android/blob/a78a52816d157ddf9f0ecfcc947a628c10b28021/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/podcast/PodcastHeaderTooltip.kt#L90 Would it be better to create a reusable tooltip component rather than repeating the code?

modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/filters/FiltersFragment.kt
```diff
@@ -183,6 +214,70 @@ class FiltersFragment :
...
+                            )
+                            .semantics {
+                                invisibleToUser()
+                            },
+                    ) {
+                        Layout(
+                            content = {
+                                FiltersTooltip(
+                                    onClickClose = { closeTooltip() },
+                                    modifier = Modifier.widthIn(max = 326.dp).padding(horizontal = 16.dp),
```
geekygecko, line 253: | Design | Implementation |
| --- | --- |
| Screenshot 2025-03-27 at 3 24 56 pm | Screenshot 2025-03-27 at 3 24 56 pm | 

I'm not sure if it's only on my test device but is it possible to line up the arrow closer to the plus icon? Should the description text also be slightly smaller?

modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/filters/FiltersTooltip.kt
mebarbosa: This shape i a triangle pointing down and I need to draw one pointing up

modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/filters/FiltersFragment.kt
mebarbosa: It's resolved now!

modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/filters/FiltersTooltip.kt
mebarbosa: Ashita and I were talking about this, and we now have a tooltip component. I've updated this PR to include it

geekygecko (approved):

PR #3867, opened 2025-04-09
Refresh existing episodes content metadata

MiSikora:
## Description

We have a report of a bug where a podcast doesn't update to display a video after an episode was changed from audio to video. See: p1744110093014929-slack-C02A333D8LQ.

I don't fully understand the current update logic:

```kotlin
if (existingPodcast.isSubscribed && isFundingUrlSame && (podcastResponse.wasCached() || podcastResponse.notModified())) {
    LogBuffer.i(LogBuffer.TAG_BACKGROUND_TASKS, "Refreshing podcast ${existingPodcast.uuid} not required as cached")
    return
}
```

This code doesn't seem correct. It prevents episodes from being updated when the podcast itself hasn’t changed—which might not be the case in most scenarios, hence the bug. The `wasCached()` and `notModified()` makes sense but I don't get the point of checking it fully. It is done after the most expensive work, which is executing a network request. So there's not much resource saving.

Also, this comment below seems misleading, and the code beneath it may be incorrect:

```kotlin
// as new episodes are added a task is run to get the content type and file size from the server file as it is more reliable
if (existingEpisode.fileType.isNullOrBlank()) {
    existingEpisode.fileType = newEpisode.fileType
}
if (existingEpisode.sizeInBytes <= 0) {
    existingEpisode.sizeInBytes = newEpisode.sizeInBytes
}
if (existingEpisode.duration <= 0) {
    existingEpisode.duration = newEpisode.duration
}
```

The comment refers to new episodes, but the code updates an existing one. It also mentions getting data from the server file, but as far as I know, `UpdateEpisodeDetailsTask` is executed only for new episodes, so this wouldn't apply here. Moreover, `UpdateEpisodeDetailsTask` already checks if `fileType` is blank before updating it, so it wouldn't overwrite a pre-existing fileType.

All of this seems off, so I removed the code that was preventing updates. That said, we have multiple different places where episodes are updated, and I’m not entirely sure this is the right place for the fix. Still, based on my analysis, this change appears to be correct—and in any case, it fixes the bug.

## Testing Instructions

1. Install the app from the `main` branch.
2. Follow [this podcast](https://pca.st/episode/eb68f9ae-1ab8-46ac-b236-3f761288fa12).
3. Play an episode from it.
4. Notice that there is video content.
5. Clear the Up Next queue.
6. Execute this query in the App Inspection `UPDATE podcast_episodes SET file_type = 'audio/mp3'`.
7. Open podcast details.
8. Plan an episode.
9. Notice there is no video content.
10. Close the app.
11. Update the app with the one from the `fix/episode-content-update` branch.
12. Open the app.
13. Go to the podcast details.
14. Wait a couple of seconds to sync with the server response.
15. Play an episode.
16. Notice that now it displays the video content.

## Checklist


CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 7.87
 -----
-
+*   Bug Fixes
+    *   Fix episode sometimes not updating their metadata.
+        ([#3789](https://github.com/Automattic/pocket-casts-android/pull/3789))
```
MiSikora, line 5:
```suggestion
        ([#3867](https://github.com/Automattic/pocket-casts-android/pull/3867))
```

PR #3872, opened 2025-04-10
[Push Notifications Revamp] Support deeplink for Staff Picks

mebarbosa:
## Description
- Supports the deep link for opening discover stuff picks. This will be used for the new Stuff Picks notification that will be implemented in the next PRs: `pktc://discover/staffpicks`
- See document with deeplinks: p1743685755349989/1743684484.596599-slack-C08KX131YRJ

## Testing Instructions
Since we don't have the notifications implemented, we can test this in command line:

1. Install the app
2. In Android Studio terminal, run: `adb shell am start -a android.intent.action.VIEW -d "pktc://discover/staffpicks"`
3. Ensure Stuff Picks screen ✅

## Checklist

### DIFF
--- a/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
+++ b/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
@@ -82,7 +82,10 @@ import au.com.shiftyjelly.pocketcasts.deeplink.StaffPicksDeepLink
 import au.com.shiftyjelly.pocketcasts.deeplink.ThemesDeepLink
 import au.com.shiftyjelly.pocketcasts.deeplink.UpgradeAccountDeepLink
 import au.com.shiftyjelly.pocketcasts.deeplink.UpsellDeepLink
+import au.com.shiftyjelly.pocketcasts.discover.util.DiscoverDeepLinkManager
+import au.com.shiftyjelly.pocketcasts.discover.util.DiscoverDeepLinkManager.Companion.STAFF_PICKS_LIST_ID
 import au.com.shiftyjelly.pocketcasts.discover.view.DiscoverFragment
+import au.com.shiftyjelly.pocketcasts.discover.view.PodcastListFragment
 import au.com.shiftyjelly.pocketcasts.endofyear.StoriesActivity
 import au.com.shiftyjelly.pocketcasts.endofyear.StoriesActivity.StoriesSource
 import au.com.shiftyjelly.pocketcasts.endofyear.ui.EndOfYearLaunchBottomSheet
@@ -245,6 +248,8 @@ class MainActivity :
     @Inject lateinit var watchSync: WatchSync
 
     @Inject lateinit var notificationHelper: NotificationHelper
+
+    @Inject lateinit var discoverDeepLinkManager: DiscoverDeepLinkManager
 
     @Inject @ApplicationScope
     lateinit var applicationScope: CoroutineScope
@@ -1354,8 +1359,15 @@ class MainActivity :
                     openImport()
                 }
                 is StaffPicksDeepLink -> {
-                    closeToRoot()
-                    openTab(VR.id.navigation_discover)
+                    val podcastListFragment = supportFragmentManager.fragments.find { it is PodcastListFragment } as? PodcastListFragment
+                    if (podcastListFragment?.listUuid != STAFF_PICKS_LIST_ID) {
+                        openTab(VR.id.navigation_discover)
+                        lifecycleScope.launch {
+                            val staffPicksList = discoverDeepLinkManager.getDiscoverList(STAFF_PICKS_LIST_ID, resources) ?: return@launch
+                            val fragment = PodcastListFragment.newInstance(staffPicksList)
+                            addFragment(fragment)
+                        }
+                    }
                 }
                 is PlayFromSearchDeepLink -> {
                     playbackManager.mediaSessionManager.playFromSearchExternal(deepLink.query)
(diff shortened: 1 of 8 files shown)


app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
```diff
@@ -1352,6 +1353,15 @@ class MainActivity :
                 is ImportDeepLink -> {
                     openImport()
                 }
+                is StaffPicksDeepLink -> {
```
mebarbosa, line 1356: I am following the same approach that we use for opening a particular filter

modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/viewmodel/DiscoverViewModel.kt
```diff
@@ -71,6 +71,7 @@ class DiscoverViewModel @Inject constructor(
             categoryFeed = null,
             isLoading = true,
             isError = false,
+            openStaffPicks = false,
```
MiSikora, line 74: Navigation is a one-off event that is a side effect and mustn't be part of the state. Having it here can lead to many ambiguities when there are many such fields and more then 1 get set to `true`. Additionally, this makes state hard to manage as it requires setting values to `false` from the UI after navigation occurs. Please use a different solution to the navigation.

modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/view/DiscoverFragment.kt
```diff
@@ -241,6 +246,19 @@ class DiscoverFragment :
                         (activity as FragmentHostListener).addFragment(fragment)
                         fragment.listener = this@DiscoverFragment
                     }
+
+                    if (state.openStaffPicks) {
+                        val discoverRows = adapter?.currentList?.filterIsInstance<DiscoverRow>()
+                        val staffPicksRow: DiscoverRow? = discoverRows?.firstOrNull { it.listUuid == "staff-picks" }
+                        staffPicksRow?.let {
+                            view.post {
```
MiSikora, line 254: Why do you need to `post()` here?

MiSikora (changes requested):

mebarbosa:
Hey @MiSikora 
I've pushed another approach for this. You can check [aa65d05](https://github.com/Automattic/pocket-casts-android/pull/3872/commits/aa65d05635f12967a3f0a59d5d2621e1ed8e52bd)

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>DiscoverDeepLinkNavigation</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>

</p>

MiSikora (commented):
I found two issues:

1. When I navigate using `adb shell am start -a android.intent.action.VIEW -d "pktc://discover/staffpicks"` and then go to the previous screen I can no longer navigate using the same deep link.
2. When I navigate using a deep link, go back, and then change device's configuration the staff pick screen is added to the back stack and is added twice. See the video below:

https://github.com/user-attachments/assets/50ccc818-8420-4994-a967-15799f129ff4

mebarbosa:
@MiSikora I've fixed it [d67720f](https://github.com/Automattic/pocket-casts-android/pull/3872/commits/d67720f8b02307fd39f034c767e3e3a84b8f0b6a)

MiSikora (commented):
Thanks for the update! I noticed there’s still a navigation-related issue:

1. Deep link to Staff Picks.  
2. Change the configuration a couple of times.  
3. The screen stacks on top of itself multiple times, requiring multiple back presses to exit.

Functionally, this solution doesn’t differ much from the previous one. The navigation destination is still managed as top-level state. The main change is that it’s now decoupled from the UI state of the Discover page.

However, the core approach to navigation still raises some concerns. Since the fragment stack is already handled internally by the `FragmentManager`, managing it externally can lead to subtle and hard-to-diagnose issues like the one here.

A more robust solution might be to lift the Discover data (or at least the Staff Picks part of it out of the `DiscoverViewModel`. That way, the flow could look like this:

1. Receive a deep link to Staff Picks.  
2. Check if the Staff Picks screen is already visible.  
3. If not, retrieve data from the Staff Picks data source.  
4. Navigate to the Staff Picks screen.

This approach eliminates the need for a separate state holder just for navigation, and it promotes a single source of truth for the Staff Picks data, Making it more reusable and easier to reason about.

Even in the current PR, navigating to Staff Picks requires passing through both `DiscoverFragment` and `DiscoverAdapter`, which adds a fair bit of coupling and makes the logic harder to follow.

That said, I understand the goal here may not be to overhaul the structure. While I still have reservations about the architectural principle being reused from the previous implementation, if you're able to resolve the bugs without adding too much complexity, I’m okay with moving forward as-is.

mebarbosa:
@MiSikora Thank you for your inputs. I refactor this and created the `DiscoverDeepLinkManager` that will fetch discover data and look for staff picks list. See the new approach [68a8c34](https://github.com/Automattic/pocket-casts-android/pull/3872/commits/68a8c34ed340d53be22d98744bc821190dc099a7)

MiSikora (approved):
Thanks for doing all the changes. I'd be great if we shared the data source between `DisoverViewModel` and this feature but it probably isn't necessary at the moment.

PR #3910, opened 2025-04-17
Fix issue with disappearing file image right after upload completes

### LINKED ISSUE #1631: Image artwork is missing on files immediately after upload
### Description

The image artwork for files temporarily disappears after uploading the file to the cloud.

### 1. Phone after upload
1. Add a new user episode file to your device with artwork
2. From the files screen (Profile tab → Files), observe the artwork is displayed by the episode
3. Upload the episode to the cloud
4. Wait for the upload to complete
5. Observe that the file you just uploaded is no longer showing the episode artwork
6. Back out of the Files screen
7. Reopen the files screen
8. Observe that the file that you just uploaded now has its artwork back.

https://github.com/Automattic/pocket-casts-android/assets/4656348/e0aa31db-a2fb-4865-a0d4-d7ad48be90ed

Samsung A13, Android 13, Pocket Casts 7.54-rc-3

sztomek:
## Description

As title says. 
During my investigation I discovered that the file upload process consits of multiple steps. To call out the most important ones:
- Upload the file itself
- Upload the image (when applicable). The image will be uploaded by using a URL that is unique to the previously uploaded file.
After the sequence completes, we sync files with the server. However at this point the server might not have `ServerFile.imageUrl` property. Image gets uploaded to an aws bucket, and I assume it requires some time until our backend gets notified about the image upload completion. We then overwrote our db record with the data we got from the server -- effectively erased the image from the records. If you refresh the files list after a few seconds, the server will send the image.
I modified the code and added a new argument to the `syncFiles` method that controls whether we should skip comparison on image-related properties when examining whether our local record needs to be updated or not.

Fixes #1631 

## Testing Instructions
1. Make sure you log in with a Plus account
2. Navigate to Profile
3. Select Files
4. Tap the + FAB
5. Browse a file and select an image to be uploaded
6. Tap Save
7. Tap the new file you've just created
8. On the bottom sheet, select Upload to cloud
9. Observe
- [ ] image remains the same after the upload completes
10. Wait for a few seconds (~5) and pull down to refresh file list
- [ ] image of the previously uploaded file flickers -- it indicates that the server now sends the `imageUrl` field that points to the aws file 

## Screenshots or Screencast 

https://github.com/user-attachments/assets/8a663b52-6de6-4f03-aabd-035bc6d67645

## Checklist
 
- [x] with different themes

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/UserEpisodeManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/UserEpisodeManager.kt
@@ -283,7 +283,7 @@ class UserEpisodeManagerImpl @Inject constructor(
         }
     }
 
-    override suspend fun syncFiles(playbackManager: PlaybackManager) {
+    private suspend fun syncFiles(playbackManager: PlaybackManager, syncArtworkChanges: Boolean) {
         val episodesToSync = userEpisodeDao.findUserEpisodesToSyncBlocking()
         if (episodesToSync.isNotEmpty()) {
             val response = withContext(Dispatchers.IO) {
@@ -337,7 +337,7 @@ class UserEpisodeManagerImpl @Inject constructor(
                     didChange = true
                 }
 
-                if (existingFile.hasCustomImage != it.hasCustomImage) {
+                if (syncArtworkChanges && existingFile.hasCustomImage != it.hasCustomImage) {
                     existingFile.hasCustomImage = it.hasCustomImage
                     didChange = true
                 }
@@ -347,7 +347,7 @@ class UserEpisodeManagerImpl @Inject constructor(
                     didChange = true
                 }
 
-                if (existingFile.artworkUrl != it.imageUrl) {
+                if (syncArtworkChanges && existingFile.artworkUrl != it.imageUrl) {
                     existingFile.artworkUrl = it.imageUrl
                     didChange = true
                 }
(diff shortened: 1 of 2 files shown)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/UserEpisodeManager.kt
```diff
@@ -283,7 +283,7 @@ class UserEpisodeManagerImpl @Inject constructor(
         }
     }
 
-    override suspend fun syncFiles(playbackManager: PlaybackManager) {
+    private suspend fun syncFiles(playbackManager: PlaybackManager, shouldIgnoreImageChanges: Boolean) {
```
MiSikora, line 286: I wonder if it wouldn't be better to use this boolean as "should sync" instead of "should ignore". The reason is that later in the body of the function it is always negated. So having a parameter like `syncArtworkChanges` simplifies it a bit.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/UserEpisodeManager.kt
```diff
@@ -412,6 +412,10 @@ class UserEpisodeManagerImpl @Inject constructor(
         }
     }
 
+    override suspend fun syncFiles(playbackManager: PlaybackManager) {
+        syncFiles(playbackManager, false)
```
MiSikora, line 416: Could you add a name to the boolean parameter for readability?

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/UserEpisodeManager.kt
sztomek: Good call, addressed it in this commit 3bdd16ba8368c14d15b0d3c8f9cec2f3f17b3173

PR #3920, opened 2025-04-22
Add data types to the payment module

MiSikora:
## Description

This PR adds data layer to the payment module and mapping from Google's Billing types to it. There are 2 main decisions that I made. If there's disagreement we can discuss them and work out whether they should be changed.

1. I map only subscription types and skip one-time purchases. We don't have any in the app and adding them would be represented better as a separate type with something like a `sealed interface` hierarchy for both subscriptions and one-time purchases.
2. Apart from monthly and yearly billing intervals, I added support for weekly ones as well While we don't have anything like that in the app I think it's important to keep. Our domain specific offers like Plus Monthly, Plus Yearly, etc., shouldn't affect the payment module too heavily. While I'm fine skipping the whole category of data like on-time purchases doing this for something like a billing period feels wrong. I think it is unlikely that we'll have one-time purchases but it's not unreasonable to imagine that our plans or some free trial periods might be based in weeks.

## Testing Instructions

There's nothing to test. Code review the API and consider whether it makes sense to you.

## Checklist

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -285,6 +285,7 @@ play-cast = "com.google.android.gms:play-services-cast-framework:22.0.0"
 play-review = "com.google.android.play:review-ktx:2.0.2"
 play-wearable = "com.google.android.gms:play-services-wearable:19.0.0"
 reorderable = "sh.calvin.reorderable:reorderable:2.4.3"
+robolectric = "org.robolectric:robolectric:4.14"
 security-lint = { module = "com.android.security.lint:lint", version.ref = "security-lint" }
 timber = "com.jakewharton.timber:timber:5.0.1"
 turbine = "app.cash.turbine:turbine:1.2.0"
--- a/modules/services/payment/src/main/kotlin/au/com/shiftyjelly/pocketcasts/payment/billing/ProductDetailsMapper.kt
+++ b/modules/services/payment/src/main/kotlin/au/com/shiftyjelly/pocketcasts/payment/billing/ProductDetailsMapper.kt
@@ -1,0 +1,186 @@
+package au.com.shiftyjelly.pocketcasts.payment.billing
+
+import au.com.shiftyjelly.pocketcasts.payment.BillingPeriod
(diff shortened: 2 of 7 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

gradle/libs.versions.toml
```diff
@@ -285,6 +285,7 @@ play-cast = "com.google.android.gms:play-services-cast-framework:22.0.0"
 play-review = "com.google.android.play:review-ktx:2.0.2"
 play-wearable = "com.google.android.gms:play-services-wearable:19.0.0"
 reorderable = "sh.calvin.reorderable:reorderable:2.4.3"
+robolectric = "org.robolectric:robolectric:4.14"
```
jangramukul, line 288: any reason behind using Robo?

MiSikora: `com.android.billingclient.api.ProductDetails` uses `android.text.TextUtils` internally.

modules/services/payment/src/main/kotlin/au/com/shiftyjelly/pocketcasts/payment/billing/ProductDetailsMapper.kt
```diff
@@ -0,0 +1,186 @@
...
+        if (designator != 'P') {
+            logWarning("Missing billing period duration designator", context)
+            return null
+        }
+        val valuePart = iso8601Duration.drop(1)
+
+        val rawCount = valuePart.takeWhile(Char::isDigit)
+        val count = rawCount.toIntOrNull()
+        if (count == null) {
+            logWarning("Invalid billing period interval count '$count'", context)
```
geekygecko, line 162:
```suggestion
            logWarning("Invalid billing period interval count '$rawCount'", context)
```

geekygecko (approved):
LGTM. I like the idea of asserting the logs.

PR #3922, opened 2025-04-23
Handles podcast subscription race condition

geekygecko:
## Description

While testing the Discover section I found an issue that if you tap on the subscribe button on the artwork and then quickly tap the artwork to open the podcast, it can open the podcast page with the podcast as not being followed. After looking into the issue it seems like in the `SubscribeManager` the podcast is being added as followed and unfollowed at the same time. 

I'm really not sure of the best approach to solve this issue. My solution does not download and add the podcast to the database on the podcast page if it is already being subscribed to. It will wait for a second and retry that five times. Please let me know if you have a better solution. 

## Testing Instructions

1. Fresh install the app
2. Go to the Discover tab
3. Scroll down to a podcast large list
4. Tap the subscribe button on the artwork and then quickly tap the artwork
5. ✅ Verify the podcast is marked as followed on the podcast page

## Video

Before

https://github.com/user-attachments/assets/84a47bfb-4c4a-4589-85f4-e463c164dd61

After

https://github.com/user-attachments/assets/eba94036-f5ae-4496-800f-9af7c33f12aa

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
     *   Add account encouragement flow
         ([#3909](https://github.com/Automattic/pocket-casts-android/pull/3909))
 *   Bug Fixes
+    *   Fix podcast follow in the Discover tab not always working
+        ([#3922](https://github.com/Automattic/pocket-casts-android/pull/3922))
     *   Fix issue with disappearing file image after upload completes
         ([#3910](https://github.com/Automattic/pocket-casts-android/pull/3910))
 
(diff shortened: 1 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

MiSikora (approved):
I’ve experienced this a couple of times in the past. I didn’t want to fix it because I believe proper synchronization would require extensive sync code analysis and potentially restructuring how we handle many things.

My only concern is that `findOrDownloadPodcastRxSingle()` is used in many places, and I’m unsure what the impact might be. Specifically, that it could delay any action using the function by 5 seconds if a podcast isn’t immediately available and is in the process of being followed.

I’m fine with merging the PR since I haven’t encountered any issues myself, but let’s consider the concern above before doing it.

geekygecko:
> My only concern is that `findOrDownloadPodcastRxSingle()` is used in many places

That's a good point. I have added a parameter so the change is localised to the podcast page. Hopefully this is good enough to fix the issue for now.

PR #3925, opened 2025-04-24
Add subscription types to the payment module

MiSikora:
## Description

This PR adds data models for Pocket Casts-specific subscription types. There are a few key assumptions:

1. All base plans (Plus Monthly, Plus Yearly, Patron Monthly, Patron Yearly) must be present.
2. A base plan has only a single pricing phase.

Our current code already somewhat assumes this, but it's not explicitly modeled in the types, which makes it harder to manage. The proposed approach makes these assumptions explicit.

I considered creating specific types for offer subscriptions. For example, we expect that Winback offers consist of a single discounted pricing phase followed by the base one. However, I don't believe that's the right approach. I think such specific handling should live in repositories or view models that deal with offer logic. This way, all subscription offers can be modeled using a single type, which helps reduce the API surface and minimizes code churn in the payment module.

One oddity you might notice is the `PaymentResult` type. It currently has only the `getOrNull()` function. This type will be expanded with functions like `map()`, `flatMap()`, etc., as the payment module evolves. For now, it's a simple container to provide a uniform API across payment-related methods. Ideally, I’d use [Arrow’s `Either` type](https://arrow-kt.io/learn/typed-errors/either-and-ior/), which is more generic. But I’m not sure if everyone is comfortable with adding it. If people are on board, we should probably introduce it in a separate step and then replace `PaymentResult`, which shouldn't be too difficult.

## Testing Instructions

Code review the API.

## Checklist

### DIFF
--- a/modules/services/payment/src/main/kotlin/au/com/shiftyjelly/pocketcasts/payment/Data.kt
+++ b/modules/services/payment/src/main/kotlin/au/com/shiftyjelly/pocketcasts/payment/Data.kt
@@ -65,3 +65,216 @@ data class BillingPeriod(
         data object Infinite : Cycle
     }
 }
+
+class SubscriptionPlans private constructor(
+    private val plans: Map<Key, PaymentResult<SubscriptionPlan>>,
+) {
+    fun getBasePlan(
+        tier: SubscriptionTier,
+        billingCycle: SubscriptionBillingCycle,
+    ): SubscriptionPlan.Base {
+        val key = Key(tier, billingCycle, offer = null)
+        // This is a safe cast because constructor is private and we validate data in the create function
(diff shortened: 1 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

sztomek (approved):
The code looks good. It took me a bit to understand the initialization and how the contracts are enforced, but it makes sense now.
The tests would notify us if a future change would "break" the assumption that the 
`create` function either yields a `Success` object with a fully initied (containing the must-have plans) `SubscriptionPlans` or a `Failure` object.

PR #3929, opened 2025-04-25
Attempt to fix NoSuchElementException crash

sztomek:
## Description
This PR attemts to fix the crash that a user has reported. See [slack](https://a8c.slack.com/archives/C02A333D8LQ/p1745423604825769)

I never managed to reproduce the issue on my end, and to be honest it made me scratch my head.
The reported stack trace looked like this:
```
java.util.NoSuchElementException: Cannot find value for key 87
	at androidx.collection.MutableLongIntMap.androidx.collection.LongIntMap.get(SourceFile:122)
                                          c0.MutableLongIntMap.get
	at au.com.shiftyjelly.pocketcasts.models.db.dao.UpNextDao$getUpNextBaseEpisodes$suspendImpl$$inlined$sortedBy$1.androidx.compose.foundation.text.selection.SelectionManager$getSelectionLayout-Wko1d7g$$inlined$compareBy$1.compare(SourceFile:45)
```

Checked our code and the `sortedBy` clause had this: `idToPosition[it.uuid]`. The brackets are translated to `operator fun Collection.get(key: K): V?` (notice that the return type is nullable)
`sortedBy` also accepts null values.
However the stackTrace was explicit on this:  `NoSuchElementException at androidx.collection.MutableLongIntMap.androidx.collection.LongIntMap.get`
I believe when we build a prod version of the app, the compiler optimises the code and the actual type of `idToPosition` map will be `MutableLongIntMap` since that's the most efficient collection to store this type of data.
But then the `get` method of this collection throws `NoSuchElementException` when there's no value for the given key [docs](https://developer.android.com/reference/kotlin/androidx/collection/LongIntMap#get(kotlin.Long))

Anyways, I changed the operator fun invocation to `Collections.getOrDefault(key: K, defaultValue: V?): V?` hoping that will behave as expected.

## Testing Instructions
1. Start playing some episodes, add some more to the up next queue
2. Kill and relaunch app
- [x] app doesn't crash at any point

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,14 @@
     *   Add account encouragement flow
         ([#3909](https://github.com/Automattic/pocket-casts-android/pull/3909))
 *   Bug Fixes
+    *   Fix crash when updating playlist
+        ([#3929](https://github.com/Automattic/pocket-casts-android/pull/3929))
     *   Fix podcast follow in the Discover tab not always working
         ([#3922](https://github.com/Automattic/pocket-casts-android/pull/3922))
-    *   Fix issue with disappearing file image after upload completes
-        ([#3910](https://github.com/Automattic/pocket-casts-android/pull/3910))
     *   Fix missing fullscreen player controls in landscape mode
         ([#3917](https://github.com/Automattic/pocket-casts-android/pull/3917))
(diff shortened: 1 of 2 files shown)


geekygecko (approved):
Thanks for looking into this crash.

PR #3941, opened 2025-04-28
Add purchase APIs to the payment module

MiSikora:
## Description

Continuation of Billing APIs improvements. This PR adds purchase related types and integrations. I also added some improvements to the `PaymentResult` type. Failure type now has enumerated code to improve decision making and `recover()` method to map from failures to other results.

## Testing Instructions

Code review is enough as there's nothing to test yet.

## Checklist

### DIFF
--- a/modules/services/payment/src/main/kotlin/au/com/shiftyjelly/pocketcasts/payment/PaymentResult.kt
+++ b/modules/services/payment/src/main/kotlin/au/com/shiftyjelly/pocketcasts/payment/PaymentResult.kt
@@ -1,30 +1,56 @@
 package au.com.shiftyjelly.pocketcasts.payment
 
 sealed interface PaymentResult<out T : Any> {
-    data class Success<out T : Any>(val value: T) : PaymentResult<T>
+    data class Success<out T : Any>(
+        val value: T,
+    ) : PaymentResult<T>
 
-    data class Failure(val message: String) : PaymentResult<Nothing>
+    data class Failure constructor(
+        val code: PaymentResultCode,
+        val message: String,
+    ) : PaymentResult<Nothing>
 }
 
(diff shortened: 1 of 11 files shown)


modules/services/payment/src/main/kotlin/au/com/shiftyjelly/pocketcasts/payment/PaymentResult.kt
```diff
@@ -1,30 +1,56 @@
...
+    data class Unknown(val code: Int) : PaymentResultCode
+}
+
+inline fun <T : Any, R : Any> PaymentResult<T>.map(block: (T) -> R) = when (this) {
     is PaymentResult.Success -> PaymentResult.Success(block(value))
     is PaymentResult.Failure -> this
 }
 
-fun <T : Any, R : Any> PaymentResult<T>.flatMap(block: (T) -> PaymentResult<R>) = when (this) {
+inline fun <T : Any, R : Any> PaymentResult<T>.flatMap(block: (T) -> PaymentResult<R>) = when (this) {
```
sztomek, line 35: ❓ why did you need to make this and the following methods `inline`? I know what `inline` does but doesn't see the benefits nor the necessity of it

sztomek (approved):
Code looks good, left a question but otherwise it's good to go.

modules/services/payment/src/main/kotlin/au/com/shiftyjelly/pocketcasts/payment/PaymentResult.kt
MiSikora: To be able to use `suspend` functions in lambdas.

sztomek: aaaah thanks, that wasn't clear at first glance

PR #3947, opened 2025-04-30
TalkBack improvements

### LINKED ISSUE #3931: Accessibility: Talkback issues
### Description

A number of issues encountered while using Talkback that we should fix:

- [x] When trying to activate the CTA, I can only do it by double tapping in the button but outside of the text "Start Free Trial". Can we make the whole button (text included) actionable?
- [x] The order of the elements being dictated doesn't seem to be right, it starts by reading the Disclaimer. I would use the top to bottom order seen in the screen (start by the heading "Everything you love...", then the segmented control Yearly vs Monthly, etc.)
- [x] This also makes it so the Disclaimer is skipped when swiping from the last item in the features list (goes directly to the CTA). Let's please change that so it goes to the Disclaimer.
- [x] The Patron card is not selectable. Can we add an area to the bit of the card that is visible to make it actionable? I would dictate it after the Disclaimer and before the CTA. As an alternative, we can also make the "pager dots" actionable if this is tricky or complicated. I would dictate either as "Change plan".

_No response_

### Did you search for existing list?

- [x] I have searched for existing issues.

sztomek:
## Description
This PR attempts to address the issues listed in #3931 
- vertical focus order is mostly correct, you can vertically traverse the main screen elements - title, period selector, subscription tier selector, cta.
- title is now focusable
- CTA focus issues have been adressed

~~However there are still 2 major issues open:~~
~~- Feature card items cannot be selected via keyboard arrows~~
~~- Dislaimer text (with the PP and TNC links) is not selectable~~

Fixes #3931 

## Testing Instructions
1. Open the upgrade screen (Profile -> Settings -> Upgrade to Plus)
2. Enable TalkBack (System Settings -> Accessbility -> TalkBack)
3. Start navigating up and down using the keyboard 

## Screenshots or Screencast 

https://github.com/user-attachments/assets/e2ca384a-934c-45c1-be9b-258ef1b57512

## Checklist
~~- [ ] I have considered whether it makes sense to add tests for my changes~~
~~- [ ] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml`~~
~~- [ ] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.~~
 
- [x] with different themes

### DIFF
--- a/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/components/ProductAmountVerticalText.kt
+++ b/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/components/ProductAmountVerticalText.kt
@@ -10,6 +10,8 @@ import androidx.compose.material.MaterialTheme
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.focus.FocusRequester
+import androidx.compose.ui.focus.focusRequester
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.text.style.TextDecoration
@@ -71,6 +73,7 @@ fun ProductAmountHorizontalText(
     verticalAlignment: Alignment.Vertical = Alignment.CenterVertically,
     secondaryTextColor: Color = MaterialTheme.theme.colors.primaryText02,
     isFocusable: Boolean = false,
+    focusRequester: FocusRequester? = null,
 ) {
     Row(
         modifier = modifier,
@@ -86,9 +89,15 @@ fun ProductAmountHorizontalText(
                     } else {
                         MaterialTheme.theme.colors.primaryText01
                     },
-                modifier = if (isFocusable) {
-                    Modifier.focusable()
-                } else Modifier,
+                modifier = if (focusRequester != null) {
+                    Modifier.focusRequester(focusRequester)
+                } else {
+                    Modifier
+                }.then(
+                    if (isFocusable) {
+                        Modifier.focusable()
+                    } else Modifier
+                ),
             )
         }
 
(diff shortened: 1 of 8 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/components/ProductAmountVerticalText.kt
```diff
@@ -69,6 +72,8 @@ fun ProductAmountHorizontalText(
     hasBackgroundAlwaysWhite: Boolean = false,
     verticalAlignment: Alignment.Vertical = Alignment.CenterVertically,
     secondaryTextColor: Color = MaterialTheme.theme.colors.primaryText02,
+    isFocusable: Boolean = false,
+    focusRequester: FocusRequester? = null,
```
MiSikora, line 76: I might be missing something but why there is `focusRequester` parameter? From what I checked it is always `null`.

modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/components/HorizontalPagerWrapper.kt
```diff
@@ -47,7 +61,11 @@ fun HorizontalPagerWrapper(
         }
     }
 
-    var pagerHeight by remember { mutableStateOf(0) }
+    val coroutineScope = rememberCoroutineScope()
+    val focusRequesters = remember { List(pageCount) { FocusRequester() } }
+    val subContentFocusRequesters = remember { List(pageCount) { FocusRequester() } }
```
MiSikora, line 66: `pageCount` must be used as a `remember` key parameter. Otherwise the list won't update with any changes and we might get `IndexOutOfBoundsException`.

MiSikora (changes requested):

modules/services/compose/src/main/java/au/com/shiftyjelly/pocketcasts/compose/components/HorizontalPagerWrapper.kt
sztomek: Another great catch, thanks!
Addressed in cad8866531897a9a126c6617e2d55125e68ff9e1

PR #3961, opened 2025-05-06
Simplify payment client testing

MiSikora:
## Description

This simplifies the testing logic of `PaymentClient`. Previously, it was designed to emulate real behavior: we had to launch the purchase flow asynchronously, and while the task was running, we had to emit a set of events like receiving or acknowledging manually. I realized that the previous approach isn't what we really wanted as it would put too much burden on higher-level tests without any real benefit.

This change automates those emissions. While we lose some flexibility in testing, it significantly simplifies the process. We no longer need to spin up a background job to monitor purchases or explicitly emit events. If we want to simulate a failure, we can simply set the appropriate `PaymentResultCode` properties before launching the purchase. By default, without setting anything, we now get a happy-path scenario.

In the future, if we want to have more control, we could change the implementation from a constant response to a queue of responses that gets depleted 1 by 1.

## Testing Instructions

Verify the code.

## Checklist

### DIFF
--- a/modules/services/payment/src/main/kotlin/au/com/shiftyjelly/pocketcasts/payment/FakePaymentDataSource.kt
+++ b/modules/services/payment/src/main/kotlin/au/com/shiftyjelly/pocketcasts/payment/FakePaymentDataSource.kt
@@ -36,7 +36,7 @@ class FakePaymentDataSource : PaymentDataSource {
         return if (loadedProductsResultCode == PaymentResultCode.Ok) {
             PaymentResult.Success(loadedProducts)
         } else {
-            PaymentResult.Failure(loadedProductsResultCode, "Error message")
+            PaymentResult.Failure(loadedProductsResultCode, "Load products error")
         }
     }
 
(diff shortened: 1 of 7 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

modules/services/payment/src/main/kotlin/au/com/shiftyjelly/pocketcasts/payment/FakePaymentDataSource.kt
```diff
@@ -10,48 +10,65 @@ import com.android.billingclient.api.QueryPurchaseHistoryParams
...
+            val billingResult = PaymentResult.Success(Unit)
+            val purchaseResult = if (purchasedProductsResultCode == PaymentResultCode.Ok) {
+                PaymentResult.Success(purchasedProducts)
+            } else {
+                PaymentResult.Failure(purchasedProductsResultCode, "Error message")
+            }
+            _purchaseResults.emit(purchaseResult)
+            billingResult
+        } else {
+            PaymentResult.Failure(billingFlowResultCode, "Error message")
```
sztomek, line 62: ⛏️ I know it's just the fake data source, but maybe we could add a more describing error message to distinguish this case from the other `Failure` case above

sztomek (approved):
Nice job 👍

PR #4002, opened 2025-05-14
[Tooling] Skip jobs if there are no relevant code changes

iangmaia:
Internal reference: paaHJt-7Ss-p2

The main goal of this PR is to optimize the wait time for PRs with minor, non-code related changes, specially the ones related to the release process.

I've updated the original script to cover build, validation and localization jobs:
- `should-skip-job.sh --job-type [build|lint]`: skips if the changes are limited to documentation, tooling, metadata, etc
- `should-skip-job.sh --job-type validation`: same as above, but also skip on localization changes. The idea is that it can be valuable to run things as a Prototype build when localization changes

## Testing
I've created the PR https://github.com/Automattic/pocket-casts-android/pull/4009 to test a few scenarios skipping jobs:
- App version bump: https://buildkite.com/automattic/pocket-casts-android/builds/11842
- ++ Localization update: https://buildkite.com/automattic/pocket-casts-android/builds/11843
- ++ Metadata update: https://buildkite.com/automattic/pocket-casts-android/builds/11844

### DIFF
--- a/.buildkite/commands/assemble-release-apk.sh
+++ b/.buildkite/commands/assemble-release-apk.sh
@@ -1,5 +1,9 @@
 #!/bin/bash -eu
 
+if .buildkite/commands/should-skip-job.sh --job-type build; then
+  exit 0
+fi
+
 "$(dirname "${BASH_SOURCE[0]}")/restore-cache.sh"
 
 echo "--- :rubygems: Setting up Gems"
--- a/.buildkite/commands/assemble-release-apk.sh
+++ b/.buildkite/commands/assemble-release-apk.sh
@@ -1,6 +1,6 @@
 #!/bin/bash -eu
 
-if .buildkite/commands/should-skip-job.sh --job-type build; then
+if "$(dirname "${BASH_SOURCE[0]}")/should-skip-job.sh" --job-type build; then
   exit 0
 fi
 
--- a/.buildkite/commands/should-skip-job.sh
+++ b/.buildkite/commands/should-skip-job.sh
@@ -37,9 +37,9 @@ fi
 # Function to display skip message and create annotation
 show_skip_message() {
   local job_type=$1
-  local message="Skipping ${BUILDKITE_LABEL:-Job} - no relevant files changed"
+  local message="Skipped ${BUILDKITE_LABEL:-Job} - no relevant files changed"
   local context="skip-$(echo "${BUILDKITE_LABEL:-$job_type}" | sed -E -e 's/[^[:alnum:]]+/-/g' | tr A-Z a-z)"
-  
+
   echo "$message" | buildkite-agent annotate --style "info" --context "$context"
   echo "$message"
 }
--- a/.buildkite/commands/should-skip-job.sh
+++ b/.buildkite/commands/should-skip-job.sh
@@ -1,0 +1,73 @@
+#!/bin/bash -eu
+
+# Usage: should-skip-job.sh --job-type [validation|build|lint]
+# --job-type validation: For jobs like unit/instrumented tests, manifest validation…
+#     Skip when changes are limited to documentation, tooling, non-code files, and localization files
+# --job-type lint: For jobs doing linting, especially ones that might include linting of localization files
+#     Skip when changes are limited to documentation, tooling, and non-code files.
+#     Does not skip if localization files changed (so that linter can run to report translation inconsistencies)
+# --job-type build: For jobs building an apk binary, e.g. Assemble, Prototype Builds…
+#     Skip when changes are limited to documentation, tooling, and non-code files
+#
+# Return codes:
+# 0 - Job should be skipped (script also handles displaying the message and annotation)
+# 1 - Job should not be skipped
+# 15 - Error in script parameters
+
+COMMON_PATTERNS=(
+  "*.md"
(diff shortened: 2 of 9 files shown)


.buildkite/pipeline.yml
```diff
@@ -25,7 +25,7 @@ steps:
         agents:
           queue: "linter"
 
-      - label: 'Lint'
+      - label: 'Android Lint'
```
mokagio, line 28: Will this require an update of the checks rules?

.buildkite/commands/assemble-release-apk.sh
```diff
@@ -1,5 +1,9 @@
 #!/bin/bash -eu
 
+if .buildkite/commands/should-skip-job.sh --job-type build; then
```
mokagio, line 3: What do you think of using this:

```suggestion
if "$(dirname "${BASH_SOURCE[0]}")/should-skip-job.sh --job-type build; then
```

I like it because it makes the path relative. I doubt we'll ever want to move the files from where they are, but I like that we _could_, seems like a good practice to have.

.buildkite/commands/should-skip-job.sh
```diff
@@ -0,0 +1,73 @@
...
+fi
+
+# Function to display skip message and create annotation
+show_skip_message() {
+  local job_type=$1
+  local message="Skipping ${BUILDKITE_LABEL:-Job} - no relevant files changed"
+  local context="skip-$(echo "${BUILDKITE_LABEL:-$job_type}" | sed -E -e 's/[^[:alnum:]]+/-/g' | tr A-Z a-z)"
+  
+  echo "$message" | buildkite-agent annotate --style "info" --context "$context"
+  echo "$message"
```
mokagio, line 44: This could be ~~simplified~~ condensed as

```suggestion
  echo "$message" | tee >(buildkite-agent annotate --style "info" --context "$context")
```

I tried a version locally, albeit not with the Buildkite agent.

trietbui85:
It seems https://appsinfrap2.wordpress.com/2025/02/12/introducing-pr_changed_files-command-to-optimize-ci-usage-in-pull-requests is private link since I cannot access it 😭

.buildkite/pipeline.yml
iangmaia: Hmm I missed this was a required check, nice catch.
I think I'll revert the change, as this isn't directly related to this PR -- calling it `Android Lint` makes it clearer but at the same time this is implicit (and we don't have more linting at the moment like `ktlint`).

iangmaia: Updated on a3abc53ab90d6f4e617353c982c493cd6331d8ba.

.buildkite/commands/should-skip-job.sh
```diff
@@ -0,0 +1,73 @@
...
+    # We'll let the job run (won't skip) if PR includes changes in localization files though
+    PATTERNS=("${COMMON_PATTERNS[@]}")
+    if pr_changed_files --all-match "${PATTERNS[@]}"; then
+      show_skip_message "$job_type"
+      exit 0
+    fi
+    exit 1
+    ;;
+  *)
+    echo "Error: Job type must be either 'validation', 'build', or 'lint'"
```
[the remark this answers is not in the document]
iangmaia: Nice suggestion -- updated on 82103cae29c8bb154be5e2c2bce51463de94e3e4.

.buildkite/commands/should-skip-job.sh
AliSoftware: @mokagio [Great minds](https://github.com/woocommerce/woocommerce-ios/pull/15086#discussion_r1953189035)… though as you can see from that comment of mine, we found out that this `tee >(buildkite-agent annotate)` didn't work after all.

[I'm still unclear why it doesn't work](https://github.com/woocommerce/woocommerce-ios/pull/15086#discussion_r1953238296), while a test like ` echo 'Hello' | tee >(say)` works as expected (both print _and_ say Hello), so I'm guessing it's something a bit specific to how `buildkite-agent annotate` reads the `stdin` that works when passed through `|` but not through `tee`?
Though a test like `echo '{"steps":[{"command":"echo Hello"}]}' | tee >(buildkite-agent pipeline upload --format yaml --dry-run)` also works as expected (prints both the input JSON to stdout and reads and print the YAML interpretation of that pipeline)… So still not sure what makes `buildkite-agent annotate` not work the same as all the others…

AliSoftware: Ah, with the help of ChatGPT, I [finally found a solution that works for this](https://buildkite.com/automattic/pocket-casts-android/builds/11876/summary/annotations?sid=0196da73-0b70-4403-b5c1-dd7f8ab03a6b#0196da73-0b99-4329-81fe-c02e9e0156fe/130-132) after all!

_(See internal discussion in p1747421916865929-slack-C02KLTL3MKM for more details)_

geekygecko:
Should we be adding extra files we don't think effect the tests to the `COMMON_PATTERNS` array? Such as 
```
  "docs/**"
  "scripts/**"
```

mokagio:
@geekygecko makes sense to me. Feel free to update the script to include them. 👍

PR #4007, opened 2025-05-15
Convert `NotifiactionsFragment` to Compose-base components

sztomek:
## Description
As titlte suggests.
I took the existing `NotificationsSettingsFragment` that inhereted from `PreferencesFragment` and blew it up.
Now we have a `NotificationsSettingsViewModel` that holds the state and interacts with `NotificationsPreferencesRepository`. The latter one is responsible to provide data and persist changes.
I used the existing comopsables from other settings screens, so I heavily rely on `SettingRow`. I also salvaged and "composified" pieces from the old fragment to reduce impact.

## Testing Instructions
1. Launch app and navigate to Settings -> Notifications
2. Play with various settings
3. Navigate away
4. Come back and see if the previous settings were properly preserved
5. You may want to apply this patch [disable_OS_version_condition_on_preferences.patch](https://github.com/user-attachments/files/20226614/disable_OS_version_condition_on_preferences.patch)

## Screenshots or Screencast 

https://github.com/user-attachments/assets/b45208e7-f86c-4d4e-8dd1-d8e307b722f0

## Checklist
- ~[ ] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml`~
- ~[ ] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.~
 
this PR does not contain UI changes, ignore these items -->
- [x] with different themes
- ~[ ] with a landscape orientation~
- [x] with the device set to have a large display and font sizex


sztomek:
I'm going to add tests in a follow-up PR!

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Errors
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please add tests for class <code>NotificationsSettingsViewModel</code> (or add <code>unit-tests-exemption</code> label to ignore this).</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please add tests for class <code>NotificationsPreferencesRepositoryImpl</code> (or add <code>unit-tests-exemption</code> label to ignore this).</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/notifications/NotificationsSettingsFragment.kt
```diff
@@ -0,0 +1,58 @@
...
+                },
+            )
+        }
+    }
+
+    override fun podcastSelectFragmentSelectionChanged(newSelection: List<String>) {
+        viewModel.onSelectedPodcastsChanged(newSelection)
+    }
+
+    override fun podcastSelectFragmentGetCurrentSelection() = runCatching { runBlocking { viewModel.getSelectedPodcastIds() } }.getOrDefault(emptyList())
```
MiSikora, line 57: If `getSelectedPodcastIds()` can throw, would it make sense to reflect that in the method name—something like `getSelectedPodcastIdsOrThrow()`? That said, I’m not sure this should be public at all. Might be cleaner if `getSelectedPodcastIds()` handled the exception internally instead if throwing is expected at all.

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/notifications/NotificationsSettingsScreen.kt
```diff
@@ -0,0 +1,143 @@
...
+import au.com.shiftyjelly.pocketcasts.settings.notifications.model.NotificationPreference
+import au.com.shiftyjelly.pocketcasts.settings.notifications.model.NotificationPreferences
+import au.com.shiftyjelly.pocketcasts.views.fragments.PodcastSelectFragment
+import au.com.shiftyjelly.pocketcasts.views.fragments.PodcastSelectFragmentSource
+
+@Composable
+internal fun NotificationsSettingsScreen(
+    onBackPressed: () -> Unit,
+    bottomInset: Dp,
+    viewModel: NotificationsSettingsViewModel,
```
MiSikora, line 43: Would it be possible to remove the `ViewModel` dependency and instead pass in a state object along with the necessary callbacks? That way we can create Compose previews, since the current setup makes that impossible.

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/notifications/NotificationsSettingsScreen.kt
```diff
@@ -0,0 +1,143 @@
...
+    onAdvancedSettingsClicked: () -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    val state: NotificationsSettingsViewModel.State by viewModel.state.collectAsState()
+
+    // Unfortunately, PodcastSelectFragment was meant to be used from another fragment that defines a toolbar.
+    // This flag is used to determine whether we should render the podcast selector inside this composable and change toolbar title and override back navigation when necessary.
+    var isShowingPodcastSelector by remember { mutableStateOf(false) }
+
+    val launcher = rememberLauncherForActivityResult(ActivityResultContracts.StartActivityForResult()) { result ->
```
MiSikora, line 53: I think we should avoid using `rememberLauncherForActivityResult` directly inside our UI composables. It would simplify this component if it just accepted a lambda like `onEpisodeRingtoneClick: () -> Unit`, and handled all the intent and result logic outside the UI layer.

Just to clarify—I’m not against using `rememberLauncherForActivityResult` itself, but if we do use it, it should be instantiated in the fragment’s `contentWithoutConsumedInsets` function, not within the UI composable.

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/notifications/NotificationsSettingsScreen.kt
```diff
@@ -0,0 +1,143 @@
...
+                    isShowingPodcastSelector = false
+                } else {
+                    onBackPressed()
+                }
+            },
+        )
+
+        Box {
+            LazyColumn(
+                modifier
```
MiSikora, line 88: The `Modifier` passed into the composable should be applied to the top-level component—in this case, the `Column`. That helps ensure any styling or layout changes are properly respected from the outside.

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/notifications/NotificationsSettingsScreen.kt
```diff
@@ -0,0 +1,143 @@
...
+                        )
+                    }
+                }
+            }
+            if (isShowingPodcastSelector) {
+                AndroidFragment(
+                    modifier = Modifier.fillMaxSize(),
+                    clazz = PodcastSelectFragment::class.java,
+                    arguments = PodcastSelectFragment.createArgs(source = PodcastSelectFragmentSource.NOTIFICATIONS),
+                )
```
MiSikora, line 139: Using APIs like this makes it impossible to generate previews, and `isShowingPodcastSelector` adds extra state that we probably don’t need to manage.

What do you think about handling this similarly to the approach I suggested with the ViewModel?

```kotlin
NotificationsSettingsScreen(
  onChoosePodcastsClick = {
    val fragment = PodcastSelectFragment.newInstance(source = PodcastSelectFragmentSource.NOTIFICATIONS)
    childFragmentManager.beginTransaction()
      .add(fragment, "podcastSelect")
      .commit()
  }
)
```

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/notifications/NotificationsSettingsViewModel.kt
```diff
@@ -0,0 +1,141 @@
...
+    internal fun loadPreferences() = viewModelScope.launch {
+        _state.update { it.copy(categories = preferenceRepository.getPreferenceCategories()) }
+    }
+
+    internal fun onPreferenceChanged(preference: NotificationPreference<*>) {
+        viewModelScope.launch {
+            when (preference.preference) {
+                NotificationPreferences.NEW_EPISODES_NOTIFY_ME -> {
+                    preferenceRepository.setPreference(preference)
+                    (preference.value as? Boolean)?.let {
```
MiSikora, line 47: The fact that the generic parameter isn't available and we have to cast it later is a bit of a code smell. If we already need to know the specific type to cast to, it kind of defeats the purpose of using generics in the first place.

Also, since actions only run when the cast is successful, this could introduce subtle bugs down the line—where things silently stop working and it’s not immediately clear why.

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/notifications/NotificationsSettingsViewModel.kt
```diff
@@ -0,0 +1,141 @@
...
+                        )
+                    }
+                }
+
+                NotificationPreferences.NEW_EPISODES_RINGTONE -> {
+                    preferenceRepository.setPreference(preference)
+                    analyticsTracker.track(AnalyticsEvent.SETTINGS_NOTIFICATIONS_SOUND_CHANGED)
+                }
+
+                else -> Unit
```
MiSikora, line 114: Using `else` in `when` statements that cover enum values should be avoided. If we add a new preference later, it's really easy to miss it because the `else` will silently handle it. It’d be better to list all cases explicitly—even if some preferences do nothing—so it’s clearer and safer to maintain.

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/notifications/NotificationsSettingsViewModel.kt
```diff
@@ -0,0 +1,141 @@
...
+                podcastManager.updateShowNotifications(it.uuid, newSelection.contains(it.uuid))
+            }
+            loadPreferences()
+        }
+    }
+
+    internal suspend fun getSelectedPodcastIds(): List<String> = viewModelScope.async(Dispatchers.Default) {
+        val uuids = podcastManager.findSubscribedBlocking().filter { it.isShowNotifications }.map { it.uuid }
+        uuids
+    }.await()
```
MiSikora, line 136: This could be simplified a bit:

```kotlin
internal suspend fun getSelectedPodcastIds(): List<String> = withContext(Dispatchers.Default) {
  podcastManager.findSubscribedBlocking()
    .filter { it.isShowNotifications }
    .map { it.uuid }
}
```

That said, a couple of things stand out—why use the `Default` dispatcher here instead of `IO`? And can we avoid using any dispatcher at all by switching to `findSubscribedNoOrder()` instead of `findSubscribedBlocking()`?

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/notifications/components/NotificationPreferenceCategory.kt
```diff
@@ -0,0 +1,223 @@
...
+import com.afollestad.materialdialogs.list.listItemsMultiChoice
+import com.afollestad.materialdialogs.list.updateListItemsMultiChoice
+import okhttp3.internal.toImmutableList
+
+@Suppress("UNCHECKED_CAST")
+@Composable
+internal fun NotificationPreferenceCategory(
+    categoryTitle: String,
+    items: List<NotificationPreference<*>>,
+    onItemClicked: (NotificationPreference<*>) -> Unit,
```
MiSikora, line 40: This is the same issue I mentioned earlier regarding generics. Suppressing the warning here isn’t really safe—we can’t guarantee that the values will actually be of these types. Relying on that assumption could easily lead to runtime issues.

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/notifications/components/NotificationPreferenceCategory.kt
```diff
@@ -0,0 +1,223 @@
...
+                                },
+                            )
+                        }
+
+                        else -> Unit
+                    }
+                }
+
+                is MultiSelectPreference<*> -> {
+                    val activity = LocalContext.current
```
MiSikora, line 116: I think there’s a typo here—shouldn’t this be `context`?

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/notifications/data/NotificationsPreferencesRepositoryImpl.kt
```diff
@@ -0,0 +1,211 @@
...
+import au.com.shiftyjelly.pocketcasts.localization.R as LR
+
+internal class NotificationsPreferencesRepositoryImpl @Inject constructor(
+    @IoDispatcher private val dispatcher: CoroutineDispatcher,
+    @ApplicationContext private val context: Context,
+    private val settings: Settings,
+    private val podcastManager: PodcastManager,
+) : NotificationsPreferenceRepository {
+
+    override suspend fun getPreferenceCategories(): List<NotificationPreferenceCategory> = withContext(dispatcher) {
```
MiSikora, line 34: Is there a specific reason this needs to run on the `IO` dispatcher?

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/notifications/data/NotificationsPreferencesRepositoryImpl.kt
```diff
@@ -0,0 +1,211 @@
...
+        when {
+            notificationCount == 0 -> context.getString(LR.string.settings_podcasts_selected_zero)
+            notificationCount == 1 -> context.getString(LR.string.settings_podcasts_selected_one)
+            notificationCount >= podcastCount -> context.getString(LR.string.settings_podcasts_selected_all)
+            else -> context.getString(
+                LR.string.settings_podcasts_selected_x,
+                notificationCount,
+            )
+        }
+    }.take(1).firstOrNull()
```
MiSikora, line 137: No need to call `take(1)` here. Also, it would be more appropriate to use `first()` instead of `firstOrNull()`, since we’re fetching a list and it’s never going to be `null`. If there are no subscribed podcasts, the list will just be empty.

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/notifications/data/NotificationsPreferencesRepositoryImpl.kt
```diff
@@ -0,0 +1,211 @@
...
+                settings.newEpisodeNotificationActions.set(value = setting, updateModifiedAt = true)
+            }
+
+            NotificationPreferences.NEW_EPISODES_RINGTONE -> {
+                val setting = expectTypedValue<String>(preference)
+                settings.notificationSound.set(value = NotificationSound(setting, context), updateModifiedAt = false)
+            }
+
+            // Add handler for new preferences here
+            else -> Timber.d("Unhandled preference received in setPreference: ${preference.preference}")
```
MiSikora, line 201: Same issue here as mentioned earlier with using `else` in a `when` statement. It makes it easy to miss new cases if we add more values later. Better to be explicit about all possible options.

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/notifications/di/NotificationSettingsModule.kt
```diff
@@ -0,0 +1,18 @@
...
+import dagger.hilt.InstallIn
+import dagger.hilt.components.SingletonComponent
+import javax.inject.Singleton
+
+@InstallIn(SingletonComponent::class)
+@Module
+abstract class NotificationSettingsModule {
+
+    @Binds
+    @Singleton
```
MiSikora, line 16: Is `@Singleton` needed here?

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/notifications/model/NotificationPreference.kt
```diff
@@ -0,0 +1,75 @@
+package au.com.shiftyjelly.pocketcasts.settings.notifications.model
+
+/**
+ * This enum defines the preferences we keep record on
+ */
+internal enum class NotificationPreferences {
```
MiSikora, line 6: The naming feels a bit off here. Something like `NotificationPreferenceType` or `NotificationTypePreferenceId` might be a bit clearer and more descriptive.

MiSikora (changes requested):
Thanks for working on this—there’s a lot of good stuff here. I especially like how the logic has been pulled out of the fragment and the UI migrated to Compose. That’s a solid step forward.

That said, I do think there are some design choices that could make future maintenance harder than necessary.

For one, I don’t think generics are being used effectively here. As it stands, they don’t provide much benefit—there’s still a lot of casting and type checking happening, which kind of defeats the purpose.

I’m also unsure about the design of `NotificationsPreferencesRepositoryImpl`, particularly the use of `Context`. Right now, it’s being used to read resources needed for the UI, but that responsibility really belongs in the UI layer.

Using `Context` like this is fine:

```kotlin
data class DataType(
  val value: Int,
) {
  fun title(context: Context) = when (value) {
    0 -> LR.string.zero_title
    1 -> LR.string.one_title
    else -> LR.string.default_title
  }
}
```

In this case, resource access remains dynamic, and we don’t need to mock or use real `Context` instances for repository tests.

Another concern is the use of `Build.VERSION` within the repository—it introduces a platform dependency that also makes testing trickier. It would be better to either abstract that out or move the responsibility to a higher layer. Personally, I think abstracting it would lead to a cleaner and more testable design.

---

I know the PR is already quite big, but I wouldn’t mind if you included overall tests and previews for `NotificationsSettingsScreen`. It’s something we should aim to include in the same PR, rather than leaving it for follow-ups.

Whether something is easy or hard to test is often a good signal of how robust the design is. And if we can’t create previews, that’s usually a sign that there’s too much going on inside the composable.

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/notifications/NotificationsSettingsScreen.kt
sztomek: I tried to do what you've suggested, but the podcast selector fragment wasn't displayed for some reason, here's my patch for evidence (based on my commit 5b3b6e0eb290bb3860d749be0ac0a65af67076ed)
[hoist_podcast_selector_to_upper_levels.patch](https://github.com/user-attachments/files/20243344/hoist_podcast_selector_to_upper_levels.patch)

we also need to change the title of the toolbar to "Select Podcasts", because `PodcastSelectFragment` doesn't display a toolbar.
of course, i could also lift the toolbar up to the fragment's code but that would only leave `NotificationsSettingsScreen` with this:
```
LazyColumn(modifier) { state.categories.forEach { item {  NotificationPreferenceCategory(...)  } } } 
```
the only difference it would make is to push the responsibility up a level

I might have misunderstood something, please let me know if i did so

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/notifications/NotificationsSettingsFragment.kt
sztomek: thanks addressed in 5b3b6e0eb290bb3860d749be0ac0a65af67076ed

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/notifications/NotificationsSettingsScreen.kt
sztomek: yup, addressed in 5b3b6e0eb290bb3860d749be0ac0a65af67076ed

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/notifications/data/NotificationsPreferencesRepositoryImpl.kt
sztomek: I believe it's a habit of mine to use IO for background operations, I assume you'd advocate for `Default` dispatcher.
If my memories are correct, IO is pooled while Default is a single threaded executor, I don't really see performance issues or benefits of using the other. Please let me know if I'm mistaken.

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/notifications/di/NotificationSettingsModule.kt
sztomek: I wanted to make sure that there will only a single instance available for the repository.
If I'm not mistaken, `SingletonComponent` is app-scoped (and there's only 1 app instance alive at any time), but the providers would create new instances of the provided repository without the `@Singleton` annotation.
LMK if i'm wrong.

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/notifications/NotificationsSettingsScreen.kt
MiSikora: You're right. My sample is wrong. Sorry about that!

The fragment would have to added to some fragment container. Which isn't a big issue because we can create one in the `NotificationsSettingsFragment` either through XML or in code. Though XML is probably better. Something like:

```xml
<?xml version="1.0" encoding="utf-8"?>
<FrameLayout
    xmlns:android="http://schemas.android.com/apk/res/android"
    android:layout_width="match_parent"
    android:layout_height="match_parent">

    <androidx.fragment.app.FragmentContainerView
        android:id="@+id/fragment_container"
        android:layout_width="match_parent"
        android:layout_height="match_parent" />

</FrameLayout>
```

```kotlin
override fun onCreateView(
    inflater: LayoutInflater,
    container: ViewGroup?,
    savedInstanceState: Bundle?,
): View {
    val binding = TODO()
    val composeView = contentWithoutConsumedInsets { TODO() }
    binding.root.addView(composeView, 0)
    return binding.rootView
}
```

I know that this might look like it's daunting or annoying but It think it is really important to avoid using Android-specific APIs in our Compose UI and keep them pure.

There are also 2 more issues with the current redesign:
- Hardware back navigation no longer works. Before when I used back button while the select podcasts screen was displayed it went back to the notifications screen. Now it goes back to the settings.
- Navigation state isn't saved. When I open podcasts selection and then change the configuration podcasts selection is no longer shown.

The original implementation didn't from these issues. I haven't tested it but it's possible that my proposal suffers from the same issues but I suspect it might be easier to solve it there.

MiSikora: There is `showToolbar` property for the input arguments of the `PodcastSelectFragment`.

sztomek:
Thanks for your review @MiSikora 
I believe I addressed most of your concerns regarding the PR.
I have some notes however.

> For one, I don’t think generics are being used effectively here. As it stands, they don’t provide much benefit—there’s still a lot of casting and type checking happening, which kind of defeats the purpose.

Went for the sealed class apprach where all preference item has a dedicated data structure. The benefit is that we no longer need to cast values, but it bloated the view level a bit.
I believe the initial approach was easier to extend, I could live with those safe casts.

>  I’m also unsure about the design of NotificationsPreferencesRepositoryImpl, particularly the use of Context. Right now, it’s being used to read resources needed for the UI, but that responsibility really belongs in the UI layer.

Ummm I don't think we can get rid of the context in the repository impl, because there is `NotificationSound` that  takes context as ctor arg, or the way we concatenate complex labels (like the Actions item summary). I've introduced `TextResource` for simpler use cases, but I don't see it as a huge win.

About the podcast selector fragment: I don't really see where your suggestion is going, as I explained [here](https://github.com/Automattic/pocket-casts-android/pull/4007#discussion_r2092902311)
I could follow your advise and make `NotificationsSettingsFragment` the host of the `PodcastSelectorFragment` but that would change the current experience since the podcast fragment doesn't have a toolbar. So I would need to also push up the toolbar and the title change + back nav logic to the settings fragment. 

The navigation and config change issues you've found are valid problems, thanks for flagging them. This commit fixed them both 8fc9eb17fba10616be1ef24d9776bdfb24a1e1fb

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/notifications/NotificationsSettingsScreen.kt
sztomek: Damn, I missed that. I just used this fragment as it was previously used on the old notification settings screen 🫠

MiSikora (commented):
Thanks for all the follow-up work — I think this is much better overall. I still see one major issue with the fragment and back-handling logic inside the `NotificationsSettingsScreen`, but I saw your comment about the upcoming change. Most of other things I commented on are rather small things.

Regarding the other points: if you disagree with any of my suggestions and feel that another approach is better (and it's not just an either/or situation), please feel free to push back and let’s discuss it. No need to just apply requested changes if you think there’s a better way. Not saying that happened here — just putting it out there in case.

>I believe the initial approach was easier to extend, I could live with those safe casts.

Since you've changed that part, I won’t go into detail unless you want to discuss it (here or elsewhere). That said, I think relying on casting like that is very risky to use from a maintainability perspective. It's especially error-prone during refactors.

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/notifications/NotificationsSettingsViewModel.kt
```diff
@@ -0,0 +1,138 @@
...
+            }
+            loadPreferences()
+        }
+    }
+
+    internal suspend fun getSelectedPodcastIds(): List<String> = runCatching {
+        viewModelScope.async(Dispatchers.Default) {
+            val uuids = podcastManager.findSubscribedBlocking().filter { it.isShowNotifications }.map { it.uuid }
+            uuids
+        }.await()
```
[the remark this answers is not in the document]
sztomek: You're right, thanks
I was shortsighted, just copied the existing code from the old fragment [source](https://github.com/Automattic/pocket-casts-android/blob/755c688acfeeefe3666beafea79a0f4249bfc3b2/modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/NotificationsSettingsFragment.kt#L192)
But didn't realise it could be rephrased like this.

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/notifications/data/NotificationsPreferencesRepositoryImpl.kt
```diff
@@ -0,0 +1,185 @@
...
+                settings.newEpisodeNotificationActions.set(value = preference.value, updateModifiedAt = true)
+            }
+
+            is NotificationPreferenceType.NotificationSoundPreference -> {
+                settings.notificationSound.set(value = preference.notificationSound, updateModifiedAt = false)
+            }
+
+            is NotificationPreferenceType.AdvancedSettings,
+            is NotificationPreferenceType.NotifyOnThesePodcasts,
+            -> Unit // these are not on us to set
```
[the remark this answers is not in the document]
sztomek: I'd just rather remove the comment. 
I could write way longer explanation like `AdvancedSettings` is handled by the OS and `NotifyOnThesePodcast` settings will be set on the `SelectPodcastsFragment`, but that would couple code with comments and that could add maintenance cost -- you'll need to keep the comments up-to-date if/when the code changes.

wdyt?

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/notifications/components/NotificationPreferenceCategory.kt
```diff
@@ -0,0 +1,206 @@
...
+                        onSave = { value ->
+                            onItemClicked(
+                                item.copy(value = value),
+                            )
+                        },
+                    )
+                }
+
+                is NotificationPreferenceType.NotificationActions -> {
+                    val activity = LocalContext.current
```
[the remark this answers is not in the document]
sztomek: oops, i missed this somehow, thanks!

modules/features/settings/src/test/java/au/com/shiftyjelly/pocketcasts/settings/settings/NotificationsSettingsViewModelTest.kt
```diff
@@ -0,0 +1,139 @@
...
+            onBlocking { getPreferenceCategories() }.doReturn(categories)
+        }
+    }
+
+    @Test
+    fun `GIVEN repository holds valid data WHEN vm is created THEN state updates`() = runTest {
+        val viewModel = createViewModel()
+
+        assertEquals(viewModel.state.value.categories, categories)
+        verify(repository).getPreferenceCategories()
```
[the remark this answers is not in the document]
sztomek: hmmm you're right. i should have only focused on the outcome, not the intrinsics in this case. thanks

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/notifications/data/NotificationsPreferencesRepositoryImpl.kt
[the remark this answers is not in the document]
MiSikora: Yup, I think it's better to remove it.

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/notifications/NotificationsSettingsViewModel.kt
```diff
@@ -0,0 +1,138 @@
...
+    internal fun onSelectedPodcastsChanged(newSelection: List<String>) {
+        viewModelScope.launch(Dispatchers.IO) {
+            podcastManager.findSubscribedBlocking().forEach {
+                podcastManager.updateShowNotifications(it.uuid, newSelection.contains(it.uuid))
+            }
+            loadPreferences()
+        }
+    }
+
+    internal suspend fun getSelectedPodcastIds(): List<String> = runCatching {
```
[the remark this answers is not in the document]
MiSikora: `await()` won't throw on its own — it can only throw if the underlying function does.  Now, whether it's worth catching database exceptions is the real question.

IMO yes, and I agree with you that it should be represented using some kind of `Result<T>` type.  Right now, throughout most of the app, we assume that database operations won't fail.  While I don't think we should generally be doing that, I also think it's better not to wrap errors ad-hoc at the call sites.  Error handling should be defined at the repository level.

We shouldn't have to remember whether a function can throw — that should be expressed through the type system and enforced. Doing things like:

```kotlin
runCatching {
  someOperation()
}.getOrElse { TODO() }
```

is bad for us when the database actually fails — we'll lose the stack trace and won't be able to advise users on what went wrong. Having a repository return `Result<T>` allows us to handle logging and any other necessary operations in one place.

So to answer your question if I think we should drop `runCatching`. At this exact moment — it honestly doesn't really matter 😅. If the database fails for some reason, we'll know about it long before users ever reach this screen, since the app will crash.

modules/features/settings/src/test/java/au/com/shiftyjelly/pocketcasts/settings/settings/NotificationsSettingsViewModelTest.kt
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    @Test
+    fun `GIVEN preference WHEN preference changes THEN repository called to persist change`() = runTest {
+        val viewModel = createViewModel()
+        val changedPreference = notifyMe.copy(isEnabled = !notifyMe.isEnabled)
+        viewModel.onPreferenceChanged(changedPreference)
+
+        verify(repository).setPreference(changedPreference)
+        verify(analytics).track(any(), any())
```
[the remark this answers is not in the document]
sztomek: done 002c352fa8fa738407cf4e1b9a7f63f2c93f2d38

sztomek:
thanks @MiSikora for the review!
the theme issue you've noticed stems from the fact that previously these radiogroup-style dialogs were handled by the `PreferenceFragment` iteself, to be more exact, it was `ListPreference` that presented the dialog with the options. 
Now that I moved everything to Compose, `SettingRadioDialogRow` represents these preferences and it takes care of rendering the dialog with the given theme colors applied.

PR #4046, opened 2025-05-26
Add loading state to podcasts page

MiSikora:
## Description

During rewrite of podcasts tab to coroutines the loading state got removed. This resulted in podcasts page showing empty state when podcasts were loading. Normally this isn't an issue but for a large amount of podcasts it can take some time before they are available. This PR fixes the problem by tracking if podcasts are loaded or not.

## Testing Instructions

1. Apply [this patch](https://github.com/user-attachments/files/20444002/empty-state.patch).
```
curl -L https://github.com/user-attachments/files/20444002/empty-state.patch | git apply
```
2. Open the app.
3. Go to the podcasts tab.
4. Close the app.
5. Open the app.
6. For the first 5 seconds you should not see the empty state message.
7. After 5 seconds you should see either the message or the podcasts you're subscribed to. 

## Checklist


MiSikora:
If we do I think we shouldn't be doing it in this beta fix. We didn't have loading state before so I'd treat it as a regular change.

PR #4047, opened 2025-05-26
[Notifications Revamp] Offers notifications

sztomek:
## Description
This PR adds the next step of the Notifications Revamp project - introducing offers notifications. The render condition is that the user doesn't have a subscription, and the display frequency is 2 weeks.
Upon tapping the notification, we should land on the subscription page, this is done via the `UpsellDeeplink`.

## Testing Instructions
1. apply patch and build app [offers_notifications.patch](https://github.com/user-attachments/files/20444118/offers_notifications.patch)
2. Log in with an account that doesn't have a subscription
3. Settings -> Notifications
- [ ] offers category is rendered and the switch is off by default
4. Enable offers
5. Wait for 10 seconds
- [ ] Notification appears
6. Tap on the notification
- [ ] Upsell page opens
7. Now log off and log in with an account that is subscribed already
8. Turn off offers then repeat steps 4-5
- [ ] no notification since you're already subscribed

## Screenshots or Screencast 

https://github.com/user-attachments/assets/efb2963c-a499-4fad-a268-880510080782

## Checklist
- ~[ ] If this is a user-facing change, I have added an entry in CHANGELOG.md~
- ~[ ] Any jetpack compose components I added or changed are covered by compose previews~
 
- [x] with different themes


MiSikora (commented):
Looks good but I think an unwanted change from the patch got submitted.

PR #4050, opened 2025-05-28
Migrate Discover small list to Compose

MiSikora:
## Description

This PR migrates the small list row in the Discover page to Compose. While that wasn't the primary goal of this PR, I included it as part of my exploration into the feasibility of a broader Discover migration.

The main objective was to increase the touch target size of the subscribe button, which is currently only `24 dp`. In this PR, I’ve increased it to the standard `48 dp`.

| Before | After |
|--------|-------|
| markup_1000002590 | markup_1000002591 |

The Compose migration — arguably the most interesting part — came with some challenges. During testing, I discovered that the pager state is shared across multiple view holders. For example, scrolling to page 3 of the "Trending" list would also scroll the "Popular in United States" list to page 3. To address this, I now save the state in the `DiscoverAdapter` and restore it during data binding. Previously, we handled this using a `Parcelable` to store the `LayoutManager`'s state. Now, I simply store an `Int` representing the currently displayed page. This workaround is necessary due to our integration with `RecyclerView`.

An alternative approach could be using a distinct `ViewHolder` for each separate section but that would sacrifice recycling benefits and introduce complexity given how the Discover section is structured.

## Testing Instructions

Test the Discover section, focusing particularly on small lists like "Trending" or "Popular in `your_country`". There should be no meaningful regressions. The only issue I observed is some initial jank when scrolling to a new page for the first time. This only occurs in debug builds and is due to Compose’s debug tooling.

## Screenshots or Screencast 

| Before | After |
| - | - |
| b | Screenshot_20250528-094350 |

## Checklist
 
- [x] with different themes

### DIFF
--- a/modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/compose/SmallListRow.kt
+++ b/modules/features/discover/src/main/java/au/com/shiftyjelly/pocketcasts/discover/compose/SmallListRow.kt
@@ -86,8 +86,8 @@ private fun SmallListRowPreview(
         SmallListRow(
             pagerState = pagerState,
             podcasts = podcasts,
-            onClickPodcast = { },
-            onClickSubscribe = { },
+            onClickPodcast = {},
+            onClickSubscribe = {},
         )
     }
 }


CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 *   Updates
     *   Upgrade to the latest version of Jetpack Compose
         ([#4049](https://github.com/Automattic/pocket-casts-android/pull/4049))
+    *   Improve accessibility of the follow button in some of the Discover sections
+        ([#4049](https://github.com/Automattic/pocket-casts-android/pull/4049))
```
MiSikora, line 7:
```suggestion
        ([#4050](https://github.com/Automattic/pocket-casts-android/pull/4050))
```

MiSikora:
> I've noticed a small glitch: when tapping the row, ripple is not displayed immediately but only upon return.

Interesting. That would on Compose. The ripple still starts on tap. You can test it by holding the row and noticing that the ripple is there. It appears that Compose pauses the ripple effect and resumes it when we come to the fragment. Or restarts it because it wasn't fully consumed and the ripple is remembered in the view holder.

Issue #3663, opened 2025-02-27
Drag and Drop: Podcasts swap around


joashrajin:
### Description

When you try to drag and drop podcasts on the podcasts page, they don't always stick to the spot you want, and the results are quite unpredicable

### Step-by-step reproduction instructions

1. Ensure you are subscribed to many podcasts (maybe over 50)
2. Set the layout to Grid View
3. Drag a podcast from somewhere in the middle of the page to a few tiles up
4. Notice that it either works or reverts back to where it was

### Screenshots or screen recording

https://github.com/user-attachments/assets/a6adc3b0-ead7-4c52-a390-afa996d1598e

### Did you search for existing bug reports?

- [x] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

Samsung S21 FE, v7.81, Android 14

sharpestmarble:
This looks like the app is just waiting for the server to update, then the server sends back down an updated order. This happens with my (large) up-next queue. I have found I can get around this by disconnecting from both wifi and mobile data, then make as many updates as I want, then re-connect to the Internet

PR #4053, opened 2025-05-28
Fix podcasts drag & drop

MiSikora:
## Description

During coroutines migration I didn't account for the drag & drop state which is kept separately from the UI state. This fixes it by applying the previous solution of updating the `adapterState` with each new UI state. Ideally this would be a part of the UI state but I don't want to make too many changes in a beta fix.

## Testing Instructions

1. Follow some podcasts.
2. Go to the podcasts tab.
3. Ordering podcasts by drag & drop should work.

## Checklist


sztomek:
Might be related to https://github.com/Automattic/pocket-casts-android/issues/3663 ?

MiSikora:
The root cause of this issue? No. But the fact that we keep `adapterState` separately from the UI state yes. I'm working on it right now. I though I might make it a part of this fix but it's a bigger issue so I'll work on it separately.

PR #4055, opened 2025-05-28
[Notification Revamp] Notification popup

sztomek:
## Description
This PR adds the bottom sheet to explain why users should enable notifications.
The sheet is displayed on the Podcasts screen when the notifications are disabled. 
It offers an option to dismiss the sheet -- in that case we won't show it again.
Otherwise we will try to request permissions if `shouldShowPermissionRationale=true`. As a last resort, we will open the system settings.
Other than that, I also added a check on the Podcast screen to check if notifications are enabled when the user taps the bell icon on a podcast that they already follow. 

See: hVHQ8RgGQf1afBWybZExD2-fi-360_5234

## Testing Instructions
1. Install app
2. Don't allow notifications
3. Select Podcasts tab
- [ ] popup is displayed
4. Tap 'Allow'
- [ ] system permission dialog displayed
5. Tap deny
6. Navigate away from the Podcasts tab then go back
- [ ] popup is displayed
7. Tap Allow again
8. Takes you to the system settings
9. Repeat step 6
10. Dismiss the popup with the X icon
11. Repeat step 6
- [ ] popup is not shown again
12. Now open a podcast and subscribe to it
13. Tap the bell icon
- [ ] snack is displayed

## Screenshots or Screencast 
Screenshot 2025-05-28 at 19 21 57
Screenshot 2025-05-28 at 19 21 57

## Checklist
- ~[ ] If this is a user-facing change, I have added an entry in CHANGELOG.md~
 
- [x] with different themes

### DIFF
--- a/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/notifications/EnableNotificationsPromptFragment.kt
+++ b/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/notifications/EnableNotificationsPromptFragment.kt
@@ -19,8 +19,6 @@ import androidx.compose.foundation.layout.windowInsetsPadding
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
-import androidx.compose.ui.input.nestedscroll.nestedScroll
-import androidx.compose.ui.platform.rememberNestedScrollInteropConnection
 import androidx.compose.ui.unit.dp
 import androidx.fragment.compose.content
 import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsEvent
@@ -63,11 +61,6 @@ internal class EnableNotificationsPromptFragment : BaseDialogFragment() {
 
     private var wasDismissedViaCloseButton = false
 
-    override fun onCreate(savedInstanceState: Bundle?) {
-        super.onCreate(savedInstanceState)
-        isCancelable = false
-    }
-
     override fun onCreateView(
         inflater: LayoutInflater,
         container: ViewGroup?,
@@ -88,34 +81,30 @@ internal class EnableNotificationsPromptFragment : BaseDialogFragment() {
                 ),
         ) {
             AppThemeWithBackground(theme.activeTheme) {
-                Box(
+                EnableNotificationsPromptScreen(
                     modifier = Modifier
-                        .nestedScroll(rememberNestedScrollInteropConnection())
                         .fillMaxSize()
-                        .windowInsetsPadding(WindowInsets.safeDrawing.only(WindowInsetsSides.Horizontal)),
-                ) {
-                    EnableNotificationsPromptScreen(
-                        modifier = Modifier.padding(
+                        .padding(
                             vertical = 16.dp,
                             horizontal = 16.dp,
-                        ),
-                        onCtaClicked = {
-                            analyticsTracker.track(AnalyticsEvent.NOTIFICATIONS_PERMISSIONS_ALLOW_TAPPED)
-                            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
-                                if (shouldShowRequestPermissionRationale(Manifest.permission.POST_NOTIFICATIONS)) {
-                                    permissionRequester.launch(Manifest.permission.POST_NOTIFICATIONS)
-                                    analyticsTracker.track(AnalyticsEvent.NOTIFICATIONS_OPT_IN_SHOWN)
-                                } else {
-                                    notificationHelper.openNotificationSettings(requireActivity())
-                                }
+                        )
+                        .windowInsetsPadding(WindowInsets.safeDrawing.only(WindowInsetsSides.Horizontal)),
+                    onCtaClicked = {
+                        analyticsTracker.track(AnalyticsEvent.NOTIFICATIONS_PERMISSIONS_ALLOW_TAPPED)
+                        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+                            if (shouldShowRequestPermissionRationale(Manifest.permission.POST_NOTIFICATIONS)) {
+                                permissionRequester.launch(Manifest.permission.POST_NOTIFICATIONS)
+                                analyticsTracker.track(AnalyticsEvent.NOTIFICATIONS_OPT_IN_SHOWN)
+                            } else {
+                                notificationHelper.openNotificationSettings(requireActivity())
                             }
-                        },
-                        onDismissClicked = {
-                            wasDismissedViaCloseButton = true
-                            dismiss()
-                        },
-                    )
-                }
+                        }
+                    },
+                    onDismissClicked = {
+                        wasDismissedViaCloseButton = true
+                        dismiss()
+                    },
+                )
             }
         }
     }


modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/notifications/EnableNotificationsPromptScreen.kt
```diff
@@ -0,0 +1,95 @@
...
+        modifier = modifier,
+        horizontalAlignment = Alignment.CenterHorizontally,
+    ) {
+        IconButton(
+            onClick = onDismissClicked,
+            modifier = Modifier.align(Alignment.End),
+        ) {
+            Icon(
+                painter = painterResource(IR.drawable.ic_close),
+                contentDescription = "close icon",
```
MiSikora: Content descriptions should be localized.

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastViewModel.kt
```diff
@@ -103,6 +110,8 @@ class PodcastViewModel
 
     val groupedEpisodes: MutableLiveData<List<List<PodcastEpisode>>> = MutableLiveData()
     val signInState = userManager.getSignInState().toLiveData()
+    private val _showNotificationSnack = MutableStateFlow<SnackBarMessage?>(null)
```
MiSikora: Wouldn't it be more suitable to make it a `SharedFlow` instead of a `StateFlow`? Current setup requires initial artificial `null` state and filtering it out when collecting the stream.

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastViewModel.kt
```diff
@@ -332,10 +341,22 @@ class PodcastViewModel
         }
     }
 
-    fun showNotifications(podcastUuid: String, show: Boolean) {
+    fun showNotifications(podcastUuid: String, podcastTitle: String, show: Boolean) {
```
MiSikora: Since we pass in the uuid there shouldn't be a need to pass in the title. It can be accessed from the local storage.

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastViewModel.kt
```diff
@@ -717,6 +742,23 @@ class PodcastViewModel
...
     }
 
+    sealed interface SnackBarMessage {
+        val message: TextResource
+        val cta: TextResource?
+
+        data class ShowNotificationsDisabledMessage(
+            override val message: TextResource,
+            override val cta: TextResource,
+            private val uniqueUuid: UUID = UUID.randomUUID(),
```
MiSikora: The UUID isn't used anywhere.

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastViewModel.kt
```diff
@@ -717,6 +742,23 @@ class PodcastViewModel
         REFRESH_BUTTON("refresh_button"),
     }
 
+    sealed interface SnackBarMessage {
+        val message: TextResource
+        val cta: TextResource?
```
MiSikora: I don't really understand why this type needs to be a `sealed` type. Can't we have a simple class that accepts both CTA label and CTA action?

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastsViewModel.kt
```diff
@@ -62,10 +64,15 @@ class PodcastsViewModel
...
-                    SuggestedFoldersState.Available
+                if (notificationHelper.hasNotificationsPermission()) {
+                    _suggestedFoldersState.value = if (folders.isEmpty()) {
+                        SuggestedFoldersState.Empty
+                    } else {
+                        SuggestedFoldersState.Available
+                    }
+                    _notificationsState.value = NotificationsPromptState.AlreadyHasPermissions
+                } else if (!settings.notificationsPromptAcknowledged.value) {
+                    _notificationsState.value = NotificationsPromptState.ShowPrompt
```
MiSikora: Using suggested folders to determine notification behavior seems likely to result in confusing logic and potential bugs. Can these two concepts be separated? Even now, this causes a bug: if I never grant notification permission, the suggested folders dialog never appears. Neither automatically nor when tapping the folders button.

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastsViewModel.kt
```diff
@@ -273,6 +283,14 @@ class PodcastsViewModel
         folderUuidObservable.accept(Optional.ofNullable(folderUuid))
     }
 
+    fun checkNotificationPermission() {
+        if (notificationHelper.hasNotificationsPermission()) {
+            _notificationsState.value = NotificationsPromptState.AlreadyHasPermissions
+        } else if (settings.notificationsPromptAcknowledged.value) {
+            _notificationsState.value = null
+        }
+    }
```
MiSikora: This function is a bit misleading. It doesn't check the permission. It updates permission state.

modules/services/images/src/main/res/drawable-xxxhdpi/android_mockup.png
MiSikora, line 1: The added assets are really big in size. Can you convert them to the WebP format? I use this script locally.

```
#!/bin/bash
set -e

QUALITY=75

usage="Create a set of WebP images from a provided image. Scripts assumes that the image is in xxxhdpi resolution.
    -h  show this help text
    -i  source image file
    -q  compression factor for RGB channels between 0 and 100 (default is 75)"

# Assign arguments
options=':h:i:q:'
while getopts $options option; do
  case "$option" in
    h) echo "$usage"; exit 0;;
    i) IMAGE=$(realpath "$OPTARG");;
    q) QUALITY=$OPTARG;;
    :) printf "Missing argument for -%s\n\n" "$OPTARG" >&2; echo "$usage" >&2; exit 1;;
   \?) printf "Illegal option: -%s\n\n" "$OPTARG" >&2; echo "$usage" >&2; exit 1;;
  esac
done

if ! command -v magick &> /dev/null; then
  echo "ImageMagick doesn't seem to be installed (or not in PATH)."
  echo "Please install it using 'brew install imagemagick' and try again."
  exit 3
fi

# Require source image file
if [ ! -f "${IMAGE}" ] ; then
  echo "-i argument must be provided\n"
  echo "$usage" >&2; exit 1
fi

if ! ((QUALITY >= 0 && QUALITY <= 100)); then
  echo "-q argument '${QUALITY}' must be a number between 0 and 100\n"
  echo "$usage" >&2; exit 1
fi

mkdir -p "./res/drawable-xxxhdpi"
mkdir -p "./res/drawable-xxhdpi"
mkdir -p "./res/drawable-xhdpi"
mkdir -p "./res/drawable-hdpi"
mkdir -p "./res/drawable-mdpi"
mkdir -p "./res/drawable-ldpi"

FILENAME=$(basename "$IMAGE" | cut -d. -f1)

magick "${IMAGE}" -scale "100%" -quality "${QUALITY}" -define "webp:lossess=true" "./res/drawable-xxxhdpi/$FILENAME.webp"
magick "${IMAGE}" -scale "75%" -quality "${QUALITY}" -define "webp:lossess=true" "./res/drawable-xxhdpi/$FILENAME.webp"
magick "${IMAGE}" -scale "50%" -quality "${QUALITY}" -define "webp:lossess=true" "./res/drawable-xhdpi/$FILENAME.webp"
magick "${IMAGE}" -scale "37.5%" -quality "${QUALITY}" -define "webp:lossess=true" "./res/drawable-hdpi/$FILENAME.webp"
magick "${IMAGE}" -scale "25%" -quality "${QUALITY}" -define "webp:lossess=true" "./res/drawable-mdpi/$FILENAME.webp"
magick "${IMAGE}" -scale "18.75%" -quality "${QUALITY}" -define "webp:lossess=true" "./res/drawable-ldpi/$FILENAME.webp"
```

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastsViewModel.kt
```diff
@@ -363,6 +382,11 @@ class PodcastsViewModel
         data object Available : SuggestedFoldersState()
     }
 
+    sealed class NotificationsPromptState {
+        data object ShowPrompt : NotificationsPromptState()
+        data object AlreadyHasPermissions : NotificationsPromptState()
```
MiSikora: The whole state with its flow seems a bit strange. `AlreadyHasPermissions` type isn't used anywhere as a read value and we use `null` that means the same in the UI layer.

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastViewModel.kt
sztomek: Good callout, thanks!

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastViewModel.kt
sztomek: yeah, it's initial purpose was to guarantee uniqueness, but it won't be necessary with the sharedflow approach now.

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastViewModel.kt
sztomek: agreed, in this form it isn't really telling. 
made some changes and `cta` is now a member of `ShowNotificationsDisabledMessage`.
I'd still keep the sealed class structure because on `PodcastFragment` the original code that displayed a snackbar 
 (when subscribed for new new episodes) used a fixed duration of 3000 ms. Meanwhile the new snacks for disabled notifications would use a different (default) duration

modules/services/images/src/main/res/drawable-xxxhdpi/android_mockup.png
sztomek: nice, i love it. thanks!

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastsViewModel.kt
sztomek: Reworked definitions and flow to better reflect what I wanted to achieve. Thakns for flagging these problems 🙏

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

sztomek:
@MiSikora thanks for the review, you've made very nice points.
I think I've addressed all of them but the ones that require design input. Asked Dom to provide feedback for the landscape mode here: p1748880766784139-slack-C08KX131YRJ 
Please take another look when you can 🙏

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/viewmodel/PodcastViewModel.kt
```diff
@@ -335,7 +341,23 @@ class PodcastViewModel
     fun showNotifications(podcastUuid: String, show: Boolean) {
         analyticsTracker.track(AnalyticsEvent.PODCAST_SCREEN_NOTIFICATIONS_TAPPED, AnalyticsProp.notificationEnabled(show))
         viewModelScope.launch {
-            podcastManager.updateShowNotifications(podcastUuid, show)
+            if (!notificationHelper.hasNotificationsPermission()) {
+                _showNotificationSnack.tryEmit(
```
MiSikora: Since we are inside a `CoroutineScope` we should use `emit()` instead of `tryEmit()`. Otherwise values aren't delivered through a `SharedFlow` that has no buffer capacity. It is the case here, resulting in the snackbar not being displayed.

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/podcast/PodcastFragment.kt
```diff
@@ -917,6 +912,27 @@ class PodcastFragment : BaseFragment() {
                 }
             }
         }
+
+        viewLifecycleOwner.lifecycleScope.launch {
+            viewModel.showNotificationSnack.flowWithLifecycle(lifecycle)
+                .filterNotNull()
```
MiSikora: This shouldn't be needed any more.

MiSikora (approved):
Looks good. Just needs one fix to handling `MutableSharedFlow` before merging.

> but the ones that require design input

Yeah, sounds good. I think if we decide to keep the bottom sheet appearance instead of the full-screen one it is important to add the drag gesture. It feels very unnatural to me to see a bottom sheet page that I cannot dismiss by dragging it down.

PR #4056, opened 2025-05-29
Do not skip CI checks when only Gradle changes are present

wzieba:
We should run build checks in cases of Gradle wrapper or dependencies changes.


ParaskP7 (approved):
LGTM, thanks for dealing with this @wzieba ! 🙇

Cc @iangmaia

PR #4113, opened 2025-06-12
Add transcript entries sanitization

MiSikora:
## Description

This PR adds transcript sanitization logic. Currently, it addresses the following issues:
- Excessive white space in the text.
- Sentences split across multiple entries. Transcripts can appear like the example below. We need to detect and concatenate such strings to improve readability. Some logic for this already exists, but it is as exhaustive and isn't applied across all types of transcripts.
```
Some start of a sentence

that is continued here. But now we have a new one

That goes here. So what should we do? We need to join them.
```
- Grouping utterances by the same speaker. This works in combination with sentence detection.
- Filtering out empty entries.

> [!note]
> Sentence detection heuristics aren't perfect, as they don't account for characters like the [Japanese Full Stop](https://www.compart.com/en/unicode/U+3002). And possibly multitude of edge cases. However, this is still an improvement over the current implementation, which handles only `.`, `!`, and `?`.

## Testing Instructions

I already tested it locally with some transcripts but for now code review should be enough. Once I integrate it with our code we can verify it further before merging the integration.

## Checklist

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/transcript/TranscripSanitization.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/transcript/TranscripSanitization.kt
@@ -5,6 +5,7 @@ import au.com.shiftyjelly.pocketcasts.models.to.TranscriptEntry
 internal fun List<TranscriptEntry>.sanitize() = map(TranscriptEntry::compactWhiteSpace)
     .joinSplitSentences()
     .joinConsecutiveSpeakers()
+    .removeRepeatedSpeakers()
     .map(TranscriptEntry::trim)
     .filter(TranscriptEntry::isNotEmpty)
 
@@ -109,6 +110,24 @@ private fun List<TranscriptEntry>.joinConsecutiveSpeakers(): List<TranscriptEntr
         entries += TranscriptEntry.Speaker(names)
     }
     return entries
+}
+
+private fun List<TranscriptEntry>.removeRepeatedSpeakers(): List<TranscriptEntry> {
+    var lastSpeaker: String? = null
+    return mapNotNull { entry ->
+        when (entry) {
+            is TranscriptEntry.Speaker -> {
+                if (lastSpeaker != entry.name) {
+                    lastSpeaker = entry.name
+                    entry
+                } else {
+                    null
+                }
+            }
+
+            is TranscriptEntry.Text -> entry
+        }
+    }
 }
 
 private fun TranscriptEntry.trim() = when (this) {
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/transcript/TranscripSanitization.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/transcript/TranscripSanitization.kt
@@ -24,7 +24,7 @@ private fun TranscriptEntry.compactWhiteSpace() = when (this) {
 }
 
 private val AnyWhiteSpace = """\s+""".toRegex()
-private val TwoOrMoreEmptySpaces = """[ \t]+""".toRegex()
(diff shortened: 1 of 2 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/transcript/TranscripSanitization.kt
```diff
@@ -0,0 +1,179 @@
...
+    is TranscriptEntry.Text -> {
+        val newValue = value
+            .replace(TwoOrMoreEmptySpaces, " ")
+            .replace(ThreeOrMoreNewLines, "\n\n")
+        copy(value = newValue)
+    }
+}
+
+private val AnyWhiteSpace = """\s+""".toRegex()
+private val TwoOrMoreEmptySpaces = """[ \t]+""".toRegex()
```
sztomek: ❓ afaik `\s` represents a single whitespace (tab included) while `\t` matches only on tabs. I was wondering why haven't you defined it as `\s{2,}` ?

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/transcript/TranscripSanitization.kt
```diff
@@ -0,0 +1,179 @@
...
+private fun TranscriptEntry.isNotEmpty() = when (this) {
+    is TranscriptEntry.Speaker -> name.isNotEmpty()
+    is TranscriptEntry.Text -> value.isNotEmpty()
+}
+
+fun String.endsAsSentence(): Boolean {
+    return EndOfSentencePunctuation.any { punctuation -> endsWith(punctuation) }
+}
+
+private val EndOfSentencePunctuation = listOf(
```
sztomek, line 147: ❓ Are you sure that this is an exhaustive set? I mean they look reasonable to me, but I'm not convinced that we may not find more ending characters (e.g. in some foreign languages).

sztomek (approved):

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/transcript/TranscripSanitization.kt
MiSikora: I'm certain it's not. I wrote more about it in the PR's description.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/transcript/TranscripSanitization.kt
MiSikora: I'm not sure I understand the question. Are you asking why `[ \t]+` over `\s{2,}`? You're correct about `{2,}` instead of `+` and I'll fix that. But `\s` wouldn't be correct as it would account for new lines as well.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/transcript/TranscripSanitization.kt
sztomek: oh, really. my bad 😓

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/transcript/TranscripSanitization.kt
sztomek: > `\s` wouldn't be correct as it would account for new lines as well.

Ah, you're right. In that case `[ \t]{2,}` would be the way to go.

PR #4155, opened 2025-06-25
Imporve vertical player experience on smaller devices

MiSikora:
## Description

Currently when there is very limited space vertical player doesn't look great. This PR address it by changing artwork's corner radius if artwork is small, hiding artwork when there is no space for or it would look bad, and decreasing font scaling of text on ads.

## Testing Instructions

Test player with various font sizes on a small device.

## Screenshots or Screencast 

| Before | After |
| - | - |
| a1 | b1 |
| a2 | b2 |
| a3 | b3 |
| a4 | b4 |
a5 | b5 |
a6 | b6 |

## Checklist
 
- [x] with different themes

### DIFF
--- a/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/PlayerHeaderFragment.kt
+++ b/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/PlayerHeaderFragment.kt
@@ -21,7 +21,6 @@ import androidx.compose.animation.fadeOut
 import androidx.compose.animation.shrinkVertically
 import androidx.compose.animation.slideInVertically
 import androidx.compose.foundation.background
-import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.PaddingValues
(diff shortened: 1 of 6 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/PlayerHeaderFragment.kt
```diff
@@ -729,6 +707,69 @@ class PlayerHeaderFragment : BaseFragment(), PlayerClickListener {
...
 
+    @Composable
+    private fun AdAndArtwork(
+        ad: BlazeAd?,
+        artworkOrVideoState: ArtworkOrVideoState,
+        isTranscriptOpen: Boolean,
+        playerColors: PlayerColors,
+        modifier: Modifier = Modifier,
+    ) {
+        SubcomposeLayout(modifier = modifier) { constraints ->
```
sztomek, line 718: TIL! Wow, never used this component before. Had to read up on the topic to catch a glimpse of what it's used for.

sztomek (approved):
Looks good across small/large devices with various font size settings.

💡 One idea to consider: instead of hiding the artwork when there's not much space available, we should maybe consider pushing it behind player control components with an alpha layer?
In case of videos it would be better to just hide it as we do it now.

MiSikora:
> One idea to consider: instead of hiding the artwork when there's not much space available, we should maybe consider pushing it behind player control components with an alpha layer?

Maybe, I’m not sure. You could bring this up with the designers to see if it's something we should pursue or if they have other ideas. Personally, I’d just make the player content scrollable, but that would require us to have a predefined maximum height for artwork and video.

PR #4195, opened 2025-07-04
[Onboarding] Add new generic upgrade variants

sztomek:
## Description
This PR is the first step of the new onboarding saga.
It introdudes the the a mocked screen that can display both variants.
As long as the FF `new_onboarding_upgrade` is enabled, you'll be able to see the screen in action when you tap the Settings->Pocket Casts Plus item.

Please note that TalkBack and accessbility were not considered in the scope of this PR, I'm planning to polish those areas in upcoming PRs.

It's important to call out that the screen is mocked, there is no real data interaction yet.

Fixes https://linear.app/a8c/issue/PCDROID-14/implement-generic-upgrade-screens

## Testing Instructions
1. Install app
2. Tap Settings -> Pocket Casts Plus

## Screenshots or Screencast 
| Variant A | Variant B | 
| --- | --- | 
| Screenshot 2025-07-04 at 20 48 33 | Screenshot 2025-07-04 at 20 48 59 |

https://github.com/user-attachments/assets/8f942ccd-504f-4262-bf54-9b782ffc013a

## Checklist
- [ ] ~If this is a user-facing change, I have added an entry in CHANGELOG.md~ -- will be added when the entire feature is merged to `main`
 
- If this PR does not contain UI changes, ignore these items -->
- [x] with different themes
- [ ] ~with a landscape orientation~
- [ ] ~with the device set to have a large display and font size~
- [ ] ~for accessibility with TalkBack~

### DIFF
--- a/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/components/UpgradePlanSelector.kt
+++ b/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/components/UpgradePlanSelector.kt
@@ -1,0 +1,211 @@
+package au.com.shiftyjelly.pocketcasts.account.onboarding.components
+
+import androidx.compose.foundation.background
+import androidx.compose.foundation.border
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.shape.CircleShape
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material.Icon
+import androidx.compose.material.MaterialTheme
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.layout.SubcomposeLayout
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.tooling.preview.PreviewParameter
+import androidx.compose.ui.unit.dp
+import au.com.shiftyjelly.pocketcasts.compose.AppThemeWithBackground
+import au.com.shiftyjelly.pocketcasts.compose.components.TextH40
+import au.com.shiftyjelly.pocketcasts.compose.components.TextH50
+import au.com.shiftyjelly.pocketcasts.compose.preview.ThemePreviewParameterProvider
+import au.com.shiftyjelly.pocketcasts.compose.theme
+import au.com.shiftyjelly.pocketcasts.images.R
+import au.com.shiftyjelly.pocketcasts.ui.theme.Theme.ThemeType
+
+@Composable
+fun UpgradePlanSelector(
+    plan: String,
+    priceAndPeriod: String,
+    isSelected: Boolean,
+    onSelected: () -> Unit,
+    modifier: Modifier = Modifier,
+    savings: String? = null,
+    pricePerWeek: String? = null,
+) {
+    SubcomposeLayout(modifier = modifier) { constraints ->
+        val savingsBadge = savings?.let {
+            val badgeMeasurable = subcompose("savings") {
+                SavingsLabel(
+                    savings = savings,
+                )
+            }
+            badgeMeasurable.first().measure(constraints)
+        }
+        val badgeHeight = savingsBadge?.height ?: 0
+
+        val row = subcompose("selector_row") {
+            PlanRow(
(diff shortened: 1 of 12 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/components/UpgradePlanSelector.kt
```diff
@@ -0,0 +1,211 @@
...
+@Composable
+fun UpgradePlanSelector(
+    plan: String,
+    priceAndPeriod: String,
+    isSelected: Boolean,
+    onSelected: () -> Unit,
+    modifier: Modifier = Modifier,
+    savings: String? = null,
+    pricePerWeek: String? = null,
+) {
```
MiSikora, line 44: This function should probably accept a `SubscriptionPlan.Base` object rather than individual data elements. 

We also have `au.com.shiftyjelly.pocketcasts.profile.winback.SubscriptionRow`, which is very similar to this component. It might be worthwhile to refactor both into a single component with customizable styling options, such as badge placement.

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/components/UpgradePlanSelector.kt
```diff
@@ -0,0 +1,211 @@
...
+    SubcomposeLayout(modifier = modifier) { constraints ->
```
MiSikora, line 45: Subcomposition isn't cheap, and we don't need it here. You can position badge with simple `Aligment` modifier inside `Box` container.

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/components/UpgradeTrialScheduleItem.kt
```diff
@@ -0,0 +1,127 @@
...
+    iconSize: Dp = 43.dp,
+    connectingRodWith: Dp = 8.dp,
+) {
+    val iconBackground = MaterialTheme.theme.colors.primaryInteractive01
+    val density = LocalDensity.current
+    val connectingRodWidthPx = density.run { connectingRodWith.toPx() }
+    val iconSizePx = density.run { iconSize.toPx() }
+    Row(
+        modifier = modifier
+            .drawWithContent {
```
MiSikora, line 50: I don't think this with `ScheduleContent` function is the right approach to this UI problem. For example, look how it looks like when there is not enough space.

Screenshot_20250707_112959

And it will be very hard or even impossible to solve it with this approach. Instead it is a great candidate for a custom layout that accepts something like this:

```kotlin
data class MyData(
  val iconId: Int,
  val title: String,
  val description: String,
)

@Composable
fun MyLayout(
  data: List<MyData>,
) {
  // Here you can use Layout() to compute all necessary data
  // such as gradient colors, spacing, etc.
  TODO()
}
```

Or if you want to restrict number of steps:

```kotlin
@Composable
fun MyLayout(
  step1: MyData,
  step2: MyData,
  step3: MyData,
)
```

Though I think that writing the code for 3 or N elements isn't any different. And making it generic for number of parameters should be safer in terms of edge cases, etc. Plus it would make it more reusable if needed.

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/upgrade/OnboardingUpgradeVariantsScreen.kt
```diff
@@ -0,0 +1,355 @@
...
+
+        Column(
+            modifier = Modifier
+                .fillMaxSize()
+                .padding(
+                    horizontal = 24.dp,
+                    vertical = 16.dp,
+                ),
+        ) {
+            val mockPlan = OnboardingSubscriptionPlan.create(
```
MiSikora, line 104: I'd suggest adding a `companion object` level property like `PlanWithOffer` `PlanWithoutOffer`. It will increase readability and be very helpful for previews or even testing. See, for example, how `SubscriptionPlan` has `Preview` properties defined.

sztomek:
Gonna close this one, in favor of the following smaller ones:

https://github.com/Automattic/pocket-casts-android/pull/4205 

https://github.com/Automattic/pocket-casts-android/pull/4204 

https://github.com/Automattic/pocket-casts-android/pull/4203 

https://github.com/Automattic/pocket-casts-android/pull/4202

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/components/UpgradePlanSelector.kt
sztomek: Yeah, the plan was to dynamically position the 'Save XX%' label and I remembered that you've done great things wiht subcomposeLayout in the past, thought it's gonna be a good time to put it into practice. 😅 

Checked `SubscriptionRow` and it uses fixed offsets to shift the y position of the label. Went with that path anyways.

PR #4197, opened 2025-07-06
Allow podcast and folder options dialogs to stay open through config changes (like screen rotation)

josephkreiser:
## Description

Allow podcast and folder options dialogs to stay open through config changes (like screen rotation), instead of dismissing them.

Instead of creating the dialog when it is shown and dismissing it on config change, click handlers can be restored by checking to see if a particular dialog fragment is in the fragment manager and if it isn't then instantiate a new one, then establish all the dialog's click handlers.

Fixes # 

## Testing Instructions

1. On the Podcasts screen, tap the three dot icon in the upper right to open the "podcast options" bottom sheet dialog
2. Rotate the screen to see the dialog remain open and not be dismissed
3. Tap the "Sort by" or "Badges" option to open the "Sort by" or "Badges" bottom sheet dialog
4. Rotate the screen to see the dialog remain open and not be dismissed
5. Select a folder to go to that folder's screen and tap the three dot icon in the upper right to open the "folder options" bottom sheet dialog
6. Rotate the screen to see the dialog remain open and not be dismissed
7. Tap the "Sort by" option to open the "Sort by" bottom sheet dialog
8. Rotate the screen to see the dialog remain open and not be dismissed

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/podcasts/FolderOptionsDialog.kt
+++ b/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/podcasts/FolderOptionsDialog.kt
@@ -19,15 +19,18 @@ class FolderOptionsDialog(
     val fragment: Fragment,
     val settings: Settings,
 ) {
+    companion object {
+        private const val FRAGMENT_FOLDER_OPTIONS_DIALOG = "folder_options_dialog"
+        private const val FRAGMENT_FOLDER_SORT_DIALOG = "folder_sort_dialog"
+    }
 
-    private var showDialog: OptionsDialog? = null
-    private var sortDialog: OptionsDialog? = null
+    private val showDialog: OptionsDialog
+    private val sortDialog: OptionsDialog
     private val fragmentManager: FragmentManager?
         get() = fragment.activity?.supportFragmentManager
 
-    fun show() {
-        val fragmentManager = fragmentManager ?: return
-        val dialog = OptionsDialog()
+    init {
+        showDialog = (fragmentManager?.findFragmentByTag(FRAGMENT_FOLDER_OPTIONS_DIALOG) as? OptionsDialog ?: OptionsDialog())
             .addTextOption(
                 titleId = LR.string.podcasts_menu_sort_by,
                 imageId = IR.drawable.ic_sort,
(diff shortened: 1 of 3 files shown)


MiSikora:
@josephkreiser Thanks for the contribution. Before I review could you sync with `main`? We updated `ktlint` to the latest version so some new checks might trigger. We've also just had a release so the changelog entry will need to be moved to `7.94`.

josephkreiser:
> @josephkreiser Thanks for the contribution. Before I review could you sync with `main`? We updated `ktlint` to the latest version so some new checks might trigger. We've also just had a release so the changelog entry will need to be moved to `7.94`.

Should be all set now

josephkreiser:
@MiSikora 
`Initializer` sounds good to me.

The first line of `createFolderOptionsDialog()` (or soon to be `createFolderOptionsDialogInitializer()`) function is: `val folder = viewModel.uiState.value.folder ?: return null`. I'm assuming the value of `folder` won't actually change, but if this were a value that could change in between config changes, we would want to recreate the `FolderOptionsDialogInitializer` right before we show a new one.  Should we treat `viewModel.uiState.value.folder` as something that could change, or assume that it won't?

MiSikora:
> Should we treat viewModel.uiState.value.folder as something that could change, or assume that it won't?

No, the folder entity can't change because `folderUuid` is a `val` in `PodcastsViewModel`. However, I didn't realize that it's initialized like this: `val folder = viewModel.uiState.value.folder ?: return null`.  

Since `folder` might not be available at creation time due to asynchronous loading, my proposal needs to be modified.

```kotlin
interface OptionsDialogInitializer {
    fun initializeDialog()

    fun showDialog()
}

class PodcastsFragment : BaseFragment() {
    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)
        initializeOptionsDialog()
    }

    private fun openOptions() {
        initializeOptionsDialog().showDialog()
    }

    private fun initializeOptionsDialog(): OptionsDialogInitializer {
        val initializer = if (folderUuid == null) {
            createPodcastOptionsInitializer()
        } else {
            createFolderOptionsInitializer()
        }
        initializer.initializeDialog()
        return initializer
    }
}
```

One more thing I didn’t mention earlier: it would be better if `PodcastsOptionsDialog` and `FolderOptionsDialog` didn’t hold strong references to `OptionsDialog` instances. We should avoid keeping fragments in memory and instead let the framework manage and destroy them as needed. This won’t be as critical with the proposal above, since the initializers won’t be fields of the `PodcastsFragment`, so there won’t be any memory leaks. But it’s still better to avoid such practices altogether.

MiSikora:
I'm closing this issue due to inactivity but please feel free to reopen it or create a new one if my comments are addressed in the followup.

PR #4199, opened 2025-07-07
Fix flaky tests in `CategoriesManagerTest`

sztomek:
## Description
As title suggests.
We've noticed that some of the unit tests are randomly failing p1751871528948729-slack-C07J5LNP4SF
Culprit was `CategoriesManagerTest`, it turned out that I made a mistake and haven't used the right dispatcher to execute suspend functions on.
There were other tests prone to the same errors, like `PlaylistManagerImplTest` and `AppLifecycleObserverTest`, so I've updated them as well.
This PR is the remedey that solves this

## Testing Instructions
Just give a pipeline a good look and expect to see green lights

## Checklist


modules/services/repositories/src/test/java/au/com/shiftyjelly/pocketcasts/repositories/categories/CategoriesManagerTest.kt
```diff
@@ -23,6 +23,9 @@ internal class CategoriesManagerTest {
     @get:Rule
     val featureFlagRule = InMemoryFeatureFlagRule()
 
+    @get:Rule
+    val coroutineRule = MainCoroutineRule()
```
MiSikora: I'm not sure I understand the need for this rule. This rule is generally necessary for `ViewModel`s because of the problematic `viewModelScope` property, which enforces `Dispatchers.Main`. 

In this case, however, the rule doesn't help on its own with the test It would be simpler to just use:

```kotlin
private val scope = CoroutineScope(UnconfinedTestDispatcher())

@After
fun tearDown() {
  scope.cancel()
}
```

MiSikora (approved):

PR #4206, opened 2025-07-08
Make AutoPlaySource constants lazy

MiSikora:
## Description

Update to Kotlin `2.2.0` changed initialization order of `object`s. Previously `AutoPlaySource.None` was initialized before `companion object`. Now it happens after. This caused an issue because `None` was resolved to `null` causing a crash.

There might be other classes and properties defined like this but I couldn't find any.

## Testing Instructions

1. Open app.
2. Go to filters.
3. App should not crash.

## Checklist
 
- [ ] with different themes


sztomek (approved):
Yup, worked as expected, app didn't crash.

💡btw this `AutoPlaySource` could have been a good candidate for an `enum class` si we could have used `enum.entries` instead of the `Constants` that caused troubles for us

MiSikora:
`AutoPlaySource` can't be converted to an enum because `PodcastOrFilter` is a dynamic type. But we can subclass it so all constants are an enum. I'll make that change.

PR #4236, opened 2025-07-16
Introduce `DoNotMockDataClass` rule

wzieba:
Closes: AINFRA-943
Internal: paqN3M-1w9-p2

## Description

This PR introduces `DoNotMockDataClass` from Slack's lint (provided by WordPress lint wrapper). It also adds current violations to the baseline - I hope that in the future we'll address them.

And, on top of that, it improves configuration of Lint in this project. From now on:
- Custom lint checks are added to _all_ Library and App modules. If checks are not added, then rules from them don't run in a module.
- `wear` and `automotive`'s lint results will now send sarif report to GitHub. These projects will check only their own codebase.
- `app`'s lint will check itself + all dependencies

The above is IMO optimal configuration:
- we benefit from performance of `checkDependencies=true` (see [Only analyze app/leaf modules](https://googlesamples.github.io/android-custom-lint-rules/user-guide.md.html))
- we don't duplicate violations from subprojects (`wear` or `automotive`'s lint results don't include subproject violations)

## Testing Instructions

Check out https://github.com/Automattic/pocket-casts-android/pull/4245 and see that lint warnings from both `app` and `wear` modules are reported.

## Screenshot

Screenshot 2025-07-16 at 19 14 45

## Checklist
 
- [ ] with different themes


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

gradle/libs.versions.toml
```diff
@@ -152,6 +153,7 @@ lifecycle-viewmodel-compose = { module = "androidx.lifecycle:lifecycle-viewmodel
 # CLI included to be picked up by dependencies update automation
 ktlint-cli = { module = "com.pinterest.ktlint:ktlint-cli", version.ref = "ktlint"}
 ktlint-compose-rules = { module = "io.nlopez.compose.rules:ktlint", version.ref = "ktlint-compose-rules"}
+wordpress-lint = { module = "org.wordpress:lint", version.ref = "wordpress-lint" }
```
geekygecko, line 156: Is there a reason this is WordPress and not Automattic, similar to our other company libraries?
Would you mind pointing me to where we can see all the lint rules? Sorry I couldn't find it.

wzieba: > Is there a reason this is WordPress and not Automattic

I don't know - it's an old project from 7 years ago.

> Would you mind pointing me to where we can see all the lint rules? 

Here's the registry: https://github.com/wordpress-mobile/WordPress-Lint-Android/blob/trunk/WordPressLint/src/main/java/org/wordpress/android/lint/WordPressIssueRegistry.kt

There's only a handful of them: for Pocket Casts, it's basically only `DoNotMockDataClass` and `DoNotMockSealedClass` (rest of them are about Java).

.buildkite/commands/lint.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -eu
+#!/bin/bash -u
```
ParaskP7, line 1: 👍

.buildkite/commands/lint.sh
```diff
@@ -27,5 +27,7 @@ else
 fi
 
 upload_sarif_to_github 'app/build/reports/lint-results-release.sarif'
+upload_sarif_to_github 'automotive/build/reports/lint-results-release.sarif'
+upload_sarif_to_github 'wear/build/reports/lint-results-release.sarif'
```
ParaskP7, line 31: Praise (❤️): Nice!

automotive/lint-baseline.xml
```diff
@@ -1,6103 +1,50 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<issues format="6" by="lint 8.7.2" type="baseline" client="gradle" dependencies="true" name="AGP (8.7.2)" variant="all" version="8.7.2">
-
-    <issue
-        id="MissingSuperCall"
```
ParaskP7, line 5: > [This way, lint is actually faster and `lint-baseline.xml` of `automotive` and `wear` don't have duplicated issues of subprojects](https://github.com/Automattic/pocket-casts-android/pull/4236/commits/d93e6b08edc4b0e93b0d3349a7b2c8847d4c5536)

Praise (❤️): Nice upgrade!

ParaskP7:
Btw and FYI @wzieba , me running `./gradlew lintDebug`, that is without `app:`, or `automotive`, or `wear` fails on the `modules/services/repositories` module due to 8 `DoNotMockDataClass` violations. Are we sure we want this behavior, is that expected? 🤔

wzieba:
`./gradlew lintDebug` runs `lintDebug` task in all submodules and these submodules aren't aware of `baseline` or `config` files.

At this moment, I don't see any value in supporting running `lintDebug` task - it's less performant and doesn't give us anything in return.

ParaskP7:
> ./gradlew lintDebug runs lintDebug task in all submodules and these submodules aren't aware of baseline or config files.

👍 

> At this moment, I don't see any value in supporting running lintDebug task - it's less performant and doesn't give us anything in return.

True, can't disagree with the performance aspect of things. 💯

Question (❓): So, maybe we can somehow restrict running `lintDebug` and at the same time warn engineers that they should instead use `<app_module>:lintDebug`, if they need to run lint for the whole project (ish, as this still will run for that specific app only, ingoring any other apps). Or to use `<library_module>:lintDebug` command instead, if they want to run lint for a specific module. Do you think it's worth it doing that, just to avoid confusion, engineers thinking that Lint is failing but actually it is not... 🤔

PS: I know for sure, I have been using `lintDebug` instead of module targeted lint command, and it would have caught me by surprise if I was to see something failing all of a sudden, just like I did here (and on DOAndroid)... 🤷

wzieba:
>  Or to use <library_module>:lintDebug command instead

This will also make the Lint fail in your case, as library module won't still have access to the configuration or baseline. It's exactly the same as with `lintDebug`.

---

We could apply config and baseline to all subprojects, but project's `lintDebug` isn't something that should be executed anyway, as it's less performative. Running `lintDebug` per project in this case is also actually incorrect. E.g. if you run `lintDebug` on `settings` module, you'll get `MissingPermission` error. It's a false negative, as the app applies this permission, only in a different module.

ParaskP7:
> This will also make the Lint fail in your case, as library module won't still have access to the configuration or baseline. It's exactly the same as with lintDebug.

For sure, of course, but at least I'll now know it fails on this library module I am targeting.

> We could apply config and baseline to all subprojects, but project's lintDebug isn't something that should be executed anyway, as it's less performative. Running lintDebug per project in this case is also actually incorrect. E.g. if you run lintDebug on settings module, you'll get MissingPermission error. It's a false negative, as the app applies this permission, only in a different module.

I understand, but what's wrong with removing the 'lintDebug' task from the list of task that can run for this project, so that it's become more explicit? Or, adding a warning that you shouldn't use this 'lintDebug' task anymore, whenever someone tries to run it? That's my main suggestion here. I think it is worth it. Is it possible you think? 🤔

PS: If you still think that not worth it, please ignore this comment.

PR #4238, opened 2025-07-17
Task/playlists

MiSikora:
## Description

This PR adds the initial entry point for Playlists. Since it’s part of our main UI, I added the ability to read feature flags in a way that preserves their value for the lifetime of the app’s process. Otherwise, we could run into consistency issues. For example, the bottom navigation bar might display the "Filters" item while the Playlists page does not, or vice versa.

## Testing Instructions

1. Open the app.  
2. Tap the "Playlists" menu item.  
3. You should see the Playlists fragment.  
4. Scroll down the page.  
5. Tap the "Playlists" menu item again.  
6. The page should scroll back to the top.  
7. Disable the feature flag in Beta Features settings.  
8. Change the device configuration (e.g., rotate the screen).  
9. The Playlists fragment should still be visible.  
10. Close the app completely, including the playback service if media is playing.  
11. Reopen the app.  
12. You should now see the Filters fragment instead of Playlists.

## Checklist


sztomek (approved):
Worked as described ✅

PR #3908, opened 2025-04-17
Add weekly price to yearly offers during winback flow

MiSikora:
## Description

This PR adds weekly price to the offers.

Designs: G22KfZwL1Gv5yyxdTheF9H-fi-1075_9457

## Testing Instructions

> [!note]
> Test this with the debug variant and [this patch](https://github.com/user-attachments/files/19793722/winback.patch) applied.

1. Create a new account and subscribe to any plan.  
2. Go to your account details.  
3. Tap on the cancellation row.  
4. Check available plans.
5. Notice that yearly offers have per week value.
6. Notice that yearly Plus has the best value sticker.

## Screenshots or Screencast 

| Before | After |
| - | - |
| b | a |

## Checklist
- [ ] ~If this is a user-facing change, I have added an entry in CHANGELOG.md~
- [x] Ensure the linter passes (`./gradlew spotlessApply` to automatically apply formatting/linting)
- [ ] ~I have considered whether it makes sense to add tests for my changes~
- [x] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml`
- [x] Any jetpack compose components I added or changed are covered by compose previews
- [ ] ~I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.~
 
#### I have tested any UI changes...

- [x] with different themes
- [x] with a landscape orientation
- [x] with the device set to have a large display and font size
- [x] for accessibility with TalkBack

PR #4240, opened 2025-07-17
[Onboarding] Wire up read data and logic

sztomek:
## Description
This PR adds the next piece to the puzzle: the new onboarding screen now works with actual data!

Fixes https://linear.app/a8c/issue/PCDROID-28/wire-up-logic-on-upgrade-screens

## Testing Instructions
1. Apply [debug-as-prod (1).patch](https://github.com/user-attachments/files/21287520/debug-as-prod.1.patch)
2. Build `debugProd` variant and install it
3. Do not log in upon opening the app
4. Navigate to Settings -> Pocket Casts Plus
5. Select a plan and tap 'Subscribe' or 'Start free trial' if you're eligible (depends on your google account, if you've ever claimed the free trial with that google account, you won't be eligible)
6. Go through subscription flow presented by the Play billing sdk
- [ ] things worked

## Screencast
https://github.com/user-attachments/assets/9f682269-0065-4426-9452-90facaffa64b

## Checklist

### DIFF
--- a/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/components/SubscriptionPlanRow.kt
+++ b/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/components/SubscriptionPlanRow.kt
@@ -41,6 +41,8 @@ import au.com.shiftyjelly.pocketcasts.compose.components.TextP40
 import au.com.shiftyjelly.pocketcasts.compose.preview.ThemePreviewParameterProvider
 import au.com.shiftyjelly.pocketcasts.compose.theme
 import au.com.shiftyjelly.pocketcasts.payment.BillingCycle
+import au.com.shiftyjelly.pocketcasts.payment.Price
+import au.com.shiftyjelly.pocketcasts.payment.PricingSchedule
 import au.com.shiftyjelly.pocketcasts.payment.SubscriptionPlan
 import au.com.shiftyjelly.pocketcasts.payment.SubscriptionTier
 import au.com.shiftyjelly.pocketcasts.ui.theme.Theme.ThemeType
@@ -49,7 +51,7 @@ import au.com.shiftyjelly.pocketcasts.localization.R as LR
 
 @Composable
 fun AvailablePlanRow(
-    plan: SubscriptionPlan.Base,
+    plan: SubscriptionPlan,
     isSelected: Boolean,
     onClick: () -> Unit,
     modifier: Modifier = Modifier,
@@ -65,11 +67,11 @@ fun AvailablePlanRow(
 
 @Composable
 fun UpgradePlanRow(
-    plan: SubscriptionPlan.Base,
+    plan: SubscriptionPlan,
     isSelected: Boolean,
     onClick: () -> Unit,
     modifier: Modifier = Modifier,
-    priceComparisonPlan: SubscriptionPlan.Base? = null,
+    priceComparisonPlan: SubscriptionPlan? = null,
 ) {
     SubscriptionPlanRow(
         plan = plan,
@@ -141,7 +143,7 @@ private data class RowConfig(
 
 @Composable
 private fun SubscriptionPlanRow(
-    plan: SubscriptionPlan.Base,
+    plan: SubscriptionPlan,
     isSelected: Boolean,
     onClick: () -> Unit,
     rowConfig: RowConfig,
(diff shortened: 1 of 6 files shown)


modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/components/SubscriptionPlanRow.kt
```diff
@@ -274,43 +276,48 @@ private fun CheckMark(
...
 }
 
-private val SubscriptionPlan.Base.pricePerMonth: Float
+private val SubscriptionPlan.pricePerMonth: Float
     get() {
         val pricePerMonth = when (billingCycle) {
-            BillingCycle.Monthly -> pricingPhase.price.amount
-            BillingCycle.Yearly -> pricingPhase.price.amount / 12.toBigDecimal()
+            BillingCycle.Monthly -> recurringPrice?.amount?.toFloat() ?: 0f
+            BillingCycle.Yearly -> (recurringPrice?.amount?.toFloat() ?: 0f) / 12f
```
MiSikora: ⚠️ When dealing with prices we should do all the manipulations using `BigDecimal` functions. Conversion to `Float` should be done only at the end for the UI purposes.

It doesn't matter much in this instance but it is a bad practice to use floating-point arithmetic when dealing with money as it can lead to rounding errors.

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/components/SubscriptionPlanRow.kt
```diff
@@ -274,43 +276,48 @@ private fun CheckMark(
...
+    is SubscriptionPlan.Base -> pricingPhase.price
+    is SubscriptionPlan.WithOffer -> pricingPhases.find { it.schedule.recurrenceMode is PricingSchedule.RecurrenceMode.Infinite }?.price
+}
+
+private val SubscriptionPlan.pricePerWeek: Float
     get() {
         val pricePerWeek = when (billingCycle) {
-            BillingCycle.Monthly -> pricingPhase.price.amount / 4.toBigDecimal()
-            BillingCycle.Yearly -> pricingPhase.price.amount / 52.toBigDecimal()
+            BillingCycle.Monthly -> (recurringPrice?.amount?.toFloat() ?: 0f) / 4f
```
MiSikora: 🛑 I missed it earlier during the code review but this calculation isn't right for price per week.

20.99 PLN per month / 4 = 5.25 PLN
20.99 PLN per month * 12 / 52 = 4.84 PLN

I made the same mistake [here](https://github.com/Automattic/pocket-casts-android/pull/3908#discussion_r2049756216).

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/components/SubscriptionPlanRow.kt
```diff
@@ -274,43 +276,48 @@ private fun CheckMark(
...
         }
-        return pricePerMonth.toFloat()
+        return pricePerMonth
     }
 
-private val SubscriptionPlan.Base.pricePerWeek: Float
+private val SubscriptionPlan.recurringPrice: Price? get() = when (this) {
+    is SubscriptionPlan.Base -> pricingPhase.price
+    is SubscriptionPlan.WithOffer -> pricingPhases.find { it.schedule.recurrenceMode is PricingSchedule.RecurrenceMode.Infinite }?.price
+}
```
MiSikora: 🛑 `recurringPrice` should never be `null`. Allowing it to be nullable can lead to inconsistent UI and UX if there are issues in our offer mapping logic or in the Play Console configuration.

There are a few possible solutions, but the most sensible one is to move `recurringPrice` to be a property of `SubscriptionPlan`, and derive it based on the selected offer. Something like:

```kotlin
sealed interface SubscriptionPlan {    
    val recurringPrice get() = when (this) {
        is Base -> pricingPhase.price
        is WithOffer -> pricingPhases.last().price // or the logic that you use
    }
}
```

This should be covered with tests for every supported offer to ensure we have the correct prices.

MiSikora (changes requested):
I couldn't test the PR. Applying the patch doesn't work.

```
git apply ~/Downloads/debug-as-prod.1.patch
error: patch failed: build.gradle.kts:319
error: build.gradle.kts: patch does not apply
```

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/components/SubscriptionPlanRow.kt
sztomek: thanks, i'll keep it in mind in the future 🙏

sztomek:
sorry, attached the wrong patch... 
[debug-as-prod copy.patch](https://github.com/user-attachments/files/21302636/debug-as-prod.copy.patch)
please try with this one.
I have also addressed all the comments on this PR

MiSikora (approved):

PR #4243, opened 2025-07-17
Rename Filter to Smart Playlist

MiSikora:
## Description

This PR renames filters to smart playlists, along with all associated types such as `PlaylistDao` and `PlaylistManager`. The goal is to reduce confusion when developing the playlists feature.

I also removed `fallbackToDestructiveMigration()` in debug builds. Keeping it is too risky, as we might not notice when migrations break (something that has happened in the past). While this change might occasionally be inconvenient during development, I believe the risk of breaking the database is not worth it.

## Testing Instructions

1. Check out a version prior to the database migration.
```
git checkout 8836801d22975569d7b091bd03f74468498ac188
```
3. Install the app.
4. Create some filters.
5. Check out the version with the database migration.
```
git checkout task/playlist-rename
```
7. Install the app.
8. Verify that your filters are still available.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/filters/DurationOptionsFragment.kt
```diff
@@ -101,8 +101,8 @@ class DurationOptionsFragment : BaseFragment() {
         val btnClose = binding.btnClose
 
         launch {
-            val playlist = playlistManager.findByUuid(requireArguments().getString(ARG_PLAYLIST_UUID)!!) ?: return@launch
-            this@DurationOptionsFragment.playlist = playlist
+            val playlist = smartPlaylistManager.findByUuid(requireArguments().getString(ARG_PLAYLIST_UUID)!!) ?: return@launch
```
sztomek, line 104: 💡 instead of `!!` you could use `.orEmpty()`

modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/filters/EpisodeOptionsFragment.kt
```diff
@@ -75,8 +75,8 @@ class EpisodeOptionsFragment :
         launch {
             val uuid = requireArguments().getString(ARG_PLAYLIST_UUID)!!
```
sztomek, line 76: 💡 i know it's not your code, but could use `.orEmpty` here as well

modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/filters/DurationOptionsFragment.kt
MiSikora: This changes the behavior. I don't want to introduce behavioral changes for old code into such a big PR with only name changes. Besides, crashing might be preferable to falling back to a default value.

PR #4261, opened 2025-07-21
Add missing `notification_open` event

sztomek:
## Description
For some reason, the `notification_open` event was never implemented during the Notifications Revamp project.
Slack convo: p1752688629524859-slack-C028JAG44VD
You can find the specs of the event in the [tracking plan](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?gid=0#gid=0) or just look at this screenshot:
Screenshot 2025-07-21 at 21 09 35
This event is quite important because we meant to track the change in volume of notification interactions with and without the FF enabled.

## Testing Instructions
1. You should see 'Tracked: noticication_open` in LogCat when:
- You tap on a notification from the Notification Revamp project. The easiest way to trigger them is to visit Settings > Developer > Notifications testing and fire random notifications
- You tap on any other notification our app displays... This one is a bit harder to test, but you should play with downloads, syncing account, etc.

## Checklist

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/notification/NotificationOpenReceiver.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/notification/NotificationOpenReceiver.kt
@@ -1,7 +1,6 @@
 package au.com.shiftyjelly.pocketcasts.repositories.notification
 
 import android.content.BroadcastReceiver
-import android.content.ComponentName
 import android.content.Context
 import android.content.Intent
 import android.os.Bundle
@@ -41,14 +40,17 @@ class NotificationOpenReceiver : BroadcastReceiver() {
                         action = intent.getStringExtra(it)
                         intent.removeExtra(it)
                     }
+
                     EXTRA_ORIGINAL_FLAGS -> {
                         flags = intent.getIntExtra(it, 0)
                         intent.removeExtra(it)
                     }
+
                     EXTRA_ORIGINAL_COMPONENT -> {
                         component = @Suppress("DEPRECATION") intent.getParcelableExtra(it)
                         intent.removeExtra(it)
                     }
+
                     EXTRA_CATEGORY -> intent.removeExtra(it)
                     else -> Unit
                 }
@@ -78,7 +80,7 @@ class NotificationOpenReceiver : BroadcastReceiver() {
 
     private fun tryLaunchIntent(intent: Intent, context: Context) {
         try {
-            context.startActivity(intent.apply {  flags = Intent.FLAG_ACTIVITY_NEW_TASK })
+            context.startActivity(intent.apply { flags = Intent.FLAG_ACTIVITY_NEW_TASK })
         } catch (t: Throwable) {
             Timber.w("Failed to launch activity for intent $intent")
         }
@@ -115,6 +117,18 @@ class NotificationOpenReceiver : BroadcastReceiver() {
         fun toPodcastIntentRelay(context: Context, intent: Intent) = Intent(context, NotificationOpenReceiver::class.java).apply {
             putExtras(intent)
             putExtra(EXTRA_CATEGORY, CATEGORY_PODCAST)
+            putExtra(EXTRA_ORIGINAL_FLAGS, intent.flags)
+            intent.component?.let {
+                putExtra(EXTRA_ORIGINAL_COMPONENT, it)
+            }
+            intent.action?.let {
+                putExtra(EXTRA_ORIGINAL_ACTION, it)
+            }
+        }
+
+        fun toDeeplinkIntentRelay(context: Context, intent: Intent) = Intent(context, NotificationOpenReceiver::class.java).apply {
+            putExtras(intent)
+            putExtra(EXTRA_CATEGORY, CATEGORY_DEEP_LINK)
             putExtra(EXTRA_ORIGINAL_FLAGS, intent.flags)
             intent.component?.let {
                 putExtra(EXTRA_ORIGINAL_COMPONENT, it)
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/notification/NotificationOpenReceiver.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/notification/NotificationOpenReceiver.kt
@@ -30,7 +30,7 @@ class NotificationOpenReceiver : BroadcastReceiver() {
             event = AnalyticsEvent.NOTIFICATION_OPENED,
             properties = buildMap {
                 put("category", category)
-            }
+            },
         )
 
         val originalIntent = Intent().apply {
@@ -47,7 +47,9 @@ class NotificationOpenReceiver : BroadcastReceiver() {
                     }
 
                     EXTRA_ORIGINAL_COMPONENT -> {
-                        component = @Suppress("DEPRECATION") intent.getParcelableExtra(it)
+                        component =
+                            @Suppress("DEPRECATION")
+                            intent.getParcelableExtra(it)
                         intent.removeExtra(it)
                     }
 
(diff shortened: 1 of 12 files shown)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/notification/NotificationOpenReceiver.kt
```diff
@@ -0,0 +1,143 @@
...
+import dagger.hilt.android.AndroidEntryPoint
+import javax.inject.Inject
+import timber.log.Timber
+
+@AndroidEntryPoint
+class NotificationOpenReceiver : BroadcastReceiver() {
+    @Inject
+    lateinit var analyticsTracker: AnalyticsTracker
+
+    override fun onReceive(context: Context?, intent: Intent?) {
```
MiSikora: ℹ️ Both `context` and `intent` cannot be `null` in the `BroadcastReceiver`. They just aren't marked with any annotations in the platform code.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/notification/NotificationOpenReceiver.kt
```diff
@@ -0,0 +1,143 @@
...
+        when (intent?.action) {
+            ACTION_REVAMP_NOTIFICATION_OPENED -> handleRevampedNotificationOpen(intent, context)
+            else -> handleRelayedNotificationOpen(intent ?: Intent(), context)
+        }
+    }
+
+    private fun handleRelayedNotificationOpen(intent: Intent, context: Context?) {
+        val category = intent.getStringExtra(EXTRA_CATEGORY)
+        if (context == null || category == null) return
```
MiSikora: ⚠️ Is there a reason we're accepting a nullable `Context?` here if we just return early when it's `null`? It might make more sense to require a non-null `Context` instead.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/notification/NotificationOpenReceiver.kt
```diff
@@ -0,0 +1,143 @@
...
+        }
+    }
+
+    companion object {
+        private const val ACTION_REVAMP_NOTIFICATION_OPENED = "au.com.shiftyjelly.pocketcasts.ACTION_REVAMPED_NOTIFICATION_OPENED"
+        private const val EXTRA_REVAMPED_TYPE = "extras.revamped.notification.type"
+        private const val EXTRA_CATEGORY = "extras.notification.category"
+        private const val EXTRA_ORIGINAL_COMPONENT = "extras.notification.component"
+        private const val EXTRA_ORIGINAL_FLAGS = "extras.notification.flags"
+        private const val EXTRA_ORIGINAL_ACTION = "extras.notification.action"
```
MiSikora: ❓ Could you prefix those with our package name? Just to avoid any potential namespace conflicts and make it easier to debug in the future when inspecting intents.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/notification/NotificationOpenReceiver.kt
```diff
@@ -0,0 +1,143 @@
...
+
+    private fun handleRelayedNotificationOpen(intent: Intent, context: Context?) {
+        val category = intent.getStringExtra(EXTRA_CATEGORY)
+        if (context == null || category == null) return
+
+        analyticsTracker.track(
+            event = AnalyticsEvent.NOTIFICATION_OPENED,
+            properties = buildMap {
+                put("category", category)
+            },
```
MiSikora:
```suggestion
            properties = mapOf("category" to category),
```

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/notification/NotificationOpenReceiver.kt
```diff
@@ -0,0 +1,143 @@
...
+
+                    EXTRA_ORIGINAL_FLAGS -> {
+                        flags = intent.getIntExtra(it, 0)
+                        intent.removeExtra(it)
+                    }
+
+                    EXTRA_ORIGINAL_COMPONENT -> {
+                        component =
+                            @Suppress("DEPRECATION")
+                            intent.getParcelableExtra(it)
```
MiSikora: 🛑 It can potentially lead to crashes. Use `IntentCompat.getParcelableExtra()` instead.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/notification/NotificationOpenReceiver.kt
```diff
@@ -0,0 +1,143 @@
...
+
+        analyticsTracker.track(
+            event = AnalyticsEvent.NOTIFICATION_OPENED,
+            properties = buildMap {
+                put("category", category)
+            },
+        )
+
+        val originalIntent = Intent().apply {
+            intent.extras?.keySet()?.forEach {
```
MiSikora: ❓ I'm a bit unclear on what this code is doing. Is it copying the original `intent` and then replacing `action`, `flags`, and `component` with the ones from the extras? If that's the goal, I think there might be a simpler way to express it. But I might be misunderstanding the intention here.


```kotlin
Intent(intent).apply {
    action = intent.getStringExtra(EXTRA_ORIGINAL_ACTION) ?: action
    flags = intent.getIntExtra(EXTRA_ORIGINAL_FLAGS, flags)
    component = IntentCompat.getParcelableExtra(intent, EXTRA_ORIGINAL_COMPONENT, ComponentName::class.java) ?: component
    removeExtra(EXTRA_CATEGORY)
}
```

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/notification/NotificationOpenReceiver.kt
```diff
@@ -0,0 +1,143 @@
...
+                    else -> Unit
+                }
+            }
+            putExtras(intent.extras ?: Bundle())
+        }
+        tryLaunchIntent(originalIntent, context)
+    }
+
+    private fun handleRevampedNotificationOpen(intent: Intent, context: Context?) {
+        if (context == null) return
```
MiSikora: ⚠️ Same nullability pattern is used here. I know we have quite a bit of this in the legacy parts of the codebase, but we should really try to avoid it going forward. 

If a function or class can't operate with a `null` value, it probably shouldn't accept `null` in the first place. Instead, the caller should handle that case. Either by not calling the function or doing something else appropriate.

It's just too easy to accidentally pass in `null` and get no feedback about it, which can lead to subtle bugs that are hard to track down.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/notification/NotificationOpenReceiver.kt
sztomek: You've got the point.
The problem with your suggested alternative is the following. I'm using explicit intents addressed to `NotificationOpenReceiver` in the broadcast factories (`toPodcastIntentRelay` and `toEpisodeIntentRelay`) like `= Intent(context, NotificationOpenReceiver::class.java`), and putting the action, flags, component into the bundle with unique keys.
In this method I'm trying to reconstruct the original intent. Component should be still read from the extras, rest could be salvaged... I'm making changes not to use the action and flags keys.

MiSikora: Sorry, I don't understand the problem that my proposal creates. It replaces `action`, `flags`, and `component` by reading them from extra properties. Do I miss something?

sztomek: Nope, it was me who misunderstood something... Sorry

PR #4272, opened 2025-07-24
Open podcast search by tapping the Discover button when the Discover page is displayed

rlparks:
## Description
Allows the user to open the podcast search quickly by tapping the Discover navigation button when the Discover page is already displayed. This is a common feature in Android apps like Google Photos, Play Store, and Immich that I miss when using Pocket Casts.

I am curious if the analytics source should be different, though technically it is launched from Discover as it currently says. Please let me know if anything is out of place. Thank you!

## Testing Instructions
1. Navigate to the Discover tab
2. Tap the Discover navigation button
3. The search fragment should open as if the user tapped the top search bar

## Screenshots or Screencast 
[Screen recording of double-tapping the Discover button and the keyboard popping up](https://github.com/user-attachments/assets/6df8c91b-b1b8-4584-8e57-3d69bf4fe48a)

## Checklist
 
- [x] with different themes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,11 @@
 7.95
 -----
-*   Bug Fixes
-    *   Fix missing file sizes on the episode page
-        ([#4262](https://github.com/Automattic/pocket-casts-android/pull/4262))
 *   Updates
     *   Open podcast search by tapping the Discover button when the Discover page is displayed
         ([#4272](https://github.com/Automattic/pocket-casts-android/pull/4272))
+* Bug Fixes
+    *   Fix missing file sizes on the episode page
+        ([#4262](https://github.com/Automattic/pocket-casts-android/pull/4262))
 
 7.94
 -----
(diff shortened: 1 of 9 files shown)


CookieyedCodes:
Thank you for adding this :)

MiSikora (changes requested):
Thank you for this change!

I checked in with the design team, and we'd like to slightly adjust the behavior. Instead of opening search immediately when tapping the Discover button, we’d prefer the following flow:

- If the page is not at the top, tapping the Discover button should scroll the page to the top.
- If the page is already at the top and the button is tapped again, then it should open the search.

To implement this, I’d suggest updating `TopScrollable.scrollToTop()` to return a `Boolean`. `true` if it scrolls to top, and `false` if the page was already at the top. This way, you can trigger the search only if `scrollToTop()` returns `false` and the Discover page is currently displayed.

For checking if the page is already scrolled to top, you can use:

- `RecyclerView.canScrollVertically(-1)`
- `LazyListState.canScrollBackward`

One last thing. Could you please update the changelog entry? The current description says it's triggered by a "double-tap," but it’s actually a "tap when the Discover page is already displayed."

rlparks:
That all makes sense; thank you for the guidance on implementing the scroll check. I'll give it my best shot :)

rlparks:
I think this should work... this is my first time using both Kotlin and Compose so I hope the `LazyListState` usage is good. Thanks for the review :)

PR #4289, opened 2025-07-29
[Onboarding] Folders upgrade

sztomek:
## Description
Introducing the first of the contextual upgrade screens: Folders.
The animation was quite tricky as it was composed of multiple animated pieces.
If you want to dig deep in the animation, [Download Principle](https://principleformac.com/) and import David's animations [from drive](https://drive.google.com/drive/u/1/folders/1WFh43s0odc_6r_j6dpfjMAvJ5JA24lrJ)
It's not ideal, but we've agreed to implemenet these animations natively on each platform. The folders anim seems  to be the trickiest. I tried to avoid using fixed sizes (besides paddings and offsets) to ensure elements will look good across diverse device configurations.
See previews and tweak modifiers.

Animation: https://video.wordpress.com/v/9DfZOPWA
Figma: PviUP0aiZctOprDuuYRwpb-fi-4999_4808

Fixes https://linear.app/a8c/issue/PCDROID-15/upgrade-screen-folders

## Testing Instructions
1. Log in with an account that doesn't have any subscriptions
2. Navigate to a screen that promotes folders (e.g. Podcasts -> Tap the folder icon on top)
3. Check animation

## Screenshots or Screencast 

https://github.com/user-attachments/assets/e641895c-3cf2-4b35-9eaf-d7bd62db7cb8

## Checklist
 
- [x] with different themes


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/upgrade/OnboardingUpgradeScreen.kt
```diff
@@ -2,6 +2,7 @@ package au.com.shiftyjelly.pocketcasts.account.onboarding.upgrade
 
 import UpgradeTrialItem
 import UpgradeTrialTimeline
+import androidx.compose.animation.ExperimentalSharedTransitionApi
```
MiSikora: ❓ Is this annotation needed? I see it used in couple of screens but none of the APIs say that we need it when I remove the annotation.

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/upgrade/contextual/Folders.kt
```diff
@@ -0,0 +1,544 @@
...
+    val color: Color,
+    val tiles: List<TileConfig>,
+)
+
+private fun randomColor() = Color(
+    red = Random.nextFloat(),
+    green = Random.nextFloat(),
+    blue = Random.nextFloat(),
+    alpha = 1f,
+)
```
MiSikora: ⛏️ Could you move this function to the bottom? It is used only for the Composable UI preview making it less important. Initially I was wondering why it is the part of the design.

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/upgrade/contextual/Folders.kt
```diff
@@ -0,0 +1,544 @@
...
+
+private val previewTiles = (1..8).map {
```
MiSikora: 💡 Perhaps we could name it `predefinedTiles` instead? `preview` conflates meaning with Compose.

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/upgrade/contextual/Folders.kt
```diff
@@ -0,0 +1,544 @@
...
+            2 -> TransformOrigin(1f, 1f)
+            3 -> TransformOrigin(0f, 1f)
+            6 -> TransformOrigin(1f, 0f)
+            7 -> TransformOrigin(0f, 0f)
+            else -> TransformOrigin(0f, 0f)
+        },
+    )
+}
+
+private val mockTile = TileConfig(
```
MiSikora: 💡 Perhaps `placeholderTile`?

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/upgrade/contextual/Folders.kt
```diff
@@ -0,0 +1,544 @@
...
+    index = -1,
+    drawableResId = null,
+    anchor = TransformOrigin(0f, 0f),
+)
+
+@Composable
+private fun previewFolders() = listOf(
```
MiSikora: 💡 Perhaps `predefinedFolders`?

MiSikora: 💡 This function could be marked with `@ReadOnlyComposable`.

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/upgrade/contextual/Folders.kt
```diff
@@ -0,0 +1,544 @@
...
+    FolderConfig(
+        folderName = "Books",
```
MiSikora: 💡 The titles here could be localized since we have easy access to resources.

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/upgrade/contextual/Folders.kt
```diff
@@ -0,0 +1,544 @@
...
+    ),
+)
+
+private val edgeFadeIndices = listOf(1, 4, 5, 8)
+
+@Composable
+fun FoldersAnimation(
+    modifier: Modifier = Modifier,
+    tiles: List<TileConfig> = previewTiles,
+    folders: List<FolderConfig> = previewFolders(),
```
MiSikora: ⚠️ Since these parameters are not dynamic, I suggest moving them out of the function definition and into the function body. Currently, the code makes assumptions, such as accessing folders with `folders[1]`. This is ok only if we can guarantee that the structure will not change. Something that’s only possible when the data cannot be passed in from the outside. Having it in the function definition introduces the risk that an element at index `1` may not exist.

This change would also allow you to make the `TileConfig` and `FolderConfig` classes private.

Alternatively, if you want to keep the APIs public, the config classes should not operate on dynamic lists but on properties like `leftFolder`, `centerFolder`, `rightFolder`, etc. Which BTW is still a correct thing to do even if the API is private. But it's not necessary in such a case since we can make assumptions about the private code.

MiSikora (approved):
I left some minor comments and highlighted one issue about the public API.

Also, one thing to consider is constraining the max height of the animation component. It looks a bit funny on tablets and foldables because it doesn't fit the screen.

Screenshot_20250729_113404

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/upgrade/contextual/Folders.kt
```diff
@@ -0,0 +1,544 @@
...
+                Tile(
+                    modifier = Modifier.aspectRatio(1f),
+                    tileConfig = it,
+                )
+            }
+            TextH30(
+                text = spec.folderName,
+                color = MaterialTheme.theme.colors.primaryInteractive02,
+                maxLines = 1,
+            )
```
MiSikora: 💡 It might make sense to use `disableFontScale` property here.

PR #4290, opened 2025-07-29
Add playlist creation page

MiSikora:
## Description

Designs: 5sC4z4Mu42LvL4MAAIbQVi-fi-764_7143

## Testing Instructions

1. Go to playlists screen.
2. Tap the create button.
3. You should see playlist creation page and pre-populated text field.
4. Verify the screen UX by providing texts and interacting with different buttons.

## Screenshots or Screencast 

https://github.com/user-attachments/assets/ea1fe499-2554-4fc6-a6ca-78ef55d283aa

## Checklist
 
- [x] with different themes


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/playlists/PlaylistsFragment.kt
```diff
@@ -48,7 +49,10 @@ class PlaylistsFragment :
                 listState = listState,
                 onCreatePlaylist = {
                     viewModel.trackCreatePlaylistClicked()
-                    Timber.i("Create playlist clicked")
+                    val fragment = childFragmentManager.findFragmentByTag("createPlaylist")
```
MiSikora:
```suggestion
                    val fragment = childFragmentManager.findFragmentByTag("create_playlist")
```

modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/playlists/create/NewPlaylistPage.kt
```diff
@@ -0,0 +1,204 @@
...
+            )
+            SmartPlaylistButton(
+                enabled = titleState.text.isNotBlank(),
+                onClick = onContinueToSmartPlaylist,
+                modifier = Modifier.fillMaxWidth(),
+            )
+            Spacer(
+                modifier = Modifier.height(24.dp),
+            )
+            RowButton(
```
sztomek, line 125: ⛏️  default `fontSize` is only `16.sp`, designs specify it as `18.sp` 
also, the letterSpacing seems a bit too wide by default

modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/playlists/create/NewPlaylistPage.kt
```diff
@@ -0,0 +1,204 @@
...
+    modifier: Modifier = Modifier,
+) {
+    val focusRequester = remember { FocusRequester() }
+    LaunchedEffect(focusRequester) {
+        focusRequester.requestFocus()
+    }
+    Column(
+        modifier = modifier,
+    ) {
+        IconButton(
```
sztomek, line 73: ❓ I know you've followed designs, but the position of the close button seems a bit off to me. We have similar buttons in the app in many places, all of them are positioned to the top right corner. Maybe that would worth discussing with designers

modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/playlists/create/NewPlaylistPage.kt
```diff
@@ -0,0 +1,204 @@
...
+                        ) {
+                            Icon(
+                                painter = painterResource(IR.drawable.ic_close_outlined),
+                                contentDescription = stringResource(LR.string.clear),
+                                tint = MaterialTheme.theme.colors.primaryIcon02,
+                            )
+                        }
+                    }
+                },
+                onImeAction = onCreateManualPlaylist,
```
sztomek, line 111: 💡 maybe it should also dismiss keyboard, because it stays open no matter how hard i spam the done button

MiSikora: It will be automatically dismissed when the screen disappears. Since it doesn't navigate anywhere it stays open for now.

modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/playlists/create/NewPlaylistPage.kt
MiSikora: If we want to change that I think this one should be addressed via global change to the button. In all Figma designs buttons have this spec but we use our built-in component regardless.

modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/playlists/create/NewPlaylistPage.kt
MiSikora: > all of them are positioned to the top right corner.

I'm not sure I see that. Screen elements like banners have buttons in the top right corner, but for full-screen pages, it's almost always in the top left corner, just like the standard toolbar navigation.

- The Smart Folders bottom sheet has a button in the top left corner.
- The current filter creation screen has a button in the top left corner.
- The Winback available plans page has a button in the top left corner.
- The Transcript page has a button in the top left corner.

The only bottom sheet I found with the button in the top right corner is the Referral page, which seems to be an outlier.

Another point is that this flow has either "X" or "<-" button. It would be weird to have the arrow in the top right corner. Or to have these buttons in different places.

Screenshot 2025-07-30 at 12 05 59

PR #4294, opened 2025-07-29
Fix race condition between navigation bar translation and player offset

MiSikora:
## Description

There is a race condition when trying to translate navigation bar after a configuration change. Moving animation to post layout fixes the issue.

Fixes PCDROID-64

## Testing Instructions

1. Play something and open the player.
2. Rotate the screen a couple of times.
3. Navigation bar should not draw on top of the player.

## Checklist


geekygecko (approved):
Worked well in my testing. Thanks for fixing this issue.

PR #4297, opened 2025-07-29
Add resistance to dismissing playlist sheet

MiSikora:
## Description

Playlists are created using a bottom sheet. Unfortunately, the swipe-to-dismiss gesture is very sensitive. Sometimes it even triggers when swiping up.

Because of our integration with Jetpack Compose, we do not have an easy way to reduce the gesture's sensitivity. Increasing friction does notwork, and relying on fling velocity is not reliable because Compose interferes with those values due to nested scroll interop issues.

However, the sheet's offset and the view's height are consistent and easy to work with. While this is not an ideal solution, I implemented a workaround where the sheet re-expands itself if the offset threshold does not reach 150 dp after flinging.

This solution is not perfect, but I prefer it over disabling the swipe gesture completely.

## Testing Instructions

1. Start playlist creation.
2. Notice that swiping it down to dismiss it gives some resistance.

## Checklist

### DIFF
--- a/modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/playlists/create/CreatePlaylistFragment.kt
+++ b/modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/playlists/create/CreatePlaylistFragment.kt
@@ -2,11 +2,9 @@ package au.com.shiftyjelly.pocketcasts.playlists.create
 
 import android.os.Bundle
 import android.view.LayoutInflater
+import android.view.View
 import android.view.ViewGroup
 import androidx.compose.foundation.layout.fillMaxSize
-import androidx.compose.foundation.layout.navigationBarsPadding
-import androidx.compose.foundation.rememberScrollState
-import androidx.compose.foundation.verticalScroll
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
 import androidx.compose.ui.Modifier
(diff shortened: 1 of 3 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

modules/services/views/src/main/java/com/google/android/material/bottomsheet/ViewPager2AwareBottomSheetBehavior.kt
```diff
@@ -20,6 +21,22 @@ open class ViewPager2AwareBottomSheetBehavior<V : View> @JvmOverloads constructo
...
+
+    fun setPreFlingInterceptor(interceptor: PreFlingInterceptor) {
+        preFlingInterceptor = interceptor
+    }
+
+    override fun onNestedPreFling(coordinatorLayout: CoordinatorLayout, child: V, target: View, velocityX: Float, velocityY: Float): Boolean {
+        val shouldDisableFling = preFlingInterceptor?.shouldInterceptFlingGesture(velocityX, velocityY) ?: false
+        return if (shouldDisableFling) {
+            preFlingInterceptor?.onFlingIntercepted(velocityX, velocityY)
+            true
```
MiSikora:
```suggestion
            false
```

modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/playlists/create/CreatePlaylistFragment.kt
```diff
@@ -82,6 +84,29 @@ class CreatePlaylistFragment : BaseDialogFragment() {
...
+                behavior.setPreFlingInterceptor(
+                    object : ViewPager2AwareBottomSheetBehavior.PreFlingInterceptor {
+                        override fun shouldInterceptFlingGesture(velocityX: Float, velocityY: Float): Boolean {
+                            val offsetPx = (view.height * (1f - behavior.calculateSlideOffset())).roundToInt()
+                            val offsetDp = offsetPx.pxToDp(requireContext())
+                            return offsetDp < 150
+                        }
+
+                        override fun onFlingIntercepted(velocityX: Float, velocityY: Float) {
+                            view.post {
```
sztomek, line 102: ❓ why do we need to post it? 
this practice is prone to errors according to my experience. imagine edge cases where the view has been detached from the hierarchy when the runnable gets executed 🤔

sztomek (approved):
Worked well in my testing!

modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/playlists/create/CreatePlaylistFragment.kt
MiSikora: It is needed because behavior can set its state to collapsed/hidden in different frames leading to collapsing anyway.

I know that posting is error-prone and can lead to bad behavior or even crashes. But what is the risk in this case?

PR #4345, opened 2025-08-07
Clear transient state when editing rules

MiSikora:
## Description

Currently, when creating a Smart Playlist, the transient state when editing the rules is never cleared. This leads to a bit of a confusing state.

1. I edit the rule.
2. I don't save it.
3. I come back to the rule.
4. The previous settings are still there instead of reflecting my currently applied rules.

This change fixes that. However the transient state of selected podcasts and episode duration limits are kept as long as the applied rule is the opposite. This is done to align behavior with the one for Filters.

## Testing Instructions

1. Start Smart Playlist creation.
2. Edit, for example, download status.
3. Save it.
4. Go back to it.
5. Change the rule.
6. Change device's configuration.
7. The rule should keep it's value.
8. Go back to the previous screen either with an "X" or system navigation.
9. Go back to the rule.
10. The rule should not use the state that was previously set.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

geekygecko (approved):

PR #4365, opened 2025-08-13
Fix notifications don't open

sztomek:
## Description
It's been reported that notifications no longer open the app when tapped.
Culprit is Android's background launch limitation.
The issue persits since we introduced the `notification_open` event tracking in #4275 .
I've added a `BroadcastReceiver` to report the analytics event and then tried to launch the original intent. However we can't launch activities while the app is in background starting with Android 12. Yay!
I've tried start a temporary foreground service to launch the activity from it, but that didn't work either 🤷 
So I ended up wiht this current solution and instead of the `BroadcastReceiver` I now have a trampoline `Activity` that reports the event then launches the original activity and finishes.

Fixes https://linear.app/a8c/issue/PCDROID-90/user-reports-issues-with-podcast-notifications-not-working

## Testing Instructions
1. Open app, grant notification permissions
2. Profile -> Settings -> Developer -> Notifications testing
3. Fire an instant notification
4. Send app to bg
5. Tap the notification
- [ ] `notification_open` is still tracked (visible in logcat if you filter 'Tracked')
- [ ] you land on the appropriate screen

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-android/milestone/99">7.95 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/notification/NotificationOpenReceiverActivity.kt
```diff
@@ -0,0 +1,119 @@
...
+import androidx.appcompat.app.AppCompatActivity
+import androidx.core.content.IntentCompat
+import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsEvent
+import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsTracker
+import dagger.hilt.android.AndroidEntryPoint
+import javax.inject.Inject
+import timber.log.Timber
+
+@AndroidEntryPoint
+class NotificationOpenReceiverActivity : AppCompatActivity() {
```
MiSikora, line 16: ⚠️ If this Activity is intended to be a transparent bootstrap, we should declare its theme in the Manifest as `android:theme="@android:style/Theme.NoDisplay"` and move all the logic to `onCreate()`. This theme ensures that there is no UI available for the Activity, and `onCreate()` must call `finish()`.

MiSikora (approved):
This works, but I think, in general, this could be solved in a more appropriate way, as we are fighting against the OS with workarounds. It should fine to merge as a beta fix but it would be great to followup with an idiomatic solution targeting the `main` branch.

Instead of launching a receiver or an invisible activity, we could simply launch `MainActivity` and let it handle everything. We should go back to using the previous `Intent` creation in `NotificationWorker` and other places.


```kotlin
private fun broadcastIntent(type: NotificationType): PendingIntent {
    return PendingIntent.getActivity(
        applicationContext,
        0,
        type.toIntent(applicationContext),
        PendingIntent.FLAG_UPDATE_CURRENT or PendingIntent.FLAG_IMMUTABLE,
    )
}
```

Introduce a new deep link type and implement it for the appropriate deep links.

```kotlin
sealed interface NotificationDeepLink : IntentableDeepLink {
    val analyticsCategory: String
    val analyticsType: String?
    
    fun toBaseIntent(context: Context): Intent
    
    override fun toIntent(context: Context) = toBaseIntent(context)
        .putExtra("au.com.shiftyjelly.pocketcasts.extras.notification.category", analyticsCategory)
        .putExtra("au.com.shiftyjelly.pocketcasts.extras.revamped.notification.type", analyticsType)
}
```

And then either launch reporting from the `MainActivity`, where this event will be received, or add a callback to the `DeepLinkFactory` type.

```kotlin
// In MainActivity
val deepLink = deepLinkFactory.create(intent)
if (deepLink is NotificationDeepLink) {
  // Send analytics
}

// In DeepLinkFactory
class DeepLinkFactory(
    val listeners: Collection<Listener> = emptySet(),
) {
  fun create(intent: Intent): DeepLink? {
    val deepLink = TODO("Current implementation")
    listeners.forEach(Listener::onDeepLinksAvailable)
    return TODO("Current implementation")
  }

  interface Listener {
    fun onDeepLinksAvailable(deepLinks: List<DeepLink>)
  }
}
```

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/notification/NotificationOpenReceiverActivity.kt
sztomek: good points, thanks!

sztomek:
thanks for the review and suggestion @MiSikora 🙏 
created a ticket that addesses your thoughts: https://linear.app/a8c/issue/PCDROID-100/rework-notification-open-tracker

PR #4369, opened 2025-08-15
Adds transcript sharing

geekygecko:
## Description

This change introduces transcript sharing to the app. At the top of the transcript from the episode card or the full screen player there is now a share button. Unfortunately, this doesn't just copy the transcript string to your clipboard as I found that this was getting clipped with even an episode 30 minutes long. 

I also fixed a minor typo in `isSuccsessful`.

Internal reference: p1755098401183599/1754958021.729059-slack-C083B1L7ZU7

## Testing Instructions
1. Find an episode with a transcript, such as the latest episode of The Daily. 
2. On the podcast page, tap on the episode row and then tap the "View transcript" button. 
3. Tap the share icon in the toolbar. 
4. ✅ Verify the file is shared and it contains a transcript.
5. Play the episode. 
6. Open the full screen player and tap the transcript button. 
7. Tap the share icon in the toolbar. 
8. ✅ Verify the file is shared.
9. ✅ Verify the analytic event was output `transcript_shared`

## Screenshots

|  |  |
| --- | --- |
| Screenshot_20250815_135433 | Screenshot_20250815_135440 | 
| Screenshot_20250815_141726 | Screenshot_20250815_140421 |

## Checklist
 
- [x] with different themes


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

modules/features/transcripts/src/main/kotlin/au/com/shiftyjelly/pocketcasts/transcripts/TranscriptSharingClient.kt
```diff
@@ -0,0 +1,15 @@
...
+
+interface TranscriptSharingClient {
+    suspend fun shareTranscript(request: SharingRequest): SharingResponse
+}
+
+internal fun SharingClient.asTranscriptClient() = object : TranscriptSharingClient {
+    override suspend fun shareTranscript(request: SharingRequest): SharingResponse {
+        return this@asTranscriptClient.share(request)
+    }
+}
```
geekygecko, line 15: I used the existing sharing logic, hopefully this was the right thing to do.

modules/features/transcripts/src/main/kotlin/au/com/shiftyjelly/pocketcasts/transcripts/di/TranscriptsModule.kt
```diff
@@ -0,0 +1,16 @@
...
+import au.com.shiftyjelly.pocketcasts.transcripts.asTranscriptClient
+import dagger.Module
+import dagger.Provides
+import dagger.hilt.InstallIn
+import dagger.hilt.components.SingletonComponent
+
+@Module
+@InstallIn(SingletonComponent::class)
+object TranscriptsModule {
+    @Provides
```
sztomek, line 14: 💡 this could be `@Singleton`

modules/features/transcripts/src/main/kotlin/au/com/shiftyjelly/pocketcasts/transcripts/TranscriptSharingClient.kt
sztomek: sounds reasonable to me

modules/features/transcripts/src/main/kotlin/au/com/shiftyjelly/pocketcasts/transcripts/di/TranscriptsModule.kt
MiSikora: It most likely shouldn't. Inexpensive, stateless objects that aren't created frequently have lower overall cost of instantiation than thread-safety locking mechanism that goes into making sure that something is a singleton instance.

MiSikora (commented):
It is probably unlikely to happen but should we disable sharing transcripts for private feeds?

geekygecko:
> It is probably unlikely to happen but should we disable sharing transcripts for private feeds?

As this feature is aimed at sharing for your own use, I might not disable this, but it's a good point. I will raise it with the team. I will also make a change in a separate PR to add a feature flag.

CookieyedCodes:
@geekygecko I don't think any private feeds are big enough to trigger the AI auto transcripts especially because patreons system generates feeds individually and every few months just my opinion tho

Issue #4116, opened 2025-06-12
Unable to change podcast order


ctdealba:
### Description

Changing the order in the podcast list view will revert after viewing a podcast detail page.

### Step-by-step reproduction instructions

1. Go to Podcasts
2. Drag and drop a podcast to change the order
3. View the detail page of a podcast
4. Return to Podcasts

### Screenshots or screen recording

https://github.com/user-attachments/assets/ab1c6e37-4cd0-4947-bed6-6e4bc6601f67

### Did you search for existing bug reports?

- [x] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

Platform: Phone - motorola - motorola edge (2022
App version: 7.90 (9344)
Android version: 14 SDK 34

PR #4390, opened 2025-08-19
Sync podcasts and folders using Protobuf

MiSikora:
## Description

This PR begins the migration of data sync from JSON to Protobuf. There are several reasons for this change:

- The current syncing code is difficult for me to understand. At some point, I’ll need to add Manual Playlist syncing, and the best way for me to handle it is by analyzing the code line by line. While doing so, I might as well improve it.
- When I benchmarked JSON vs. Protobuf (pdeCcb-4AQ-p2), the JSON payload was ~3× larger. This is a significant amount of data. Especially with Manual Playlists coming, which may sync a large volume of data. (p1755569658553379-slack-C093RV9N8DR)  
- Migrating away from JSON sync will allow us to drop a lot of legacy code once `PodcastSyncProcess` is removed.  
- `PodcastSyncProcess` is a mix of blocking code, RxJava, and coroutines. In the new approach, I want to rely solely on coroutines.  

### Migration Plan
- While I work on the migration, syncing will remain behind a disabled feature flag so that day-to-day development isn’t disrupted.  
- Once migration is complete, I’ll enable the feature flag in debug builds and connect it to Firebase.  
- After release to beta, I’ll perform a gradual rollout to monitor for sync issues.  
- If everything looks good, I’ll remove the old syncing code, leaving only the Protobuf-based syncing. If not I'll either need to make updates or retrofit manual playlists syncing to JSON. Though with the better understanding of the sync process it should be easier than in the current moment.

The current code is a bit crude. I haven’t added any logging yet, but I plan to improve this later. For now, I’m focused on the core application logic. Once that’s complete, I’ll add logging and better exception handling.  

Relates to PCDROID-119

## Testing Instructions

> [!NOTE]  
> Episode sort order doesn't sync due to https://github.com/Automattic/pocket-casts-sync-api/pull/1191

1. Sign in on devices A and B with the same account.  
2. Sync the devices.  
3. On both devices, go to Beta Features and enable the **`Sync data using Protobuf`** feature flag.  
4. Verify that subscribed podcasts sync.  
5. Verify that "skip first" and "skip last" podcast settings sync.  
6. Verify that the home grid sort order syncs.  
   - Note: Drag & drop syncing may be problematic due to #4116.  
7. Verify that folders sync.  
8. Verify that the folder sort order syncs.  
   - Note: Drag & drop may also be problematic for the same reason.  

## Checklist

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/data/PodcastSync.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/data/PodcastSync.kt
@@ -33,8 +33,8 @@ import kotlinx.coroutines.withContext
 internal class PodcastSync(
     private val podcastManager: PodcastManager,
     private val playbackManager: PlaybackManager,
+    private val missingPodcastsSemaphore: Semaphore,
 ) {
-    private val semaphore = Semaphore(permits = 10)
 
     suspend fun fullSync(serverPodcasts: List<UserPodcastResponse>) {
         val localPodcasts = podcastManager.findSubscribedNoOrder()
@@ -49,7 +49,7 @@ internal class PodcastSync(
         coroutineScope {
             localMissingUuids.mapNotNull(serverPodcastMap::get).forEach { serverPodcast ->
                 launch {
-                    semaphore.withPermit {
+                    missingPodcastsSemaphore.withPermit {
                         syncPodcast(serverPodcast)
                     }
                 }
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/data/DataSyncProcess.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/data/DataSyncProcess.kt
@@ -14,6 +14,7 @@ import java.time.Instant
 import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.sync.Semaphore
 import timber.log.Timber
 
 class DataSyncProcess(
@@ -23,7 +24,7 @@ class DataSyncProcess(
     private val playbackManager: PlaybackManager,
     private val settings: Settings,
 ) {
-    private val podcastSync = PodcastSync(podcastManager, playbackManager)
+    private val podcastSync = PodcastSync(podcastManager, playbackManager, missingPodcastsSemaphore = Semaphore(permits = 10))
     private val folderSync = FoldersSync(folderManager)
 
     suspend fun sync(): Result<Unit> {


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          4 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>DataSyncProcess</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>FoldersSync</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>PodcastSync</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>

</p>

modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/featureflag/Feature.kt
```diff
@@ -170,6 +170,14 @@ enum class Feature(
         hasFirebaseRemoteFlag = true,
         hasDevToggle = true,
     ),
+    PROTO_DATA_SYNC(
+        key = "proto_data_sync",
+        title = "Sync data using Protobuf",
+        defaultValue = true,
```
MiSikora:
```suggestion
        defaultValue = false,
```

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/PodcastDao.kt
```diff
@@ -245,6 +246,16 @@ abstract class PodcastDao {
     @Query("UPDATE podcasts SET sync_status = :syncStatus WHERE subscribed = 1")
     abstract suspend fun updateAllSubscribedSyncStatus(syncStatus: Int)
 
+    @Query("UPDATE podcasts SET sync_status = :syncStatus WHERE uuid IN (:uuids)")
+    protected abstract suspend fun updateAllSyncStatusUnsafe(syncStatus: Int, uuids: Collection<String>)
+
+    @Transaction
+    open suspend fun updateAllSyncStatus(syncStatus: Int, uuids: Collection<String>) {
+        uuids.chunked(AppDatabase.SQLITE_BIND_ARG_LIMIT - 1).forEach { chunk ->
```
sztomek, line 254: ❓ comparing this to `FolderDao.deleteAll`, why is `-1` necessary?

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/AppDatabase.kt
```diff
@@ -164,6 +164,8 @@ abstract class AppDatabase : RoomDatabase() {
     }
 
     companion object {
+        const val SQLITE_BIND_ARG_LIMIT = 999
```
sztomek, line 168: ❓ is there a physical limitation of SQLite that I didn't know of? Maybe it would worth adding an expl comment just in case.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/data/PodcastSync.kt
```diff
@@ -0,0 +1,178 @@
...
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.sync.Semaphore
+import kotlinx.coroutines.sync.withPermit
+import kotlinx.coroutines.withContext
+
+internal class PodcastSync(
+    private val podcastManager: PodcastManager,
+    private val playbackManager: PlaybackManager,
+) {
+    private val semaphore = Semaphore(permits = 10)
```
sztomek: ⚠️  I might be reading it wrong, but this `Semaphore` is only used in `fullSync`.
I think it's a bit error prone to have this semaphore inited with 10 permits, while the `PodcastSync` instance may be used for multiple `fullSync` calls.
Right now fullSync is not meant to run in parellel and you create new PodcastSync instances for the syncs, but this may change in the future 🤷 
Also, there's no guarantee that `localMissingUuids.mapNotNull(...).forEach { ... }` will only yield <10 items.
Or did i miss something?

sztomek (commented):
The code looks good, but I'm experiencing some odd behaviors around podcasts sync.

I'm logged in with the same account on my Pixel 6 and on an emulator, bot have the FF enabled.
On this account I have 70+ followed podcasts.
On the emulator, I've tried unfollowing a podcast, but it takes a lot of time until the tile disappears from the podcasts grid. And after ~30 seconds, when I pull-refreshed the grid, it appeared again (meanwhile I did nothing on the Pixel 6).
I think it shouldn't have reappeared.

https://github.com/user-attachments/assets/ec1da305-89df-41dc-8c06-faf5421d2427

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/AppDatabase.kt
MiSikora: It is a standard SQLite limit - https://sqlite.org/limits.html. I'll add a link to SQLite documentation.

> To prevent excessive memory allocations, the maximum value of a host parameter number is SQLITE_MAX_VARIABLE_NUMBER, which defaults to 999 for SQLite versions prior to 3.32.0 (2020-05-22) or 32766 for SQLite versions after 3.32.0.

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/PodcastDao.kt
MiSikora: Because query `UPDATE podcasts SET sync_status = :syncStatus WHERE uuid IN (:uuids)` uses `:sync_status` as a binding argument leaving 998 arguments.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/data/PodcastSync.kt
MiSikora: I didn't think of it because it is an `internal` class. But I'll move the semaphore to `DataSyncProcess` and inject it here.

I don't understand this part:

> Also, there's no guarantee that `localMissingUuids.mapNotNull(...).forEach { ... }` will only yield <10 items.
Or did i miss something?

The point of 10 permits is for cases where the expression has more than 10 items. What is the problem?

MiSikora:
> The code looks good, but I'm experiencing some odd behaviors around podcasts sync.

Thanks for catching that. I'll take a look. I haven't considered that pull-to-refresh might be doing something here related to syncing.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/data/PodcastSync.kt
sztomek: ummm so 10 concurrent execution of the critical section is what you've planned for?

MiSikora: We use 5 currently.

https://github.com/Automattic/pocket-casts-android/blob/ceb9b511acda8c8429c0f9e3632aab860001bf56/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/PodcastSyncProcess.kt#L210-L219

MiSikora:
Ok so the issue with unfollowing was resolved in 53fa7e99edc1aea554597292ac2b8141d4193ece. I mistakenly changed the delete episodes function to be suspending instead of fire-and-forget. This caused unfollowing to wait until downloaded episodes were deleted.

Thanks for catching the bug! 🙏

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/AppDatabase.kt
MiSikora: [d47c4b1](https://github.com/Automattic/pocket-casts-android/pull/4390/commits/d47c4b19acb346617ab51161273006b2a8e453de)

sztomek (approved):
Worked well, thanks for fixing the issue.

PR #4417, opened 2025-08-26
[Onboarding] Respect user's interests on Recommendations

sztomek:
## Description
Prioritize selected interest categories on the recommendations page.
When the user has selected at least 3 interest categories, we should promote them to the top of the recommendations list. Promoted categories will be followed by the regular recommendations (Featured, Trending, and the rest of discover categories)
If there aren't any interests selected, or the FF is disabled, we should return the usual recommendations (Featured, Trending, discover categories)
Bonus: Added tests to verify this behavior

Fixes https://linear.app/a8c/issue/PCDROID-131/adapt-to-interests-on-recommendations

## Testing Instructions
1. build `debug` so the FF will be enabled
2. On the intro carousel, select Get started
3. On the interests screen, try two scenarios:
- A - tap Not now
- B - select at least 3 then continue
4. Observe recommendations

## Screenshots or Screencast 

https://github.com/user-attachments/assets/af76ee8a-f53b-4cd7-a154-3506a05d38e6

## Checklist

### DIFF
--- a/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/recommendations/OnboardingInterestsViewModel.kt
+++ b/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/recommendations/OnboardingInterestsViewModel.kt
@@ -5,11 +5,11 @@ import androidx.lifecycle.viewModelScope
 import au.com.shiftyjelly.pocketcasts.repositories.categories.CategoriesManager
 import au.com.shiftyjelly.pocketcasts.servers.model.DiscoverCategory
 import dagger.hilt.android.lifecycle.HiltViewModel
-import javax.inject.Inject
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
+import javax.inject.Inject
 import au.com.shiftyjelly.pocketcasts.localization.R as LR
 
 @HiltViewModel
@@ -29,7 +29,10 @@ class OnboardingInterestsViewModel @Inject constructor(
         fetchCategories()
     }
 
-    fun skipSelection() {
+    fun skipSelection(doWhenFinished: () -> Unit) {
+        viewModelScope.launch {
+            doWhenFinished()
+        }
     }
 
     fun updateSelectedCategory(category: DiscoverCategory, isSelected: Boolean) {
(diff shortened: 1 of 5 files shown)


modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/recommendations/OnboardingInterestsViewModel.kt
```diff
@@ -29,7 +29,11 @@ class OnboardingInterestsViewModel @Inject constructor(
         fetchCategories()
     }
 
-    fun skipSelection() {
+    fun skipSelection(doWhenFinished: () -> Unit) {
+        viewModelScope.launch {
+            categoriesManager.setInterestCategories(emptySet())
+            doWhenFinished()
+        }
```
MiSikora: ⚠️ This is an erroneous pattern when communicating between UI and ViewModel. It can lead to memory leaks, missing events, or even crashes. See this code:

```kotlin
viewModelScope.launch {
  longOperation()
}
uiCallback()
```

When this code is triggered and the UI is destroyed while `longOperation` is running, the coroutine won’t be cancelled. It also might not be cancelled when a `ViewModel` is destroyed if `longOperation` is not cooperative. This causes a memory leak when UI references are captured in the `doWhenFinished` callback. After that, the callback executes for a UI component that is no longer part of the view hierarchy. Depending on what `doWhenFinished` does, this will most often lead to problems.  

Fortunately, this can be easily changed in this instance without introducing any of the communication patterns for one-off events. Since `setInterestCategories` is not a suspending function and not an expensive call, there’s no need to launch it in a coroutine or use a callback. Everything can be executed synchronously from the UI.  

```kotlin
viewModel.skipSelection()
onShowRecommendations()
```

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/recommendations/OnboardingInterestsViewModel.kt
```diff
@@ -52,16 +56,21 @@ class OnboardingInterestsViewModel @Inject constructor(
         }
     }
 
-    fun saveInterests() {
+    fun saveInterests(doWhenFinished: () -> Unit) {
```
MiSikora: The communication issue applies here as well.

modules/features/account/src/test/java/au/com/shiftyjelly/pocketcasts/account/onboarding/recommendations/OnboardingRecommendationsStartPageViewModelTest.kt
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    @Test
+    fun `should prioritize interest categories when FF is on and interests are set`() = runTest {
+        FeatureFlag.setEnabled(Feature.NEW_ONBOARDING_RECOMMENDATIONS, true)
+        whenever(categoriesManager.interestCategories).thenReturn(MutableStateFlow(mockCategories.takeLast(3).toSet()).asStateFlow())
+
+        val viewModel = createViewModel()
+        val state = viewModel.state.value
+        // assert(state.sections.take(3).map { it.title }.all { it in mockCategories.takeLast(3).map { it.title } })
```
MiSikora: ⛏️ Can be removed.

MiSikora (approved):

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/recommendations/OnboardingInterestsViewModel.kt
sztomek: You're right, thanks for pointing out!

PR #4421, opened 2025-08-26
Keep the GitHub Release of first beta as draft

AliSoftware:
So that the Release Manager can first smoke-test it before publishing it to the public

See https://linear.app/a8c/issue/AINFRA-1108/make-beta-gh-releases-publishing-safer

## Description

Fixes # 

## Checklist
 
- [ ] with different themes

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -494,6 +494,8 @@ platform :android do
     version = version_name_current
     build_code = build_code_current
     is_beta = beta_version?(version)
+    is_first_beta = version.end_with?('-rc-1')
+
     unless skip_prechecks
       # Match branch names that begin with `release/`
       # Skip this check on CI because that action relies on CI env vars, but when this lane is called as part of `code_freeze` the branch changed since the CI build started, so that wouldn't work
@@ -535,7 +537,10 @@ platform :android do
       release_assets << signed_apk_artifact_path
     end
 
-    create_gh_release(version: version, prerelease: is_beta, release_assets: release_assets.compact) if create_gh_release
+    # - If this is the first beta release, keep the GitHub release as draft so the release manager has time to smoke-test it before publishing it.
+    # - For any subsequent intermediate beta, we can publish it immediately though
+    # - For the final release, also create the GitHub Release as a draft so that we only publish it after Google has approved it and the release is live
+    create_gh_release(version: version, prerelease: is_beta, is_draft: is_first_beta || !is_beta, release_assets: release_assets.compact) if create_gh_release
   end
 
   # Builds and bundles the given app
@@ -673,14 +678,18 @@ platform :android do
   #
   # @param version [Hash<String>] The version to create. Expects keys "name" and "code"
   # @param prerelease [Bool] If true, the GitHub Release will have the prerelease flag
+  # @param is_draft [Bool|nil] If true, the GitHub Release will be created as a draft.
+  #        If nil, the GitHub Release will be published immediately if `prerelease` is true (beta version),
+  #        and created as a draft if `prerelease` is false (final version).
+  # @param release_assets [Array<String>] The assets to attach to the GitHub Release
   #
-  private_lane :create_gh_release do |version:, prerelease: false, release_assets: []|
+  private_lane :create_gh_release do |version:, prerelease: false, is_draft: nil, release_assets: []|
     create_github_release(
       repository: GITHUB_REPO,
       version: version,
       release_notes_file_path: EXTRACTED_RELEASE_NOTES_PATH,
       prerelease: prerelease,
-      is_draft: !prerelease,
+      is_draft: is_draft || !prerelease,
       release_assets: release_assets.join(',')
     )
   end


fastlane/Fastfile
```diff
@@ -494,6 +494,8 @@ platform :android do
     version = version_name_current
     build_code = build_code_current
     is_beta = beta_version?(version)
+    is_first_beta = version.end_with?('-rc-1')
```
MiSikora: I think this is ok for now but will the the non-manual solution be more robust? Because what will happen in case of a crash is:

1. `-rc-1` crashes.
2. We don't publish it.
3. We fix it.
4. We build `-rc-2` and it gets published without creating a draft first.

At this stage it hopefully shouldn't crash and devs were diligent enough to test their fixes with the release builds across all platforms before triggering a new beta. But Murphy's law says otherwise.

AliSoftware: Right.

How would RC-2 be created though? Would you use the "➕ Add Beta" button in ReleasesV2 to create it? If so, how would we know to consider this RC-2 to still be a beta that needs to still be draft with manual publishing and not an intermediate beta later in the process that can be auto-published?

Or would you create that new beta to fix RC-1 in some other way in ReleasesV2, that would allow us to know it's still one we'd need to publish as draft?

If there is no way to distinguish between "a beta build that is done as part as trying to fix the beta from the code freeze until we get it right" and "a regular intermediate beta later in the release process", that means we might have to create _all_ GitHub releases as draft first and add a task in the release scenario to publish it after manual smoke testing in all cases (not just first beta but also all intermediate betas).
I'd be ok with this, but it's just that the initial request in Linear was mentioning not needing that for intermediate betas and only for the first one… but I'm wondering how we can distinguish the cases then 🤔

MiSikora: Yeah, I would probably use the MC tool to create a new RC. I'm not claiming to have a solution to the problem, just pointing this out.

I'm not opposed to making sure that each beta is first published as a draft. Initially I though it might be overzealous to do it but if there is no good way to distinguish such builds from first beta and makes it easier on your end I think it is a viable option.

AliSoftware: Ok, let's make every beta create the GH release as draft (and add a step in ReleasesV2 to publish it when ready), even on Intermediate betas then.

AliSoftware:
I have now updated that PR to:
 - Have all GitHub Releases be created as draft (for all betas, as well as for the final release)
 - Have the `release-pipelines/update-rollout.yml` CI pipeline be able to _also_ call the `publish_github_release` lane if `$PUBLISH_DRAFT_GITHUB_RELEASE == true`
 - Get rid of the `release-pipelines/publish_release.yml` now that this feature is integrated in the `update-rollout.yml` pipeline

The ideas is that, on the ReleasesV2 scenario, we'll now be able to piggy-back on the existing scenario Tasks that were calling `update_rollout` to also publish the Draft GitHub Release at the same time:
 - For the tasks that was updating the rollout of betas from 0% to 20% (during code freeze or for intermediate betas) after having smoke-tested the APKs, those will now also pass `PUBLISH_DRAFT_GITHUB_RELEASE=true` to the CI job so that it also publishes the `X.Y-rc-N` GitHub (pre-)Release that was created as draft
 - For the tasks that are about updating the rollout of an already-published-with-partial-rollout build to higher percentages (e.g. rollout from 20% or 50% to 100%), those would still keep `PUBLISH_DRAFT_GITHUB_RELEASE=false`, as there are no draft GitHub Release to publish in those cases
 - For the "Release" Monday task about updating the rollout of the production build from 1% (from when the build was submitted for review on Friday) to 10% after Google has approved the build over the weekend, we also pass `PUBLISH_DRAFT_GITHUB_RELEASE=true` to the CI job so that it also publishes the (non-prerelease) `X.Y` GitHub Release that was created as draft on the previous Friday.

iangmaia (approved):
👍

AliSoftware:
> 👍 

Actually @iangmaia FYI after some good insights from @MiSikora in 191438-ghe-Automattic/wpcom/files/0e3f3f0a43f661ad286daf1462dc90a6e835fd3f#r183074 I'm planning to adjust this PR a bit to have the `update-rollouts.yml` pipeline create 2 separate jobs/`steps` (one for each lane to be run) instead of running both lanes in the same single job, so that in the (hopefully unlikely and rare case) `fastlane update_rollout` call fails, at least the call to `publish_github_release` will still be executed. So I might request another review for you after I've made that change 😛  (I'll work on it right now)

This is going to slightly increase the time this build will take—especially because we only have one agent in the `tumblr-metal` queue, so only one job can run at a time on that queue meaning the 2 jobs will be sequential not in parallel—but I don't think that will amount to that much of an overhead and would still be worth it to protect us against the failure case.

.buildkite/release-pipelines/update-rollouts.yml
```diff
@@ -8,9 +8,6 @@
 steps:
   - label: "🚂 Update Rollouts"
     command: |
-      echo '--- 🤖 Use bot for Git operations'
-      source use-bot-for-git
-
```
AliSoftware, line 13: See rationale in commit message of e08cd021e74d4cddee6de55cc7775452575de0e3

TL;DR the update or rollouts doesn't do any `git push`, so this shouldn't be necessary in practice

.buildkite/release-pipelines/update-rollouts.yml
```diff
@@ -26,3 +23,19 @@ steps:
       manual:
         allowed: false
         reason: "Always prefer re-triggering a new build from ReleaseV2 rather than retrying the individual job from Buildkite"
+
+  - label: ":github: Publish Draft GitHub Release"
+    if: build.env("PUBLISH_DRAFT_GITHUB_RELEASE") == "true"
+    command: |
+      echo '--- :ruby: Setup Ruby Tools'
+      install_gems
+
```
AliSoftware, line 36: See rationale in commit message of e08cd021e74d4cddee6de55cc7775452575de0e3 for not calling `use-bot-for-git` nor calling `bundle exec fastlane configure_apply`

TL;DR This job doesn't do any `git push` and nor does it use any secret from `.configure-files/`, as it's only using the GitHub API (and the `GITHUB_TOKEN` is provided as a CI secret, not as an encrypted file managed by `configure`)

.buildkite/release-pipelines/update-rollouts.yml
```diff
@@ -26,3 +23,19 @@ steps:
...
+      echo '--- 🚀 Publish Draft GitHub Release'
+      bundle exec fastlane publish_gh_release skip_confirm:true track:"$TRACK"
+    plugins: [$CI_TOOLKIT]
+    agents:
+      queue: "tumblr-metal"
```
AliSoftware, line 41: Given that this job doesn't do any `git push`, technically it doesn't really need to run on a trusted agent like `tumblr-metal` like other `release-pipelines/*` do.

So in practice I guess we could also consider making this "Publish Draft Github Release" job use the `android` queue instead of `tumblr-metal` (and actually, same for the `Update Rollouts` job above)

I decided to keep using `tumblr-metal` queue here anyway:
 - Both for consistency with all other `release-pipelines/*` running on that trusted agent
 - But also because the jobs that are run on our `tumblr-metal` agent are typically very short jobs that terminate fast, while most jobs on the `android` queue would take longer to finish (e.g. jobs running Unit Tests), and thus take longer to free the agents for the next jobs in the queue like the ones from this pipeline.
    - Sure on one hand we have only one agent in the `tumblr-metal` queue so we can only run one job on that `tumblr-metal` queue at a time, while we have a lot of agents in our `android` queue (with auto-scaling enabled too).
    - But on the other hand, the wait time for jobs targeting `tumblr-metal` is currently very short (because most jobs it runs are short ones), and it already has the repo cloned locally which makes the `checkout` stage of the job very fast (we do have git caching on our `android` agents too, but not as efficient since it still requires downloading cache from S3 and EC2 instances can be scaled down between jobs requiring cache to be redownloaded, etc)

So overall I think that even if using a trusted agent (like `tumblr-metal`) is no longer a _requirement_ for those 2 jobs anymore, it still makes sense to keep using them in this context anyway.

wdyt @iangmaia ?

fastlane/Fastfile
```diff
@@ -420,27 +420,29 @@ platform :android do
...
-  # This lane publishes a release on GitHub and creates a PR to backmerge the current release branch into the next release/ branch
+  # This lane publishes a release on GitHub. If `track` is `production`, it will also delete the release branch.
   #
   # @param [Boolean] skip_confirm (default: false) If set, will skip the confirmation prompt before running the rest of the lane
+  # @param [String] track (default: 'production') The Google Play track the release was published to in Google Play Console. Must be either `beta` or `production`.
+  #        If 'production': The GitHub Release will be expected to have a title like `X.Y`, and this lane will also delete the release branch after publishing the GitHub Release.
+  #        If 'beta': The draft GitHub Release will be expected to have a title like `X.Y-rc-N`. The release branch will NOT be deleted for those.
   #
-  lane :publish_release do |skip_confirm: false|
+  lane :publish_gh_release do |skip_confirm: false, track: 'production'|
```
AliSoftware, line 430: ℹ️ Note: I had to rename this lane from `publish_github_release` (the name I chose to rename it to in the first iterations of that PR) to `publish_gh_release` because Fastlane doesn't allow you to have a lane with the same name as an existing action (and indeed we have a fastlane action already using the name `publish_github_release`… and in fact that's the action that this lane calls on line 450)

.buildkite/release-pipelines/update-rollouts.yml
iangmaia: Makes sense 👍

.buildkite/release-pipelines/update-rollouts.yml
iangmaia: > consistency with all other release-pipelines/* running on that trusted agent

> the wait time for jobs targeting tumblr-metal is currently very short (because most jobs it runs are short ones), and it already has the repo cloned locally which makes the checkout stage of the job very fast

> So overall I think that even if using a trusted agent (like tumblr-metal) is no longer a requirement for those 2 jobs anymore, it still makes sense to keep using them in this context anyway

I agree 👍
It would be nice if at some point we had a queue focused on release jobs (which is in practice what `tumblr-metal` is currently doing), so these questions would always have a simple direct answer.

.buildkite/release-pipelines/update-rollouts.yml
iangmaia: ⚠️ Oh, wait a second. It does a `Fastlane::Helper::GitHelper.delete_remote_branch_if_exists!(current_branch)`, so it will require write git permission, right? 🤔

AliSoftware: Ooooh, great catch!! 🙇

iangmaia: So I believe indeed `remove_branch_protection` and `publish_github_release` will indeed only require the GitHub API `GITHUB_TOKEN`, but `Fastlane::Helper::GitHelper.delete_remote_branch_if_exists!` will require push permission.

AliSoftware: Should be fixed in bdbf8bf92e89feb6f95fd9979c52e58a1326a2f4

iangmaia (approved):
🚀

AliSoftware:
I just had a thought that merging this so we can adopt it for the next release scenario `7.97`… could have risked breaking the tasks that will still have to be run for current scenario of `7.96`… But in practice I think that's ok to merge this to `main` even before `7.96` is done because:

 - When [the "Update Rollouts" tasks of the `7.96` scenario](https://mc.a8c.com/releases-v2/release.php?product=PCAndroid&version=7.96#task-b200aec9-4b7e-4eaf-aab2-b5923d761a6c-fragment) will be run:
    - It will trigger the `release-pipelines/update-rollouts.yml` pipeline on the `main` branch—thus using the updated .yml from this PR that'd be merged to `main` by then—but without passing the `PUBLISH_DRAFT_GITHUB_RELEASE` env var at all to that call.
    - But that's not a problem, because the changes we made to `release-pipelines/update-rollouts.yml` in this PR will make it behave the same as before that PR when it's triggered without the `PUBLISH_DRAFT_GITHUB_RELEASE` env var being provided. So even if that task will run the pipeline from `main`, it should still do what's needed for `7.96`
 - When the ["Publish the release" task of the `7.96` scenario](https://mc.a8c.com/releases-v2/release.php?product=PCAndroid&version=7.96#task-ff427540-bbe5-43ed-a1f0-6d9a3e5656d0-fragment) will be run:
    - It will trigger the `release-pipelines/publish-release.yml` pipeline… which won't exist anymore on `main`… but luckily that task triggers a build for this pipeline on the `release/*` branch, not on `main`. And on that `release/7.96` branch that `publish-release.yml` pipeline will still exist, so that should still work

So given that I think it's ok to merge this into main and reset the `7.97` scenario to use that new approach, as it shouldn't disrupt the remaining tasks of the `7.96` scenario anyway.

PR #4425, opened 2025-08-27
[Onboarding] Add `popularity` to categories

sztomek:
## Description
We've updated the `categories_v2.json` to include the optional `popularity` field with `int` type.
We should only display categories for interests where `popularity` is present, and the interests should be sorted by the popularity value.
This PR does that.

Ctxt: p1756212471672129/1755858078.317389-slack-C0932TFPUDC

+ bonus: fixed a small animation glitch on the category pills during their animation: their horizontal edges may got clipped in case of long rows.

Fixes https://linear.app/a8c/issue/PCDROID-136/update-data

## Testing Instructions
1. build `debug` version of the app so the FF will be enabled
2. Get started
3. Check the [categories json](https://static.pocketcasts.net/discover/json/categories_v2.json) and compare what you see on the ui with what the server sends.

## Screenshots or Screencast 
(animation fix)
| Before | After |
| --- | --- |  
| <video src="https://github.com/user-attachments/assets/c99a41bb-1d1c-4cde-b8b3-7a6a8e97e0e6"/> | <video src="https://github.com/user-attachments/assets/fa470630-6d50-467c-af24-3d5eed70ef0c" /> |

## Checklist


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/categories/CategoriesManager.kt
```diff
@@ -93,6 +94,12 @@ class CategoriesManager @Inject constructor(
         }
     }.distinctUntilChanged()
 
+    val availableInterestCategories = discoverCategories.asStateFlow()
+        .map { categories ->
+            categories.filter { it.popularity != null }
+                .sortedBy { it.popularity }
+        }
```
MiSikora: `.asStateFlow()` is unnecessary here since the flow is being later mapped and loses its State type. If you want/need to keep it as State it can be `discoverCategories.map().stateIn(scope, /* appropriate configuration */)`

geekygecko (approved):

PR #4436, opened 2025-08-29
Fix syncing individual podcasts

MiSikora:
## Description

This change prevents the sync process from failing if subscribing to a podcast fails. Instead, syncing will continue, matching the behavior of the old sync process. The issue was discovered during staging sync, where some podcasts were missing, but it is entirely possible for such calls to just fail.

Relates to PCDROID-119

## Testing Instructions

Code review should be enough.

## Checklist


geekygecko (approved):
My dodgy staging account synced. Thanks for fixing this.

PR #4450, opened 2025-09-03
Create deeplink for beta settings

sztomek:
## Description
Small utility/QoL change: adding a new deeplink that opens the beta settings screen.
It would be useful to toggle some FFs on the Prod app, alas this PR.

Fixes https://linear.app/a8c/issue/PCDROID-144/add-secret-deeplink-to-beta-settings

## Testing Instructions
1. Download a notes app that supports markups (e.g. Obisidian)
2. Add this entry: `[Beta settings](pktc://beta_settings)` 
3. Save the note and tap the link
4. ✅  app should open the beta settings

Or alternatively via adb:
`adb shell am start -a android.intent.action.VIEW -d pktc://beta_settings`

## Screenshots or Screencast 

https://github.com/user-attachments/assets/7356974b-d9fe-4ad2-830c-f21a351710c9

## Checklist

### DIFF
--- a/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/deeplink/DeepLinkFactoryTest.kt
+++ b/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/deeplink/DeepLinkFactoryTest.kt
@@ -1166,4 +1166,14 @@ class DeepLinkFactoryTest {
 
         assertEquals(DeveloperOptionsDeeplink, deeplink)
     }
+
+    @Test
+    fun betaSettings() {
+        val intent = Intent(ACTION_VIEW)
+            .setData(Uri.parse("pktc://beta_settings"))
+
+        val deeplink = factory.create(intent)
+
+        assertEquals(BetaSettingsDeeplink, deeplink)
+    }
 }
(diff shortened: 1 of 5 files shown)


MiSikora (commented):
This won't work properly. Once a feature flag has `hasFirebaseRemoteFlag` set to `true`, it won't respond to toggling in production because of how `FirebaseRemoteFeatureProvider` is implemented.

Even if this is changed, I'm not sure about allowing users to change feature flags in production just by visiting the `pktc://beta_settings` link. Could you clarify which workflow we want to make easier or simplify? That way, we can discuss other solutions instead of giving users a skeleton key in production.

sztomek:
Nice, adding @geekygecko to the conversation

geekygecko:
> Could you clarify which workflow we want to make easier or simplify?

I would like to be able to turn on a feature flag in the release version of the app so I can test it while using it as my everyday device. For example, I want to try keeping the media session active even when playback has stopped to see if the controls in my car aren't greyed out. I wonder if this would be solved by your suggestion to start using Firebase builds. 

Thanks for making this change @sztomek , hopefully I haven't wasted your time.

MiSikora:
> I wonder if this would be solved by your suggestion to start using Firebase builds.

I see what you mean. Firebase builds probably wouldn't be convenient for the scenario you described. As far as I can tell, the approach in this PR is indeed the most straightforward way to enable that. That said, I do want to note my concern about making it possible to manipulate feature flags in production builds from within the app. From my perspective, that isn't something that should be available.

Regardless of that point, this change currently doesn't work because of the `FeatureProvider` behavior I mentioned earlier. I think we should address that first before merging. Otherwise feature flag values can't actually be changed in release builds.

geekygecko:
> From my perspective, that isn't something that should be available.

I think you are right. Let's close this PR for now.

PR #4475, opened 2025-09-10
Do not cleanup manual playlists data

MiSikora:
## Description

This adapts our removal podcast and episode clean up logic to account for episodes in manual playlists.

Closes PCDROID-143

## Testing Instructions

1. Test with `debugProd`.
2. Start the app in a clean state without being signed in.
3. Follow the [Radiolab](https://pca.st/radiolab) podcast.
4. Create a manual playlist.
5. Add some Radiolab episodes to the playlist.
6. Unfollow Radiolab.
7. Return to the playlist.
8. Episodes should still be available.
9. Create an account.
10. Go to Discover.
11. Follow any podcast other than Radiolab.
12. Execute the query.
```sql
UPDATE podcasts SET added_date = 1600000000000 WHERE uuid IS 'f5b97290-0422-012e-f9a0-00163e1b201c'
```
13. Go to the Profile tab and sync the account.
14. Return to the playlists.
15. Episodes should still be available.

## Checklist

### DIFF
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/PlaylistDao.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/PlaylistDao.kt
@@ -136,6 +136,26 @@ abstract class PlaylistDao {
 
     @Query(
         """
+        SELECT DISTINCT episode.podcast_uuid
+        FROM playlists AS playlist
+        JOIN manual_playlist_episodes AS episode ON episode.playlist_uuid IS playlist.uuid
+        WHERE playlist.deleted IS 0 AND playlist.manual IS NOT 0
+    """,
+    )
+    abstract suspend fun getPodcastsAddedToManualPlaylists(): List<String>
+
+    @Query(
+        """
+        SELECT DISTINCT episode.episode_uuid
+        FROM playlists AS playlist
+        JOIN manual_playlist_episodes AS episode ON episode.playlist_uuid IS playlist.uuid
+        WHERE playlist.deleted IS 0 AND playlist.manual IS NOT 0
+    """,
+    )
+    abstract suspend fun getEpisodesAddedToManualPlaylists(): List<String>
+
+    @Query(
+        """
         SELECT manual_episode.*
         FROM playlists AS playlist
         JOIN manual_playlist_episodes AS manual_episode ON manual_episode.playlist_uuid IS playlist.uuid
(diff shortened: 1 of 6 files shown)


modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/PlaylistDao.kt
```diff
@@ -134,6 +134,26 @@ abstract class PlaylistDao {
     @Query("SELECT episode_uuid FROM manual_playlist_episodes WHERE playlist_uuid IS :playlistUuid")
     abstract suspend fun getManualPlaylistEpisodeUuids(playlistUuid: String): List<String>
 
+    @Query(
+        """
+        SELECT DISTINCT episode.podcast_uuid
+        FROM playlists AS playlist
+        JOIN manual_playlist_episodes AS episode ON episode.playlist_uuid IS playlist.uuid
+        WHERE playlist.deleted IS 0 AND playlist.manual IS NOT 0
```
geekygecko, line 142: As `manual` is a boolean is there a reason it's not just comparing to 1?

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/PodcastManagerImpl.kt
```diff
@@ -80,30 +82,26 @@ class PodcastManagerImpl @Inject constructor(
...
-                podcast.autoArchiveAfterPlaying = AutoArchiveAfterPlaying.defaultValue(context)
-                podcast.autoArchiveInactive = AutoArchiveInactive.Default
-                podcast.autoArchiveEpisodeLimit = null
-                podcast.overrideGlobalArchive = false
-                podcast.folderUuid = null
-                podcastDao.updateSuspend(podcast)
-            } else {
-                // if they aren't signed in, just blow it all away
-                podcastDao.delete(podcast)
-                episodeDao.deleteAll(episodes)
```
geekygecko, line 105: Has this been removed, as the podcast could still be being displayed on the page?

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/PlaylistDao.kt
MiSikora: Both could be used. `IS NOT 0` is a safer comparison because it works in cases where `manual` for whatever reason has value greater than `1`. In SQLite it is till represented as integer not a boolean. Do you want me to change it?

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/PodcastManagerImpl.kt
MiSikora: I'm not sure I understand the question. Which page?

But I removed it because it interfered with manual playlists and episodes that are added from such podcasts.

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/PlaylistDao.kt
geekygecko: If you prefer this way, then that's fine with me. I was thinking it should never be another value so we could simplify the condition.

PR #4480, opened 2025-09-11
Handle archived episodes in manual playlists

MiSikora:
## Description

This adds ability to show and hide archived episodes in manual playlists.

Designs: 5sC4z4Mu42LvL4MAAIbQVi-fi-2606_25672

Closes PCDROID-114

## Testing Instructions

1. Create a manual playlist.
2. Add episodes to the playlist.
3. Archive some episodes.
4. The archive status bar should be visible.
5. Tap the archive bar button to toggle between showing and hiding archived episodes.
6. Archive all episodes.
7. Set the playlist to hide archived episodes.
8. The archived empty state should be visible.
9. Tap the CTA button.
10. All episodes should be visible.
11. Archive some episodes.
12. Toggle the playlist to show archived episodes.
13. Tap the "Play All" button.
14. All archived episodes should become unarchived.

## Screenshots or Screencast 

| Show archived | Hide archived | All archived |
| - | - | - |
| image | image | image |

## Checklist
 
- [x] with different themes

### DIFF
--- a/modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/playlists/component/PlaylistHeader.kt
+++ b/modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/playlists/component/PlaylistHeader.kt
@@ -113,9 +113,9 @@ internal data class PlaylistHeaderData(
     val isNoEpisodeShown = metadata.totalEpisodeCount != 0 && metadata.displayedEpisodeCount == 0
 
     val isNoArchivedEpisodeShown = metadata.archivedEpisodeCount > 0 &&
-            metadata.displayedEpisodeCount == 0 &&
-            !metadata.isShowingArchived &&
-            metadata.totalEpisodeCount == metadata.archivedEpisodeCount
+        metadata.displayedEpisodeCount == 0 &&
+        !metadata.isShowingArchived &&
+        metadata.totalEpisodeCount == metadata.archivedEpisodeCount
 }
 
 internal data class PlaylistHeaderButtonData(


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

modules/services/localization/src/main/res/values/strings.xml
```diff
@@ -974,6 +974,14 @@
...
     <string name="playlist_episode_unavailable_description">The podcast creator deleted this episode. It will stay in your playlist until you remove it.</string>
     <string name="playlist_episode_unavailable_cta">Remove from playlist</string>
+    <string name="playlist_episodes_archived_count">%1$d archived</string>
+    <string name="playlist_episodes_archived_title">No Episodes</string>
+    <plurals name="playlist_episodes_archived_description">
+        <item quantity="one">Your episode in this playlist has been archived</item>
+        <item quantity="other">All %1$d episodes of this playlist have been archived</item>
+    </plurals>
+    <string name="playlist_episodes_hide_archived">Hide Archived</string>
+    <string name="playlist_episodes_show_archived">Show Archived</string>
```
geekygecko: We have the same label on the podcast page with archived lowercase. It seems to match the Android OS not using capitals, so should we copy that?

```
<string name="show_archived">Show archived</string>
```

https://m3.material.io/foundations/content-design/style-guide/ux-writing-best-practices

Image

modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/playlists/component/PlaylistHeader.kt
```diff
@@ -432,25 +489,43 @@ private fun PlaylistSearchBar(
...
+            )
+            TextP50(
+                text = buttonLabel,
+                fontWeight = FontWeight.Medium,
+                color = MaterialTheme.theme.colors.primaryIcon01,
+                modifier = Modifier.clickable(
+                    indication = null,
+                    interactionSource = null,
+                    onClick = onClickButton,
                 ),
```
geekygecko: I believe this button is missing a press state.

MiSikora: It's deliberate.

modules/services/localization/src/main/res/values/strings.xml
MiSikora: Sorry, but it is too frustrating to figure out on my own when should I use capitalized label versions and when should I use not capitalized ones. There are many instances like this where we already have copies that are not capitalized and in designs for new screens they are capitalized. I think it would be best to highlight this to designers. I'll update this here but it becomes increasingly problematic.

MiSikora:
> My only comment is about the product decision after using it. It feels strange that if I’ve toggled on showing archived episodes, pressing “Play All” automatically unarchives everything.

Why does it feel strange? This acts in the same way as if you multi-selected multiple episodes and added them to the queue from the multi-select toolbar. Or if you tap "download all" in the podcast page. It doesn't skip archived episodes. If someone doesn't want to unarchive archived episodes they shouldn't show them when they tap "Play All", no?

modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/playlists/component/PlaylistHeader.kt
MiSikora: Oh, I see that on the podcast page it has one. But that one is too large. In any case I'll add one.

geekygecko:
> Why does it feel strange? This acts in the same way as if you multi-selected multiple episodes and added them to the queue from the multi-select toolbar. 

With multi-select you are choosing to pick the ones that haven't been played, and the ones that have. But with Play All, it's assuming you want to play the archived episodes. 

> Or if you tap "download all" in the podcast page. It doesn't skip archived episodes.

Good point. 🤔 

Maybe the real issue is that I don't think archived episodes should be in a manual playlist. Up Next lists don't have this behaviour, and these feel like a similar concept. I should probably be discussing this with the product team and not on a PR. Ignore me. Thanks for responding.

MiSikora:
> With multi-select you are choosing to pick the ones that haven't been played, and the ones that have. But with Play All, it's assuming you want to play the archived episodes.

Not necessarily. When you tap select all you don't really know what's down the list.

CookieyedCodes:
Just my 2 cents, but going off my user intuition, 
To me Play all, should behave like auto play and only play podcasts that are visible at the time of playing with a pop up saying do you want to wipe up next with a check box to make it a default behavior & if a list is empty a pop-up "do you want to unarchive played episodes & play all", 

multiselect also dose tend to only select episodes which is shown -yes even with all-, weather that be with episode search or with show/hide, I do want custom playlists to have archived podcasts but at least shown/hidden so the list is mainted & but I don"t want to relisten to episodes until a list is potentially finished as I want list hopping to be maintained

I think the download all, unarchives as downloading episodes unarchives them 😅

PR #4513, opened 2025-09-19
[CI] Add FAD Prototype Builds for `main` Branch

ParaskP7:
Closes: [AINFRA-1127](https://linear.app/a8c/issue/AINFRA-1127)

---

## Description

Goals:
- **Main**: This PR adds a new Firebase App Distribution (FAD) prototype builds job on CI, to run on-demand for when such builds are triggered on the `main` branch (usually after a PR gets merged). 18a1534622714c82fce60789d1fc68371911ab8c + dd714f30b9accf38cf59eae32121733789f32dd2
    - See build in action: [build](https://buildkite.com/automattic/pocket-casts-android/builds/13655/steps/canvas)
        - [Trigger](https://buildkite.com/automattic/pocket-casts-android/builds/13655/steps/canvas?sid=0199768b-6e6b-462d-8146-7d7f67f2127c) + [Prototype Builds (FAD)](https://buildkite.com/automattic/pocket-casts-android/builds/13655/steps/canvas?sid=0199768b-6e6e-45a4-9070-82152afdc458) + [FAD Releases](https://console.firebase.google.com/project/a8c-apps/appdistribution/app/android:au.com.shiftyjelly.pocketcasts/releases)
    - See build no action: [build](https://buildkite.com/automattic/pocket-casts-android/builds/13657/steps/canvas)
        - No trigger is available because we've now switched this functionality to `main` only. 86c540c98a16470b47f35fe623b328358cdc29b2
- **Secondary**: This PR also makes the existing (S3) prototype builds job run on-demand. 1ad4d96ae22e0a5cf8326171529a13e8a141fb20
    - See build in action: [build](https://buildkite.com/automattic/pocket-casts-android/builds/13655/steps/canvas)
        - [Trigger](https://buildkite.com/automattic/pocket-casts-android/builds/13655/steps/canvas?sid=0199768b-6e68-4377-9d20-217658cef68e) + [Prototype Builds](https://buildkite.com/automattic/pocket-casts-android/builds/13655/steps/canvas?sid=0199768b-6e6a-4573-9aaa-93387d9c4b6f)
    - See build no action: [build](https://buildkite.com/automattic/pocket-casts-android/builds/13657/steps/canvas)
        - [Trigger](https://buildkite.com/automattic/pocket-casts-android/builds/13657/steps/canvas?sid=019976c4-8cd9-4e03-ae39-9acf6b2c55c3) + Click and see `Share Prototype Builds?` dialog

---

#### Main Goal

[...]

PR #4518, opened 2025-09-22
Add prototype build type

MiSikora:
## Description

This PR introduces a new build type that uses all release optimizations but also includes development tooling, such as feature toggles. Unlike the `release` build type, it is not obfuscated.

Unfortunately, I couldn’t limit `prototype` to just the `:app` module and rely on the `release` matching fallback, because we use `BuildConfig` without any dependency injection, and properties in different modules reference their own generated static fields.

@ParaskP7 I decided to use the production app ID to enable communication with Play Services. I'm not sure if we’ll need to add a new `google-services.json` just for deployment to FAD. But if we do we can add the file under `./app/src/prototype/`, using the configuration exported from the A8C organization. I believe we can add it just like we handle configuration files for debug builds, without requiring `.mobile-secrets`.

Relates to AINFRA-1127

## Testing Instructions

1. `./gradlew :app:assemblePrototype && adb install ./app/build/outputs/apk/prototype/app-prototype.apk`
2. Open the app. You should have access to the debug panels in the settings.
3. `./gradlew :app:assembleRelease && adb install ./app/build/outputs/apk/release/app-release.apk`.
4. Open the app. You should not have access to the debug panels in the settings.

## Checklist

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -213,19 +213,6 @@ subprojects {
         checkReleaseBuilds = false
     }
 
-    val serverMainUrlProd = "\"https://refresh.pocketcasts.com\""
-    val serverApiUrlProd = "\"https://api.pocketcasts.com\""
-    val serverCacheUrlProd = "\"https://cache.pocketcasts.com\""
-    val serverCacheHostProd = "\"cache.pocketcasts.com\""
-    val serverStaticUrlProd = "\"https://static.pocketcasts.com\""
-    val serverSharingUrlProd = "\"https://sharing.pocketcasts.com\""
-    val serverListUrlProd = "\"https://lists.pocketcasts.com\""
-    val serverListHostProd = "\"lists.pocketcasts.com\""
-    val serverShortUrlProd = "\"https://pca.st\""
-    val serverShortHostProd = "\"pca.st\""
-    val serverWebPlayerHostProd = "\"play.pocketcasts.com\""
-    val webBaseHostProd = "\"pocketcasts.com\""
-
     plugins.withType<BasePlugin>().configureEach {
         configure<BaseExtension> {
             compileOptions {
@@ -256,18 +243,18 @@ subprojects {
                 buildConfigField("String", "META_APP_ID", "\"${project.property("metaAppId")}\"")
                 buildConfigField("String", "APPS_FLYER_KEY", "\"${project.property("appsFlyerKey")}\"")
 
-                buildConfigField("String", "SERVER_MAIN_URL", serverMainUrlProd)
-                buildConfigField("String", "SERVER_API_URL", serverApiUrlProd)
-                buildConfigField("String", "SERVER_CACHE_URL", serverCacheUrlProd)
-                buildConfigField("String", "SERVER_CACHE_HOST", serverCacheHostProd)
-                buildConfigField("String", "SERVER_STATIC_URL", serverStaticUrlProd)
-                buildConfigField("String", "SERVER_SHARING_URL", serverSharingUrlProd)
-                buildConfigField("String", "SERVER_SHORT_URL", serverShortUrlProd)
-                buildConfigField("String", "SERVER_SHORT_HOST", serverShortHostProd)
-                buildConfigField("String", "SERVER_WEB_PLAYER_HOST", serverWebPlayerHostProd)
-                buildConfigField("String", "WEB_BASE_HOST", webBaseHostProd)
-                buildConfigField("String", "SERVER_LIST_URL", serverListUrlProd)
-                buildConfigField("String", "SERVER_LIST_HOST", serverListHostProd)
+                buildConfigField("String", "SERVER_MAIN_URL", "\"https://refresh.pocketcasts.com\"")
+                buildConfigField("String", "SERVER_API_URL", "\"https://api.pocketcasts.com\"")
+                buildConfigField("String", "SERVER_CACHE_URL", "\"https://cache.pocketcasts.com\"")
+                buildConfigField("String", "SERVER_CACHE_HOST", "\"cache.pocketcasts.com\"")
+                buildConfigField("String", "SERVER_STATIC_URL", "\"https://static.pocketcasts.com\"")
+                buildConfigField("String", "SERVER_SHARING_URL", "\"https://sharing.pocketcasts.com\"")
+                buildConfigField("String", "SERVER_SHORT_URL", "\"https://pca.st\"")
+                buildConfigField("String", "SERVER_SHORT_HOST", "\"pca.st\"")
+                buildConfigField("String", "SERVER_WEB_PLAYER_HOST", "\"play.pocketcasts.com\"")
+                buildConfigField("String", "WEB_BASE_HOST", "\"pocketcasts.com\"")
+                buildConfigField("String", "SERVER_LIST_URL", "\"https://lists.pocketcasts.com\"")
+                buildConfigField("String", "SERVER_LIST_HOST", "\"lists.pocketcasts.com\"")
 
                 testInstrumentationRunner = project.property("testInstrumentationRunner") as String
                 testApplicationId = "au.com.shiftyjelly.pocketcasts.test${project.name.replace("-", "_")}"


build.gradle.kts
```diff
@@ -253,6 +255,19 @@ subprojects {
                 buildConfigField("String", "META_APP_ID", "\"${project.property("metaAppId")}\"")
                 buildConfigField("String", "APPS_FLYER_KEY", "\"${project.property("appsFlyerKey")}\"")
 
+                buildConfigField("String", "SERVER_MAIN_URL", serverMainUrlProd)
```
ParaskP7: Praise (❤️): Nice deduplication here! 🥇

build.gradle.kts
```diff
@@ -321,37 +336,20 @@ subprojects {
...
-                    buildConfigField("String", "WEB_BASE_HOST", webBaseHostProd)
-                    buildConfigField("String", "SERVER_LIST_URL", serverListUrlProd)
-                    buildConfigField("String", "SERVER_LIST_HOST", serverListHostProd)
-
                     signingConfig = signingConfigs.getByName("debug")
                 }
 
-                named("release") {
+                maybeCreate("prototype").apply {
+                    isDebuggable = false
```
ParaskP7: Questions (❓):

1. This is what we really want here, right? 🤔
2. Because `isDebuggable` is `false` by default, we set it to `false` here explicitly to make sure reader of this build type can quickly reason about ti, correct? 🤔

app/build.gradle.kts
```diff
@@ -38,6 +38,20 @@ android {
             manifestPlaceholders["appIcon"] = "@mipmap/ic_launcher_3"
         }
 
+        maybeCreate("prototype").apply {
```
ParaskP7, line 41: Question (❓): Why not add the `isMinifyEnabled` + `isShrinkResources` + `proguardFiles ` configuration on the root level `build.gradle.kts` file, just like it is done for the rest, to keep things consistent, there might be a reason I am not aware of: 🤔


`build.gradle.kts` (root)
```kts
    plugins.withType<AppPlugin>().configureEach {
        configure<BaseAppModuleExtension> {
            lint(configureLint)

            buildTypes {
                ...
                maybeCreate("prototype").apply {
                    isMinifyEnabled = true
                    proguardFiles.addAll(
                        listOf(
                            getDefaultProguardFile("proguard-android-optimize.txt"),
                            rootProject.file("proguard-rules.pro"),
                            rootProject.file("proguard-rules-no-obfuscate.pro"),
                        ),
                    )
                    isShrinkResources = true
                }
                ...
            }
        }
```

`build.gradle.kts` (app)
```kts
    buildTypes {
        ...
        named("prototype") {
            manifestPlaceholders["appIcon"] = "@mipmap/ic_launcher_prototype"
        }
        ...
    }
```

app/build.gradle.kts
```diff
@@ -38,6 +38,20 @@ android {
...
+            manifestPlaceholders["appIcon"] = "@mipmap/ic_launcher_prototype"
+
+            isMinifyEnabled = true
+            isShrinkResources = true
+            proguardFiles.addAll(
+                listOf(
+                    getDefaultProguardFile("proguard-android-optimize.txt"),
+                    rootProject.file("proguard-rules.pro"),
```
ParaskP7: Question (❓): Do we need  to add this `proguard-rules.pro` if we then (kind of) overwrite that with the next `proguard-rules-no-obfuscate.pro` file? 🤔

modules/features/profile/src/prototype/res/xml/authenticator.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<account-authenticator xmlns:android="http://schemas.android.com/apk/res/android"
+    android:accountType="au.com.shiftyjelly.pocketcasts.pocketcasts"
+    android:icon="@mipmap/ic_launcher_prototype"
+    android:label="@string/app_name"
+    android:smallIcon="@mipmap/ic_launcher_prototype" />
```
ParaskP7, line 6: Praise (❤️): Thanks for adding this new icon and making this `prototype` build stand-out with its own icon. 🥇

modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/SettingsFragment.kt
```diff
@@ -69,7 +69,7 @@ class SettingsFragment : BaseFragment() {
                             @Suppress("DEPRECATION")
                             activity?.onBackPressed()
                         },
-                        isDebug = BuildConfig.DEBUG,
+                        isDebug = BuildConfig.DEBUG || BuildConfig.IS_PROTOTYPE,
```
ParaskP7, line 72: Info (ℹ️): Noted! (enabling `Developer` settings)

modules/features/shared/src/main/java/au/com/shiftyjelly/pocketcasts/shared/AppLifecycleObserver.kt
```diff
@@ -100,7 +100,7 @@ class AppLifecycleObserver(
     }
 
     private fun setupFeatureFlags() {
-        val providers = if (BuildConfig.DEBUG) {
+        val providers = if (BuildConfig.DEBUG || BuildConfig.IS_PROTOTYPE) {
```
ParaskP7, line 103: Info (ℹ️): Noted! (enabling `Feature Flags`)

modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/featureflag/Feature.kt
```diff
@@ -2,6 +2,8 @@ package au.com.shiftyjelly.pocketcasts.utils.featureflag
 
 import au.com.shiftyjelly.pocketcasts.helper.BuildConfig
 
+private val isDebugOrPrototypeBuild = BuildConfig.DEBUG || BuildConfig.IS_PROTOTYPE
```
ParaskP7, line 5: Info (ℹ️): Noted! (enabling `Feature Flags`)

build.gradle.kts
MiSikora: 1. Yes. Setting `isDebuggable = true` sets `BuildConfig.DEBUG` to `true`. Unfortunately there is a lot of legacy code that shouldn't be enabled in prototype builds that relies on that flag. Like Timber's debug logging, OkHttp logging, VM policies, etc. I don't want the build to suffer from performance due to debug stuff.
2. TIL. I'll remove the line.

ParaskP7: 1. `...like Timber's debug logging, OkHttp logging, VM policies, etc.` gotcha, thanks and 👍 !
2. 👍 + 🙏

app/build.gradle.kts
MiSikora: I kind of hate top level configurations because builds suffer during configuration phase. Especially in large projects, which Pocket Casts fortunately isn't. And I know that I created it in the first place but it was rather cleanup of the old Groovy build configuration. But I'll move it like you suggest.

app/build.gradle.kts
MiSikora: We don't overwrite it. We only append the new `-dontobfuscate` rule. I still want to have all the shrinking rules that we use in production builds.

app/build.gradle.kts
ParaskP7: > But I'll move it like you suggest.

No need to move it @MiSikora , I was just curious, especially if you folks have plans to move that top level configuration in the future. 💯 

PS: My main worry is that people wouldn't know that there are 2 places for such configuration, update one without the other, not knowing which is which and what to add where, etc. I was mainly worried about consistency here... 😊

MiSikora:
>     1. `./gradlew assembleRelease`: ❌  Running this fails for all modules fails, but also it fails on `main`. This is due to `account_activity.xml`. Is that known and expected, something we don't want to fix (I ask in general, not in this PR)? 🤔

I have no clue but I doubt it is known or expected. I never use top level assemble commands. I wouldn't stress about it.

>     1. `./gradlew assemblePrototype`: ❌ This too fails but for a different reason (`AndroidManifest.xml`). Is this expected? 🤔
> 
>     3. `./gradlew automotive:assemblePrototype`: ❌ Manifest merger failed with multiple errors. Is this expected? 🤔
> 
>     4. `./gradlew wear:assemblePrototype`: ❌ Manifest merger failed with multiple errors. Is this expected? 🤔

For `2` and `3` I'll filter out variants to make it clear that these projects don't have it. I don't know if it'll fix `1`. But if it doesn't I won't be spending too much time on it.

---

Thanks for the review. I'll address your comments in the next push 🙏

app/build.gradle.kts
ParaskP7: Just for my FYI, what will this newly appending `-dontobfuscate` rule do to the previously added rule configuration, wouldn't it somehow override it? 🤔 (FYI: I am not the best ProGuard user, need to read the docs, thus asking here to understand it quicker... 😊)

ParaskP7:
Thanks so much for the quick reply @MiSikora ! 🙇 🥇

> I have no clue but I doubt it is known or expected. I never use top level assemble commands. I wouldn't stress about it.

🤝 let's not stress about it, not at least on this PR.

FYI: I just used those top level assemble commands because it saves me the trouble to wait on running multiple commands. But, I know that sometime this is not expected, an example is `lint` and the latest changes we did there, making it a must to run the individual `list` commands instead of this one top level.

> For 2 and 3 I'll filter out variants to make it clear that these projects don't have it.

This makes sense, being explicit should be a nice win here, thanks! 🙏 

> I don't know if it'll fix 1. But if it doesn't I won't be spending too much time on it.

Again 🤝 let's not stress about it, not at least on this PR.

MiSikora:
@ParaskP7 I fixed most of the issues. I didn't fix the top level `assemblePrototype` but I didn't spend any brainpower on it.

build.gradle.kts
```diff
@@ -253,6 +243,19 @@ subprojects {
                 buildConfigField("String", "META_APP_ID", "\"${project.property("metaAppId")}\"")
                 buildConfigField("String", "APPS_FLYER_KEY", "\"${project.property("appsFlyerKey")}\"")
 
+                buildConfigField("String", "SERVER_MAIN_URL", "\"https://refresh.pocketcasts.com\"")
```
ParaskP7, line 246: Praise (❤️): Another good thing done, thanks! 😄

app/build.gradle.kts
MiSikora: No, it just means the code won’t be obfuscated. All the performance-related shrinking rules such as removing dead or unnecessary code, inlining, and so on will still run. Anything R8 might mistakenly touch is protected by our `proguard-rules.pro` configuration. The additional `-dontobfuscate` rule from the new file ensures that if the code in this build crashes, we’ll get readable stack traces in our logs instead of something like `p5.i.a.s.a`.

ParaskP7: Aaaah, okay, okay, okay, gotcha, cool, nice, thanks @MiSikora , resolved! 🙇

automotive/build.gradle.kts
```diff
@@ -41,6 +41,12 @@ android {
     }
 }
 
+androidComponents {
+    beforeVariants { builder ->
+        builder.enable = builder.buildType != "prototype"
```
ParaskP7, line 46: Info (ℹ️): Noted, thanks! 🥇 

`./gradlew automotive:assemblePrototype`
```
FAILURE: Build failed with an exception.

* What went wrong:
Cannot locate tasks that match 'automotive:assemblePrototype' as task 'assemblePrototype' not found in project ':automotive'.
```

wear/build.gradle.kts
```diff
@@ -51,6 +51,12 @@ android {
     }
 }
 
+androidComponents {
+    beforeVariants { builder ->
+        builder.enable = builder.buildType != "prototype"
```
ParaskP7, line 56: Info (ℹ️): Noted, thanks! 🥇 

`./gradlew wear:assemblePrototype`
```
FAILURE: Build failed with an exception.

* What went wrong:
Cannot locate tasks that match 'wear:assemblePrototype' as task 'assemblePrototype' not found in project ':wear'.
```

app/build.gradle.kts
MiSikora: I updated it. It makes sense here for sure as there are no plans to restructure our Gradle files and Gradle is sometimes arcane knowledge. But I don't think that consistency is always a good argument because if you have something bad in your code you will make it consistently bad. :)

ParaskP7: > I updated it. It makes sense here for sure as there are no plans to restructure our Gradle files and Gradle is sometimes arcane knowledge.

Thanks! 👍

> But I don't think that consistency is always a good argument because if you have something bad in your code you will make it consistently bad.

😄 for sure, I hear you...

ParaskP7:
Awesome stuff @MiSikora , thanks for this work, I have nothing more to add here. Let's merge this and I'll then create another project for this new variant on our Firebase instance to test with FAD and #4513, I really hope this is exactly what you all need! 🙇 ❤️ 🚀

PR #4550, opened 2025-09-30
[Improved Search] Implement autocomplete

sztomek:
## Description
This PR implements the new autocomplete API.
Specs: pdeCcb-aQE-p2

The results are not dispalyed on the UI just yet, for now it only incorporates autocomplete in the ordinary search flow.
Autocomplete is intantaneous, it doesn't debounce term changes as the regular search does.
You should see `/autocomplete/search` requests for each keystroke you made in your LogCat / Network activity monitor.

Fixes https://linear.app/a8c/issue/PCDROID-169/wire-up-autocomplete

## Testing Instructions
1. Install app and go through onboarding
2. On the discover tab, tap the search bar
3. Start typing something
4. Observe LogCat or network monitor and look for `GET /autocomplete/search` requests

## Screenshots or Screencast 
Screenshot 2025-09-30 at 11 04 27

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>SearchAutoCompleteManagerImpl</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>

</p>

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/to/SearchAutoCompleteItem.kt
```diff
@@ -0,0 +1,6 @@
+package au.com.shiftyjelly.pocketcasts.models.to
+
+sealed class SearchAutoCompleteItem {
+    data class Term(val term: String)
+    data class Podcast(val uuid: String, val title: String, val author: String)
```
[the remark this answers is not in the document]
MiSikora: 🚫  This suggestion is correct. `SearchAutoCompleteItem` cannot be created without being inherited from but we use it in results.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/search/SearchAutoCompleteManager.kt
```diff
@@ -0,0 +1,9 @@
+package au.com.shiftyjelly.pocketcasts.repositories.search
+
+import au.com.shiftyjelly.pocketcasts.models.to.SearchAutoCompleteItem
+import io.reactivex.Observable
+
+interface SearchAutoCompleteManager {
+    suspend fun autoCompleteSearch(term: String): List<SearchAutoCompleteItem>
+    fun autoCompleteSearchRxObservable(term: String): Observable<List<SearchAutoCompleteItem>>
```
MiSikora: ℹ️ I think this should be moved out of this interface since we want to write the new code using coroutines only. The `suspend -> Rx` conversion can be handled at the call site using `rxObservable`, as you’re already doing in the implementation. Adding functions like this bloats the interface unnecessarily.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/search/SearchAutoCompleteManagerImpl.kt
```diff
@@ -0,0 +1,35 @@
...
+    private val searchService: SearchService,
+    private val settings: Settings,
+) : SearchAutoCompleteManager {
+    override suspend fun autoCompleteSearch(term: String): List<SearchAutoCompleteItem> {
+        val response = searchService.autoCompleteSearch(query = term, terms = TERM_LIMIT, podcasts = PODCAST_LIMIT, language = settings.discoverCountryCode.value)
+        return response.results.map {
+            when (it) {
+                is AutoCompleteResult.TermResult -> SearchAutoCompleteItem.Term(term = it.value)
+                is AutoCompleteResult.PodcastResult -> SearchAutoCompleteItem.Podcast(uuid = it.value.uuid, title = it.value.title, author = it.value.author)
+            } as SearchAutoCompleteItem
```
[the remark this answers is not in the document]
MiSikora: 🚫 This comment is kind of correct. The cast is necessary only because `Term` and `Podcast` do not inherit from the base class.

modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/di/SearchModule.kt
```diff
@@ -0,0 +1,53 @@
...
+import retrofit2.converter.protobuf.ProtoConverterFactory
+
+@Qualifier
+@Retention(AnnotationRetention.BINARY)
+annotation class SearchMoshi
+
+
+@Module
+@InstallIn(SingletonComponent::class)
+class SearchModule {
```
MiSikora: ❓ I'm having trouble understanding the approach here. Why do we deviate so much from the configuration in the `ServersModule`? I'll spread my questions over couple of comments.

modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/di/SearchModule.kt
```diff
@@ -0,0 +1,53 @@
...
+    @Provides
+    @Singleton
+    @SearchMoshi
+    internal fun provideMoshi(): Moshi {
+        return Moshi.Builder()
```
MiSikora: ❓Do we need a new instance of `Moshi` over adding adapter factory to it?

modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/di/SearchModule.kt
```diff
@@ -0,0 +1,53 @@
...
+                PolymorphicJsonAdapterFactory.of(AutoCompleteResult::class.java, "type")
+                    .withSubtype(AutoCompleteResult.TermResult::class.java, "term")
+                    .withSubtype(AutoCompleteResult.PodcastResult::class.java, "podcast")
+            )
+            .build()
+    }
+
+    @Provides
+    @Singleton
+    internal fun provideApiRetrofit(@Cached okHttpClient: OkHttpClient, @SearchMoshi moshi: Moshi): Retrofit {
```
MiSikora: ℹ️ Providing a Retrofit instance in our codebase without a qualifier instance isn't unfortunately safe. It will be injected every time anyone requests `Retrofit` without any annotation. Which is an easy thing to do, and which will result in easy to miss bugs.

modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/di/SearchModule.kt
```diff
@@ -0,0 +1,53 @@
...
+        return Retrofit.Builder()
+            .addConverterFactory(ProtoConverterFactory.create())
+            .addConverterFactory(MoshiConverterFactory.create(moshi))
+            .baseUrl(Settings.SEARCH_API_URL)
+            .client(okHttpClient)
+            .build()
+    }
+
+    @Provides
+    internal fun provideSearchService(retrofit: Retrofit): SearchService = retrofit.create(SearchService::class.java)
```
MiSikora: ⚠️ This should be marked with `@Singleton`. It's expensive to create retrofit's interfaces.

modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/search/SearchService.kt
```diff
@@ -0,0 +1,14 @@
...
+
+import retrofit2.http.GET
+import retrofit2.http.Query
+
+interface SearchService {
+    @GET("/autocomplete/search")
+    suspend fun autoCompleteSearch(
+        @Query("q") query: String,
+        @Query("termsLimit") terms: Int,
+        @Query("podcastsLimit") podcasts: Int,
```
MiSikora: ❓ Could we use the full names for the parameters as well? Instead of `terms`, `termsLimit`? `autoCompleteSearch(terms = 5, podcasts = 6)` doesn't read clearly.

modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/discover/AutoCompleteSearch.kt
```diff
@@ -0,0 +1,9 @@
+package au.com.shiftyjelly.pocketcasts.servers.discover
+
+import au.com.shiftyjelly.pocketcasts.models.to.SearchAutoCompleteItem
+
+data class AutoCompleteSearch(
+    val searchTerm: String = "",
+    val error: Throwable? = null,
+    val results: List<SearchAutoCompleteItem> = emptyList()
+)
```
MiSikora: ❓ Shouldn't this class be also a sealed type? Without a sealed type we run into exhaustiveness issues, which otherwise are guarded by the compiler.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/search/SearchAutoCompleteManager.kt
sztomek: gotcha, I just tried to stay consistent with `PodcastManager` that offers the same viarants of its methods.

modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/discover/AutoCompleteSearch.kt
sztomek: Created this class to the exiting patterns like `GlobalServerSearch`, `EpisodeSearch`, `PodcastSearch`.
But sure, why not

sztomek:
>> Autocomplete is intantaneous, it doesn't debounce term changes as the regular search does.

> Why's that the case? While performing a search, especially a network based one, a debounce of 150-300 ms is very useful. Otherwise it opens and closes a connection on every keystroke, which is quite wasteful in terms of battery.

That was the idea that steered me not to have debounce:
pdeCcb-aqh-p2#comment-8043
Screenshot 2025-09-30 at 12 27 44

Added 200ms as you've requested

MiSikora:
> Added 200ms as you've requested

Yeah, I've just seen the summary from the meeting that says no debounce. Personally, I think it is a bad idea to have no debounce due to energy consumption. I think it's best to do whatever you think is right here but it would be good to clarify it with the team. Let me know if you want me to chime in since I'm the one who's opposed to this requirement.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/search/SearchAutoCompleteManager.kt
MiSikora: Yeah, I think it's a hard balance between being consistent and moving away from old APIs. If possible I think we should keep Rx in the legacy code. This will make things like migrations and maintenance easier.

modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/di/ServersModule.kt
```diff
@@ -78,6 +81,11 @@ class ServersModule {
             .add(AccessToken::class.java, AccessToken.Adapter)
             .add(RefreshToken::class.java, RefreshToken.Adapter)
             .add(BlazeAdLocation::class.java, BlazeAdLocationMoshiAdapter())
+            .add(
+                PolymorphicJsonAdapterFactory.of(AutoCompleteResult::class.java, "type")
+                    .withSubtype(AutoCompleteResult.TermResult::class.java, "term")
+                    .withSubtype(AutoCompleteResult.PodcastResult::class.java, "podcast")
```
MiSikora: 💡 I would consider putting it as a property of a companion object for `AutoCompleteResult`. This way it will be more visible if any changes are needed due to new search result types, etc.

Alternatively we could consider adding `moshi-sealed` to automate adapter generation - https://github.com/ZacSweers/MoshiX/tree/main/moshi-sealed. I used it in the past for some projects and it provides some nice safety features.

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchHandler.kt
```diff
@@ -101,6 +106,34 @@ class SearchHandler @Inject constructor(
         .toObservable()
         .map { podcasts -> podcasts.map(Podcast::uuid).toHashSet() }
 
+    private val autoCompleteResults = searchQuery
+        .subscribeOn(Schedulers.io())
+        .map { it.string.trim() }
+        .debounce {
+            Observable.timer(200L, TimeUnit.MILLISECONDS)
+        }
```
MiSikora: ℹ️ I think this can be expressed as `.debounce(200L, TimeUnit.MILLISECONDS)`. The overload that uses `Observable` is useful when there is a need for different debounce values based on the input. Something like:

```kotlin
debounce { query ->
    if (query.isEmpty()) {
        Observable.just(1L)
    } else {
        Observable.timer(200L, TimeUnit.MILLISECONDS)
    }
}
```

MiSikora (approved):
Looks good, thanks for the update!

PR #4565, opened 2025-10-02
Stop widget crash

geekygecko:
## Description

A common crash in Sentry and the Google Play console seems to be linked to the widget. 

```
BadParcelableException
Failure retrieving array; only received 8 of 68
android.os.BadParcelableException: Failure retrieving array; only received 37 of 43
    at android.content.pm.BaseParceledListSlice.<init>(BaseParceledListSlice.java:117)
    at android.content.pm.ParceledListSlice.<init>(ParceledListSlice.java:42)
    at android.content.pm.ParceledListSlice.<init>
    at android.content.pm.ParceledListSlice$1.createFromParcel(ParceledListSlice.java:80)
    at android.content.pm.ParceledListSlice$1.createFromParcel(ParceledListSlice.java:78)
    at android.os.Parcel.readTypedObject(Parcel.java:4340)
    at com.android.internal.appwidget.IAppWidgetService$Stub$Proxy.getInstalledProvidersForProfile(IAppWidgetService.java:1323)
    at android.appwidget.AppWidgetManager.getInstalledProvidersForProfile(AppWidgetManager.java:1279)
    at android.appwidget.AppWidgetManager.getInstalledProviders(AppWidgetManager.java:1224)
    at androidx.glance.appwidget.GlanceAppWidgetManager.addAllReceiversAndProvidersToPreferences(GlanceAppWidgetManager.kt:272)
    at androidx.glance.appwidget.GlanceAppWidgetManager.getState(GlanceAppWidgetManager.kt:109)
```

There are no references to the Pocket Casts packages, but it seems to come from the `GlanceAppWidgetManager.getState` method. This change wraps some calls to this class with the hope that it will reduce the crashes. Unfortunately, I can't repeat this crash, but we should be able to see if it's effective by a decrease in the Sentry errors. 

## Testing Instructions

Just out the widgets to make sure they still update.

## Checklist

### DIFF
--- a/modules/features/widgets/src/main/kotlin/au/com/shiftyjelly/pocketcasts/widget/PlayerWidgetManager.kt
+++ b/modules/features/widgets/src/main/kotlin/au/com/shiftyjelly/pocketcasts/widget/PlayerWidgetManager.kt
@@ -6,6 +6,7 @@ import androidx.glance.appwidget.GlanceAppWidgetManager
 import androidx.glance.appwidget.updateAll
 import au.com.shiftyjelly.pocketcasts.models.entity.BaseEpisode
 import au.com.shiftyjelly.pocketcasts.repositories.di.ApplicationScope
+import au.com.shiftyjelly.pocketcasts.utils.log.LogBuffer
 import au.com.shiftyjelly.pocketcasts.widget.data.ClassicPlayerWidgetState
 import au.com.shiftyjelly.pocketcasts.widget.data.LargePlayerWidgetState
 import au.com.shiftyjelly.pocketcasts.widget.data.MediumPlayerWidgetState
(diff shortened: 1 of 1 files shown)


MiSikora (commented):
I don't think this is this the right approach. It will only hide the crash and make the widgets have either invalid state or make them unusable. I mentioned something like this briefly in this post pbCcaG-1Nt-p2.

If you feel strongly about this change I won't block the PR but let's discuss it first.

geekygecko:
> I don't think this is this the right approach.

Thanks for sharing your post, it makes sense. I do feel that this is slightly different, as the code that is crashing is outside of our control and seems likely to be fixed by Google eventually. Letting the app crash doesn't fix the issue, and there doesn't seem to be a safe way to do anything else. But I don't feel strongly about it and respect your opinion. Maybe let's chat about it first.

MiSikora:
> as the code that is crashing is outside of our control and seems likely to be fixed by Google eventually.

But we don't know that. We don't know what's causing the problem. Maybe we're doing something wrong while communicating with Glance APIs. Maybe there are things that fix this even if it is a Google's issue. For example [here](https://gist.github.com/pyricau/5f864374c1b1a028a1a8dfe25b6761df) is a fix to a different `BadParcelableException` that was caused by the OS. I think it's hard to give definitive statements when we don't know the root cause of the issue.

> Letting the app crash doesn't fix the issue, and there doesn't seem to be a safe way to do anything else.

AFAIK we're not getting any reports from the users so this might be an imperceptible crash. But my point is that catching the exception isn't safe either as it might lead to an incorrect widget state. Which might or might not be even worse.

geekygecko:
I'm going to close this PR for now.

PR #4571, opened 2025-10-05
[Improved Search] Show predictive suggestions

sztomek:
## Description
This PR displays the search suggestions on the search screen.
Analytics and the recently announced language code unification is not part of the scope.

Designs: 9aiIqQo7ibm7R1AyIYr8fH-fi-1_1574
P2: pdeCcb-aqh-p2

Fixes https://linear.app/a8c/issue/PCDROID-170/update-search-screen

## Testing Instructions
1. Clean install the app
2. Press 'Get started' on onboarding, try filtering some podcasts on the recommendations step
- [ ] search still works there
3. Once you're done with onboarding, tap 'Discover'
4. Tap the search bar
5. Start typing
- [ ] suggestions appear as you type
6. Select a suggestion
- [ ] real search kicks in and you'll see podcasts and episodes
- [ ] tapped term will be saved as history item
7. On the suggestions, tap the '+' icon on any podcast suggestion
- [ ] verify that the podcasts followed status updates after a short period of time

Please also verify:
- Search functionality remains intact with the FF turned OFF
- Search on podcasts tab still works with/without the FF

## Screenshots or Screencast 

https://github.com/user-attachments/assets/314ee3f8-8370-4670-8fbd-286276f9ef8f

## Checklist
 
- [x] with different themes


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchAutoCompleteResultsPage.kt
```diff
@@ -0,0 +1,286 @@
...
+                .align(Alignment.Center),
+        ) {
+            CircularProgressIndicator()
+        }
+        LazyColumn(
+            modifier = Modifier.nestedScroll(nestedScrollConnection),
+            contentPadding = PaddingValues(bottom = bottomInset),
+        ) {
+            results.forEachIndexed { index, item ->
+                item(key = item.hashCode(), contentType = "content-${item.javaClass}") {
```
MiSikora: I have a couple of questions and comments regarding the `key` property.

- ❓ From what I understand about the feature it isn't a dynamic list that gets manipulated. We load it and display it. Does the key help here?
- ❓ How many items do we expect here to be present? Right now it shows 4 suggestions and 4 results. Is it expected to go into dozens or more? If it is then `hashCode()` isn't a good `key` candidate as hash collision can lead to crashes. If it is not then `key` isn't really important since we're displaying only a couple of items.

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchAutoCompleteResultsPage.kt
```diff
@@ -0,0 +1,286 @@
...
+                    .padding(start = 8.dp)
+                    .clickable(onClick = onFollow),
+            )
+        }
+    }
+}
+
+@Composable
+private fun EpisodeRow(
+    item: Any,
```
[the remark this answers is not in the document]
MiSikora: ❓I'm also a bit confused about this. I understand that it is a mock function but why does it accept `Any` instead of the episode type?

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchFragment.kt
```diff
@@ -248,26 +282,56 @@ class SearchFragment : BaseFragment() {
                 }
             }
         }
+
+        binding.searchSuggestions.apply {
+            ViewCompositionStrategy.DisposeOnViewTreeLifecycleDestroyed
+            setContent {
```
MiSikora: ℹ️ `ViewCompositionStrategy.DisposeOnViewTreeLifecycleDestroyed` invocation is a no-op. It should be `setContentWithViewCompositionStrategy()`.

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchHandler.kt
```diff
@@ -101,50 +105,83 @@ class SearchHandler @Inject constructor(
...
+        .flatMapLatest { query ->
             if (query.isEmpty()) {
-                Observable.just<AutoCompleteSearch>(AutoCompleteSearch.Success(searchTerm = query, results = emptyList()))
-            } else {
-                rxObservable<List<SearchAutoCompleteItem>> {
-                    autoCompleteManager.autoCompleteSearch(term = query)
-                }.subscribeOn(Schedulers.io())
-                    .map<AutoCompleteSearch> {
-                        AutoCompleteSearch.Success(
+                flow {
```
MiSikora: ℹ️ This can be replaced with `flowOf()`.

MiSikora (commented):
Apart from using `hashCode()` as a key, I don't there are any issues with the code itself. However, I found a few UI and UX issues worth addressing:

### Incorrect highlight range

Currently, the implementation highlights the first `N` characters of the word but the search term can appear anywhere within the result.

Image

### Small follow tap target

The follow button is quite small, making it easy to misclick.

Image

### Button highlight doesn’t extend edge to edge

Image

### Follow state is not responsive

There’s about a 2-second delay between tapping the follow button and seeing the followed state.

https://github.com/user-attachments/assets/f250759c-7c8e-4f0e-978a-acf8400d884d

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchAutoCompleteResultsPage.kt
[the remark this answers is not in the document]
sztomek: Good question 😓 
The autocomplete API currently doesn't return episodes, but on the designs there are some.
I created a UI component nevertheless, without the actual type. I know I could have defined the class by looking at what data the UI needs, but didn't do so for some reason 🤔 
Anyways, I've addressed this on my current branch for the search results UI changes, so I'd not touch this here if you don't mind.

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchAutoCompleteResultsPage.kt
sztomek: hi
> ❓ From what I understand about the feature it isn't a dynamic list that gets manipulated. We load it and display it. Does the key help here?

you're right. removed the keys.

> ❓ How many items do we expect here to be present? Right now it shows 4 suggestions and 4 results. Is it expected to go into dozens or more? If it is then hashCode() isn't a good key candidate as hash collision can lead to crashes. If it is not then key isn't really important since we're displaying only a couple of items.

4 items are currently what we expect to display from each section. removed the key anyways.

sztomek:
thanks for your review @MiSikora 
I've addressed all your concerns in 77ea7d9b3a34b4112017fd3a8071361d93c7cf0b
also answered your questions, i believe.
Pls take another look 👀

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchAutoCompleteResultsPage.kt
```diff
@@ -0,0 +1,297 @@
...
+            Icon(
+                painter = painterResource(IR.drawable.ic_add_black_24dp),
+                contentDescription = null,
+                tint = MaterialTheme.theme.colors.secondaryText02,
+                modifier = Modifier
+                    .padding(start = 8.dp)
+                    .size(32.dp)
+                    .clickable(onClick = onFollow),
+            )
+        }
```
MiSikora, line 219: ℹ️ Now the icons feel a bit too large.

Image

I'd suggest sticking to the standard sizing provided by the Material Design with `minimumInteractiveComponentSize()` and `IconButton`. Something like: 

```kotlin
if (item.isSubscribed) {
    Icon(
        modifier = Modifier.minimumInteractiveComponentSize(),
        painter = painterResource(IR.drawable.ic_check),
        contentDescription = null,
        tint = MaterialTheme.theme.colors.support02,
    )
} else {
    IconButton(onClick = onFollow) {
        Icon(
            painter = painterResource(IR.drawable.ic_add_black_24dp),
            contentDescription = null,
            tint = MaterialTheme.theme.colors.secondaryText02,
        )
    }
}
```

This way we can get expected touch targets and a circle touch indicator.

Image

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchHandler.kt
```diff
@@ -101,50 +106,80 @@ class SearchHandler @Inject constructor(
...
+                        results = emptyList(),
+                    ),
+                )
             } else {
-                rxObservable<List<SearchAutoCompleteItem>> {
-                    autoCompleteManager.autoCompleteSearch(term = query)
-                }.subscribeOn(Schedulers.io())
-                    .map<AutoCompleteSearch> {
-                        AutoCompleteSearch.Success(
+                flowOf(autoCompleteManager.autoCompleteSearch(term = query))
```
MiSikora: 🚫 I suggested `flowOf()` only for the empty case because this can cause a crash now. Turn on the airplane mode and type in anything. It happens because exception is thrown not while emitting but before a `Flow` is even created so the `.catch()` doesn't have a chance to interact with it. This should be either changed back to `flow { emit() }` or `autoCompleteManager.autoCompleteSearch(term = query)` should be surrounded with a `try/catch`.

PR #4580, opened 2025-10-07
Add i18n headers

MiSikora:
## Description

This adds`X-App-Language` and `X-User-Region` headers to our HTTP requests.

Closes PCDROID-179

## Testing Instructions

1. Start the app.
2. Check for the headers either in the Logcat or in the Network Inspection.
3. The headers should match your configuration.
4. Go to Discover and scroll to the bottom.
5. Change your region.
6. Value should be reflected in the `X-User-Region` header.
7. Change device's or app's language from the system settings.
8. Value should be reflected in the `X-App-Language` header.

## Checklist

### DIFF
--- a/modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/di/InterceptorModule.kt
+++ b/modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/di/InterceptorModule.kt
@@ -1,10 +1,13 @@
 package au.com.shiftyjelly.pocketcasts.servers.di
 
+import android.content.Context
 import au.com.shiftyjelly.pocketcasts.preferences.AccessToken
+import au.com.shiftyjelly.pocketcasts.preferences.Settings
 import au.com.shiftyjelly.pocketcasts.servers.BuildConfig
 import au.com.shiftyjelly.pocketcasts.servers.CleanAndRetryInterceptor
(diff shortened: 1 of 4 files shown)


modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/di/InterceptorModule.kt
```diff
@@ -215,9 +248,12 @@ object InterceptorModule {
 
     @Provides
     @Player
-    fun providePlayerInterceptors(): List<OkHttpInterceptor> {
+    fun providePlayerInterceptors(
+        @I18nInterceptor i18nInterceptor: Interceptor,
+    ): List<OkHttpInterceptor> {
         return buildList {
             add(publicUserAgentInterceptor.toClientInterceptor())
+            add(i18nInterceptor.toClientInterceptor())
```
geekygecko, line 257: I noticed that the headers are being sent to third parties when an episode is played. For privacy reasons, we should limit these requests to our own servers only.

geekygecko (approved):

modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/di/InterceptorModule.kt
MiSikora: Hmm, good point. I haven't thought about privacy.

PR #4594, opened 2025-10-09
Fix podcast title sort with accents

geekygecko:
## Description

This change fixes the issue that when sorting on the podcasts tab by name, the podcasts that start with a letter with an accent mark are in the wrong position.

Fixes PCDROID-185

## Testing Instructions

1. Follow a few podcasts so it's possible to see if a sort work
2. Follow the podcast "Área de Trabalho" https://pca.st/tr7ycedb
3. Tap the toolbar three dots -> Sort by -> Name
4. ✅ Verify the podcast is at the top of the list
5. Follow the podcast "Ça s'explique" https://pca.st/4ARm
6. ✅ Verify the podcast is in the correct position

## Screenshots 

Screenshot_20251009_152719

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@
         ([#4555](https://github.com/Automattic/pocket-casts-android/pull/4555))
     *   Fix archived episodes not being downloaded when added to queue
         ([#4570](https://github.com/Automattic/pocket-casts-android/pull/4570))
+    *   Fix podcast title sort with accents   
+        ([#4594](https://github.com/Automattic/pocket-casts-android/pull/4594))
 
 7.99
 -----


modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/type/PodcastsSortType.kt
```diff
@@ -12,9 +12,9 @@ import com.squareup.moshi.JsonReader
 import com.squareup.moshi.JsonWriter
 import java.util.Locale
 
-private val replaceTheRegex = "^the ".toRegex(RegexOption.IGNORE_CASE)
+private val replaceTheRegex = "^the ".toRegex()
```
geekygecko, line 15: There is no need to ignore the case for this regex as the title is always lowercased first.

PR #4598, opened 2025-10-09
Google login tweaks

sztomek:
## Description
Small tweaks related to Google login flows on watch and phone.
Context: pdeCcb-aXl-p2#comment-8395

We no longer show the error toast when the user cancelled the credential request on the phone.
On watch we now truncate the email in the middle if the user was trying to log in with a free account.

Screenshot_20251009_160729

## Testing Instructions
Checking the code is enough

## Checklist

### DIFF
--- a/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/MainActivity.kt
+++ b/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/MainActivity.kt
@@ -350,7 +350,7 @@ private fun WearApp(
                 val email = if (signInState.email.length > 16) {
                     buildString {
                         append(signInState.email.substring(0, 6))
-                        append("...")
+                        append("…")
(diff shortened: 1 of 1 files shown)


wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/MainActivity.kt
```diff
@@ -347,7 +347,16 @@ private fun WearApp(
                 if (popped) {
                     ScrollToTop.initiate(navController)
                 }
-                val message = stringResource(LR.string.log_in_free_acccount, signInState.email)
+                val email = if (signInState.email.length > 16) {
+                    buildString {
+                        append(signInState.email.substring(0, 6))
+                        append("...")
```
MiSikora:
```suggestion
                        append("…")
```

We can use ellipsis `…` instead of 3 dots `...`.

PR #4604, opened 2025-10-10
[Improved Search] Wire up combined search API 

sztomek:
## Description
This PR adds the `search/combined` API and wires it up for the improved results screen.
It also includes some design updates (see referenced slack thread) and functional improvements:
- Local search lists folders
- Episode results now properly track playback status
- Filter chips are now floating above the results with a faded background

Fixes https://linear.app/a8c/issue/PCDROID-171/filter-search-results
Slack: p1760036879103319-slack-C09ENPJ5D2R
Designs: 9aiIqQo7ibm7R1AyIYr8fH-fi-1_35

## Testing Instructions
1. Open Discover tab
2. Tap search bar
3. Select a suggestion or hit enter to land on the results screen
- [ ] filtering works
- [ ] filter chips float above the list
4. Select an episode for playback
- [ ] play button updates its state properly
5. Follow some podcasts
- [ ] followed state updates
7. Now sign in with a Plus account that has some folders
8. Navigate to Podcasts and tap the magnifier glass in the toolbar
9. Enter a query that would yield some folder results
- [ ] folders and local podcasts appear the top of the list
10. Filtering also works in local search mode
11. Now Disable Imroved search results FF
12. Play the previous scenarios to see if search still works

## Screenshots or Screencast 

Notice the filters tab bar (solid bg, content slips below the tab bar. + faded right edge)
Screenshot_20251021_151536

https://github.com/user-attachments/assets/194eadf0-03c0-4f70-85aa-431b1428d16c

## Checklist

### DIFF
--- a/modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/component/SearchResultFilters.kt
+++ b/modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/component/SearchResultFilters.kt
@@ -23,16 +23,18 @@ import au.com.shiftyjelly.pocketcasts.compose.preview.ThemePreviewParameterProvi
 import au.com.shiftyjelly.pocketcasts.compose.theme
 import au.com.shiftyjelly.pocketcasts.ui.theme.Theme
 
+private val defaultFadeConfig = FadeConfig.Default.copy(
+    showStartFade = false,
+    showEndFade = true,
+)
+
 @Composable
 fun SearchResultFilters(
     items: List<String>,
     selectedIndex: Int,
     onFilterSelect: (Int) -> Unit,
     modifier: Modifier = Modifier,
-    fadeConfig: FadeConfig = FadeConfig.Default.copy(
-        showStartFade = false,
-        showEndFade = true,
-    ),
+    fadeConfig: FadeConfig = defaultFadeConfig,
 ) {
     FadedLazyRow(
         fadeConfig = fadeConfig,
@@ -69,14 +71,16 @@ private fun SearchFilterPill(
                         shape = CircleShape,
                     )
                 } else {
-                    Modifier.border(
-                        width = 1.dp,
-                        color = MaterialTheme.theme.colors.primaryIcon02,
-                        shape = CircleShape,
-                    ).background(
-                        color = MaterialTheme.colors.background,
-                        shape = CircleShape,
-                    )
+                    Modifier
+                        .border(
+                            width = 1.dp,
+                            color = MaterialTheme.theme.colors.primaryIcon02,
+                            shape = CircleShape,
+                        )
+                        .background(
+                            color = MaterialTheme.colors.background,
+                            shape = CircleShape,
+                        )
                 },
             )
             .clickable(onClick = onSelect)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          4 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>ImprovedEpisodeRowViewModel</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>ImprovedSearchManagerImpl</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-android/milestone/105">7.100 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/component/ImprovedSearchEpisodeResultRow.kt
```diff
@@ -65,26 +71,30 @@
...
+    episodePlaybackFlow: Flow<EpisodePlaybackData?>,
     modifier: Modifier = Modifier,
-    fetchEpisode: (suspend (EpisodeItem) -> BaseEpisode?)? = null,
+    fetchEpisode: (suspend (ImprovedSearchResultItem.EpisodeItem) -> BaseEpisode?)? = null,
 ) {
     val baseEpisode: BaseEpisode? by produceState(null) {
         value = fetchEpisode?.invoke(episode)
     }
+    val playbackData by episodePlaybackFlow
+        .map { if (it?.playingEpisodeUuid == episode.uuid) it else null }.collectAsState(null)
```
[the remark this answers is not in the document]
MiSikora: 🚫 Yes, this is correct. This will result on re-creating the flow on every recomposition.

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/component/ImprovedSearchEpisodeResultRow.kt
```diff
@@ -159,7 +170,18 @@ private fun ImprovedSearchEpisodeResultRow(
                     }
                 },
                 update = { playButton ->
-                    playButton.setButtonType(episode, buttonType = PlayButtonType.PLAY, color = buttonColor, null)
+                    val theEpisode = episode.apply {
+                        playing = playbackData != null
+                        playingStatus = if (playbackData != null) EpisodePlayingStatus.IN_PROGRESS else EpisodePlayingStatus.NOT_PLAYED
+                        playedUpToMs = playbackData?.playbackPosition ?: 0
+                    }
```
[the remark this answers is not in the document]
sztomek: you wish buddy, `episode` is `BaseEpisode` which is an interface so copy is not an option

modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/buttons/PlayButton.kt
```diff
@@ -125,12 +126,13 @@ class PlayButton @JvmOverloads constructor(
     }
 
     fun setButtonType(episode: BaseEpisode, buttonType: PlayButtonType, @ColorInt color: Int, fromListUuid: String?) {
-        if (buttonType == this.buttonType && episode.uuid == this.episodeUuid && this.buttonColor == color) {
+        if (buttonType == this.buttonType && episode.uuid == this.episodeUuid && this.buttonColor == color && episode.playedUpToMs == this.playedUpToMs) {
```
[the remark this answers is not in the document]
sztomek: maybe that was the point?

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/component/ImprovedSearchEpisodeResultRow.kt
```diff
@@ -93,13 +174,21 @@ private fun ImprovedSearchEpisodeResultRow(
...
     playButtonListener: PlayButton.OnClickListener,
     modifier: Modifier = Modifier,
-    episode: BaseEpisode? = null,
 ) {
+    val viewModel: ImprovedEpisodeRowViewModel = hiltViewModel(
+        key = episodeUuid,
+        creationCallback = { factory: ImprovedEpisodeRowViewModel.Factory ->
+            factory.create(episodeUuid = episodeUuid, podcastUuid = podcastUuid)
+        },
+    )
```
MiSikora, line 188: 🚫 This approach causes several issues.

The `hiltViewModel()` function itself is fine, but it shouldn’t be used at such a granular level. It belongs in top-level composables. For example, in a navigation function or a fragment entry point.

Currently, this setup creates a new `ViewModel` for every row and starts a flow subscription that isn’t cancelled until the entire screen is dismissed. Each flow is collected inside the `ViewModel`’s scope, which means that in the worst-case scenario, you could end up with 50 active subscriptions at the same time. This configuration should be hoisted and provided either as a state object or through a specific flow that stops collecting once the composable leaves the composition.

Another problem with using `hiltViewModel()` this way is that Compose Preview doesn’t work. When attempting to preview it in the IDE, the following message appears:

> `This preview uses a ViewModel. ViewModels often trigger operations not supported by Compose Preview, such as database access, I/O operations, or network requests. You can read more about preview limitations in our external documentation.`

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/component/SearchResultFilters.kt
```diff
@@ -29,7 +31,11 @@ fun SearchResultFilters(
     onFilterSelect: (Int) -> Unit,
     modifier: Modifier = Modifier,
 ) {
-    LazyRow(
+    FadedLazyRow(
+        fadeConfig = FadeConfig.Default.copy(
+            showStartFade = false,
+            showEndFade = true,
+        ),
```
MiSikora: 💡 This could be extracted to a top level value so the object isn't recreated on every recomposition.

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/component/SearchResultFilters.kt
```diff
@@ -67,6 +73,9 @@ private fun SearchFilterPill(
                         width = 1.dp,
                         color = MaterialTheme.theme.colors.primaryIcon02,
                         shape = RoundedCornerShape(100),
+                    ).background(
+                        color = MaterialTheme.colors.background,
+                        shape = RoundedCornerShape(100),
```
MiSikora: 💡 It can be a `CircleShape` instead.

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchViewModel.kt
```diff
@@ -51,43 +46,50 @@ class SearchViewModel @Inject constructor(
...
+                    }
                 }
-
-                if (!FeatureFlag.isEnabled(Feature.IMPROVED_SEARCH_SUGGESTIONS) && it is SearchUiState.SearchOperation.Loading) {
-                    saveSearchTerm(it.searchTerm)
+            } else {
+                searchHandler.searchResults.collect {
                     showSearchHistory = false
+                    if (_state.value is SearchUiState.OldResults) {
+                        _state.value = SearchUiState.OldResults(operation = it as SearchUiState.SearchOperation<SearchResults.SegregatedResults>)
```
MiSikora: ❓ How do we know that the `it as SearchUiState.SearchOperation<SearchResults.SegregatedResults` casting won't fail?

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchViewModel.kt
```diff
@@ -131,28 +133,23 @@ class SearchViewModel @Inject constructor(
...
-            _state.update {
-                (it as SearchUiState.Results).copy(
+            _state.update { state ->
+                (state as SearchUiState.ImprovedResults).copy(
                     selectedFilterIndex = ResultsFilters.entries.indexOf(filter),
+                    operation = (state.operation as? SearchUiState.SearchOperation.Success)?.let { success ->
+                        success.copy(
+                            results = success.results.copy(filter = filter),
+                        )
+                    } ?: state.operation,
```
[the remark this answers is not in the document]
MiSikora: ⚠️ This sounds right. Could we check the current operation first? If it is success that update the state with updated results and if it is not do nothing?

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchViewModel.kt
```diff
@@ -131,28 +133,23 @@ class SearchViewModel @Inject constructor(
...
                 mapOf(
-                    "source" to "discover",
+                    "source" to source,
                     "filter" to filter.name,
                 ),
             )
-            _state.update {
-                (it as SearchUiState.Results).copy(
+            _state.update { state ->
+                (state as SearchUiState.ImprovedResults).copy(
```
MiSikora: ❓ Similar question to the one before. How do we know this casting won't fail?

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/search/ImprovedSearchManagerImpl.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    override suspend fun combinedSearch(term: String): List<ImprovedSearchResultItem> {
+        val response = combinedSearchService.combinedSearch(CombinedSearchRequest(term))
+        return response.results.map {
+            when (it) {
+                is CombinedResult.PodcastResult -> ImprovedSearchResultItem.PodcastItem(
+                    uuid = it.uuid,
+                    title = it.title,
+                    author = it.author.orEmpty(),
+                    isFollowed = false, // to be determined later
```
MiSikora: ❓ Does it mean a separate PR?

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/component/ImprovedSearchEpisodeResultRow.kt
```diff
@@ -149,19 +238,43 @@ private fun ImprovedSearchEpisodeResultRow(
...
+                                state.playbackState.positionMs
+                            } else {
+                                state.episode.playedUpToMs
+                            }
+                            durationMs = if (state.playbackState.episodeUuid == episodeUuid) {
+                                state.playbackState.durationMs
+                            } else {
+                                state.episode.durationMs
+                            }
+                        }
```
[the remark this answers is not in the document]
MiSikora: 🚫 This comment is correct. This `update()` function should update only the state of the Android View. Not the upstream state. Doing this can lead to issues like a missing UI state due to no recomposition, etc. If the upstream state needs to be updated it should be done there at an appropriate time.

MiSikora (changes requested):
The feature overall works as described. In my testing I found 2 usability issues.

1. ⚠️ It takes quite some time for the play button to show. This makes the experience a bit jarring. I'd ask the design team if this is fine to proceed with.

https://github.com/user-attachments/assets/45fe59a6-44e5-43cf-b218-c014dbe03d83

2. ℹ️ There seems to be some state mismanagement when navigating to a podcast or an episode. Not a blocker though it would be nice to fix it.

https://github.com/user-attachments/assets/cb26f3ac-9c69-4ceb-9441-a892afc51599

Here are the steps to replicate the issue:

1. Open the search page.
2. Type in something.
3. Commit the search result.
4. Open a podcast.
5. Go back.
6. The page first shows the suggestions screen and then the results.

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchHandler.kt
```diff
@@ -257,6 +314,57 @@ class SearchHandler @Inject constructor(
...
+                            emit(SearchUiState.SearchOperation.Success(searchTerm = query, results = SearchResults.ImprovedResults(results = localResults, filter = ResultsFilters.TOP_RESULTS)))
+                        }
+                        val apiResults = improvedSearchManager.combinedSearch(query).map {
+                            if (it is ImprovedSearchResultItem.PodcastItem) {
+                                it.copy(isFollowed = subscribedUuids.contains(it.uuid))
+                            } else {
+                                it
+                            }
+                        }
+                        val combinedResults = (localResults + apiResults).distinctBy { it.uuid }
```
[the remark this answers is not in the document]
MiSikora: 💡 This sound like a good suggestions.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/search/ImprovedSearchManagerImpl.kt
sztomek: nope, it will determined later in the code :) i'm just assigning the default false here

MiSikora: Could you update the comment to reflect that? I wouldn't think that it means that the value is synced somewhere else in the code.

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/component/ImprovedSearchEpisodeResultRow.kt
sztomek: Hmm i've introduced this ViewModel approach in this commit: eb534db1999ca0a778b5d0cae311a2a1cdabf218
Before that I collected from flows in the episode row composables and produced the state via `produceState`.
That approach would prevent the issue with a large number of parallel executions as the coroutineScope of the composable will be cancelled when it leaves the composition.
However, my Copilot friend gave me an alert that got me thinkin on better ways:
SCR-20251022-jptz

How would you do this?

MiSikora: Depending on what's best for this scenario you can do something like:

```kotlin
val value by remember(episode.uuid) { 
  flow.map().filter().anyOperation() 
}.collectAsState(initialValue)
```

Or have a caching mechanism in a `ViewModel` like here:

https://github.com/Automattic/pocket-casts-android/blob/edd8a153bd647d4d6626846c1dc36d7a7bbd4208/modules/features/filters/src/main/java/au/com/shiftyjelly/pocketcasts/playlists/manual/AddEpisodesViewModel.kt#L92-L101

sztomek:
> ⚠️ It takes quite some time for the play button to show. This makes the experience a bit jarring. I'd ask the design team if this is fine to proceed with.

Yeah, I noticed that as well. But I'm afraid there's nothing I can do... In order to make an episode playable (and to make the `PlayButton` function as intended) we should fetch podcast and episode details first that is an async process. I copied the current logic from the `PodcastViewModel` that displays episode rows with the same `PlayButton`

~Now at a second glance, there may be a way around if I just consturcted a fake `BaseEpisode` with just the mandatory data that `PlayButton.setButtonType` expects...~
Nope, the playback won't work because the `PlayButtonListener.onPlayClicked` looks like this:
```
    override fun onPlayedClicked(episodeUuid: String) {
        launch {
            episodeManager.findEpisodeByUuid(episodeUuid)?.let { episode ->
                episodeManager.markAsNotPlayedBlocking(episode)
            }
        }
    }
```
and `episodeManager.findEpisodeByUuid` won't find the episode that you've never saw before.
Of course, I could provide another `PlayButton.OnClickListener` that makes sure to fetch the podcast first so its episodes will be available but that seems like a long way to go

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/search/ImprovedSearchManagerImpl.kt
sztomek: I'll just remove the comment then

MiSikora:
> Yeah, I noticed that as well. But I'm afraid there's nothing I can do... In order to make an episode playable (and to make the `PlayButton` function as intended) we should fetch podcast and episode details first that is an async process. I copied the current logic from the `PodcastViewModel` that displays episode rows with the same `PlayButton`

Could we instead show a placeholder or an empty space with the size of the button? Not necessarily in this PR but I think this should be addressed in some way. Perhaps even, as you say, create a listener that fetches the episode details. In general I feel like we need to clarify this with designers.

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchViewModel.kt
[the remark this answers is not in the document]
sztomek: gotcha, made the changes

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchHandler.kt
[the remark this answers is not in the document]
sztomek: 🤔 I don't think so. To my best knowledge, folders, podcasts, episodes have uinque ids. Will apply the change nevertheless because it wouldn't hurt.

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchViewModel.kt
sztomek: you're right, cast was unnecessary.

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/component/ImprovedSearchEpisodeResultRow.kt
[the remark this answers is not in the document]
sztomek: taken care of, thanks

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchHandler.kt
```diff
@@ -142,22 +149,66 @@ class SearchHandler @Inject constructor(
...
+                            )
                         }
-                    },
+                    }
+                }
+
+                val suggestions = buildList {
+                    addAll(remoteResults.filterIsInstance<SearchAutoCompleteItem.Term>())
+                    addAll(localResults)
+                    addAll(remoteResults.filter { (it !is SearchAutoCompleteItem.Term) && (onlyRemote || !subscribedUuids.contains((it as? SearchAutoCompleteItem.Podcast)?.uuid.orEmpty())) })
```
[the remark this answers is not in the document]
sztomek: lol, you suggested this exact change on the first place 👍

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/component/SearchResultFilters.kt
```diff
@@ -28,8 +29,13 @@ fun SearchResultFilters(
     selectedIndex: Int,
     onFilterSelect: (Int) -> Unit,
     modifier: Modifier = Modifier,
+    fadeConfig: FadeConfig = FadeConfig.Default.copy(
+        showStartFade = false,
+        showEndFade = true,
+    ),
```
MiSikora: ℹ️ This still creates an object on every function invocation. It needs to be a `val` stored outside of the function to reduce allocations.

modules/features/search/src/main/java/au/com/shiftyjelly/pocketcasts/search/SearchHandler.kt
```diff
@@ -257,6 +314,57 @@ class SearchHandler @Inject constructor(
...
+                            emit(SearchUiState.SearchOperation.Success(searchTerm = query, results = SearchResults.ImprovedResults(results = localResults, filter = ResultsFilters.TOP_RESULTS)))
+                        }
+                        val apiResults = improvedSearchManager.combinedSearch(query).map {
+                            if (it is ImprovedSearchResultItem.PodcastItem) {
+                                it.copy(isFollowed = subscribedUuids.contains(it.uuid))
+                            } else {
+                                it
+                            }
+                        }
+                        val combinedResults = (localResults + apiResults).distinctBy { it::class to it.uuid }
```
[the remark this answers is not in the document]
sztomek: guess who gave this amazing suggestion? YOU!

PR #4632, opened 2025-10-23
Normalize search inputs in the database

MiSikora:
## Description

Currently, when performing a search, we operate on the regular title text fields. This can be problematic because the inputs aren’t normalized. Any accented or non-ASCII letters require an exact match. Meaning, for example, we won’t find `ł` when typing `l`, or `œ` when typing `oe`.

Sometimes we handle a very basic normalization in the code, but sometimes we don’t. When the search is performed on the database side, however, there’s no reliable way to handle it. In this PR I opted for using `unidecode` which is a common library that can be used for that purpose backported to many languages.

There are a few strategies to address this in SQLite, most of which rely on custom functions or collators. Unfortunately, the native Android SQLite library doesn’t support binding external functions. While we could use Requery, which does expose bindings, that would be a separate project in itself. Moreover, we’d likely need to implement native code functions to handle normalization efficiently since JNI interop can lead to performance issues, especially when searching over large data sets.

A common alternative is to store a normalized field directly in the database. This approach has the added advantage of integrating nicely with Full Text Search in the future, should we decide to improve local search capabilities further. Additionally, we can reuse the same normalized fields for searches performed in code, ensuring consistent behavior across all layers.

I handled all tables except for the `podcast_episodes` table during the database migration. Some users have hundreds of thousands of episodes in their databases. In my testing, migrating a database with 1,000,000 episodes on a Pixel 6 took about one minute. All while the app was locked.

To address this, I added a batched worker that runs during the app version migration. It updates episodes in batches of 50,000, which took around two minutes to migrate all episodes. This approach allows users to continue interacting with the app during the migration process.

One surprising piece of code in this PR is the following pattern:

```kotlin
data class Entitiy(
    @ColumnInfo(name = "title") var title: String = "",
) : {
    @ColumnInfo(name = "clean_title")
    var cleanTitle: String = ""
        get() = title.unidecode()
        internal set
}
```

It’s an ugly solution, but a necessary one. Most of our entities are mutable, and for this reason, `cleanTitle` cannot be a constructor property. Otherwise, if we mutate the `title` somewhere in the code, the change wouldn’t be reflected in `cleanTitle`, leading to discrepancies and incorrect search results.

Ideally, we’d make all classes immutable and move the properties to the constructors. Similar to how it’s done in the `ManualPlaylistEpisode` instance. However, this would be a significant refactor and can’t be done ad-hoc due to the potential side effects and wide-reaching implications.

## Testing Instructions

1. Install the app from the main branch.
2. Follow some podcasts and create playlists with non-ascii names.
3. Checkout the `mehow/task/normalized-search` branch.
4. Apply [this patch](https://github.com/user-attachments/files/23123452/version.patch).
```sh
curl -L https://github.com/user-attachments/files/23123452/version.patch | git apply
```
5. Upgrade the app.
6. Open the app.
7. Open the App Inspection.
8. Execute this query and verify the clean titles.
```sql
SELECT title, cleanTitle FROM podcast_episodes
```
9. Execute this query and verify the clean titles.
```sql
SELECT title, clean_title FROM playlists
```

## Checklist

### DIFF
--- a/modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/search/KmpSearch.kt
+++ b/modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/search/KmpSearch.kt
@@ -1,7 +1,6 @@
 package au.com.shiftyjelly.pocketcasts.utils.search
 
 import au.com.shiftyjelly.pocketcasts.utils.extensions.removeAccents
-import java.text.Normalizer
 
 fun String.kmpSearch(pattern: String): List<Int> {
     return KmpSearch(pattern).search(this)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          3 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>EpisodeTitlesNormalizationWorker</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-android/milestone/106">7.101</a>. This milestone is due in less than 2 days.<br />
    </tr>
  </tbody>
</table>

</p>

modules/services/utils/src/test/java/au/com/shiftyjelly/pocketcasts/utils/UnidecodeTest.kt
```diff
@@ -0,0 +1,110 @@
...
+            listOf("ḟ", "ƒ", "ḟ", "Ḟ", "Ƒ", "Ḟ") to "f",
+            listOf("ğ", "ģ", "ĝ", "Ğ", "Ģ", "Ĝ") to "g",
+            listOf("ĥ", "ḧ", "ḩ", "Ĥ", "Ḧ", "Ḩ") to "h",
+            listOf("í", "ì", "î", "Í", "Ì", "Î") to "i",
+            listOf("ĵ", "ǰ", "ȷ", "Ĵ", "Ɉ", "ȷ") to "j",
+            listOf("ķ", "ḱ", "ǩ", "Ķ", "Ḱ", "Ǩ") to "k",
+            listOf("ĺ", "ļ", "ł", "Ĺ", "Ļ", "Ł") to "l",
+            listOf("ḿ", "ṁ", "ṃ", "Ḿ", "Ṁ", "Ṃ") to "m",
+            listOf("ń", "ñ", "ň", "Ń", "Ñ", "Ň") to "n",
+            listOf("ó", "ò", "ö", "Ó", "Ò", "Ö") to "o",
```
sztomek, line 25: ⚠️ in hungarian, there is another letter `ő` and `Ő`

modules/services/utils/src/test/java/au/com/shiftyjelly/pocketcasts/utils/UnidecodeTest.kt
```diff
@@ -0,0 +1,110 @@
...
+            listOf("ṕ", "ṗ", "ƥ", "Ṕ", "Ṗ", "Ƥ") to "p",
+            listOf("ɋ", "ʠ", "ʠ", "Ɋ", "ʠ", "ℚ") to "q",
+            listOf("ŕ", "ř", "ṙ", "Ŕ", "Ř", "Ṙ") to "r",
+            listOf("ś", "š", "ş", "Ś", "Š", "Ş") to "s",
+            listOf("ţ", "ť", "ṫ", "Ţ", "Ť", "Ṫ") to "t",
+            listOf("ú", "ù", "ü", "Ú", "Ù", "Ü") to "u",
```
sztomek, line 31: ⚠️ in hungarian, there is another letter `ű` and `Ű`

sztomek (commented):
Works very well!

Image
Image

modules/services/utils/src/test/java/au/com/shiftyjelly/pocketcasts/utils/UnidecodeTest.kt
MiSikora: The list wasn't meant to be exhaustive. For example for the letter `a` alone there is like 60+ different characters. It is more of a sanity check test. But let me know if you want me to add these two letters. I can follow up in a separate PR.

PR #4652, opened 2025-10-27
Fix Firebase remote config init

sztomek:
## Description
We've noticed that our prod versions won't launch with the new onboarding screens during a clean install, however the feature flags should be enabled for version 7.98 and above!
It was tricky to find the root cause because it is an async/timing issue.
Sometimes the config takes more time to load and by the time our code logic that decides where to route executes, it's too late.
This PR addresses this issue by halting the control flow of `MainActivity` until the firebase config is activated and fetched.
To encapsulate the init logic, I've created `FirebaseRemoteFeatureInitializer` and `MainActivity.onCreate` will take it into account.
Another small improvement this PR contains is the additional of the `splashScreen` APIs that solve some of our initial screen flickering during clean app launch.

## Testing Instructions
1. Apply this patch [prototype_to_see_firebase_remote_config_in_action.patch](https://github.com/user-attachments/files/23167127/prototype_to_see_firebase_remote_config_in_action.patch)
2. Build Prototype version
3. Launch the app
- [ ] You should land on the new onboarding screens
4. Complete onboarding
5. Kill and re-launch the app
- [ ] Things work as before

## Checklist
- [x] If this is a user-facing change, I have added an entry in CHANGELOG.md
- [x] Ensure the linter passes (`./gradlew spotlessApply` to automatically apply formatting/linting)
- [ ] ~I have considered whether it makes sense to add tests for my changes~
- [ ] ~All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml`~
- [ ] ~Any jetpack compose components I added or changed are covered by compose previews~
- [ ] ~I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.~

PR #4662, opened 2025-10-28
Make sure Firebase RemoteConfig is fetched before launching onboarding

sztomek:
We've noticed that our prod versions won't launch with the new onboarding screens during a clean install, however the feature flags should be enabled for version 7.98 and above!
It was tricky to find the root cause because it is an async/timing issue.
Sometimes the config takes more time to load and by the time our code logic that decides where to route executes, it's too late.
This PR addresses this issue by halting the control flow of `MainActivity` until the firebase config is activated and fetched.
To encapsulate the init logic, I've created `FirebaseRemoteFeatureInitializer` and `MainActivity.onCreate` will take it into account.
Another small improvement this PR contains is the additional of the `splashScreen` APIs that solve some of our initial screen flickering during clean app launch.

## Testing Instructions
1. Apply this patch 
[prototype_to_see_firebase_remote_config_in_action_-_updated2.patch](https://github.com/user-attachments/files/23506622/prototype_to_see_firebase_remote_config_in_action_-_updated2.patch)

2. Build Prototype version
3. Launch the app
- [x] You should land on the new onboarding screens
4. Complete onboarding
5. Kill and re-launch the app
- [x] Things work as before

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>FirebaseRemoteFeatureInitializer</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>

</p>

modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/featureflag/FirebaseRemoteFeatureInitializer.kt
```diff
@@ -0,0 +1,31 @@
...
+
+    private val isInitialized = AtomicBoolean(BuildConfig.DEBUG || BuildConfig.IS_PROTOTYPE)
+
+    suspend fun initialize() {
+        if (isInitialized.get()) return
+
+        val result = try {
+            val success = firebaseRemoteConfig.fetchAndActivate().await()
+            Timber.i("Firebase feature flag refreshed")
+            success
```
geekygecko:
```suggestion
            withTimeout(3000L) {
                val success = firebaseRemoteConfig.fetchAndActivate().await()
                Timber.i("Firebase feature flag refreshed")
                success
            }
```
Do you think we should add a timeout in case there are any issues calling Firebase? So the user doesn't get stuck on the splash screen?

app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
```diff
@@ -414,23 +419,43 @@ class MainActivity :
...
+        val splashScreen = installSplashScreen()
+
         super.onCreate(savedInstanceState)
         theme.setupThemeForConfig(this, resources.configuration)
         enableEdgeToEdge(navigationBarStyle = theme.getNavigationBarStyle(this))
         bottomSheetTag = savedInstanceState?.getString(SAVEDSTATE_BOTTOM_SHEET_TAG)
 
         playbackManager.setNotificationPermissionChecker(this)
 
+        var isReady = false
```
geekygecko:
```suggestion
        var isReady = settings.hasCompletedOnboarding()
```
Should we only wait if the user hasn't completed the onboarding?

app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
```diff
@@ -414,23 +419,43 @@ class MainActivity :
...
+                }
+            }
+        }
+    }
+
+    private fun onFirebaseInitComplete(savedInstanceState: Bundle?) {
         val showOnboarding = !settings.hasCompletedOnboarding() && !syncManager.isLoggedIn()
-        // Only show if savedInstanceState is null in order to avoid creating onboarding activity twice.
-        if (showOnboarding && savedInstanceState == null) {
+        if (showOnboarding) {
```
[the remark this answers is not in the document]
MiSikora: @sztomek I believe this deletion led to the issue I described [here](https://github.com/Automattic/pocket-casts-android/pull/4652#pullrequestreview-3384187341). Can't verify for sure as I'm at the airport but you can test by starting the onboarding process and toggling configuration by changing OS dark mode.

sztomek:
@geekygecko @MiSikora 
updated the PR + rebased over main. the patch has been also updated.
please do some testing as I did and let me know if you find something that should be addressed! 🙏

modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/featureflag/FirebaseRemoteFeatureInitializer.kt
```diff
@@ -0,0 +1,38 @@
...
+@Singleton
+class FirebaseRemoteFeatureInitializer @Inject constructor(
+    private val firebaseRemoteConfig: FirebaseRemoteConfig,
+) {
+
+    private companion object {
+        const val DEFAULT_TIMEOUT = 3_000L
+    }
+
+    private val isInitialized = AtomicBoolean(BuildConfig.DEBUG || BuildConfig.IS_PROTOTYPE)
```
MiSikora: ❓ I'm not sure I understand why `AtomicBoolean` is used here. Where its atomicity is being used?

MiSikora (changes requested):
🚫 [The issue](https://github.com/Automattic/pocket-casts-android/pull/4662/files#r2471912708) with multiple onboarding pages stacking I reported here still happens.

modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/featureflag/FirebaseRemoteFeatureInitializer.kt
sztomek: the purpose was to guarantee that the init block runs only once. replaced with mutex in fa5dbc564121aebceb8a31cee63520f39c2d584a

sztomek:
> [The issue](https://github.com/Automattic/pocket-casts-android/pull/4662/files#r2471912708) with multiple onboarding pages stacking I reported still happens.

Multiple instaneces of the MainActivity spawn because there's a config change. I believe it's either `WindowCompat.setDecotFitsSystemWindows` or the `installSplashScreen` that triggers that.
However `repeatOnLifecycle(Lifecycle.State.STARTED)` is supposed to guarantee that the init only runs once, on the latest instance
LMK what else to change

MiSikora:
> Multiple instaneces of the MainActivity spawn because there's a config change. I believe it's either `WindowCompat.setDecotFitsSystemWindows` or the `installSplashScreen` that triggers that.

Multiple instances of `MainActivity` were not spawning. Multiple instances of `OnboardingActivity` were. The cause was the missing `savedInstanceState == null` guard. During startup `OnboardingActivity` was created and added to the back stack. After a configuration change, it was created again and added on top of the back stack a second time.

> However `repeatOnLifecycle(Lifecycle.State.STARTED)` is supposed to guarantee that the init only runs once, on the latest instance

That’s not the case. We shouldn’t run one-time setup from `STARTED`. It will re-run on every start/stop cycle, which is what’s causing the crash now.

https://github.com/user-attachments/assets/cd9f8aff-8503-491a-a0ff-f2bb3f34f006

```
java.lang.IllegalArgumentException: Maximum number of items supported by BottomNavigationView is 5. Limit can be checked with BottomNavigationView#getMaxItemCount()
	at com.google.android.material.navigation.NavigationBarMenu.addInternal(NavigationBarMenu.java:83)
	at androidx.appcompat.view.menu.MenuBuilder.add(MenuBuilder.java:475)
	at androidx.appcompat.view.SupportMenuInflater$MenuState.addItem(SupportMenuInflater.java:547)
	at androidx.appcompat.view.SupportMenuInflater.parseMenu(SupportMenuInflater.java:223)
	at androidx.appcompat.view.SupportMenuInflater.inflate(SupportMenuInflater.java:142)
	at com.google.android.material.navigation.NavigationBarView.inflateMenu(NavigationBarView.java:544)
	at au.com.shiftyjelly.pocketcasts.ui.MainActivity.onFirebaseInitComplete(MainActivity.kt:499)
	at au.com.shiftyjelly.pocketcasts.ui.MainActivity.access$onFirebaseInitComplete(MainActivity.kt:235)
	at au.com.shiftyjelly.pocketcasts.ui.MainActivity$onCreate$2$1$1.invokeSuspend(MainActivity.kt:460)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:34)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:100)
	at android.os.Handler.handleCallback(Handler.java:1041)
	at android.os.Handler.dispatchMessage(Handler.java:103)
	at android.os.Looper.dispatchMessage(Looper.java:315)
	at android.os.Looper.loopOnce(Looper.java:251)
	at android.os.Looper.loop(Looper.java:349)
	at android.app.ActivityThread.main(ActivityThread.java:9041)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:593)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:929)
	Suppressed: kotlinx.coroutines.internal.DiagnosticCoroutineContextException: [StandaloneCoroutine{Cancelling}@24fa77c, Dispatchers.Main.immediate]
	at au.com.shiftyjelly.pocketcasts.ui.MainActivity.onFirebaseInitComplete(MainActivity.kt:499)
	at au.com.shiftyjelly.pocketcasts.ui.MainActivity.access$onFirebaseInitComplete(MainActivity.kt:235)
	at au.com.shiftyjelly.pocketcasts.ui.MainActivity$onCreate$2$1$1.invokeSuspend(MainActivity.kt:460)
```

Please revert to the previous approach:

* Gate onboarding setup with `if (savedInstanceState == null)`.
* Run initialization from `CREATED` lifecycle, not from `STARTED`. But more broadly there shouldn't be any need to call `repeatOnLifecycle()`. We don't want to repeat that code. We want it to run once. And the code won't be called if the activity is destroyed because lifecycle scope will be canceled anyway.

This ensures `OnboardingActivity` is only added once and the init code runs exactly once per creation.

sztomek:
@MiSikora 
you were right about the `Lifecycle.STARTED`, I went back to `CREATED`. thank you.
However, I'm clearly seeing 2 `MainActivity` instances. Added extra logs so you can see it too 
[firebase_-_prototype_-_logs_added.patch](https://github.com/user-attachments/files/23525971/firebase_-_prototype_-_logs_added.patch), just filter for "===" in logcat.
I'm afraid adding back the 'safeguard' `savedInstanceState==null` check won't solve anything, but! Since there first `MainActivity` instance gets destroyed and recreated with `savedInstanceState != null`.
I also noticed on your video that you're not running the prototype build.

I'm pasting the logs i'm seeing with my patch applied for reference:
[prototype-logs_logcat.txt](https://github.com/user-attachments/files/23526160/prototype-logs_logcat.txt) (rename it to `.logcat`)

MiSikora:
Can you describe a scenario that results in two instances of `MainActivity` being alive at the same time? I assume you mean something like this:

1. Open the app.
2. Switch the system dark theme setting.

This does not lead to two instances being alive at the same time. You will see `onCreate()` being called twice for two different `MainActivity` instances, but one of them is destroyed almost immediately. This happens because of how theming in this app works: it recreates `MainActivity` with the new theme applied.

https://github.com/Automattic/pocket-casts-android/blob/4177bae2c4327ffa716aa58d98a30660bf3401cf/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt#L1633-L1636

https://github.com/Automattic/pocket-casts-android/blob/4177bae2c4327ffa716aa58d98a30660bf3401cf/modules/services/ui/src/main/java/au/com/shiftyjelly/pocketcasts/ui/theme/Theme.kt#L236

You can use a different type of configuration change to see that it is only recreated once, because the theme is not being reapplied.

As for the safeguard, I guess it no longer works reliably because of the asynchronous Firebase call. However, we still want to create `OnboardingActivity` and add it to the back stack only once during startup. We still need to rely on `savedInstanceState`, because it is the only transient mechanism that survives process death. But now we need to put a custom variable in the bundle to know whether it has been shown or not.

The current solution results in two bugs:

- In debug builds, multiple instances of `OnboardingActivity` stack on top of each other. The solution should be stable regardless of the build type and should not result in this even if those are debug builds.
- In prototype/production builds, `OnboardingActivity` can open at random moments:
  - During configuration changes.
  - When minimizing and maximizing the app.
  - When going back from full-screen video playback.
  - Probably in some other scenarios.


https://github.com/user-attachments/assets/7c16434b-0572-445f-9f36-2ba8cd434db6

sztomek:
Closing this one. Will get back to it from a fresh POV once I'm back from vac

PR #4664, opened 2025-10-28
Support for AI agents

geekygecko:
Now that we are using GitHub Copilot to review our pull requests, it would be helpful to be able to add new rules. For example, in the recent pull request, I completely forgot that we are meant to use WebP images instead of PNGs. By adding the `copilot-instructions.md` file, we should be able to get this prompt in the future (if I forget again 🤦). 

I've also added an `AGENTS.md` file. It seems like a good idea if people are using Claude, Gemini, or GitHub Copilot locally. 

Initially, I thought we might be able to use `AGENTS.md` without `copilot-instructions.md`, but the following documentation suggests it might not be fully supported.

https://docs.github.com/en/copilot/concepts/prompting/response-customization#about-repository-custom-instructions

> Agent instructions, which are similar to repository-wide custom instructions, but are currently not supported by all Copilot features.
> These are specified in files called AGENTS.md, CLAUDE.md, or GEMINI.md.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -26,6 +26,15 @@
 
 ## Image Resources
 
+- Add images to `modules/services/images` only
+
+### Vector Graphics Preferred
+
+- **Prefer vector graphics (.svg)** over rasterized formats when possible
+  - Scalable to any screen density without quality loss
+  - Smaller file size for simple graphics
+  - Use for icons, logos, and simple illustrations
+
 ### WebP Format Required
 
 - **New images**: Always use `.webp` format for better compression and smaller APK
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -160,6 +160,13 @@ Core Services (model, servers)
 
 ### Image Resources
 
+- Add images to `modules/services/images` only
+
+**Prefer vector graphics (.svg) over rasterized formats when possible**:
+- Scalable to any screen density without quality loss
+- Smaller file size for simple graphics
+- Use for icons, logos, and simple illustrations
+
 **Prefer WebP format over PNG**:
 - New images should be `.webp` format for better compression and smaller APK size
 - PNG acceptable only for:


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

bjtitus:
Just a suggestion, Claude [doesn't support](https://github.com/anthropics/claude-code/issues/6235) `AGENTS.md` but `CLAUDE.md` instead. You might consider adding `CLAUDE.md` with `@AGENTS.md` inside to link the two.

geekygecko:
> Just a suggestion, Claude [doesn't support](https://github.com/anthropics/claude-code/issues/6235) `AGENTS.md` but `CLAUDE.md` instead. You might consider adding `CLAUDE.md` with `@AGENTS.md` inside to link the two.

Thanks @bjtitus, I have added that link.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,42 @@
...
+
+- **TODO only**: Use `TODO` for temporary notes, NEVER use `FIXME` (not allowed)
+- **Nullable handling**: Prefer safe calls (`?.`) and Elvis operator (`?:`) over `!!`
+
+## String Resources
+
+- Add strings to `modules/services/localization` only
+
+## Image Resources
+
```
sztomek, line 28: 💡 should specify `modules/services/images` as the drawable assets locations

sztomek: 💡 prefer using vector graphics (`.svg`s) over rasterized formats when possible

geekygecko: These are great suggestions, thanks.

PR #4695, opened 2025-11-06
[Playback] Top shows story

sztomek:
## Description
Adds the top shows story.
Now we can change the control colors (X button and story progress indicator) as well.
Removed obsolete resources

Figma: lH66LwxxgG8btQ8NrM0ldx-fi-4393_3273
Animation: https://video.wordpress.com/v/wdErJ6XQ

Fixes https://linear.app/a8c/issue/PCDROID-225/story-3-top-5-shows

## Testing Instructions
1. Sign in with an account that has playback data
2. Step until you reach the top shows story
3. Inspect

## Screenshots or Screencast 

https://github.com/user-attachments/assets/7ec562c5-a10a-4441-b18b-c032100227cf

## Checklist

### DIFF
--- a/modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/ui/TopShowsStory.kt
+++ b/modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/ui/TopShowsStory.kt
@@ -120,6 +120,7 @@ internal fun TopShowsStory(
             }
         }
         ShareStoryButton(
+            modifier = Modifier.padding(bottom = 18.dp),
             story = story,
             controller = controller,
             onShare = onShareStory,


modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/ui/StoriesPage.kt
```diff
@@ -261,6 +262,7 @@ private fun Stories(
 internal fun BoxScope.TopControls(
     pagerState: PagerState,
     progress: Float,
+    colors: List<Color>,
```
MiSikora: Instead of passing in list of colors could pass in a single colors and hoist the `colors[page]` logic to the top? It would be also good to change it to `colors.getOrNull(index)`. It avoids duplication and potentially risky `colors[index]` where index is not available

MiSikora (commented):
Looks like the bottom padding around the share button is missing.

Image

modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/ui/StoriesPage.kt
sztomek: thanks, addressed it here 1a78e73215a077d8ccb9c7b7e787e1740c5ac5fd

sztomek:
Addressed button padding here de9c019b1e562c0bc6fc5c941ea99a025aa2e2a4
I think in the long run we should hoist the button up a level so we could universally control when and where it should appear. But I'd wait with it until all the stories are updated

MiSikora (approved):

PR #4742, opened 2025-11-17
Analytics event if Playback screenshots

geekygecko:
## Description

This change adds an analytic so we can consider if people are using the phone screenshot more than our own share button. 

Fixes https://linear.app/a8c/issue/PCDROID-306/consider-tracking-screenshots-as-sharing-events

## Testing Instructions

1. Log in with an account that has playback data
2. Open playback
3. Go to a story with the share button
4. Use the phone hardware buttons to take a screenshot
5. ✅ Verify the dialog in the app offers to share use our image 
6. ✅ Verify the analytic `end_of_year_story_shared` is in the logs with the properties `year`, `story`, `activity`, and `from=button`.

## Checklist
 
- [x] with different themes

### DIFF
--- a/modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/EndOfYearViewModel.kt
+++ b/modules/features/endofyear/src/main/java/au/com/shiftyjelly/pocketcasts/endofyear/EndOfYearViewModel.kt
@@ -1,5 +1,6 @@
 package au.com.shiftyjelly.pocketcasts.endofyear
 
+import android.app.Activity
 import androidx.compose.runtime.Immutable
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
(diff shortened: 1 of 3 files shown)


sztomek (commented):
Works fine when you take screenshot with the hardware keys combo.
However if you prefer the Recents app's Screenshot button, that would go unnoticed by our app. 
This screencap is to demonstrate those scenarios:
https://github.com/user-attachments/assets/5de4d7b9-d8b7-4c20-af28-0ca5ab592f0a

Gave it a quick look and tried to extend the scope of `StoriesActivity.screenshotDetector` to register in `onStart` and unregister in `onDestroy` but that didn't really solve the issue 😞

geekygecko:
> However if you prefer the Recents app's Screenshot button, that would go unnoticed by our app.

I've taken a look at supporting this, but I don't think it's possible. I also found this in the documentation.
> In Android 14, the system API only detects a screenshot if the user performs a specific combination of hardware button presses. 
https://developer.android.com/about/versions/14/features/screenshot-detection

PR #4762, opened 2025-11-20
Enable Playback feature flag and change story duration

geekygecko:
## Description

This makes three changes.
- Enables the Playback feature flag with Firebase Remote Config
- Adds the feature to the changelog
- Changes the story duration to 7 seconds to match the Figma spec "Stories automatically skip after 7s."

## Testing Instructions

1. Log in with a user that has playback data
2. ✅ Verify the stories only last for 7 seconds

## Checklist
 
- [x] with different themes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 8.1
 -----
+*   New Features
+    *   Add Playback 2025 so you can get a recap of your listening habits from this year
+        ([#4762](https://github.com/Automattic/pocket-casts-android/pull/4762))
 *   Updates
     *   Add periodic requests for the app rating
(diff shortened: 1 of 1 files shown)


sztomek (approved):
Looks good to me. 
My only question/concern is if we should take the same approach with FFs that iOS does: to set them true when rolling them out to prod and rely on Firebase to disable them if something goes bad?

geekygecko:
> My only question/concern is if we should take the same approach with FFs that iOS does: to set them true when rolling them out to prod and rely on Firebase to disable them if something goes bad?

I will leave it set to false by default for this release, but we should discuss whether our Firebase rollout approach needs to be consistent across platforms.

Issue #1719, opened 2024-01-19
Horizontal artworks are cropped in the player view


MiSikora:
### Description

When a podcast episode has a horizontal artwork it is not displayed correctly. It is cropped instead of being fitted. I did some investigation and the bug was introduced with the fix in #1665. However, digging a little bit more it turns out that `coil.transform.RoundedCornersTransformation` is at fault. Commenting out this line fixes the issue.

https://github.com/Automattic/pocket-casts-android/blob/17d049450a9419d8aa6f9f84fabf67b6704e2b7c/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/PlayerHeaderFragment.kt#L372

Supposedly we can fix it with a custom transformer mentioned here https://github.com/coil-kt/coil/issues/421#issuecomment-640133205. The linked implementation is no longer valid as the API changed but it can be probably used as an entry point for our implementation if I solve it this way.

I also tried updating Coil but that didn't help.

### Step-by-step reproduction instructions

1. Play [this episode](https://pca.st/episode/c7bc3a08-17b7-48c8-b5b9-dbdfd63211f4).
2. Open up the player.
3. You'll see a cropped artwork.

### Screenshots or screen recording

| With transformation | Without transformation |
| - | - |
| ![Screenshot_20240119-124053](https://github.com/Automattic/pocket-casts-android/assets/30936061/71d89d21-e92b-41cd-906e-5c1697da70bf) | ![Screenshot_20240119-124814](https://github.com/Automattic/pocket-casts-android/assets/30936061/4bf3864f-0417-473d-8142-8e75d7dd450d) |

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

Google Pixel 6, Android 13, Pocket Casts `7.55-rc-2 (9184)`

Inrumpo:
Possibly connected to #1660 ?

MiSikora:
> Possibly connected to https://github.com/Automattic/pocket-casts-android/issues/1660 ?

Yes, it looks like the same issue. Thanks!

PR #4821, opened 2025-12-04
Migrate Coil from 2.7.0 to 3.3.0

geekygecko:
## Description

This change migrates to the new Coil 3 library. It includes some of the following changes:
- The Compose methods have been updated to be restartable and skippable which should improve their performance.
- By default, Coil 3.x does not include support for loading images from the network. This is to avoid forcing a large networking dependency on users who want to use their own networking solution or do not need network URL support (e.g. only loading images from disk).
- Improved Compose @Preview rendering and support for custom preview behavior via LocalAsyncImagePreviewHandler.

[Coil migration guide](https://coil-kt.github.io/coil/upgrading_to_coil3/)

Fixes https://linear.app/a8c/issue/PCDROID-357/migrate-to-coil-3

## Testing Instructions

- [ ] Test artwork in the widgets
- [ ] Test artwork in the notification shade
- [ ] Test episode artwork [Previous PR](https://github.com/Automattic/pocket-casts-android/issues/1719)
- [ ] Test artwork 
- [ ] Test Wear OS
- [ ] Test Automotive OS

## Screenshots

| | | |
| --- | --- | --- |
| Screenshot_20251204_172517 | Screenshot_20251204_172746 | Screenshot_20251204_172827 | 

## Checklist
 
- [x] with different themes

### DIFF
--- a/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/ProfileCircleView.kt
+++ b/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/ProfileCircleView.kt
@@ -19,6 +19,7 @@ import coil3.Image
 import coil3.imageLoader
 import coil3.request.ImageRequest
 import coil3.request.transformations
+import coil3.target.Target
 import coil3.toBitmap
 import coil3.transform.CircleCropTransformation
 import au.com.shiftyjelly.pocketcasts.ui.R as UR
@@ -158,7 +159,7 @@ class ProfileCircleView @JvmOverloads constructor(
     }
 
     private fun loadImageFromUri(uri: Uri, size: Int) {
-        val target = object : coil3.target.Target {
+        val target = object : Target {
             override fun onError(error: Image?) {
             }
             override fun onSuccess(result: Image) {
--- a/modules/services/ui/src/main/java/au/com/shiftyjelly/pocketcasts/ui/images/ThemedImageTintTransformation.kt
+++ b/modules/services/ui/src/main/java/au/com/shiftyjelly/pocketcasts/ui/images/ThemedImageTintTransformation.kt
@@ -14,6 +14,7 @@ import au.com.shiftyjelly.pocketcasts.ui.extensions.getThemeColor
 import au.com.shiftyjelly.pocketcasts.ui.theme.Theme
 import coil3.size.Size
 import coil3.transform.Transformation
+import coil3.Bitmap as CoilBitmap
 
 class ThemedImageTintTransformation(context: Context) : Transformation() {
     private val isActive = Theme.isImageTintEnabled(context)
@@ -32,7 +33,7 @@ class ThemedImageTintTransformation(context: Context) : Transformation() {
         return id.hashCode()
     }
 
-    override suspend fun transform(input: coil3.Bitmap, size: Size): coil3.Bitmap {
+    override suspend fun transform(input: CoilBitmap, size: Size): CoilBitmap {
         val result = Bitmap.createBitmap(input.width, input.height, Bitmap.Config.ARGB_8888)
         if (!isActive) return input
 


modules/features/widgets/src/main/kotlin/au/com/shiftyjelly/pocketcasts/widget/ui/ClassicPlayer.kt
```diff
@@ -167,11 +166,8 @@ private fun Cover(
         LaunchedEffect(episode.uuid, useEpisodeArtwork) {
             val requestFactory = PocketCastsImageRequestFactory(context).smallSize()
             val request = requestFactory.create(episode.toBaseEpisode(), useEpisodeArtwork)
-            var drawable: Drawable? = null
-            while (drawable == null) {
-                drawable = context.imageLoader.execute(request).drawable
-            }
```
geekygecko, line 173: I wasn't quite sure why there was a while loop here, and it works without it. Please let me know if you know if I need to add it back in.

modules/features/widgets/src/main/kotlin/au/com/shiftyjelly/pocketcasts/widget/ui/EpisodeImage.kt
```diff
@@ -74,11 +73,10 @@ internal fun EpisodeImage(
                 cornerRadius = if (isSystemCornerRadiusSupported) 0 else 6,
             )
             val request = requestFactory.create(episode.toBaseEpisode(), useEpisodeArtwork)
-            var drawable: Drawable? = null
-            while (drawable == null) {
-                drawable = context.imageLoader.execute(request).drawable
+            val image = context.imageLoader.execute(request).image
+            episodeBitmap = withContext(Dispatchers.Default) {
+                image?.toBitmap()
             }
```
geekygecko, line 79: Same as above, I wasn't sure why this while loop is needed.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/images/PocketCastsImageRequestFactory.kt
```diff
@@ -88,7 +90,7 @@ data class PocketCastsImageRequestFactory(
         is RequestType.Podcast -> data(context)
         is RequestType.PodcastEpisode -> data(context)
         is RequestType.UserEpisode -> data()
-        is RequestType.FileOrUrl -> filePathOrUrl.toHttpUrlOrNull() ?: File(filePathOrUrl)
+        is RequestType.FileOrUrl -> filePathOrUrl
```
geekygecko, line 93: I had to change this to get images on Automotive loading.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/images/PodcastImageColorAnalyzer.kt
```diff
@@ -21,7 +22,7 @@ class PodcastImageColorAnalyzer @Inject constructor(
         val request = requestFactory.createForPodcast(uuid)
         val bitmap = when (val result = imageLoader.execute(request)) {
             is SuccessResult -> result.memoryCacheKey?.let { key ->
-                imageLoader.memoryCache?.get(key)?.bitmap
+                imageLoader.memoryCache?.get(key)?.image?.toBitmap()
```
geekygecko, line 25: The toBitmap does take a size. I'm not sure if it's worth it but if the color analyzer only needs a small image it could save resources to load it with smaller dimensions.

modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/images/RoundedCornersTransformation.kt
geekygecko, line 1: I tested the episode from this PR, and it seems to work now. 
https://github.com/Automattic/pocket-casts-android/issues/1719

wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/NowPlayingChip.kt
```diff
@@ -87,9 +91,24 @@ private fun Content(
...
+                        ChipDefaults.Height.toPx().toInt()
+                    }
+                    PocketCastsImageRequestFactory(context).themed()
+                        .create(episode, useEpisodeArtwork)
+                        .newBuilder()
+                        .size(
+                            width = imageWidthPx,
+                            height = imageHeightPx,
+                        )
+                        .build()
```
geekygecko, line 109: The Wear OS now playing chip with the artwork as the background wasn't working. The scale of crop was being ignore and the whole image was squished into the chip.

modules/features/widgets/src/main/kotlin/au/com/shiftyjelly/pocketcasts/widget/ui/ClassicPlayer.kt
MiSikora: It is because `execute()` can fail and yield `null`. So it needs to be looped until success. We could add a small delay between requests if you'd like to give it some breathing room in case of server failures, no network available, etc.

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/ProfileCircleView.kt
```diff
@@ -157,10 +158,10 @@ class ProfileCircleView @JvmOverloads constructor(
     }
 
     private fun loadImageFromUri(uri: Uri, size: Int) {
-        val target = object : Target {
-            override fun onError(error: Drawable?) {
+        val target = object : coil3.target.Target {
```
MiSikora: ⛏️ Missing import.

modules/services/ui/src/main/java/au/com/shiftyjelly/pocketcasts/ui/images/ThemedImageTintTransformation.kt
```diff
@@ -32,7 +32,7 @@ class ThemedImageTintTransformation(context: Context) : Transformation {
         return id.hashCode()
     }
 
-    override suspend fun transform(input: Bitmap, size: Size): Bitmap {
+    override suspend fun transform(input: coil3.Bitmap, size: Size): coil3.Bitmap {
```
MiSikora: ⛏️ This could be imported `as CoilBitmap`.

modules/features/widgets/src/main/kotlin/au/com/shiftyjelly/pocketcasts/widget/ui/ClassicPlayer.kt
geekygecko: I tested this by downloading episodes in my Up Next, going into aeroplane mode and then playing an episode until the next one. The image loader returns the placeholder image instead of null.
I'm not going to add this back in for now and instead create another pull request with a change. It might be better to have a work manager job regularly refreshing the widget instead.

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/ProfileCircleView.kt
geekygecko: Thanks for this suggestion. I have made this change.

modules/services/ui/src/main/java/au/com/shiftyjelly/pocketcasts/ui/images/ThemedImageTintTransformation.kt
geekygecko: Thanks, I have also made this change.

PR #4827, opened 2025-12-05
When following a podcast cache the smaller podcast images

geekygecko:
## Description

This change improves the image caching when you follow a new podcast. Instead of just caching the large image it also caches the medium and small versions. This also combines some of the code to build the static image URL.

## Testing Instructions

1. Follow a new podcast
2. Go into aeroplane mode
3. ✅ Check the new release smart playlist shows the artwork
4. Add a widget
5. Tap play on one of the episode
6. ✅ Check the widget has the artwork

## Checklist
 
- [x] with different themes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 8.3
 -----
+*   New Features
+    *   Allow Wear OS playback through the watch speaker
+        ([#4830](https://github.com/Automattic/pocket-casts-android/pull/4830))
 *   Updates
     *   When following a podcast cache the smaller podcast images
         ([#4827](https://github.com/Automattic/pocket-casts-android/pull/4827))
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -76,7 +76,7 @@ coil-compose = { module = "io.coil-kt.coil3:coil-compose", version.ref = "coil" 
 coil-network-okhttp = { module = "io.coil-kt.coil3:coil-network-okhttp", version.ref = "coil" }
 
 # Compose
-compose-activity = "androidx.activity:activity-compose:1.12.0"
+compose-activity = "androidx.activity:activity-compose:1.12.1"
 compose-animation = { module = "androidx.compose.animation:animation" }
 compose-animation-graphics = { module = "androidx.compose.animation:animation-graphics" }
 compose-bom = { module = "androidx.compose:compose-bom", version.ref = "compose" }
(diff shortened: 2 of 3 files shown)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/SubscribeManager.kt
```diff
@@ -133,14 +135,19 @@ class SubscribeManager @Inject constructor(
...
-        }.onErrorComplete()
+    private fun cacheArtworkRxCompletable(podcast: Podcast) = rxCompletable {
+        try {
+            val urls = PodcastImage.getArtworkUrls(uuid = podcast.uuid, isWearOS = Util.isWearOs(context))
+            urls.forEach { url ->
+                val request = ImageRequest.Builder(context)
+                    .data(url)
+                    .memoryCachePolicy(CachePolicy.DISABLED)
+                    .build()
+                context.imageLoader.execute(request)
```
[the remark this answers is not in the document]
geekygecko: It seemed to work with `execute` when I tested it, but I have taken the advice and changed it to `enqueue`.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/podcast/SubscribeManager.kt
```diff
@@ -133,14 +135,19 @@ class SubscribeManager @Inject constructor(
...
-        }.onErrorComplete()
+    private fun cacheArtworkRxCompletable(podcast: Podcast) = rxCompletable {
+        try {
+            val urls = PodcastImage.getArtworkUrls(uuid = podcast.uuid, isWearOS = Util.isWearOs(context))
+            urls.forEach { url ->
+                val request = ImageRequest.Builder(context)
+                    .data(url)
+                    .memoryCachePolicy(CachePolicy.DISABLED)
+                    .build()
+                context.imageLoader.enqueue(request)
```
MiSikora: ❓ Why is it wrapped in a `Completable` if it enqueues?

geekygecko: I was thinking that the podcast subscribe flow doesn't need to wait for the image caching but maybe it would be best so to reduce the concurrent networking. I have changed it back to `enqueue`.

MiSikora: I don't think it is a problem to do it in the background. But in that case it shouldn't be wrapped in a `Completable`. `Completable` informs that the operation will be finished by the time it emits the completion signal.

geekygecko: Yes that make sense, thanks.

PR #4882, opened 2026-01-06
Ensure onboarding feature flag is loaded from remote source

sztomek:
## Description
This PR aims to cover an edge case we've identified last year.
Feature initialization happens in an async manner (when we're talking about prod builds) inside the ctor of `FireabaseRemoteFeatureProvider`. If a flag is queried while the remote config is being fetched, we simply return the default value. 
That resulted the very awkward scenario with the onboarding feature launch -- people who just installed the app saw the old onboarding despite we had the flags properly configured on Firebase. The default value of `Feature.NEW_ONBOARDING_ACCOUNT_CREATION` was false at this time, we changed it later to true so the issue was seemingly solved.

This is my last attempt to fix the issue ultimately. My initial attemtps were:
- use an `androidx.startup.Initializer` to init firebase provider. that was a dead end beacuse the initializer is not blocking the app launch flow
- change logic on the `MainActivity` to init the feature providers in `onCreate` before doing anything else. that did not work fine either, i had many issues with how we launch the onboarding activity. multiple onboarding activities could be created or none of them were launched at all.

So i finally went with this approach and created a suspend function that will wait until all the providers are inited. Yeah, `FeatureProvider.awaitInitialization` has been also introduced for this purpose.

This isn't ideal either, because you'll need to 'know' that a feature flag is expected to be called early in the app launch flow...
The bulletproof path would be to rework our whole app launch flow and stay of splash until everything gets initialized. 

## Testing Instructions
1. Uninstall the app
2. Apply this patch [onboarding-flag.patch](https://github.com/user-attachments/files/24458249/onboarding-flag.patch) that sets the default value to false and lets us use debug as prod. 
3. Launch the app and verify if you're seeing the new onboarding.

## Checklist

### DIFF
--- a/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
+++ b/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
@@ -177,7 +177,6 @@ import au.com.shiftyjelly.pocketcasts.ui.helper.FragmentHostListener
 import au.com.shiftyjelly.pocketcasts.ui.helper.NavigationBarColor
 import au.com.shiftyjelly.pocketcasts.ui.helper.StatusBarIconColor
 import au.com.shiftyjelly.pocketcasts.ui.theme.Theme
-import au.com.shiftyjelly.pocketcasts.ui.theme.ThemeColor
 import au.com.shiftyjelly.pocketcasts.utils.Network
 import au.com.shiftyjelly.pocketcasts.utils.Util
 import au.com.shiftyjelly.pocketcasts.utils.featureflag.Feature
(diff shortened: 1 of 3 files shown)


geekygecko:
@sztomek I would like to ask Michał to review this once he is back from AFK.

MiSikora (commented):
What do you think about adjusting this slightly? Instead of exposing `FeatureFlag.isEnabledWithRemote()`, we could expose `FeatureFlag.awaitRemoteProviders()`.

The main issue I see with `isEnabledWithRemote()` is that it does not compose well with the rest of the API surface, such as `isEnabledFlow()`, `isEnabledForUser()`, and `isExclusiveToPatron()`. Introducing an explicit `awaitRemoteProviders()` call would allow all feature-flag accessors to share the same initialization mechanism.

With this approach, usage would allow more flexibility.

```kotlin
produceState<Boolean?>(null) {
  FeatureFlag.awaitRemoteProviders()
  FeatureFlag.isEnabled(Feature.Foo) // Or any other getter
}

val enabledFlow = flow {
  FeatureFlag.awaitRemoteProviders()
  emitAll(FeatureFlag.isEnabledFlow(Feature.Foo))
}
```

PR #4905, opened 2026-01-19
Only pause playback after 200ms on AUDIOFOCUS_LOSS_TRANSIENT_CAN_DUCK

sztomek:
## Description
In our current implementation we're pausing playback it the moment when we detect a focus loss. However it might be a good idea to only do it after a 200ms-long grace period.
This PR does that.
Plus adds tests for FocusManager.

Fixes PCDROID-369

## Testing Instructions
I don't know any reliable way to test it in practice, but we have logs in place so maybe we can collect data from users.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 8.5
 -----
-
+*   Updates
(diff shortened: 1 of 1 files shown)


geekygecko:
@MiSikora Thanks for reviewing this PR.

PR #4932, opened 2026-02-02
Distribute app Prototype Build from PRs via FAD

mokagio:
## Description

See https://linear.app/a8c/issue/AINFRA-1741/distribute-pr-prototype-builds-via-fad-in-pocket-casts-android

After internal discussion we established that:

- Prototype Builds are useful, but only in limited cases. Therefore, we'll keep them opt-in in both `main` and PRs
- It's okay to distributed the app build via Firebase App Distribution (FAD) and store the Wear and Automotive builds in S3

## Testing Instructions

Observe optional Prototype Builds in Buildkite and verify that the app build can be installed via FAD. The Wear and Automotive builds should be unaffected.

## Screenshots or Screencast 

<table>
<tr>
<td>

Screenshot_20260204-135825

</td>
<td>

Screenshot_20260204-135922

</td>
</tr>
</table>

## Checklist
- [x] If this is a user-facing change, I have added an entry in CHANGELOG.md — N.A.

 
N.A.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -695,7 +695,14 @@ platform :android do
         build_type: other_builds_type,
         properties: gradle_properties
       )
-      other_apks << lane_context[SharedValues::GRADLE_ALL_APK_OUTPUT_PATHS].find { |p| p.include?("/#{platform}/") }
+
+      generated_apk_paths = lane_context[SharedValues::GRADLE_ALL_APK_OUTPUT_PATHS]
+
+      platform_apk = generated_apk_paths&.find { |p| p.include?("/#{platform}/") }
+
+      UI.user_error!("Could not find generated APK for '#{platform}' platform in the GRADLE_ALL_APK_OUTPUT_PATHS list.") if platform_apk.nil?
+
+      other_apks << platform_apk
     end
 
     s3_comments = []


.buildkite/pipeline.yml
```diff
@@ -88,35 +88,19 @@ steps:
           - "**/build/outputs/apk/**/*"
 
     ##########
-    # Optional Prototype Builds for PR branches, distributed via S3
+    # Optional Prototype Builds for PR and main branches
+    # Builds all modules, uploads to S3, and uploads app to FAD
```
mokagio:
```suggestion
    # Optional Prototype Builds
    #
    # Builds all modules, uploads app to FAD and Wear and Automotive to S3
```

.buildkite/pipeline.yml
```diff
@@ -88,35 +88,19 @@ steps:
...
     ##########
   - group: Prototype Builds
-    if: "build.pull_request.id != null"
+    if: "build.pull_request.id != null || build.branch == 'main'"
     steps:
       - input: "🚀 Generate Prototype Builds?"
-        prompt: Generate Prototype Builds on the PR?
+        prompt: Generate Prototype Builds? (uploads to S3 and FAD)
```
mokagio:
```suggestion
        prompt: Generate Prototype Builds?
```

.buildkite/pipeline.yml
```diff
@@ -88,35 +88,19 @@ steps:
...
         key: prototype_builds_triggered
-      - label: ":github: Prototype Builds"
+      - label: ":rocket: Prototype Builds"
         depends_on: prototype_builds_triggered
         command: |
           ".buildkite/commands/prototype-build.sh"
```
mokagio:
```suggestion
        command: .buildkite/commands/prototype-build.sh
```

.buildkite/pipeline.yml
```diff
@@ -88,35 +88,19 @@ steps:
...
         key: prototype_builds_triggered
-      - label: ":github: Prototype Builds"
+      - label: ":rocket: Prototype Builds"
         depends_on: prototype_builds_triggered
-        command: |
-          ".buildkite/commands/prototype-build.sh"
+        command: .buildkite/commands/prototype-build.sh
         plugins: [ $CI_TOOLKIT ]
         artifact_paths:
           - "**/build/outputs/apk/**/*"
```
mokagio, line 106:
```suggestion
          - "**/build/outputs/apk/**/*"
        # No notify node to avoid GitHub checks status being stuck on yellow.
        # This step being optional, chances are it will never be triggered.
```

fastlane/Fastfile
```diff
@@ -637,18 +637,25 @@ platform :android do
...
+        reuse_identifier: 'app-prototype-build-link',
+        body: '🚧 Prototype builds will be available soon'
+      )
+    end
 
-    prototype_build_type = 'DebugProd'
+    build_config = if pr_number
+                     { task: 'assemble', build_type: 'DebugProd' }
+                   else
+                     { task: 'app:assemble', build_type: 'Prototype' }
```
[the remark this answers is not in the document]
mokagio: This is intentional.

fastlane/Fastfile
```diff
@@ -636,25 +636,69 @@ platform :android do
...
+    UI.user_error!("'FIREBASE_APP_DISTRIBUTION_ACCOUNT_KEY' must be defined as an environment variable.") unless firebase_account_key
+
+    pr_number = pull_request_number
+    if pr_number
+      comment_on_pr(
+        project: GITHUB_REPO,
+        pr_number: pr_number,
+        reuse_identifier: 'app-prototype-build-link',
+        body: '🚧 Prototype builds will be available soon'
+      )
```
mokagio: Image

mokagio:
The prototype build so configured failed with:

> The APK package name 'au.com.shiftyjelly.pocketcasts.debug' does not match your Firebase app's package name 'au.com.shiftyjelly.pocketcasts'. Change the APK package name to 'au.com.shiftyjelly.pocketcasts' and retry the upload.

Will need to revisit the different build types between main and PRs...

build.gradle.kts
```diff
@@ -372,9 +372,7 @@ subprojects {
                     applicationIdSuffix = ".debug"
                 }
 
-                maybeCreate("debugProd").apply {
-                    applicationIdSuffix = ".debug"
-                }
+                maybeCreate("debugProd")
```
mokagio: @geekygecko @MiSikora is this okay?

Builds on `main` use `Prototype` but builds on PRs use `DebugProd`. Firebase is configured with `au.com.shiftyjelly.pocketcasts`, so to avoid [upload rejections](https://buildkite.com/automattic/pocket-casts-android/builds/15248/steps/canvas?sid=019c1cc1-feab-4a21-87f6-d81de57a50c7#019c1cc1-ff19-4b44-a666-1e823c91c611/L2109) I thought we'd use the same package name for both.

If having different application ids for those build type is important, then we could use the same build type for all Firebase builds. In that case, please suggest which. Thanks!

Claude tells me:

```
  ┌────────────┬──────────────┬───────────────────────────────────────────────┐
  │ Build Type │ Minification │                    Purpose                    │
  ├────────────┼──────────────┼───────────────────────────────────────────────┤
  │ DebugProd  │ No           │ Fast builds, production servers               │
  ├────────────┼──────────────┼───────────────────────────────────────────────┤
  │ Prototype  │ Yes          │ Pre-release testing (catches proguard issues) │
  └────────────┴──────────────┴───────────────────────────────────────────────┘
```

MiSikora: This change will break the debug app in a couple of places, including our Firebase integration and the deep links configuration.

I wouldn’t mind if we published the `prototype` variant also from PRs for the Android app, if that helps address the issue. As long as it’s done on demand and builds aren't publicly accessible, which I understand to be the case.

fastlane/Fastfile
```diff
@@ -632,29 +632,72 @@ platform :android do
...
+      task: 'app:assemble',
+      build_type: app_build_type,
+      properties: {
+        'skipSentryProguardMappingUpload' => true
+      }
     )
 
-    prototype_build_type = 'DebugProd'
+    app_apk = lane_context[SharedValues::GRADLE_ALL_APK_OUTPUT_PATHS].first
+    UI.user_error!("No app APK found in Gradle output.") unless app_apk
```
dangermattic: <table data-meta="generated_by_pr-check">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/Automattic/pocket-casts-android/blob/090b35a63debdbd9862aedb8ce009226e20fa979/fastlane/Fastfile#L655"></span>Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols.</td>
    </tr>
  </tbody>
</table>

fastlane/Fastfile
```diff
@@ -664,52 +709,20 @@ platform :android do
 
       install_url = "#{PROTOTYPE_BUILD_DOMAIN}/#{upload_path}"
 
-      prototype_build_details_comment(
+      s3_comments << prototype_build_details_comment(
         app_display_name: get_app_display_name(apk_path: apk_path),
+        app_icon: ':amazon-s3:',
```
mokagio, line 721: Otherwise, it would default to Firebase which is not accurate:

Image

vs

Image

fastlane/Fastfile
```diff
@@ -632,29 +632,74 @@ platform :android do
     )
   end
 
-  # Builds a prototype build and uploads it to S3
+  # Builds prototype builds and uploads to FAD (app) and S3 (wear/automotive)
```
mokagio:
```suggestion
  # Compiles prototype builds and uploads to FAD (app) and S3 (wear/automotive)
```

build.gradle.kts
mokagio: Thanks @MiSikora . I addressed this by building the `Prototype` type from both PRs and main branch, https://github.com/Automattic/pocket-casts-android/pull/4932/changes#diff-dff4b99d4e651ab9d7597876ec8dbe92aa934e42c0fb55f4aadd6c106fc96688R642-R649

fastlane/Fastfile
```diff
@@ -632,29 +632,74 @@ platform :android do
...
+    other_platforms = %w[wear automotive]
+
+    other_apks = []
+    other_platforms.each do |platform|
+      gradle(
+        task: "#{platform}:assemble",
+        build_type: other_builds_type,
+        properties: gradle_properties
+      )
+      other_apks << lane_context[SharedValues::GRADLE_ALL_APK_OUTPUT_PATHS].find { |p| p.include?("/#{platform}/") }
```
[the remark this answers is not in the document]
mokagio: Addressed in 3f592d2

fastlane/Fastfile
```diff
@@ -632,29 +632,81 @@ platform :android do
...
+      )
+
+      generated_apk_paths = lane_context[SharedValues::GRADLE_ALL_APK_OUTPUT_PATHS]
+
+      platform_apk = generated_apk_paths&.find { |p| p.include?("/#{platform}/") }
+
+      UI.user_error!("Could not find generated APK for '#{platform}' platform in the GRADLE_ALL_APK_OUTPUT_PATHS list.") if platform_apk.nil?
+
+      other_apks << platform_apk
+    end
```
AliSoftware, line 706: I was about to suggest to use `other_apks = other_platforms.map do …` instead of using `each` then doing `other_apks << platform_apk` inside the loop.

But then I realized that this would increase the indentation level a bit much, so maybe using `.each` and `<<` are OK after all 😛

mokagio: Ah, that's true. However, I decided to use `each` over `map` because `map` should not perform side effects. Doing `x = [...].map { ... }` is neater than the `each` alternative, but it didn't sit well with me. Call me pedantic, ah-ah.

AliSoftware: Oh very true and I 100% agree with this. I just didn't realize there were side effects done by the loop at first but now I see the obvious `gradle` call! (that I missed at first when being too focused on this `each` vs `map`) 😅

So all good to keep `each` indeed

PR #4968, opened 2026-02-09
[Q1 Quality] WearOS - Implement connectivity warnings

sztomek:
## Description
This PR adds a fullscreen notification when we detects connection loss. The notification is debounced, so frequent changes of connectivity won't be too disturbing. Users can dismiss the notification by tapping it or by waiting 4 seconds when it will be gone automatically. 
I also added an inline warning display to the player screen when streaming an episode that's not downloaded.

Fixes PCDROID-409

## Testing Instructions
1. Install the wear app on your wear device
2. Log in
3. Bring device to offline state
4. Verify fullscreen notification is displayed
5. Either wait for 4 seconds or tap the screen to dismiss the noti
6. Get back online
7. Start playing an episode that's not downloaded
8. Bring device to offline state
9. Verify that the player screen displays a warning.
10. Get back online
11. Verify that the player shows no warning

## Screenshots or Screencast 
Screenshot_20260209_165757_debug
Screenshot_20260209_165832_debug

## Checklist

### DIFF
--- a/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/WearMainActivityViewModel.kt
+++ b/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/WearMainActivityViewModel.kt
@@ -29,7 +29,7 @@ import kotlinx.coroutines.TimeoutCancellationException
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.asStateFlow
-import kotlinx.coroutines.flow.sample
+import kotlinx.coroutines.flow.debounce
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.reactive.asFlow
@@ -79,7 +79,7 @@ class WearMainActivityViewModel @Inject constructor(
         viewModelScope.launch {
             var previousConnectivityState: Boolean? = null
             connectivityStateManager.isConnected
-                .sample(CONNECTIVITY_DEBOUNCE_MS)
+                .debounce(CONNECTIVITY_DEBOUNCE_MS)
                 .collect { isConnected ->
                     val shouldShowNotification = previousConnectivityState != null &&
                         previousConnectivityState != isConnected &&
--- a/wear/src/test/kotlin/au/com/shiftyjelly/pocketcasts/wear/networking/ConnectivityStateManagerTest.kt
+++ b/wear/src/test/kotlin/au/com/shiftyjelly/pocketcasts/wear/networking/ConnectivityStateManagerTest.kt
@@ -76,6 +76,7 @@ class ConnectivityStateManagerTest {
             networks = listOf(buildNetworkStatus(NetworkType.Wifi)),
         )
         val connectivityStateManager = createConnectivityStateManager()
+        testScheduler.runCurrent()
 
         connectivityStateManager.isConnected.test {
             assertEquals(true, awaitItem())
@@ -90,6 +91,7 @@ class ConnectivityStateManagerTest {
             networks = listOf(buildNetworkStatus(NetworkType.Cell)),
         )
         val connectivityStateManager = createConnectivityStateManager()
+        testScheduler.runCurrent()
 
         connectivityStateManager.isConnected.test {
             assertEquals(true, awaitItem())
@@ -107,6 +109,7 @@ class ConnectivityStateManagerTest {
             ),
         )
         val connectivityStateManager = createConnectivityStateManager()
+        testScheduler.runCurrent()
 
         connectivityStateManager.isConnected.test {
             assertEquals(true, awaitItem())
@@ -124,6 +127,7 @@ class ConnectivityStateManagerTest {
             ),
         )
         val connectivityStateManager = createConnectivityStateManager()
+        testScheduler.runCurrent()
 
         connectivityStateManager.isConnected.test {
             assertEquals(true, awaitItem())
@@ -154,6 +158,7 @@ class ConnectivityStateManagerTest {
             networks = listOf(buildNetworkStatus(NetworkType.Wifi)),
         )
         val connectivityStateManager = createConnectivityStateManager()
+        testScheduler.runCurrent()
 
         connectivityStateManager.isConnected.test {
             assertEquals(true, awaitItem())
@@ -174,6 +179,7 @@ class ConnectivityStateManagerTest {
             networks = listOf(buildNetworkStatus(NetworkType.Wifi)),
         )
         val connectivityStateManager = createConnectivityStateManager()
+        testScheduler.runCurrent()
 
         connectivityStateManager.isConnected.test {
             assertEquals(true, awaitItem())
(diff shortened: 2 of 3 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/networking/ConnectivityStateManager.kt
```diff
@@ -0,0 +1,31 @@
...
+
+@OptIn(ExperimentalHorologistApi::class)
+@Singleton
+class ConnectivityStateManager @Inject constructor(
+    networkRepository: NetworkRepository,
+    @ApplicationScope coroutineScope: CoroutineScope,
+) {
+    val isConnected: StateFlow<Boolean> = networkRepository.networkStatus
+        .map { networks ->
+            val networkTypes = networks.networks.map { it.networkInfo.type }.toSet()
```
MiSikora: ℹ️ We can avoid creating intermediate collection with this.

```kotlin
val networkTypes = networks.networks.mapTo(mutableSetOf()) { it.networkInfo.type }
```

wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/component/ConnectivityNotificationOverlay.kt
```diff
@@ -0,0 +1,110 @@
...
+) {
+    ScreenScaffold(
+        timeText = null,
+        modifier = modifier,
+    ) {
+        val currentOnDismiss by rememberUpdatedState(onDismiss)
+        LaunchedEffect(Unit) {
+            delay(4.seconds)
+            currentOnDismiss()
+        }
```
MiSikora: ❓ Would it make sense to use here `DisposableEffect` with an event observer that matches the start lifecycle? Otherwise someone might not even notice the warning if the app is minimized or something like that.

wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/WearMainActivityViewModel.kt
```diff
@@ -68,6 +75,24 @@ class WearMainActivityViewModel @Inject constructor(
                     _state.update { it.copy(signInState = signInState) }
                 }
         }
+
+        viewModelScope.launch {
+            var previousConnectivityState: Boolean? = null
+            connectivityStateManager.isConnected
+                .debounce(CONNECTIVITY_DEBOUNCE_MS)
```
MiSikora: ❓ Shouldn't it be `sample` instead of `debounce`? If network changes its state often we might never get an event.

wear/src/test/kotlin/au/com/shiftyjelly/pocketcasts/wear/networking/ConnectivityStateManagerTest.kt
```diff
@@ -0,0 +1,192 @@
...
+    @Test
+    fun `isConnected is false when no networks available`() = runTest {
+        networkStatusFlow.value = Networks(
+            activeNetwork = null,
+            networks = emptyList(),
+        )
+        advanceUntilIdle()
+
+        assertEquals(false, connectivityStateManager.isConnected.value)
+    }
```
MiSikora: 🚫 This test setup is incorrect.

1. `MainCoroutineRule()` is created as an `@Rule`, but it isn’t actually used for anything.
2. `testScope` is created using the dispatcher from the rule, but there’s no need for that.
3. The scope passed into `ConnectivityStateManager` is not used in the tests. Instead, a different scope is used. This makes things like `advanceUntilIdle` not operating on the underlying coroutine scope and is rather a chance that it works.

What you most likely want is something like this:

```kotlin
private val scope = TestScope()

@Before
fun setup() {
  connectivityStateManager = ConnectivityStateManager(networkRepository, testScope.backgroundScope)
}

@Test
fun `foo bar`() = testScope.runTest {
  // Test setup
  
  yield() // Instead of advanceUntilIdle()

  // Assertion
}
```

However, since `ConnectivityStateManager` exposes a `Flow`, it would be more appropriate to test it using Turbine:

```kotlin

@Test
fun `foo bar`() = testScope.runTest {
  connectivityManager.isConnected.test {
    skipItems(1) // Initial value
    
    // Change in networkStateFlow

    assertEquals(someValue, awaitItem())

    cancel()
  }
```

wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/component/ConnectivityNotificationOverlay.kt
sztomek: good idea, thanks!

wear/src/test/kotlin/au/com/shiftyjelly/pocketcasts/wear/networking/ConnectivityStateManagerTest.kt
sztomek: thanks, valid points. addressed this as well!

wear/src/test/kotlin/au/com/shiftyjelly/pocketcasts/wear/networking/ConnectivityStateManagerTest.kt
```diff
@@ -0,0 +1,210 @@
...
+        networkStatusFlow.value = Networks(
+            activeNetwork = null,
+            networks = emptyList(),
+        )
+        val connectivityStateManager = createConnectivityStateManager()
+
+        yield()
+        connectivityStateManager.isConnected.test {
+            assertEquals(false, awaitItem())
+        }
```
MiSikora: ℹ️ When testing with Turbine `yield()` isn't necessary because `awaitItem()` suspends until a new value arrives. If an old value is present it is better to explicitly omit it using `skipItems(1)` for clearer intentions. So a typical test with Turbine of a `StateFlow` looks like this:

```kotlin
fun test() = runTest {
  flow.test {
    skipItems(1)

    TODO("Some code to trigger the flow value propagation")
    assertEquals(expected, awaitItem())

    cancel()
  }
}
```

wear/src/test/kotlin/au/com/shiftyjelly/pocketcasts/wear/networking/ConnectivityStateManagerTest.kt
```diff
@@ -0,0 +1,210 @@
...
+    fun `isConnected is false when no networks available`() = testScope.runTest {
+        networkStatusFlow.value = Networks(
+            activeNetwork = null,
+            networks = emptyList(),
+        )
+        val connectivityStateManager = createConnectivityStateManager()
+
+        yield()
+        connectivityStateManager.isConnected.test {
+            assertEquals(false, awaitItem())
```
MiSikora, line 53: ℹ️ You need to `cancel()` these at the end because they are `StateFlow`s. Otherwise Turbine won't close and the test will hang forever or time out with an error.

PR #4981, opened 2026-02-12
Improve download episode worker

MiSikora:
## Description

This PR introduces several improvements to the episode download worker:

- Updates the worker progress data from `true` to `false` once the work is complete. This allows correct differentiation between queued and in-progress states in cases of failures or cancellation due to constraint changes.
- Removes the cancellation detection logic. Based on testing, I found it unnecessary. `WorkManager` does not propagate failures when a task is cancelled, and the only way for a user to cancel a download is by cancelling the work itself (not the underlying HTTP call).
- Includes the run attempt count in the work info output. This helps identify and cancel long-queued downloads.
- Includes the work info ID in the work info output. This also helps with cancelling long-queued downloads.
- Adds a retry mechanism in case of download failures.

Relates to PCDROID-429

## Testing Instructions

Code review my changes.

## Checklist
 
- [x] with different themes

### DIFF
--- a/modules/services/repositories/src/test/java/au/com/shiftyjelly/pocketcasts/repositories/download/DownloadWorkInfoTest.kt
+++ b/modules/services/repositories/src/test/java/au/com/shiftyjelly/pocketcasts/repositories/download/DownloadWorkInfoTest.kt
@@ -275,30 +275,6 @@ class DownloadWorkInfoTest {
     }
 
     @Test
-    fun `map work info's run attempt count`() {
-        val workInfo = WorkInfo(
-            id = UUID.randomUUID(),
-            state = WorkInfo.State.ENQUEUED,
-            runAttemptCount = 10,
-            tags = setOf(DownloadEpisodeWorker.episodeWorkerName("episode-id")),
-        )
-
-        val downloadInfo = DownloadEpisodeWorker.mapToDownloadWorkInfo(workInfo)
-
-        assertEquals(
-            DownloadWorkInfo.Pending(
-                id = workInfo.id,
-                episodeUuid = "episode-id",
-                runAttemptCount = 10,
-                isWifiRequired = false,
-                isPowerRequired = false,
-                isStorageRequired = false,
-            ),
-            downloadInfo,
-        )
-    }
-
-    @Test
     fun `map work info with power constraint`() {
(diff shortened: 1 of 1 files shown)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/download/DownloadEpisodeWorker.kt
```diff
@@ -419,11 +440,11 @@ private const val WAIT_FOR_WIFI_KEY = "wait_for_wifi"
...
-internal const val IS_WORK_STARTED_KEY = "is_work_started"
+internal const val IS_WORK_EXECUTING = "is_work_executing"
 
 // Output keys
 internal const val DOWNLOAD_FILE_PATH_KEY = "download_file_path"
 internal const val ERROR_MESSAGE_KEY = "error_message"
 
 private val OUT_OF_STORAGE_MESSAGES = setOf("no space", "not enough space", "disk full", "quota")
-internal const val CANCELLED_MESSAGE = "___EPISODE_DOWNLOAD_CANCELLED___"
+private const val MAX_ATTEMPT_COUNT = 3
```
geekygecko, line 450: It looks like the max retries used to be 5 is there a reason it's now 3? https://github.com/Automattic/pocket-casts-android/blob/b76b1f59812ad1ba23f600e9b0d38c24f30d33be/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/download/task/DownloadEpisodeTask.kt#L83

MiSikora: I think it's too much before we surface issues to the user and is potentially one of the sources of frustrations. I'd like to change it and see how it goes from there.

Retry 1 → 30 seconds
Retry 2 → 60 seconds
Retry 3 → 120 seconds
Retry 4 → 240 seconds
Retry 5 → 480 seconds

MiSikora: I'm going to merge but let me know if you think we should change it back to 5.

PR #4985, opened 2026-02-13
Pin faraday >= 1.10.5 for security fix

mokagio:
## Summary

- Pins `faraday` to `~> 1.10, >= 1.10.5` to address a security vulnerability.
- See [faraday#1665](https://github.com/lostisland/faraday/pull/1665) for the fix.
- Tracked in [AINFRA-1976](https://linear.app/a8c/issue/AINFRA-1976/track-security-alert-upgrade-fastlane-to-use-faraday-2x).

## Test plan

- [ ] CI passes — no functional change, only a transitive dependency version bump.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

*Posted by Claude (Opus 4.6) on behalf of @mokagio with approval.*

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,3 +20,7 @@ gem 'google-apis-sheets_v4', '~> 0.26'
 #
 # See https://github.com/ruby/openssl/issues/949
 gem 'openssl', '~> 4.0'
+
+# Security: https://github.com/lostisland/faraday/pull/1665
+# Faraday 2.0 is not compatible with Fastlane
+gem 'faraday', '~> 1.10', '>= 1.10.5'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -93,7 +93,7 @@ GEM
     drb (2.2.3)
     emoji_regex (3.2.3)
     excon (0.112.0)
-    faraday (1.10.4)
+    faraday (1.10.5)
       faraday-em_http (~> 1.0)
       faraday-em_synchrony (~> 1.0)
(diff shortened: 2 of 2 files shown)


AliSoftware:
Interesting that CI didn't pass on this PR when Claude opened it (due to Danger complaining about missing labels/milestone) yet your Claude coordinating agent didn't pick up that failure to loop back on it and fix it itself 🤔 
(I'm the one who just added the labels and milestone to the PR to make Danger happy)

@mokagio  I'm thus curious if your instructions in the coordination agent prompt suggest the agent to wait for the CI to finish passing then check the result and loop ("agentic loop")? or was the instructions suggesting that this part would be done by a human and not each subagent? Or the "check for CI" part to validate was not even mentioned in the instructions you gave to your agent at all (hence it not checking)?

mokagio:
Thanks for the review @AliSoftware . I've done so many experiments recently that I'm not 100% sure at what stage I was when this run. But based on the timestamp, I'd say I didn't have agentic loop in mind and was instead operating under the assumption that it should be up to each repo to specify how to open PRs (even if just by reading `Dangerfile` which turned out to be suboptimal). Recently, I've been instructing agents to check the CI status and address Danger issues. In fact, I was considering creating a dedicated agent just for that, imagining the worker agent that implemented the code and opened the PR would hand over to the one tuned for checking GItHub checks / CI status.

---

Update: To not forget: https://linear.app/a8c/issue/AINFRA-2009/pr-followup-agent

PR #5026, opened 2026-02-23
[Predictive Back] Enhance in-app back animations

sztomek:
## Description
This PR adds the same animations that the system uses for predictive back. I did a thorough research on this topic and it turned out that we can't rely the system's predictive back animation when navigating back from our screens, because:
  - Android's system animations work on window/activity transitions (the system compositor handles cross-activity animations)
  - Fragment transitions happen within the same window - Android can't peek at "the previous activity" because there isn't one
  - The app must manually make Fragment B visible and animate both fragments in coordination.
  - BottomNavigator maintains 5 separate fragment stacks
  - MainActivity handles 7 different back scenarios with different animations:
  -- Fragment navigation → Scale + fade with peek
  -- Player bottom sheet → Scale to 80% then collapse
  -- Bottom sheets (UpNext) → System collapse animation (can use default)
  -- Modals → Custom slide animations
  -- Selection mode → No animation, just exit mode
So the only viable way forward was to create custom transitions and hook them up on the appropriate places.
I noticed a nasty bug that took a lot of time to get rid of. It was specific to the Profile>Bookmarks screen. I had to
split it into two (BookmarksDialogFragment for Player, BookmarksNavigationFragment for Profile).
Why? The original BookmarksContainerFragment was trying to be both a bottom sheet dialog (modal context) and a navigation fragment (backstack context). This caused navigation confusion - the back gesture didn't know which behavior to use. Separating them made each context explicit and predictable.

Fixes PCDROID-462

## Testing Instructions
Smoke test the app, pay attention to the back navigation.

## Screenshots or Screencast 

https://github.com/user-attachments/assets/66839629-4424-4392-bb26-027895123010

## Checklist

### DIFF
--- a/modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/helper/PredictiveBackAnimator.kt
+++ b/modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/helper/PredictiveBackAnimator.kt
@@ -1,6 +1,7 @@
 package au.com.shiftyjelly.pocketcasts.views.helper
 
 import android.view.View
+import kotlin.math.sqrt
 
 /**
  * Utility for applying predictive back gesture animations to views.
@@ -34,11 +35,11 @@ object PredictiveBackAnimator {
         /** Default alpha reduction for current view (30% = 0.7 alpha) */
         const val ALPHA_AMOUNT = 0.3f
 
-        /** Default scale increase for previous view (5% = starts at 0.95 scale) */
-        const val SCALE_AMOUNT_REVERSE = 0.05f
+        /** Default scale increase for previous view (2% = starts at 0.98 scale) */
+        const val SCALE_AMOUNT_REVERSE = 0.02f
 
-        /** Default alpha increase for previous view (50% = starts at 0.5 alpha) */
-        const val ALPHA_AMOUNT_REVERSE = 0.5f
+        /** Default alpha increase for previous view (15% = starts at 0.85 alpha) */
+        const val ALPHA_AMOUNT_REVERSE = 0.15f
 
         /** Default target scale for animation end (85%) */
         const val TARGET_SCALE = 0.85f
@@ -103,11 +104,13 @@ object PredictiveBackAnimator {
             view.setLayerType(View.LAYER_TYPE_HARDWARE, null)
         }
 
+        val easedProgress = sqrt(progress)
+
         val baseScale = 1f - scaleAmount
         val baseAlpha = 1f - alphaAmount
 
-        val scale = baseScale + (scaleAmount * progress)
-        val alpha = baseAlpha + (alphaAmount * progress)
+        val scale = baseScale + (scaleAmount * easedProgress)
+        val alpha = baseAlpha + (alphaAmount * easedProgress)
 
         view.scaleX = scale
         view.scaleY = scale
--- a/modules/services/views/src/test/java/au/com/shiftyjelly/pocketcasts/views/helper/PredictiveBackAnimatorTest.kt
+++ b/modules/services/views/src/test/java/au/com/shiftyjelly/pocketcasts/views/helper/PredictiveBackAnimatorTest.kt
@@ -64,9 +64,21 @@ class PredictiveBackAnimatorTest {
     fun `applyProgressReverse at 0 progress starts at reduced scale and alpha`() {
         PredictiveBackAnimator.applyProgressReverse(mockView, progress = 0f)
 
-        verify(mockView).scaleX = 0.95f
-        verify(mockView).scaleY = 0.95f
-        verify(mockView).alpha = 0.5f
+        verify(mockView).scaleX = 0.98f
+        verify(mockView).scaleY = 0.98f
+        verify(mockView).alpha = 0.85f
+    }
+
+    @Test
+    fun `applyProgressReverse at 0_25 progress applies decelerate easing`() {
+        // sqrt(0.25) = 0.5, so easedProgress = 0.5
+        // scale = 0.98 + (0.02 * 0.5) = 0.99
+        // alpha = 0.85 + (0.15 * 0.5) = 0.925
+        PredictiveBackAnimator.applyProgressReverse(mockView, progress = 0.25f)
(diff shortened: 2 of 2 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

CookieyedCodes:
If this is implemented I wouldn't be against a setting for fast, medium, long animations & no animations, some people like an instant response

sztomek:
thanks @MiSikora, i've ticketed the bug you discoered here https://linear.app/a8c/issue/PCDROID-466/bug-broken-back-nav as it doesn't seem to be an easy/obvious one. I already sinked some time into it

sztomek:
@MiSikora 
let me respond to your observations

>1.  The predictive animation for the player sheet appears a bit janky. Notice how it jumps and loses the fade transition especially at the top of the screen. It may be better to remove it in this context.

I've disabled the custom transition on the player bottom sheet, so now you should see the default one.

> 2. The initial scaling of the previous fragment on the back stack feels visually abrupt.

Polished the animation params so now it seems to be smoother to my eyes.

> 3.  Predictive back doesn't work for me during the onboarding flow. I don't think it is necessary to add it there, especially given how convoluted the code is. I'm highlighting it just in case.

Yeah, it doesn't work for me either, already spent some time on it without success. I'd lean towards just letting it go 👋 


Please check the app again and review the code if you feel like these efforts are justified.

modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/helper/PredictiveBackTransition.kt
```diff
@@ -0,0 +1,132 @@
...
+        }
+    }
+
+    /**
+     * Completes the transition animation.
+     */
+    @MainThread
+    fun finish() {
+        try {
+            seekController?.animateToEnd()
```
MiSikora, line 62: ❓ Should the `seekControler` be cleared here?

modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/helper/PredictiveBackTransition.kt
```diff
@@ -0,0 +1,132 @@
...
+        }
+    }
+
+    /**
+     * Cancels the transition and resets to the initial state.
+     */
+    @MainThread
+    fun cancel() {
+        try {
+            seekController?.currentFraction = 0f
```
MiSikora, line 74: ❓ Should the `seekControler` be cleared here?

modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/helper/PredictiveBackTransition.kt
```diff
@@ -0,0 +1,132 @@
...
+    private val container: ViewGroup,
+    private val duration: Long = 300,
+) {
+    private var seekController: androidx.transition.TransitionSeekController? = null
+
+    /**
+     * Starts a seekable transition with scale and fade effects.
+     */
+    @MainThread
+    fun start(backEvent: BackEventCompat) {
```
MiSikora, line 29: ℹ️ We can remove `backEvent` as it isn't used in the implementation.

modules/services/views/src/test/java/au/com/shiftyjelly/pocketcasts/views/helper/PredictiveBackTransitionTest.kt
```diff
@@ -0,0 +1,184 @@
...
+import org.junit.Test
+import org.mockito.Mockito.mock
+
+/**
+ * Tests for [PredictiveBackTransition].
+ *
+ * Tests the seekable transition-based animations for predictive back gestures.
+ * Note: These are basic smoke tests since TransitionSeekController requires a real Android environment.
+ */
+class PredictiveBackTransitionTest {
```
MiSikora, line 16: ⚠️ I’m having trouble understanding the value these tests are providing.

`PredictiveBackTransition` is a thin wrapper around `TransitionSeekController`. The tests verify that calling the wrapper methods does not throw. That’s a very low-signal assertion and doesn’t give us meaningful correctness of behavior.

In general, tests are most valuable when they validate observable behavior or interactions. Otherwise, they add maintenance cost without increasing confidence in the implementation.

I’d lean toward removing them. If there’s particular coverage they’re meant to provide, I’m happy to revisit.

modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/helper/ActivityBackHelper.kt
```diff
@@ -0,0 +1,172 @@
+package au.com.shiftyjelly.pocketcasts.views.helper
+
+import android.view.View
+import androidx.activity.BackEventCompat
+import androidx.activity.ComponentActivity
+import androidx.activity.OnBackPressedCallback
+import androidx.navigation.NavController
+
+object ActivityBackHelper {
```
MiSikora, line 9: ❓ I’m not sure where this feedback best fits, so I’ll leave it here.

`ActivityBackHelper`, `ComposeNavigationBackHelper`, and `PredictiveBackExtensions` all appear to provide a similar category of functionality, but their API shapes are quite different.

- `ActivityBackHelper` is both an object with functions and a set of extension functions that delegate to that object. And that object has extension functions inside as well.
- `ComposeNavigationBackHelper` is an object exposing functions.
- `PredictiveBackExtensions` is a collection of extension functions.

Why they don’t follow a consistent design approach? Is there a specific constraint or usage pattern that drove these differences? If not, it may be worth aligning them to a single, consistent API style to reduce cognitive overhead and improve discoverability.

sztomek:
We agreed not to include these animations in the app. internal discussion: p1773860264187239/1772474323.176949-slack-C05RR9P9RAT

PR #5076, opened 2026-03-04
[Player] Use adaptive buffer size

sztomek:
## Description
This PR adds a new feature flag `adaptive_buffer` that controls whether we should use increased buffer size when streaming an episode, while respecting user's choice of data conservation (if non-metered networks only setting is ON).
On wear, we preserve the existing buffer size since that platform is quite constrained by resources.

Fixes PCDROID-478

## Testing Instructions
just review the app and smoke test playback on all platforms

## Screenshots or Screencast 
SCR-20260304-kpzm

## Checklist


sztomek:
The original idea was about to check bitrate of the stream and increase the buffer amount on higher bit density. This change alone won't make much difference, so I'm closing this PR.

PR #5208, opened 2026-04-10
Enable metered network data warning by default

SaiSrujan-SoftSuave:
## Description
Enables **"Warn before using data"** by default to prevent unintended mobile data usage.

Previously, this setting was disabled by default, which led to users accidentally consuming large amounts of mobile data. In some cases, users assumed that **"Only on unmetered WiFi"** under Auto Download applied globally, which caused confusion.

### Changes
- Enable **"Warn before using data"** by default
- Add a **deep link to the setting** from the warning bottom sheet, allowing users to easily disable it if desired
- Improve overall clarity and prevent accidental mobile data usage

Slack Reference: p1749029285430589-slack-C028JAG44VD  
iOS issue: Automattic/pocket-casts-ios#3227  

---

## Testing Instructions
1. Install or reset the app (fresh state)
2. Navigate to **Storage & Data Use** settings  
3. Verify that:
   - ✅ "Warn before using data" is **enabled by default**
4. Turn off WiFi (use mobile data)
5. Open any podcast episode  
6. Tap **Download**
7. Verify:
   - Warning bottom sheet is shown  
   - Bottom sheet includes option/link to open the relevant setting  
8. Tap on the settings link from the warning  
9. Verify:
   - User is deep-linked to **"Warn before using data"** setting  
10. Disable the setting and retry download
11. Verify:
   - No warning is shown when disabled  

---

### After Fix
<p align="center">
  <video src="https://github.com/user-attachments/assets/7813c43e-f92d-48b7-af4d-1fe3a7738586" width="300" controls></video>
</p>

---

## Checklist
- [ ] Ensure the linter passes (`./gradlew spotlessApply`)  
- [ ] Any Jetpack Compose components I added or changed are covered by previews  
- [ ] I have updated (or requested update for) analytics tracking if needed  

#### UI Testing
- [ ] Tested with different themes  
- [ ] Tested in landscape orientation  
- [ ] Tested with large display and font size  
- [ ] Tested accessibility with TalkBack

### DIFF
--- a/modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/viewmodel/StorageSettingsViewModel.kt
+++ b/modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/viewmodel/StorageSettingsViewModel.kt
@@ -467,7 +467,7 @@ class StorageSettingsViewModel @Inject constructor(
         )
 
         data class StorageDataWarningState(
-            val isChecked: Boolean = false,
+            val isChecked: Boolean = true,
             val onCheckedChange: (Boolean) -> Unit,
         )
     }
--- a/modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/SettingsImpl.kt
+++ b/modules/services/preferences/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/SettingsImpl.kt
@@ -436,7 +436,7 @@ class SettingsImpl @Inject constructor(
 
     override val warnOnMeteredNetwork = UserSetting.BoolPref(
         sharedPrefKey = Settings.PREFERENCE_WARN_WHEN_NOT_ON_WIFI,
-        defaultValue = false,
+        defaultValue = true,
         sharedPrefs = sharedPreferences,
     )
 


MiSikora (changes requested):
Thank you for the contribution, but we cannot simply change this setting to `true`. If existing installations are updated without having explicitly set this value beforehand, the behavior will change for users. The change must be backward compatible. Specifically, the default should be set to `true` only for new installations. This can be implemented, for example, here:

https://github.com/Automattic/pocket-casts-android/blob/640060f3152d2ef6f7d92b13d9bec4f2c9504005/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/jobs/VersionMigrationsWorker.kt#L210-L214

geekygecko:
We might have to hold off merging this pull request while I check with the product team, as this change affects both auto downloads and pressing play on an episode. We may need to find a better approach as from my understanding it means anyone on mobile data will now be prompted to confirm streaming every time they tap play.

Internal discussion: p1776039525061859/1749029285.430589-slack-C028JAG44VD

MiSikora:
Thank you @SaiSrujan-SoftSuave again for the PR. But after consulting this change again with the product team we decided to not have this setting enabled by default.

PR #5227, opened 2026-04-15
Make sure archived episode status is preserved across sync

sztomek:
## Description
A user reported that downloaded episodes in their Up Next queue were becoming unplayable — the app showed them as downloaded but the actual files were missing from disk, resulting in FileNotFoundException errors. Log analysis revealed two bugs working together: first, when a user re-downloads a previously auto-archived episode, the `setReadyForDownload` SQL query un-archives it locally but doesn't set `archived_modified`, so the server never learns about the un-archival and re-archives the episode on the next sync, deleting the file again. Second, `EpisodeSync.processIncrementalResponse()` uses Room's `@Update` (which writes ALL columns) after firing off an async download cancellation — creating a race where stale in-memory `episode_status=4` overwrites the `resetDownloadStatus()` that already set it to 0, leaving the DB saying "Downloaded" while the file is gone.
 We fixed the sync loop by adding conditional `CASE WHEN archived = 1` guards to `setReadyForDownload` that set `archived_modified, last_archive_interaction_date, and exclude_from_episode_limit` only when the episode is actually archived — mirroring the existing unarchiveBlocking() pattern and ensuring the un-archival syncs to the server without creating spurious sync traffic for fresh downloads. We eliminated the race condition by replacing `updateAll()` with a new `updateAllSyncFields()` method that only writes the 14 sync-relevant columns, never touching local-only fields like `episode_status or downloaded_file_path`. As defense-in-depth, we added ENOENT detection in `PlaybackManager.onPlayerError` that uses `downloadQueue.cancel().join()` to atomically reset the stale download state, then retries via streaming — with `forcePlayerSwitch` guaranteeing `loadCurrentEpisode` re-reads fresh state from the DB before creating the player.

Fixes PCDROID-538 https://linear.app/a8c/issue/PCDROID-538/downloaded-files-gets-lost

## Testing Instructions
  1. Subscribe to any podcast, let an episode auto-download
  2. Play it to completion — confirm "Download deleted" appears in logs (auto-archive fires)
  3. Go back to the episode and re-download it manually
  4. Open the app fresh to trigger a sync (or pull-to-refresh on the podcast list)
  5. Before fix: the sync re-archives and deletes the file; playback fails with ENOENT
  6. After fix: the sync should NOT re-archive because archived_modified was set, and the episode remains playable

## Checklist

### DIFF
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/EpisodeDao.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/EpisodeDao.kt
@@ -13,6 +13,7 @@ import au.com.shiftyjelly.pocketcasts.models.db.AppDatabase
 import au.com.shiftyjelly.pocketcasts.models.db.helper.QueryHelper
 import au.com.shiftyjelly.pocketcasts.models.db.helper.UuidCount
 import au.com.shiftyjelly.pocketcasts.models.entity.BaseEpisode
+import au.com.shiftyjelly.pocketcasts.models.entity.ChapterIndices
 import au.com.shiftyjelly.pocketcasts.models.entity.EpisodeDownloadFailureStatistics
 import au.com.shiftyjelly.pocketcasts.models.entity.Podcast
 import au.com.shiftyjelly.pocketcasts.models.entity.PodcastEpisode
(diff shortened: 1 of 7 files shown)


modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/EpisodeDao.kt
```diff
@@ -580,6 +642,9 @@ abstract class EpisodeDao {
         UPDATE podcast_episodes
         SET
           archived = 0,
+          archived_modified = CASE WHEN archived = 1 THEN :archivedModified ELSE archived_modified END,
+          last_archive_interaction_date = CASE WHEN archived = 1 THEN :archivedModified ELSE last_archive_interaction_date END,
+          exclude_from_episode_limit = CASE WHEN archived = 1 THEN 1 ELSE exclude_from_episode_limit END,
```
MiSikora, line 647: ❓ Why do we modify the `exclude_from_episode_limit` column?

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/EpisodeDao.kt
```diff
@@ -580,6 +642,9 @@ abstract class EpisodeDao {
         UPDATE podcast_episodes
         SET
           archived = 0,
+          archived_modified = CASE WHEN archived = 1 THEN :archivedModified ELSE archived_modified END,
```
MiSikora, line 645: ℹ️ We should cover this change with tests in `EpisodeDaoDownloadStatusTest`.

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/EpisodeDao.kt
sztomek: `setReadyForDownload` sets `archived = 0`, effectively unarchiving the episode. `unarchiveBlocking()` (line 484) already sets `exclude_from_episode_limit = 1` for the same reason — it prevents `checkPodcastForEpisodeLimitBlocking()` from immediately re-archiving the episode the user just acted on (the limit check filters out episodes with `excludeFromEpisodeLimit = true`).

The `CASE WHEN archived = 1` guard ensures the flag is only touched when the episode was actually archived, leaving it as-is for fresh (non-archived) downloads.

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/EpisodeDao.kt
sztomek: yeah, done in 562fccc70b06141426467bc809e62cb20a368dd0

MiSikora (approved):

PR #5229, opened 2026-04-16
Update outdated GitHub Actions

iangmaia:
Fixes AINFRA-2297

## Summary
- Replace deprecated `gradle/gradle-build-action@v2` with `gradle/actions/dependency-submission@v4`

## Test plan
- [ ] Verify dependency submission workflow runs successfully on next push to main

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/.github/workflows/submit-gradle-dependencies.yml
+++ b/.github/workflows/submit-gradle-dependencies.yml
@@ -16,8 +16,4 @@ jobs:
           distribution: 'temurin'
           java-version: '17'
       - name: Setup Gradle to generate and submit dependency graphs
-        uses: gradle/gradle-build-action@v2
-        with:
-          dependency-graph: generate-and-submit
-      - name: Generate the dependency graph which will be submitted post-job
-        run: ./gradlew :app:dependencies :automotive:dependencies :wear:dependencies
+        uses: gradle/actions/dependency-submission@v4


.github/workflows/submit-gradle-dependencies.yml
```diff
@@ -16,8 +16,4 @@ jobs:
           distribution: 'temurin'
           java-version: '17'
       - name: Setup Gradle to generate and submit dependency graphs
-        uses: gradle/gradle-build-action@v2
-        with:
-          dependency-graph: generate-and-submit
-      - name: Generate the dependency graph which will be submitted post-job
```
iangmaia, line 22: Note:

Old pattern (`gradle-build-action@v2`):  
1. Step 1: Configures Gradle to record dependency info (dependency-graph: generate-and-submit)
2. Step 2: You manually run a Gradle command (`./gradlew :app:dependencies :automotive:dependencies :wear:dependencies`) to actually trigger the recording

New pattern (`gradle/actions/dependency-submission@v4`):
- Does everything in one step, running its own internal Gradle build that automatically discovers and resolves dependencies for all projects in the build. No need to enumerate :app, :automotive, :wear etc.

This is also consistent with how the repos that had already migrated use it (e.g. woocommerce-android which only adds filtering params (dependency-graph-include-projects) to limit the scope, not expand it.

If a repo needs to restrict which projects get submitted, we can pass `dependency-graph-include-projects` (like `woocommerce-android` does), but the default of "all projects" is what I believe is correct here.

twstokes (approved):

PR #5258, opened 2026-04-28
Fix missing episode duration when RSS feed lacks length

### LINKED ISSUE #5228: 8.10-rc-2 (9423) episodes with no length set in the rss are shown without any playtime, even after download
### Description

These episodes have always historically shown the episode length once downloaded, but this has recently stopped (in 8.9).

Subscribe to https://archive.org/download/rss_vfbaxxdzgi/vfbaxxdzgi

Download episodes

Image

Pixel 8, android 16, 8.10-rc2

joashrajin:
## Description

Fixes #5228

The downloads rewrite in #5017 removed the post-download `MediaExtractor`
pass that populated `episode.duration` when the RSS `<enclosure length>`
attribute was missing or empty. As a result, episodes from feeds that
omit the length (e.g. `<enclosure url="..." length="" type="audio/mpeg"/>`)
show no playtime even after they finish downloading, and they don't
contribute to the Up Next "time left" total. Last working version: 8.6.

### Changes
- Re-add a `fixMissingDuration` helper in `DownloadEpisodeWorker` that
  reads `MediaFormat.KEY_DURATION` from the downloaded file via
  `MediaExtractor` and persists it through
  `EpisodeManager.updateDurationBlocking`.
- Call it on every successful download. `updateDurationBlocking`
  already short-circuits when the extracted value is invalid, matching
  the pre-rewrite behavior.
- Wrap extraction in try/finally and a top-level catch so a malformed
  file never propagates an exception out of the worker.

## Testing Instructions

1. Subscribe to a feed whose `<enclosure>` element has an empty
   `length` attribute. The reporter's example:
   `https://archive.org/download/rss_vfbaxxdzgi/vfbaxxdzgi`
2. Download an episode.
3. Verify that once the download finishes, the episode shows a real
   playtime (e.g. `12m`, `1h 23m`) on the podcast list, episode card,
   and Up Next, instead of a blank duration.
4. Add the episode to Up Next and verify the bottom-bar "time left"
   total includes its duration.
5. As a regression check: download a normal episode (RSS includes a
   valid `length` and/or `<itunes:duration>`) and verify the playtime
   displayed matches what was shown before download (no flicker /
   regression to wrong value).
6. Force a download to fail (airplane mode mid-download) and confirm
   the failure path still works — duration extraction should not run.

## Screenshots or Screencast

_To be added — repro on emulator with the linked archive.org feed._

## Checklist

- [x] with different themes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 8.12
 -----
+*   Bug Fixes
+    *   Restore episode duration after download for feeds without an RSS length attribute
+        ([#5258](https://github.com/Automattic/pocket-casts-android/pull/5258))
 
 
 8.11
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/di/RepositoryModule.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/di/RepositoryModule.kt
@@ -23,6 +23,8 @@ import au.com.shiftyjelly.pocketcasts.repositories.chromecast.CastManagerImpl
 import au.com.shiftyjelly.pocketcasts.repositories.download.DownloadManager
 import au.com.shiftyjelly.pocketcasts.repositories.download.DownloadQueue
 import au.com.shiftyjelly.pocketcasts.repositories.download.DownloadStatusObserver
+import au.com.shiftyjelly.pocketcasts.repositories.download.MediaDurationExtractor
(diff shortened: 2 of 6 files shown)


joashrajin:
Working on addressing the Copilot feedback, but it seems to work so far: 
image


Will run more tests later before needing a review, @MiSikora

joashrajin:
@MiSikora it is ready to be reviewed :)

MiSikora (approved):
Looks great! Thanks for the fix.

Issue #1002, opened 2023-05-26
Add option to disable seek slider on lock screen


sew3521:
### What problem does this address?

Accidentally seeking to a new position in the podcast if you miss a button on the screen or if the phone unlocks in your pocket.

### What is your proposed solution?

I would really appreciate an option to either hide the seek bar on the lock screen while keeping the playback control buttons or an option to disable seeking using the seek bar on the lock screen.

### Did you search for existing feature suggestions?

- [X] I have searched for existing feature suggestions.

maral:
I need this so much! I'm one of the odd conservative fellows who run with their phone in hand and the screen sometimes becomes active and I skip to the end of the podcast. It skipped two whole podcasts in two seconds before I noticed 🙈 This is really needed.

MiSikora:
Yes, contributions are welcome.

OwaisKhatri:
Hi Team,

I just checked, this one is still open and isn't done.
Anyone can please confirm that, which solution needs to be done?

1. Hide seek bar from notification when the device is locked.
2. Disable seek bar in the notification.

CookieyedCodes:
@OwaisKhatri I am gonna guess both are linked actually but it probably makes more sense to disable the lock screen one, are you gonna work on this?

OwaisKhatri:
Yeah sure, I'll disable the seek bar in the notification.
Will share the PR by tomorrow.

CookieyedCodes:
Gud gud, id say the setting toggle make sense under notifications 🤔🤔

OwaisKhatri:
@MiSikora I have raised the [PR](https://github.com/Automattic/pocket-casts-android/pull/5209) for this issue, please review it.
One thing, that needs your attention for event tracker. I wasn't able to add the event for the new option added in the settings screen.

sew3521:
I just wanted to thank y'all for trying to make this happen. It will really help with a common issue I have with Pocket Casts. I was surprised when I saw it being worked since I figured it was a feature that did not have enough demand.

PR #5261, opened 2026-04-29
Add option to disable seek slider on lock screen

OwaisKhatri:
## Description
This PR is raised to resolve the issue while accidentally seeking to a new position in the podcast.

Fixes # [1002](https://github.com/Automattic/pocket-casts-android/issues/1002)

## Testing Instructions
1. Go to Settings Screen.
2. In settings, go to general section.
3. In general, go to Player section.
4. Enable the option (Enable lock screen scrubbing).
5. Play any episode. Seek slider will not change its position.

## Screenshots or Screencast 
Screenshot_20260429-184641

## Checklist

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/MediaSessionManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/MediaSessionManager.kt
@@ -23,7 +23,6 @@ import androidx.media.utils.MediaConstants.PLAYBACK_STATE_EXTRAS_KEY_MEDIA_ID
 import androidx.media3.common.util.UnstableApi
 import androidx.media3.session.CommandButton
 import androidx.media3.session.MediaLibraryService
-import androidx.media3.session.MediaSession
 import androidx.media3.session.SessionCommand
 import androidx.media3.session.SessionError
 import au.com.shiftyjelly.pocketcasts.analytics.SourceView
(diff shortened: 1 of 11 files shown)


OwaisKhatri:
@MiSikora Create a new PR for the issue [1002](https://github.com/Automattic/pocket-casts-android/issues/1002), due to contributor conflicts from my end.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/MediaSessionManager.kt
```diff
@@ -542,6 +556,11 @@ class MediaSessionManager(
                         updateMedia3CustomLayout()
                         media3Service?.triggerNotificationUpdate()
                     }
+                    Util.isAndroidAutoConnectedFlow(context).collect { autoConnected ->
```
MiSikora, line 559: Calling `collect()` here leak memory. Collecting is never cancelled unless the scope is cancelled. This means that if I play 3 episodes I will have 3 active subscriptions to `isAndroidAutoConnectedFlow()`.

MiSikora (commented):
@OwaisKhatri There is an issue in the current implementation that I've highlighted in a comment. I'll be unavailable until Monday, so if you fix it before then, I won't be able to review it immediately.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/MediaSessionManager.kt
OwaisKhatri: Will it be ok to use `Util.isAndroidAutoConnectedFlow(context).first()`?
The terminal operator that returns the first element emitted by the flow and then cancels flow's collection. Throws NoSuchElementException if the flow was empty.

Will need to handle the exception, if the flow is empty and we can wrap it in try/catch. Does this makes sense?

PR #5296, opened 2026-05-11
[RSM] Chat with Episode

obsantos:
## Description

Adds Episode Chat behind the `episode_chat` feature flag.

This introduces a new chat feature module, an Episode Chat entry point on episode details for episodes with transcripts, and a Plus/Patron gated chat experience with a paywall for free users. The chat supports persisted conversation history, assistant replies from the episode chat endpoint, retry/error handling, network state handling, clearing chat history, and quote responses.

Also adds optional playable quote support behind `episode_chat_playable_quotes`, including quote segment playback and restoring the previous playback state after the quote finishes or is stopped.

Includes Room schema updates for persisted episode chats/messages and server/repository wiring for the `/mobile/episode/chat` flow.

## Testing Instructions

1. Build and install a debug or prototype build.
2. Enable the `episode_chat` feature flag.
3. Open an episode that has a transcript.
4. Verify the Episode Chat banner appears below the transcript card.
5. With a free account, tap the banner and verify the Episode Chat paywall opens.
6. With a Plus/Patron account, tap the banner and verify the chat screen opens.
7. Send a message and verify the assistant response appears.
8. Close and reopen chat, then verify the conversation history is still present.
9. Turn off network connectivity and verify the chat handles the offline/error state.
10. Re-enable network connectivity and verify retry works.
11. Use the clear chat action and verify the conversation resets.
12. Enable `episode_chat_playable_quotes`, receive a response with a quote, and verify tapping the quote plays the segment and restores previous playback.

## Screenshots or Screencast 

| Paywall | Chat |
| -- | -- |
| Screenshot_20260511-182637 |  |

## Checklist
 
- [ ] with different themes


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/featureflag/Feature.kt
```diff
@@ -264,6 +264,22 @@ enum class Feature(
         hasFirebaseRemoteFlag = false,
         hasDevToggle = true,
     ),
+    EPISODE_CHAT(
+        key = "episode_chat",
+        title = "Episode Chat",
+        defaultValue = isDebugOrPrototypeBuild,
+        tier = FeatureTier.Plus(),
+        hasFirebaseRemoteFlag = false,
```
zwarm: @obsantos - did you intend for this to not be turned on/off remotely?

zwarm (commented):
.

zwarm (commented):
## Code Review

Verified against the diff. Here are the issues I found (excluding the `hasFirebaseRemoteFlag` flag item which has already been posted).

---

### High Priority

**1. `api(projects.modules.features.chat)` in podcasts module should be `implementation`**

`modules/features/podcasts/build.gradle.kts` — `chat` types are not part of `podcasts`'s public API, so this widens the compile classpath unnecessarily and will trigger recompilation of downstream modules when the `chat` module changes. The dependency analyser flagged this too.

**2. Quote messages are included in the conversation history sent to the backend**

In `ChatManagerImpl.sendMessage()`, the history is built via `textOrNull()` which returns text for `ChatMessage.Quote`, and `ChatRole.Quote.apiRole = "assistant"`. So quote cards are sent back to the model as assistant turns. The test in `ChatManagerImplTest` explicitly asserts this:

```kotlin
ConversationMessage(role = "assistant", content = "Earlier quote")
```

This may be intentional — worth confirming with the backend before merge, since re-sending quotes as context could produce duplicated or confused AI responses.

**3. `OnboardingUpgradeSource.UNKNOWN` in `ChatPaywallFragment`**

`modules/features/chat/src/main/kotlin/au/com/shiftyjelly/pocketcasts/chat/ChatPaywallFragment.kt`

```kotlin
OnboardingFlow.Upsell(OnboardingUpgradeSource.UNKNOWN)
```

Using `UNKNOWN` means analytics won't attribute upgrades coming from the chat paywall. A dedicated `EPISODE_CHAT` source (or similar) should be added.

---

### Medium Priority

**4. Wrong string used in `ChatPaywallPage` body text**

`modules/features/chat/src/main/kotlin/au/com/shiftyjelly/pocketcasts/chat/ui/ChatPaywall.kt`

```kotlin
TextH50(
    text = stringResource(LR.string.transcript_generated_paywall_description),
    ...
)
```

This is the transcript paywall description string, not a chat-specific one. The chat paywall should have its own description string.

**5. `awaitEndOfQuote()` terminates early if the user manually changes tracks**

`modules/features/chat/src/main/kotlin/au/com/shiftyjelly/pocketcasts/chat/ChatViewModel.kt`

```kotlin
state.episodeUuid != episodeUuid -> true
```

If the user manually switches to another track mid-quote, this condition fires and the quote ends — which then triggers `restorePreviousPlayback()`, interrupting the newly selected track and restarting the old episode. Consider only restoring playback if the episode change was caused by the quote itself, not by the user.

**6. `ChatSuggestionChips` exists but is not used**

`modules/features/chat/src/main/kotlin/au/com/shiftyjelly/pocketcasts/chat/ui/ChatSuggestionChips.kt`

Dead code in a new module is confusing. Either wire it up or remove it and add it back when suggestions are implemented.

**7. `checkNotNull` on missing transcript surfaces as a generic server error**

`modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/chat/ChatManagerImpl.kt`

```kotlin
val transcript = checkNotNull(selectTranscript(episodeUuid)) {
    "Transcript URL is required to send episode chat messages"
}
```

If a transcript disappears between the user opening chat and sending a message, this throws `IllegalStateException`, which is caught as `ChatError.ServerError` in the ViewModel and shown as a generic error. A more specific error state would be clearer for the user.

---

### Low Priority / Nits

**8. `parseTimestampMs` has no unit tests**

`modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/chat/Timestamp.kt`

This function handles three time formats (`HH:MM:SS`, `HH:MM:SS,mmm`, `HH:MM:SS.mmm`) and returns null for malformed input. It's a pure function — easy to test and valuable to cover given the variety of timestamp formats the backend might return.

**9. `coroutines.reactive` and `coroutines.rx2` flagged as unused by dependency analyser**

`modules/features/chat/build.gradle.kts`

The `projectHealth` task flagged these as unused dependencies. Worth verifying whether they're actually needed; if not, remove them.

**10. `episode-chat-pr-review.md` committed to repo root**

This review notes file is part of the diff and should not ship. Please delete it before merge.

modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/featureflag/Feature.kt
```diff
@@ -264,6 +264,22 @@ enum class Feature(
         hasFirebaseRemoteFlag = false,
         hasDevToggle = true,
     ),
+    EPISODE_CHAT(
+        key = "episode_chat",
+        title = "Episode Chat",
+        defaultValue = isDebugOrPrototypeBuild,
+        tier = FeatureTier.Plus(),
```
zwarm, line 279: `EPISODE_CHAT` has `hasFirebaseRemoteFlag = false`, which means it cannot be remotely toggled in production. `EPISODE_CHAT_PLAYABLE_QUOTES` correctly has `hasFirebaseRemoteFlag = true`. If chat is ever intended to be enabled in production, this should be `true` as well — otherwise there is no way to roll it out or kill-switch it via Firebase.

obsantos: solved

modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/featureflag/Feature.kt
obsantos: yes, added them to Firebase

obsantos:
>Code Review

1. Solved ✅ 
2. Included on purpose. Want the chat to have full context in case the user asks a follow up regarding the quote.
3. Will leave Unknown for now. Will follow up on future PR when the event is added to Event Horizon
4. The message is generic enough that it works for this case too
5. Will fix in a follow up PR as it might change logic quite a bit
6. Solved/Removed ✅ 
7. Uncommon edge case. Generic error is fine for now
8. Solved/Added unit tests ✅ 
9. Solved ✅ 
10. Solved/Removed ✅ 

cc @zwarm

zwarm (approved):
Thanks for the changes @obsantos . It looks good now. 
👍

modules/features/chat/src/main/kotlin/au/com/shiftyjelly/pocketcasts/chat/ui/ChatInputBar.kt
```diff
@@ -0,0 +1,130 @@
...
+) {
+    val keyboardController = LocalSoftwareKeyboardController.current
+
+    Column(
+        horizontalAlignment = Alignment.CenterHorizontally,
+        modifier = modifier
+            .background(theme.background)
+            .navigationBarsPadding(),
+    ) {
+        Divider(color = theme.divider, thickness = 0.5.dp)
```
geekygecko: Should we get rid of this or give it some padding underneath as it doesn't look quite right touching the input box. 

Image

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/episode/EpisodeFragment.kt
```diff
@@ -579,31 +584,77 @@ class EpisodeFragment : BaseFragment() {
...
+                                                } else {
+                                                    if (parentFragmentManager.findFragmentByTag("episode_chat_paywall") == null) {
+                                                        val fragment = ChatPaywallFragment.newInstance(episodeUuid, podcastUuid)
+                                                        fragment.show(parentFragmentManager, "episode_chat_paywall")
+                                                    }
+                                                }
+                                            },
+                                        ),
+                                )
+                            }
```
geekygecko, line 676: When we added the ‘View Transcripts’ button, we received negative feedback from users that it pushed the episode description further down the screen. Would it be worth rethinking how users enter the chat so we can avoid a similar issue happening again?

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/EpisodeChatDao.kt
```diff
@@ -0,0 +1,35 @@
...
+    abstract suspend fun insertMessage(message: EpisodeChatMessage)
+
+    @Query("SELECT * FROM episode_chat_messages WHERE episode_uuid = :episodeUuid ORDER BY created_at ASC")
+    abstract fun observeMessages(episodeUuid: String): Flow<List<EpisodeChatMessage>>
+
+    @Query("SELECT * FROM episode_chat_messages WHERE episode_uuid = :episodeUuid ORDER BY created_at ASC")
+    abstract suspend fun getMessages(episodeUuid: String): List<EpisodeChatMessage>
+
+    @Query("DELETE FROM episode_chat_messages WHERE episode_uuid = :episodeUuid")
+    abstract suspend fun deleteMessagesByEpisode(episodeUuid: String)
```
geekygecko, line 34: Are the chats cleared up when an episode is removed or automatically after a certain period of time?

modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/podcast/EpisodeChatModels.kt
```diff
@@ -0,0 +1,30 @@
+package au.com.shiftyjelly.pocketcasts.servers.podcast
+
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+
+@JsonClass(generateAdapter = true)
+data class EpisodeChatRequest(
+    @Json(name = "transcript_url") val transcriptUrl: String,
```
geekygecko, line 8: Could any url be passed here? Would it be better for the server to build this using the podcast and episode uuid?

geekygecko (approved):
Thanks for making this change!

modules/features/chat/src/main/kotlin/au/com/shiftyjelly/pocketcasts/chat/ChatFragment.kt
```diff
@@ -0,0 +1,124 @@
...
+
+    private fun showOptionsDialog() {
+        OptionsDialog()
+            .addTextOption(
+                titleId = LR.string.chat_clear,
+                imageId = IR.drawable.ic_delete,
+                click = { viewModel.clearChat() },
+            )
+            .show(parentFragmentManager, "chat_options")
+    }
```
geekygecko, line 113: If you open the toolbar options menu when the keyboard is open, this happens.

modules/features/chat/src/main/kotlin/au/com/shiftyjelly/pocketcasts/chat/ui/ChatInputBar.kt
obsantos: good catch! this is supposed to be shown when in offline as a warning, but without the message the divider gets too close to the input field. Fixed!

modules/features/chat/src/main/kotlin/au/com/shiftyjelly/pocketcasts/chat/ChatFragment.kt
obsantos: Fixed!

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/EpisodeChatDao.kt
obsantos: Before only on user action. But I just changed it here ([commit](https://github.com/Automattic/pocket-casts-android/pull/5296/changes/f82b27d372d14c3fdbfdc7b910ff1a226ec0d3dd)) to use a cascade relation to the episode uuid, so if the episode is deleted, the chat and messages too

modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/podcast/EpisodeChatModels.kt
obsantos: In theory yea, any url can be passed. This was raised in the [backend PR](https://github.com/Automattic/pocket-casts-podcast-api/pull/331#discussion_r3206294172) too and we put some measures in place.

Ideally we would pass the podcast/episode uuid only. We were actually doing that before but we reverted it because of some issues accessing the S3 bucket.

One benefit from accessing the url directly is that it's a Cloudflare's CDN url so its cached and we save bucket hits (less egress = $$), but i understand the concerns 😄 

We can move this convo to Slack and discuss options. As long as we are able to find the transcript for both generated and public episodes on server side, I also rather send the ids instead of urls.

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/episode/EpisodeFragment.kt
obsantos: Yep, agree! looks like @sztomek [RSM project](https://github.com/Automattic/pocket-casts-android/pull/5276) is also adding a new option here, so we would end up with 3, which is even less ideal.

We can get a designer in the loop for suggestions 👍

obsantos: For now made the buttons a bit more compact/smaller. I thought we could put them horizontally but worried about translations in certain locales might turn them into 2 lines and make it ugly. Will open a discussion 🧵  on slack in the meanwhile

Screenshot_20260519-131524

PR #5301, opened 2026-05-13
Fix transcript banner overlapping episode description

lromero16:
## Description

Fixes #PCDROID-326

The \"View Transcript\" banner was overlapping the episode description (show notes WebView) in the episode detail screen.

**Root cause**: `AnimatedNonNullVisibility` uses `expandVertically` by default, which animates the Compose layout height from 0 to full size. When embedded in a `ConstraintLayout`, the View system doesn't receive `requestLayout()` on every animation frame — so during the expand, `webViewShowNotes` stays positioned as if `episodeTranscript` has height 0, causing the overlap.

**Fix**: Override `enter`/`exit` transitions to `fadeIn()`/`fadeOut()` at the `EpisodeFragment` call site. This avoids height animation entirely; Compose reports the full height immediately, and `ConstraintLayout` positions the show notes correctly from the first frame.

## Testing Instructions
1. Open an episode that has a transcript available
2. Observe the episode detail screen — the \"View Transcript\" banner should appear above the episode description without overlapping it
3. Dismiss and reopen the episode to verify consistent behaviour

## Checklist

- [ ] with different themes

### DIFF
--- a/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/episode/EpisodeFragment.kt
+++ b/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/episode/EpisodeFragment.kt
@@ -16,6 +16,8 @@ import android.webkit.WebResourceRequest
 import android.webkit.WebView
 import android.webkit.WebViewClient
 import android.widget.Toast
+import androidx.compose.animation.fadeIn
+import androidx.compose.animation.fadeOut
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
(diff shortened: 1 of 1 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

geekygecko (commented):
Thanks for investigating a fix for this issue.

I wasn’t able to properly test this on my side as I can’t currently reproduce the error.

One thing I did notice is that the text no longer slides down the screen and instead jumps into place. This doesn’t look quite as smooth as before, which seems to be a downside of removing `expandVertically()`. Are there any alternative approaches we could try that would still avoid the issue while keeping the smoother animation?

Image

[Screen_recording_20260514_134719.webm](https://github.com/user-attachments/assets/170a4ed4-f5b1-4b59-861f-1bda929643f3)


<br/>

Can you please follow the checklist in the PR description template and sign the CLA.

PR #5320, opened 2026-05-21
Fix Up Next scroll-to-top crash on long queues

joashrajin:
## Description

The following crash is something I could replicate:

https://github.com/user-attachments/assets/f5e9cb71-9efb-4fd3-b68e-e43a91b0fe4d

The lower "Up Next" button calls `RecyclerView.quickScrollToTop()`, which installs a `LinearSmoothScroller` whose `calculateSpeedPerPixel` computed:

```kotlin
val pixelsInRange = scrollRange * scrollOffset / scrollRange.toFloat()
```

The leading `scrollRange * scrollOffset` is `Int * Int`, evaluated before the divide-by-float. On long Up Next queues both values can be tens of thousands and their product overflows `Int.MAX_VALUE`, producing a negative `pixelsInRange`. That cascades into a negative speed-per-pixel, a negative duration handed to `Action.update(…, duration, interpolator)`, and a fatal:

> `IllegalStateException: If you provide an interpolator, you must set a positive duration`

The reporter's debug log confirms this path — the crash trace ends in a `Choreographer` frame callback dispatching a `LinearSmoothScroller$Action.validate()`, and the user had 332 episodes in Up Next at the time.

This PR:
- Extracts the speed-per-pixel math into an `internal` top-level function so it can be unit-tested.
- Replaces `scrollRange * scrollOffset / scrollRange.toFloat()` (mathematically equivalent to `scrollOffset` ignoring overflow) with `scrollOffset.toFloat().coerceAtLeast(1f)`. The floor at `1f` guarantees a positive speed (and therefore a positive duration), even if some future input falls outside the early-return guards.
- Adds a unit test that exercises a `scrollOffset` × `scrollRange` combination that would have overflowed `Int.MAX_VALUE` under the old implementation.

Fixes PCDROID-417 (https://linear.app/a8c/issue/PCDROID-417/android-up-next-lower-button-causes-fatal-crash)

## Testing Instructions
1. Build and install the app.
2. Add ~20+ episodes to the Up Next queue (enough that the queue scrolls past a screenful).
3. Open the player, then open Up Next via the lower "Up Next" button (the one with the text label, not the badge button).
4. Scroll down through the queue for 5+ seconds so a large `scrollOffset` accumulates.
5. Tap the lower "Up Next" button again.
6. - [ ] Verify the queue smoothly scrolls back to the top with no crash.
7. - [ ] Repeat a few times to confirm the crash never reappears.
8. - [ ] Tap the lower "Up Next" button while already at the top — verify nothing breaks (regression check).

The new `RecyclerViewTest` also exercises the math directly:

```
./gradlew :modules:services:views:testDebugUnitTest --tests "au.com.shiftyjelly.pocketcasts.views.extensions.RecyclerViewTest"
```

## Screenshots or Screencast
N/A — bug fix, no UI changes.

## Checklist
- [x] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml` (n/a)
- [x] Any jetpack compose components I added or changed are covered by compose previews (n/a)
- [x] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics. (n/a)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 -----
 *   Bug Fixes
     *   Fix crash when scrolling to the top of a long Up Next queue
-        ([#5315](https://github.com/Automattic/pocket-casts-android/pull/5315))
+        ([#5320](https://github.com/Automattic/pocket-casts-android/pull/5320))
 
 8.12
 -----
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 8.13
 -----
-
+*   Bug Fixes
+    *   Fix crash when scrolling to the top of a long Up Next queue
+        ([#5315](https://github.com/Automattic/pocket-casts-android/pull/5315))
 
 8.12
 -----
--- a/modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/extensions/RecyclerView.kt
+++ b/modules/services/views/src/main/java/au/com/shiftyjelly/pocketcasts/views/extensions/RecyclerView.kt
@@ -38,21 +38,23 @@ fun RecyclerView.quickScrollToTop() {
         override fun getVerticalSnapPreference() = LinearSmoothScroller.SNAP_TO_START
 
         override fun calculateSpeedPerPixel(displayMetrics: DisplayMetrics): Float {
-            val scrollOffset = computeVerticalScrollOffset()
-            val scrollRange = computeVerticalScrollRange()
-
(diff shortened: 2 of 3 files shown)


joashrajin:
Moving this off my fork to get Buildkite CI to run — superseded by #5327.

PR #5328, opened 2026-05-22
Fix AAOS steering-wheel skip buttons after Android 14 update

joashrajin:
## Description

Fixes https://linear.app/a8c/issue/PCDROID-560/aaos-steering-wheel-skip-forwardback-buttons-unresponsive-in-pocket

After the Media3 migration, the Automotive app did not advertise directional next/previous player commands. GM's car-media service therefore treated skip as unavailable and fell back to `KEYCODE_MEDIA_STOP` (`keyCode=86` in the affected Chevrolet Equinox EV log). The media-button handler also did not recognise `KEYCODE_MEDIA_FAST_FORWARD` / `KEYCODE_MEDIA_REWIND`, while AAOS `KEYCODE_MEDIA_NEXT` / `KEYCODE_MEDIA_PREVIOUS` events went through the 250 ms headphone multi-tap queue.

This change advertises directional commands only from the Automotive app, handles the directional keycode variants directly, and keeps mobile, Bluetooth-headphone, and Wear OS behavior unchanged.

### Changes

- **Automotive-only directional commands** — grant `COMMAND_SEEK_TO_NEXT` / `COMMAND_SEEK_TO_PREVIOUS` from `Media3SessionCallback`, `PocketCastsForwardingPlayer`, and the transport-only unknown-controller path when `Util.isAutomotive(...)` is true. This prevents the ambiguous STOP fallback without adding duplicate controls on Wear OS.
- **`Media3SessionCallback.onMediaButtonEvent`** — handle `KEYCODE_MEDIA_FAST_FORWARD` and `KEYCODE_MEDIA_REWIND` as direct skip-forward / skip-back, alongside the existing dedicated skip keycodes.
- **Automotive NEXT/PREVIOUS bypass** — route steering-wheel NEXT/PREVIOUS directly to skip, deliberately bypassing headphone action settings and preserving the current play/pause state.
- **Held-button protection** — consume repeated `ACTION_DOWN` events without applying another fixed-length skip.
- **Shared helpers** — use `launchSkipForward()` / `launchSkipBackward()` for the direct skip paths.

### Why scope the commands to Automotive?

An earlier revision advertised next/previous commands unconditionally. That caused Wear OS to render a duplicate previous-track / next-track pair alongside the existing skip-forward / skip-back buttons ([screenshot](https://github.com/Automattic/pocket-casts-android/pull/5328#discussion_r3286574602)). The final implementation keeps the baseline transport command set for phone and Wear, and extends it only for the Automotive app.

`KEYCODE_MEDIA_STOP` is not mapped directly to a skip direction because STOP contains no forward/back information. Advertising directional commands on AAOS gives the vehicle service an unambiguous route instead.

### Automated tests

Run:

```bash
./gradlew :modules:services:repositories:testDebugUnitTest \
  --tests "au.com.shiftyjelly.pocketcasts.repositories.playback.Media3SessionCallbackTest" \
  --tests "au.com.shiftyjelly.pocketcasts.repositories.playback.Media3LibrarySessionCallbackTest" \
  --tests "au.com.shiftyjelly.pocketcasts.repositories.playback.PocketCastsForwardingPlayerTest"
```

Verify:

- Automotive session, forwarding-player, and unknown-controller paths expose next/previous commands.
- Phone/Wear paths continue to exclude those commands.
- Automotive empty state continues to exclude transport commands.
- FAST_FORWARD / REWIND and Automotive NEXT / PREVIOUS invoke the configured fixed-length skip.
- Automotive NEXT / PREVIOUS do not read headphone action settings or resume paused playback.
- Repeated key-down events are consumed without triggering an additional skip.

### Manual — AAOS emulator

1. Build the Automotive variant: `./gradlew :automotive:assembleDebug`.
2. Install and launch Pocket Casts on the AAOS emulator or an affected Android 14 AAOS vehicle.
3. Start playback of any episode.
4. Send `adb shell input keyevent 87` for NEXT, `88` for PREVIOUS, `90` for FAST_FORWARD, and `89` for REWIND.
5. Verify each directional input skips immediately, with no 250 ms delay and no `Media3: stop` log.
6. Verify a held key does not apply repeated fixed-length jumps.
7. Verify on-screen playback controls still work normally.

### Manual — mobile regression check

1. Run the mobile `debug` variant.
2. Connect Pixel Buds or another Bluetooth headset with multi-tap controls.
3. Verify double-tap / triple-tap still resolves through `headphoneControlsNextAction` / `headphoneControlsPreviousAction` after the multi-tap window.

### Manual — Wear OS regression check

1. Run the Wear OS variant and connect to a paired watch.
2. Verify the original skip-forward / skip-back pair remains, with no duplicate previous-track / next-track controls.

## Checklist

- [ ] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml` *(no string changes)*
- [ ] Any jetpack compose components I added or changed are covered by compose previews *(no Compose changes)*
- [ ] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics. *(no analytics changes)*

> [!NOTE]
> 112 focused tests across `Media3SessionCallbackTest`, `Media3LibrarySessionCallbackTest`, and `PocketCastsForwardingPlayerTest` pass locally with JDK 21 on current `main` (2026-08-07). `spotlessCheck` also passes.


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/Media3SessionCallback.kt
```diff
@@ -209,37 +211,78 @@ internal class Media3SessionCallback(
...
+                    scope.launch {
+                        try {
+                            playbackManager.skipBackwardSuspend(
+                                sourceView = source,
+                                jumpAmountSeconds = settings.skipBackInSecs.value,
+                            )
+                        } catch (e: Exception) {
+                            Timber.e(e, "Skip backward failed")
+                        }
+                    }
```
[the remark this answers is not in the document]
joashrajin: Good call — addressed in [`3db789c`](https://github.com/Automattic/pocket-casts-android/pull/5328/commits/3db789cb0). Extracted `launchSkipForward()` / `launchSkipBackward()` private helpers and collapsed the four duplicate call sites (SKIP_FORWARD/FAST_FORWARD, SKIP_BACKWARD/REWIND, and the automotive NEXT/PREVIOUS pair) into one-liners. No behaviour change — net –10 lines.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PocketCastsForwardingPlayer.kt
```diff
@@ -164,6 +164,11 @@ class PocketCastsForwardingPlayer(
                 Player.COMMAND_PLAY_PAUSE,
                 Player.COMMAND_SET_MEDIA_ITEM,
                 Player.COMMAND_SEEK_IN_CURRENT_MEDIA_ITEM,
+                // Advertise skip-to-next/previous so external controllers (notably AAOS
+                // steering-wheel buttons) route through seekToNext()/seekToPrevious()
+                // instead of falling back to STOP. See PCDROID-560.
+                Player.COMMAND_SEEK_TO_NEXT,
+                Player.COMMAND_SEEK_TO_PREVIOUS,
```
geekygecko: Enabling these actions on add devices can have issues like now the Wear OS has previous track and skip backwards.

Image

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/Media3SessionCallback.kt
```diff
@@ -209,37 +211,42 @@ internal class Media3SessionCallback(
...
+                    launchSkipForward()
+                    return true
+                }
+
+                KeyEvent.KEYCODE_MEDIA_PREVIOUS -> {
+                    launchSkipBackward()
+                    return true
+                }
+            }
+        }
```
geekygecko, line 262: This is a great idea.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PocketCastsForwardingPlayer.kt
joashrajin: Good catch, thanks for the screenshot. Fixed in [`2708c28`](https://github.com/Automattic/pocket-casts-android/pull/5328/commits/2708c28a7) — dropped the `COMMAND_SEEK_TO_NEXT` / `COMMAND_SEEK_TO_PREVIOUS` advertisement entirely (both in `PocketCastsForwardingPlayer.getAvailableCommands()` and `TRANSPORT_PLAYER_COMMANDS`), and restored the `swapPlayer excludes seek to next and previous` test to its original assertion.

The reported PCDROID-560 bug is keycode-based — the user's log showed `Media3 media button event: keyCode=86` (KEYCODE_MEDIA_STOP), so AAOS dispatched the steering-wheel press as a key event. The keycode handlers in this PR (FAST_FORWARD / REWIND + automotive NEXT / PREVIOUS bypass) are what actually fix it; the command advertisement was belt-and-suspenders for any AAOS implementation that might route via Media3 commands instead of key events. If we ever see one that needs the command route, we can re-add it scoped per-controller via `ConnectionResult` in `onConnect` so Wear OS isn't affected.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/Media3SessionCallback.kt
```diff
@@ -209,37 +211,42 @@ internal class Media3SessionCallback(
...
-                    } catch (e: Exception) {
-                        Timber.e(e, "PiP skip forward failed")
-                    }
-                }
+            // PiP skip buttons and the FAST_FORWARD/REWIND media keys (used by some
+            // AAOS steering-wheel implementations) always skip forward/back, bypassing
+            // headphone control settings. See PCDROID-560.
+            KeyEvent.KEYCODE_MEDIA_SKIP_FORWARD,
+            KeyEvent.KEYCODE_MEDIA_FAST_FORWARD,
+            -> {
```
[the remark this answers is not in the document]
joashrajin: The PR description was stale — updated. The advertisement was removed deliberately in [`2708c28`](https://github.com/Automattic/pocket-casts-android/pull/5328/commits/2708c28a7) after @geekygecko flagged that exposing those commands unconditionally caused Wear OS to render duplicate skip controls ([screenshot](https://github.com/Automattic/pocket-casts-android/pull/5328#discussion_r3286574602)).

You're right that AAOS controllers will continue to send fallback keycodes — that's the intended outcome now. The reported PCDROID-560 case is keycode-based (`Media3 media button event: keyCode=86`), and the FAST_FORWARD/REWIND + automotive NEXT/PREVIOUS handlers in this PR catch every fallback path. If a future AAOS implementation ever routes only via Media3 commands, we'd re-add the advertisement scoped per-controller via `ConnectionResult` in `onConnect` so Wear OS stays clean.

modules/services/repositories/src/test/java/au/com/shiftyjelly/pocketcasts/repositories/playback/Media3SessionCallbackTest.kt
```diff
@@ -436,6 +439,72 @@ class Media3SessionCallbackTest {
         )
     }
 
+    // --- FAST_FORWARD / REWIND keycode tests (PCDROID-560) ---
+
+    @Test
+    fun `KEYCODE_MEDIA_FAST_FORWARD calls skipForwardSuspend directly`() = runTest {
+        mockSkipSettings()
```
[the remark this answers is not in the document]
joashrajin: The `onConnect returns ... with player commands` assertions are correct — `SEEK_TO_NEXT` / `SEEK_TO_PREVIOUS` are intentionally **not** advertised, per the Wear OS regression flagged on [#discussion_r3286574602](https://github.com/Automattic/pocket-casts-android/pull/5328#discussion_r3286574602). PR description has been updated to reflect this.

Also restored the matching test name `available commands expose only core controls` in `PocketCastsForwardingPlayerTest` (was inadvertently left as the temporary `…and skip-to-next-previous` form from the earlier revision) — see [`e17441b`](https://github.com/Automattic/pocket-casts-android/pull/5328/commits/e17441bc3).

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PocketCastsForwardingPlayer.kt
joashrajin: Is this better @geekygecko or do you have any other suggestions?

PR #5359, opened 2026-05-28
Request reviews from apps-infra-tooling on Dependabot PRs

mokagio:
The intent of this config change is to get folks in the new @Automattic/apps-infra-tooling team notified when PRs such as #5357 happen so that we can act on them without distracting the product devs.


geekygecko (approved):
Thanks for doing this.

PR #5367, opened 2026-06-01
Attempt to fix RefreshTokenExpiredException crash

geekygecko:
## Description

When the OkHttp token interceptor requested an access token and the refresh token had expired, `TokenHandler.getAccessToken()` threw a `RefreshTokenExpiredException`. That exception propagated out of the interceptor and looked like it was causing app crashes in Sentry. 

This change is behind a remote feature flag so we can disable it quickly if needed.

## Testing Instructions

This is tricky one to test so it's covered by the new `TokenInterceptorTest` unit test.

## Checklist

- [x] with different themes

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/SyncAccountManagerImpl.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/SyncAccountManagerImpl.kt
@@ -120,6 +120,9 @@ open class SyncAccountManagerImpl @Inject constructor(
                 } else {
                     AccessToken(token)
                 }
+            } catch (e: RefreshTokenExpiredException) {
+                LogBuffer.i(LogBuffer.TAG_BACKGROUND_TASKS, "Could not get token because the refresh token has expired")
+                throw e
             } catch (e: Exception) {
                 LogBuffer.e(LogBuffer.TAG_BACKGROUND_TASKS, e, "Could not get token")
                 throw e // Rethrow the exception so it carries on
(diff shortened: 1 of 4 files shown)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/SyncAccountManagerImpl.kt
```diff
@@ -120,6 +120,9 @@ open class SyncAccountManagerImpl @Inject constructor(
                 } else {
                     AccessToken(token)
                 }
+            } catch (e: RefreshTokenExpiredException) {
```
sztomek, line 123: ⛏️ instead of catching the exception here and then re-throw it again, i would just simply log it it in line 118.

sztomek: oh nevermind, that would be caught in the next `catch (e: Exception)` block

modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/di/InterceptorModule.kt
```diff
@@ -109,17 +112,30 @@ object InterceptorModule {
             return builder.build()
         }
 
+        fun getAccessToken(): AccessToken? {
+            return if (FeatureFlag.isEnabled(Feature.INTERCEPTOR_REFRESH_TOKEN_FALLBACK)) {
+                try {
```
sztomek, line 117: 💡 could be swapped to a `runCatching { ... }.getOrNull()`

sztomek (approved):

modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/di/InterceptorModule.kt
geekygecko:
```
                runCatching {
                    runBlocking { tokenHandler.getAccessToken() }
                }.onFailure { error ->
                    if (error !is RefreshTokenExpiredException) {
                        throw error
                    }
                }.getOrNull()
```
This ended up being more lines. I might just leave it like it is for now.

PR #5371, opened 2026-06-01
Fix Profile settings gear being unreachable in landscape

joashrajin:
## References
Fixes [PCDROID-582](https://linear.app/a8c/issue/PCDROID-582/tablet-landscape-giftguest-pass-icon-overlaps-settings-gear)

## What it does
Fixes the Profile screen so the Settings (gear) icon stays tappable in landscape. The referrals (gift) icon and its tooltip previously overlapped the gear, swallowing its taps — most visible on tablets in landscape.

Keeps the gift and Settings icons **grouped on the right** in landscape (unchanged layout) and instead anchors the referrals tooltip to `TipPosition.TopEnd` in landscape, so the tooltip bubble extends **left, away from the Settings icon** — leaving the gear clickable. (Approach suggested by @geekygecko in review.)

## Changes
- In `ReferralsIconWithTooltip.kt`, use `TipPosition.TopEnd` for the referrals tooltip when the device is in landscape (and keep `TopStart` in portrait). In landscape the gift sits just left of the gear, so anchoring the tooltip to its end makes the bubble grow away from the gear instead of over it.

## Checklist
- [ ] I have reviewed my own changes
- [x] This PR adds fewer than 500 lines of code

### Prerequisite
- You must be **signed in to a Pocket Casts account** (one eligible for the referrals offer) to reproduce. The gift/referrals icon and its tooltip only appear when logged in — that icon/tooltip is what overlapped the gear. Logged out, there's nothing to overlap it.

### How to test
- Build and install on a tablet (or tablet emulator); a phone in landscape works too.
- Sign in to a Pocket Casts account (see Prerequisite).
- Open the **Profile** tab and rotate to **landscape** so the "Gift 2 months…" tooltip shows.
- [ ] Verify the gift and Settings icons are grouped at the right, and the tooltip bubble extends **left** of the gift (not over the gear).
- [ ] Tap the **Settings gear** — confirm Settings opens (it's no longer covered by the tooltip).
- Rotate back to **portrait**.
- [ ] Confirm the tooltip still anchors normally (TopStart) and the gear still opens Settings.


joashrajin:
Before vs After: https://linear.app/a8c/issue/PCDROID-582/tablet-landscape-giftguest-pass-icon-overlaps-settings-gear-icon

geekygecko (commented):
Thanks for looking at it fixed for this. I wonder if it looks a little bit nicer with the gift icon on the right in landscape. How about the following change? It moves the anchor to the right side of the tooltip in landscape and also means the settings icon is clickable. 

Image

Here's the git diff patch you could apply.

[landscape.patch](https://github.com/user-attachments/files/28499424/landscape.patch)

joashrajin:
> Thanks for looking at it fixed for this. I wonder if it looks a little bit nicer with the gift icon on the right in landscape. How about the following change? It moves the anchor to the right side of the tooltip in landscape and also means the settings icon is clickable.
> 
> Image
> Here's the git diff patch you could apply.
> 
> [landscape.patch](https://github.com/user-attachments/files/28499424/landscape.patch)

Thanks Phil! Can do!

geekygecko (approved):
I have added a change log entry. It might be worth pointing your AI agent to the pull request template that normally would get created if you create the pull request through GitHub interface. 

https://github.com/Automattic/pocket-casts-android/blob/main/.github/pull_request_template.md?plain=1

PR #5380, opened 2026-06-03
Attempt to fix ExoPlayer crash

sztomek:
## Description
Noticed this crash in 8.13-rc-1, see ticket for details.

Claude's investigation:
`ExoPlayer.seekTo()` internally accesses playbackInfo.periodId.periodUid and passes it to `PlaylistTimeline.getPeriodByUid()`, which casts it to `Pair`. If the UID is a plain `Object` (not yet wrapped in a Pair by the PlaylistTimeline), it throws `ClassCastException`.                                           
Why it happens: SimplePlayer.prepare() calls player.setMediaSource(source) then player.prepare(). The prepare() call is asynchronous — it kicks off work on ExoPlayer's internal playback thread. But our code immediately sets prepared = true and proceeds to playIfAllowed(), which calls player.seekTo(). At this point, ExoPlayer's internal playbackInfo may not yet have consistent period UIDs — the internal thread hasn't finished processing, and the masking state between the application thread and the playback thread can be momentarily inconsistent. We're seeking before the player's timeline is actually ready for timeline-dependent operations.

How the fix works:
  1. Deferred seek: Before seeking, we check player.playbackState. If it's STATE_READY or STATE_BUFFERING, the timeline is initialized and seeking is safe. If it's still STATE_IDLE (the state right after prepare() is called but before the internal thread has processed it), we register a one-shot listener and seek
   when the player transitions to BUFFERING/READY. This respects Media3's async prepare contract.
  2. Try-catch safety net: Even with the deferred seek, there may be other code paths where a seek reaches the ExoPlayer at a bad time (e.g., a Bluetooth controller sending a seek command through the MediaSession during player initialization). The ClassCastException catch prevents those from crashing.

**UPDATE**
after a discussion with @geekygecko we have decided not to include this fix in 8.13 as it only affected 2 users in the beta cycle. i will change target to `main` and milestone to 8.14

Fixes PCDROID-591 https://linear.app/a8c/issue/PCDROID-591/attempt-to-fix-classcastexception-crash-of-exoplayer

## Testing Instructions
I was not able to reproduce it :(
Just smoke test playback, seeking, resuming from previous position on cold app start.

## Checklist

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SimplePlayer.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SimplePlayer.kt
@@ -136,12 +136,11 @@ class SimplePlayer(
             Toast.makeText(context, "Unable to seek. File headers appear to be invalid.", Toast.LENGTH_SHORT).show()
         } else {
             val currentPlayer = player ?: return
-            val state = currentPlayer.playbackState
-            // attempting to fix this crash: https://linear.app/a8c/issue/PCDROID-591/attempt-to-fix-classcastexception-crash-of-exoplayer
-            if (state == Player.STATE_READY || state == Player.STATE_BUFFERING) {
+            // https://linear.app/a8c/issue/PCDROID-591/attempt-to-fix-classcastexception-crash-of-exoplayer
+            if (currentPlayer.isReadyForSeek()) {
                 seekPlayer(currentPlayer, positionMs)
             } else {
-                currentPlayer.addListener(object : Player.Listener {
+                val listener = object : Player.Listener {
                     override fun onPlaybackStateChanged(playbackState: Int) {
                         if (playbackState == Player.STATE_READY || playbackState == Player.STATE_BUFFERING) {
                             currentPlayer.removeListener(this)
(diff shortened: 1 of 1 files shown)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SimplePlayer.kt
```diff
@@ -135,8 +135,42 @@ class SimplePlayer(
...
+                        if (playbackState == Player.STATE_READY || playbackState == Player.STATE_BUFFERING) {
+                            currentPlayer.removeListener(this)
+                            seekPlayer(currentPlayer, positionMs)
+                        } else if (playbackState == Player.STATE_IDLE || playbackState == Player.STATE_ENDED) {
+                            currentPlayer.removeListener(this)
+                        }
+                    }
+                }
+                currentPlayer.addListener(listener)
+                if (currentPlayer.isReadyForSeek()) {
```
geekygecko: Could there be an issue if the user changes episodes before isReadyForSeek is true?

sztomek: When the user changes episodes, `handleStop()` calls `player.stop()` which synchronously transitions the player to `STATE_IDLE`. The listener catches STATE_IDLE (line 148-149) and removes itself without performing the seek. This cleanup happens before release() is called, so the listener never fires a stale seek on the old player instance.
So in theory everything should be fine.

sztomek: as per or discussion, i made another change to only attach the listener when the player is idle 211872d83111faccc69af5a5e9ada4bc58e534c1

PR #5393, opened 2026-06-05
Remove podcast sort changes feature flag

geekygecko:
## Description

Removes the `PODCASTS_SORT_CHANGES` feature flag now that the podcasts sort changes have fully shipped. The flag defaulted to `true`, so this enables the behaviour permanently and deletes the now-dead branches:

- The **Recently Played** sort option is now always shown in the podcasts sort options dialog (previously filtered out when the flag was off).
- `PodcastsSortType.fromServerId` / `fromClientIdString` no longer filter out `RECENTLY_PLAYED` based on the flag.
- The recently-played sort tooltip is now gated only on the user setting, not the flag.
- The `PODCASTS_SORT_CHANGES` entry is removed from `Feature.kt`.

This is a tech-debt cleanup with no intended change in user-facing behaviour, since the flag already defaulted to enabled.

## Testing Instructions

1. Open the **Podcasts** tab.
2. Tap the options/overflow menu and choose **Sort by**.
3. Confirm the **Recently Played** option is present in the list of sort orders.
4. Select **Recently Played** and confirm podcasts re-sort accordingly.
5. Confirm the recently-played sort tooltip still appears/dismisses according to the existing user setting.

## Checklist

- [x] with different themes


sztomek (approved):
Code looks good, sorting worked as expected!

PR #5419, opened 2026-06-12
Add trial CTA copy A/B/C test on the upgrade screen

yaelirub:
## Description

Solves POC-667

Adds a 3-arm (A/B/C) ExPlat experiment that tests the wording of the free-trial CTA button on the Plus upgrade screens. Today the button reads a generic "Start Free Trial"; we want to test whether stating the trial length and/or reframing it more invitingly increases trial starts.

**Experiment:** `pocketcasts_android_trial_cta_copy_ab_test` (configured in ExPlat)

| Variation | CTA copy |
|---|---|
| `control` | Start Free Trial (current) |
| `start_30_day_trial` | Start your 30-day Free Trial |
| `try_30_days_free` | Try 30 days for free |

Copy only changes when the offer resolves to a free trial (`SubscriptionOffer.Trial`); non-trial users are unaffected. Gated behind the `TRIAL_CTA_COPY_AB_TEST` Firebase-backed kill-switch (default off in release) plus the global `EXPLAT_EXPERIMENT` flag. Threaded through the shared `OnboardingSubscriptionPlan.ctaButtonText(...)`, so it covers all three trial-eligible upgrade surfaces: the onboarding paywall, the onboarding upgrade screen variant, and the profile/account-details upgrade card.

## Testing Instructions

1. Enable the `TRIAL_CTA_COPY_AB_TEST` feature flag (Beta Features dev screen) on a debug build.
2. With the experiment assigning a variation (or via a local override of the resolved treatment), open a Plus upgrade surface that offers a free trial — e.g. onboarding upgrade, or Profile → Account → upgrade card.
3. Confirm the CTA button reflects the assigned arm:
   - `control` → "Start Free Trial"
   - `start_30_day_trial` → "Start your 30-day Free Trial"
   - `try_30_days_free` → "Try 30 days for free"
4. Confirm non-trial offers (intro/referral/winback) are unchanged ("Subscribe to …").

Debug builds use fake subscription plans, so the paywall renders on an emulator without Google Play billing.

## Screenshots or Screencast

All three arms verified on-device (emulator); can attach captures on request.

## Checklist

- [ ] with different themes


modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/upgrade/OnboardingSubscriptionPlan.kt
```diff
@@ -178,11 +179,28 @@ data class OnboardingSubscriptionPlan private constructor(
...
+    ) = if (isRenewingSubscription) {
         stringResource(LR.string.renew_your_subscription)
     } else {
         when (key.offer) {
-            SubscriptionOffer.Trial -> stringResource(LR.string.profile_start_free_trial)
+            SubscriptionOffer.Trial -> when (val treatment = trialCtaCopyTreatment) {
+                null -> stringResource(LR.string.profile_start_free_trial)
+                else -> {
+                    val discountedPhase = requireNotNull(discountedPricingPhase)
+                    val recurringPeriods = (discountedPhase.schedule.recurrenceMode as RecurrenceMode.Recurring).value
```
sztomek: ❓ are we sure that `schedule.recurrenceMode` cast to `RecurrenceMode.Recurring` will always succeed? i feel `val recurringPeriods = (discountedPhase.schedule.recurrenceMode as? RecurrenceMode.Recurring).value` would be a safer bet

modules/services/analytics/src/main/java/au/com/shiftyjelly/pocketcasts/analytics/experiments/TrialCtaCopyTreatment.kt
```diff
@@ -0,0 +1,28 @@
...
+
+import au.com.shiftyjelly.pocketcasts.utils.featureflag.Feature
+import au.com.shiftyjelly.pocketcasts.utils.featureflag.FeatureFlag
+
+enum class TrialCtaCopyTreatment(val treatmentName: String) {
+    START_30_DAY_TRIAL("start_30_day_trial"),
+    TRY_30_DAYS_FREE("try_30_days_free"),
+    ;
+
+    companion object {
```
sztomek: 💡 could limit visibility of this object to `private` as the `ExperimentProvider.getTrialCtaCopyTreatment()` extension is the only code path that uses it

sztomek (approved):
The code looks good overall, I had some small comments to consider.

@yaelirub please try to use our PR template in to future to keep them consistent. Also the CI checks are failing, please fix them

modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/onboarding/upgrade/OnboardingSubscriptionPlan.kt
yaelirub: @sztomek  Good call — switched to a safe cast in bab0114ee. It now reads the recurrence count via `as?` and falls back to the generic "Start Free Trial" copy if the trial duration can't be derived, so there's no `ClassCastException` risk.

(Small note: `(… as? RecurrenceMode.Recurring).value` wouldn't compile since `.value` would be on a nullable, so I used `?.value` plus a null-guard fallback.)

For context, the cast is actually guaranteed in practice — `OnboardingSubscriptionPlan.create()` rejects any trial plan whose phase isn't `Recurring`, and the sibling `offerBadgeText` / `planDescriptionText` make the same assumption — but the defensive version is harmless, so happy to take it. The sibling casts (lines ~119/157) still use the unchecked form; left them as-is since they're outside this PR's scope, but glad to harden them in a follow-up if we want consistency.

modules/services/analytics/src/main/java/au/com/shiftyjelly/pocketcasts/analytics/experiments/TrialCtaCopyTreatment.kt
yaelirub: @sztomek Good point — done in d24a767b1 (rebased as f0d0dd94a). Since `getTrialCtaCopyTreatment()` was the only caller, I removed the companion `fromTreatmentName()` entirely and inlined the lookup into the extension, so there's no extra public API surface to expose.

(`private` on the companion member itself wouldn't have worked here — the extension is top-level, and a `private` companion member isn't visible to it since `private` is class-scoped rather than file-scoped. Inlining sidesteps that.)

yaelirub:
@sztomek , I updated the PR description based on the template and addressed yours and Copilots review comments. Seems like all checks pass now.

yaelirub:
@copilot resolve the merge conflicts in this pull request

sztomek:
@yaelirub thank you, everything looks great now! pressing merge...

PR #5430, opened 2026-06-15
[HLS] Block downloads of HLS-only episodes and add HLS segment-error fallback

sztomek:
## Description
This PR adds 2 changes:
- disables download option when only HLS stream is available as HLS file is just a manifest -- this is mostly theorical case.
- adds enhancements around error hanlding policy (when the FF is enabled) for hls streaming

Fixes PCDROID-610 https://linear.app/a8c/issue/PCDROID-610/update-error-handling-while-streaming-hls

## Testing Instructions
Just review the code

## Checklist


modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/episode/EpisodeFragment.kt
```diff
@@ -467,6 +467,7 @@ class EpisodeFragment : BaseFragment() {
                                 LR.string.podcasts_download_download,
                             ),
                         )
+                        binding.btnDownload.isVisible = !state.episode.isHlsOnly
```
geekygecko, line 470: Just an idea. I wonder if hiding the download button is the right approach and a better approach might be to informing the user with a toast message instead if they try to download it.

geekygecko (approved):

modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/episode/EpisodeFragment.kt
sztomek: Thanks, good thing to question! I went with hiding rather than a toast on purpose:
 - `isHlsOnly` is a permanent property of the episode (its only media is an HLS manifest), not a temporary state — so the download can never succeed. Hiding fits an action that's never possible, whereas a toast-on-tap fits transient failures like no network. A Download button with a size label that only ever errors would be misleading.
 - It's a mostly-theoretical/rare case, so these users have never seen a download button for these episodes — nothing feels missing.
 - We also already reject HLS-only downloads in `DownloadManager.canDownload()`, so the toast isn't needed as a safety net.

Happy to revisit if we start seeing real HLS-only feeds where users expect a download affordance.

PR #5432, opened 2026-06-16
[Dependency Analysis] Run `configure_apply` in the Dependency Analysis Scheduled Job

ParaskP7:
Closes: [AINFRA-2502](https://linear.app/a8c/issue/AINFRA-2502)

---

## Description

The scheduled **Dependency Analysis** pipeline (`.buildkite/schedules/dependency-analysis.yml`, which runs `./gradlew buildHealth`) has been failing since [#5389](https://github.com/Automattic/pocket-casts-android/pull/5389) with:

```
Execution failed for task ':tv:processDebugGoogleServices'
> File google-services.json is missing.
```

`buildHealth` resolves every module's compiled output, and for application modules that transitively runs `process<Variant>GoogleServices`. The existing application modules (`app`, `automotive`, `wear`) were served from the remote build cache, so their Google Services tasks never executed. The new `tv` application module — to which [#5389](https://github.com/Automattic/pocket-casts-android/pull/5389) added the `com.google.gms.google-services` plugin — had no cache entry, so Gradle actually executed `:tv:processDebugGoogleServices`, which requires a real `google-services.json`.

This scheduled job is the only `$CI_TOOLKIT` Gradle job that never ran `configure_apply`, so no `google-services.json` files were ever provisioned and the task failed. (The `tv/google-services.json` mapping itself was already added to `.configure` in #5389 — it just was never applied for this job.)

This PR adds `install_gems` and `bundle exec fastlane run configure_apply` before `./gradlew buildHealth`, matching the secrets setup already used by the lint and unit-test jobs. `configure_apply` provisions all four `google-services.json` files, so the analysis succeeds even on a cache miss.

---

## Testing Instructions

This is a CI-only change, so the reliable verification is a real run of the scheduled pipeline on this branch:

1. Run `./gradlew buildHealth` locally and verify it completes successfully.
2. Using the `New Build` 🟢 CI button for [PCAndroid](https://buildkite.com/automattic/pocket-casts-android), test this standalone dependency analysis job. (Last successful test run: [17133](https://buildkite.com/automattic/pocket-casts-android/builds/17133))
    - FYI: Start a **New Build** with:
        - **Message**: `Trigger Standalone "Dependency Analysis" Job`
        - **Commit**: `HEAD`
        - **Branch**: `da/add-configure-apply-to-dependency-analysis-in-ci`
        - **Environment Variables**: `PIPELINE=schedules/dependency-analysis.yml`
        - **Clean Checkout**: unchecked
    - PS: Confirm the build passes and the `:tv:processDebugGoogleServices` step no longer reports `File google-services.json is missing.`

---

## Screenshots or Screencast

N/A — CI configuration change with no user-facing or UI impact.

---

## Checklist
- [ ] If this is a user-facing change, I have added an entry in CHANGELOG.md _(N/A — CI-only change)_
- [ ] I have considered whether it makes sense to add tests for my changes _(N/A — pipeline config)_
- [ ] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml` _(N/A)_
- [ ] Any jetpack compose components I added or changed are covered by compose previews _(N/A)_
- [ ] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics. _(N/A — no analytics changes)_

_(N/A — this PR contains no UI changes)_

### DIFF
--- a/.buildkite/schedules/dependency-analysis.yml
+++ b/.buildkite/schedules/dependency-analysis.yml
@@ -8,6 +8,13 @@ steps:
   - label: "dependency analysis"
     command: |
       .buildkite/commands/restore-cache.sh
+
+      echo "--- :rubygems: Setting up Gems"
+      install_gems
+
+      echo "--- :closed_lock_with_key: Installing Secrets"
+      bundle exec fastlane run configure_apply
+
       echo "--- 📊 Analyzing"
       ./gradlew buildHealth
     plugins: [$CI_TOOLKIT]


mokagio (approved):
Looking good

Image

mokagio:
Taking over merging this immediately to ship the fix as soon as possible.

ParaskP7:
Thank you @mokagio ! 🙇❤️🚀

wzieba:
> The existing application modules (app, automotive, wear) were served from the remote build cache, so their Google Services tasks never executed.

I don't think it's correct. 

If you fresh clone this project and don't enable build cache, you can still build `app`. It's because `app` provides a fake, debug only `google-services.json`.

I think we should provide a fake `google-services.json` as well: Pocket Casts is open source project, so I think we should keep it easy to build. Until now, contributors could simply clone the project and build it.

What I think we should do:
1. Provide a fake `google-services.json` to `tv` module
2. Revert this PR as not needed anymore (dependency analysis should not require secrets conceptually)

ParaskP7:
Thanks for double-checking on that @wzieba , much appreciated! 🙇 

> I don't think it's correct.
> 
> If you fresh clone this project and don't enable build cache, you can still build app. It's because app provides a fake, debug only google-services.json.

I actually (quickly) checked on that, but couldn't see such configuration, thus went with the `remote build cache` idea... 😞 :facepalm: 😞

I now see the `debug` and `debugProd` source-sets having `google-services.json`, which helps them build the `debug` variant of the `app` and other such modules, except `tv`, hmmm, and if we make sure that `buildHealth` only runs for `debug` and `debugProd` builds, which it currently does, then this indeed solves it. 👍

> I think we should provide a fake google-services.json as well: Pocket Casts is open source project, so I think we should keep it easy to build. Until now, contributors could simply clone the project and build it.

I agree, and you're right, this project, as is, with a fresh clone, cannot be build. 💯 

> What I think we should do:
> 
> 1. Provide a fake google-services.json to tv module
> 2. Revert this PR as not needed anymore (dependency analysis should not require secrets conceptually)

I've just reopened [AINFRA-2502](https://linear.app/a8c/issue/AINFRA-2502) and will do exactly that, to solve the clone problem as well, thanks again! 👍 🙇

ParaskP7:
> I agree, and you're right, this project, as is, with a fresh clone, cannot be build. 💯

FYI: I think we should do the same with DOAndroid at some point, because it too can't be build from a fresh clone and thus uses `install_gems` and `bundle exec fastlane run configure_apply` for dependency analysis, and everywhere else for that matter, which threw me off here and I immediately went and did this approach of mine, making this actually worse and not solving the underlying problem, then thinking `remote build cache` and (again) auto-trusting AI on it too much... 😞 🤷 😞 🤦 😞 _(I'll create a Linear issue for DOAndroid, long overdue)_

wzieba:
No worries, it's not like we didn't made mistakes before AI in the toolkit! 😄 I just noticed this and something wasn't clicking for me, hence I decided to check more deeply.

> FYI: I think we should do the same with DOAndroid at some point, because it too can't be build from a fresh clone

Yeah - I never decided to handle this because DOAndroid is closed-source, so I thought it's not necessary. But ideally, I think it's always good for the project to be build without any secrets so 👍 from me. (and _dependency analysis should not require secrets conceptually_ opinion upholds for DOAndroid as well)

ParaskP7:
> No worries, it's not like we didn't made mistakes before AI in the toolkit! 😄

It is so easy for me to make (silly) mistakes these days, just by depending on AI doing the research, and going with the flow. I do need to stop depending on AI too much and triple check things going forward, double checking isn't enough for me anymore, especially for seemingly easy tasks that you think, oh well, easy, AI go do it (I'll do something else), then boom... 💥 😅 

> I just noticed this and something wasn't clicking for me, hence I decided to check more deeply.

Much, much, much appreciated! 🙇 x 🙇 ^ 🙇

> Yeah - I never decided to handle this because DOAndroid is closed-source, so I thought it's not necessary. But ideally, I think it's always good for the project to be build without any secrets so 👍 from me. (and dependency analysis should not require secrets conceptually opinion upholds for DOAndroid as well)

💯 and same here, then I couldn't use GitHub Actions to build me the project, during my AI experiments with DOAndroid, and that drew the line for me, it is about time and we should have that the default, no matter if a project is open or closed source.

PS: Thanks for the 👍 !

PR #5443, opened 2026-06-17
[AI Chapters] Update tracked events

sztomek:
## Description
This branch updates EventHorizon to latest version so we can now track chapter events with their new source property.
Generated chapters should have events reported with `source=generated` while episodes with curated chapters should have `source=native_media`.
I had to update the database schema to get an accurate picture on the chapters' source.

Fixes PCDROID-612 https://linear.app/a8c/issue/PCDROID-612/update-chapters-related-tracks-events

## Testing Instructions
1. Make sure `generated_chapters` is OFF
2. Find a podcast that has chapters like the Buzzcast 
3. Open one of its episodes with chapters
4. Notice the chapters tab on the episode details page
5. Open logcat and filter to "Event:" 
6. Start playing with chapters, disable a few, select a chapter, open fullscreen player and navigate between chapters with the < and > arrows.
7. Observe chapter-related events, they should all have source=native_media
8. Now enable the FF and restart the app
9. Find a podcast with generated chapters, e.g. The American Life
10. Do the same steps 3-7, but now you should see source=generated

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>ChapterOriginTypeConverter</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>

</p>

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/AppDatabase.kt
```diff
@@ -1458,6 +1460,12 @@ abstract class AppDatabase : RoomDatabase() {
             database.execSQL("ALTER TABLE bookmarks ADD COLUMN ai_summary_modified INTEGER")
         }
 
+        val MIGRATION_131_132 = addMigration(131, 132) { database ->
+            database.execSQL("ALTER TABLE episode_chapters ADD COLUMN origin INTEGER NOT NULL DEFAULT 0")
+            database.execSQL("UPDATE episode_chapters SET origin = 3 WHERE is_embedded = 1")
+            database.execSQL("UPDATE episode_chapters SET origin = 4 WHERE is_generated = 1")
```
geekygecko, line 1470: It doesn't seem like `is_embedded` and `is_generated` are needed as well as `origin`. Should we drop the old columns?

geekygecko (approved):
Looks good, only one minor comment. 

> Observe chapter-related events, they should all have source=native_media

Why are they all native_media? Podcasts like "The Changelog" include their own chapters in the feed XML, shouldn't they be `ShowNotes`? Or if we aren't sure yet until after the migration `Unknown`? 

The description spoke about the property `source` but the code is using `origin`, I assume it changed at a later date.

sztomek:
Thanks! Addressed your comments in `Address PR changes` (also merged `main` in to resolve the schema/migration conflict with the HLS work — both branches had landed on DB v132, so the chapters migration is now `MIGRATION_132_133`).

> Why are they all native_media? Podcasts like "The Changelog" include their own chapters in the feed XML, shouldn't they be ShowNotes?

The origin is set per source, so it's not always `native_media` — that line in the testing steps was just describing what I happened to see with Buzzcast. The reason embedded podcasts report `native_media` is the precedence in `ChapterDao.replaceAllChapters`: chapters embedded in the media file (read by the player) override show-notes chapters. So Buzzcast (chapters in the file) ends up `native_media`, a feed-XML-only podcast like The Changelog stays `show_notes`, podcast-index URLs are `podcast_index`, and rows that were already in the DB before the migration stay `unknown` until they're refetched. Will fix up the testing steps to be clearer.

> The description spoke about `source` but the code is using `origin`

Yep — the analytics property got renamed `source` → `origin` after I wrote the description. The code/property is `origin`.

> It doesn't seem like `is_embedded` and `is_generated` are needed as well as `origin`. Should we drop the old columns?

Good call — done. `MIGRATION_132_133` now back-fills `origin` from the old flags and then drops both columns, and there's a migration test covering the back-fill + a 5k-row batch.

PR #5446, opened 2026-06-18
Wear OS performance improvements

geekygecko:
## Description

This PR addresses ANRs and UI jank in the Wear OS app by moving expensive work off the main thread and removing blocking calls during startup and screen rendering.

**Startup (`PocketCastsWearApplication`)**
- Replaced the `runBlocking { ... }` block in `setupApp()` with non-blocking work. Notification channels, the app lifecycle observer, the playback-service toggle, and version migrations still run synchronously (migrations must complete before the UI reads any state), but the expensive playback/storage setup, account/download monitors, and stats workers are now deferred onto an injected `@ApplicationScope` coroutine instead of blocking `onCreate()`.
- This is the primary ANR fix — `onCreate()` no longer blocks on playback manager setup and storage option enumeration.

**Show notes parsing (`EpisodeViewModel` / `EpisodeScreen`)**
- HTML show notes were being parsed (`parseAsHtml()`) on the main thread inside the composable on every render. Parsing now happens once in the ViewModel flow via `.map { ... }.flowOn(DefaultDispatcher)` when the notes change, and the screen consumes a plain `String?` (`showNotesText`) instead of `ShowNotesState`.

**Podcast screen (`PodcastViewModel`)**
- Moved the episode/podcast flow computation onto `DefaultDispatcher` with `.flowOn(...)` so it no longer runs on the main thread.

**Settings & connectivity (`SettingsViewModel`, `ConnectivityStateManager`, `WearMainActivityViewModel`)**
- Removed the blocking `userManager.getSignInState().blockingFirst()` call from `SettingsViewModel`'s initial state; `signInState` is now nullable and populated reactively, with the UI rendering nothing until the first emission.
- Changed the default `isConnected` value to `true` (in both `ConnectivityStateManager` and `WearMainActivityViewModel`) so the UI doesn't briefly flash an offline/disconnected state before the real connectivity status arrives.
- Stopped forcing `showLoggingInScreen = true` on the already-logged-in sync path.

## Testing Instructions

1. Build and install the Wear OS app on a watch (or emulator)
2. Cold-start the app and confirm it launches correctly when logged in and out.
3. Confirm playback still works (start an episode, ensure the playback service starts and audio plays).
4. Open an episode detail screen with HTML show notes and confirm the notes render correctly and scrolling/expanding is smooth.
5. Open a podcast screen and confirm episodes load.
6. Open Settings and confirm the sign-in/log-out chip appears once state loads, and the data warning / background refresh toggles reflect the correct values.

## Checklist

- [x] with different themes

### DIFF
--- a/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/PocketCastsWearApplication.kt
+++ b/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/PocketCastsWearApplication.kt
@@ -101,6 +101,7 @@ class PocketCastsWearApplication :
 
         notificationHelper.setupNotificationChannels()
         appLifecycleObserver.setup()
+        PlaybackServiceToggle.ensureCorrectServiceEnabled(application)
 
         // Apply migrations before the UI starts, so the app never reads pre-migration state.
         VersionMigrationsWorker.performMigrations(
@@ -111,8 +112,6 @@ class PocketCastsWearApplication :
 
         // Defer the expensive playback/storage setup and monitors off the main thread to avoid blocking onCreate (ANRs).
         applicationScope.launch {
-            PlaybackServiceToggle.ensureCorrectServiceEnabled(application)
-
             playbackManager.setup()
             val storageChoice = settings.getStorageChoice()
             if (storageChoice == null) {
--- a/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/PocketCastsWearApplication.kt
+++ b/wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/PocketCastsWearApplication.kt
@@ -6,6 +6,7 @@ import androidx.work.Configuration
 import au.com.shiftyjelly.pocketcasts.BuildConfig
 import au.com.shiftyjelly.pocketcasts.analytics.AnalyticsController
 import au.com.shiftyjelly.pocketcasts.analytics.experiments.ExperimentProvider
+import au.com.shiftyjelly.pocketcasts.coroutines.di.ApplicationScope
 import au.com.shiftyjelly.pocketcasts.crashlogging.InitializeRemoteLogging
 import au.com.shiftyjelly.pocketcasts.preferences.Settings
 import au.com.shiftyjelly.pocketcasts.repositories.download.DownloadStatusObserver
@@ -30,9 +31,8 @@ import dagger.hilt.android.HiltAndroidApp
 import java.io.File
 import java.util.concurrent.Executors
 import javax.inject.Inject
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.runBlocking
-import kotlinx.coroutines.withContext
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
 import timber.log.Timber
 
 @HiltAndroidApp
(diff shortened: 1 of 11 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/networking/ConnectivityStateManager.kt
```diff
@@ -26,6 +26,6 @@ class ConnectivityStateManager @Inject constructor(
         .stateIn(
             scope = coroutineScope,
             started = SharingStarted.Eagerly,
-            initialValue = false,
+            initialValue = true,
```
geekygecko, line 29: The connectivity icon was showing as disconnected when the app started, until the actual connection state had been determined. I think it would be better to assume the app is connected initially, which keeps the icon hidden until the real state is known.

wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/PocketCastsWearApplication.kt
```diff
@@ -94,39 +97,40 @@ class PocketCastsWearApplication :
...
+        VersionMigrationsWorker.performMigrations(
+            context = application,
+            settings = settings,
+            moshi = moshi,
+        )
+
+        // Defer the expensive playback/storage setup and monitors off the main thread to avoid blocking onCreate (ANRs).
+        applicationScope.launch {
+            playbackManager.setup()
+            val storageChoice = settings.getStorageChoice()
```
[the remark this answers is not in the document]
geekygecko: I have tried to address this feedback.

wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/episode/EpisodeViewModel.kt
```diff
@@ -191,12 +199,18 @@ class EpisodeViewModel @Inject constructor(
...
                 inUpNext = isInUpNext(upNext, episode),
                 tintColor = getTintColor(episode, podcast, theme),
-                showNotesState = showNotesState,
+                showNotesText = showNotesText,
                 errorData = getErrorData(episode),
             )
         }.stateIn(viewModelScope, SharingStarted.WhileSubscribed(), State.Empty)
     }
 
+    private fun parseShowNotes(showNotesState: ShowNotesState): String? = if (showNotesState is ShowNotesState.Loaded) {
```
sztomek: 💡 equivalent to `(showNotesState as? ShowNotesState.Loaded)?.showNotes?.parseAsHtml()?.toString()`

sztomek (approved):
Thanks for making these improvements, I hope we're going to see positive uptick on wear health metrics!

wear/src/main/kotlin/au/com/shiftyjelly/pocketcasts/wear/ui/episode/EpisodeViewModel.kt
geekygecko: Thanks for the suggestion, I'm making this change.

PR #5449, opened 2026-06-18
[Synced Transcripts] Fix gaps in highlighting compared to iOS

sztomek:
## Description
This PR aims to bring more closer highlighting to follow the actual audio stream. Some issues have been found during beta testing (pdeCcb-cwm-p2#comment-9187) and I've did a close review and identied some areas of improvement. For example we had a bug on android in the highlighting logic when the cue reached over multiple transcript paragraphs -- that's fixed now. The sampling frequency has been reduced as well so we are now closely following the playback and fingerprints -- similarly to iOS. Lastly, we now highlight the tapped paragraph when the playback is paused.

Fixes PCDROID-613 https://linear.app/a8c/issue/PCDROID-613/improve-android-behavior

## Testing Instructions
1. Open an episode with pre/mid/post-ads (like The Daily)
2. Start listening to the episode with transcripts open and observe highlighting, you should see no (at least much more less gaps) in highlighting as long as ads aren't playing
3. Pause playback
4. Scroll down in transcripts
5. Tap on a paragraph
6. Notice the paragraph gets highlighted
7. Also do a smoke test around highlighting

## Checklist

### DIFF
--- a/modules/features/transcripts/src/main/kotlin/au/com/shiftyjelly/pocketcasts/transcripts/ui/TranscriptPage.kt
+++ b/modules/features/transcripts/src/main/kotlin/au/com/shiftyjelly/pocketcasts/transcripts/ui/TranscriptPage.kt
@@ -297,7 +297,7 @@ private fun HighlightEffect(
                     when (val outcome = resolveHighlight(transcript.entries, posMs, fingerprintTimingManager, cueIndexHolder[0])) {
                         is HighlightOutcome.Show -> {
                             cueIndexHolder[0] = outcome.entryIndex
-                            latestOnHighlightChanged(HighlightState(entryIndex = outcome.entryIndex, wordIndex = outcome.wordIndex))
+                            latestOnHighlightChanged(HighlightState(entryIndex = outcome.entryIndex))
                             wasHighlighting = true
                         }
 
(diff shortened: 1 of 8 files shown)


modules/features/transcripts/src/main/kotlin/au/com/shiftyjelly/pocketcasts/transcripts/ui/TranscriptPage.kt
```diff
@@ -269,59 +287,99 @@ private fun HighlightEffect(
...
             latestOnHighlightChanged(HighlightState())
+            val episode = playbackManager.getCurrentEpisode()
             while (true) {
-                withFrameNanos { _ ->
-                    val posMs = playbackState?.positionMs ?: return@withFrameNanos
-                    val currentRefTime = fingerprintTimingManager.matchedReferenceTime(forPlaybackTimeMs = posMs)
-                    if (currentRefTime == null) {
+                withFrameNanos { }
+                val posMs = if (episode != null) {
+                    playbackManager.getCurrentTimeMs(episode)
```
geekygecko, line 319: As this happens on every frame with a 60 Hz display device this would get called 60 times a second.  Would it be worth reducing this to say 100ms, or does it need to be that regular?

geekygecko (approved):
It worked well most of the time.

There is one issue where the transcript becomes unhighlighted after tapping Play, but I don’t think that should block the PR from being merged. It’s a little difficult to demo because you can't hear the emulator audio during the recording, but around the 30 second mark in the video you can see that the audio is playing while the transcript text is no longer highlighted.

https://github.com/user-attachments/assets/4169f981-5101-46fd-9580-a952ba1c1470

modules/features/transcripts/src/main/kotlin/au/com/shiftyjelly/pocketcasts/transcripts/ui/TranscriptPage.kt
sztomek: this is the way on iOS as well, so i'd rather keep it as it is.

sztomek:
> There is one issue where the transcript becomes unhighlighted after tapping Play, but I don’t think that should block the PR from being merged. It’s a little difficult to demo because you can't hear the emulator audio during the recording, but around the 30 second mark in the video you can see that the audio is playing while the transcript text is no longer highlighted.

thanks for flagging this, i just checked the described scenario and was able to repro it on my end. will address it in an upcoming PR so we won't block the beta release with such minor things

PR #5457, opened 2026-06-22
Sort Up Next by time remaining instead of duration

geekygecko:
## Description

Follow-up to #5399, which added the **Shortest to longest** / **Longest to shortest** Up Next sort options ordering by an episode's full duration. To match the [iOS implementation](https://github.com/Automattic/pocket-casts-ios) (`Change to sort by time remaining`), these options now sort by the episode's **time remaining** (`duration - playedUpTo`) rather than its total duration.

Why: time remaining better reflects a play queue — a long episode that's almost finished now sorts ahead of a short, unplayed one, consistent with iOS.

## Testing Instructions

1. Use either the `debug` or `debugProd` variant (the `up_next_duration` flag defaults on for these builds).
2. Add several episodes to Up Next with differing durations, and partially play some so they have different amounts of time remaining (include at least one episode with an unknown/zero duration — the Pocket Casts test podcast has one).
3. Open the Up Next screen and tap the sort option.
4. Select **Shortest to longest** and confirm episodes are ordered by ascending time remaining (a nearly-finished long episode appears before a short unplayed one).
5. Select **Longest to shortest** and confirm the reverse ordering.

## Checklist

- [x] with different themes

### DIFF
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/type/UpNextSortType.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/type/UpNextSortType.kt
@@ -60,9 +60,6 @@ enum class UpNextSortType(
     ;
 
     private companion object {
-        private val BaseEpisode.hasNoDuration get() = duration <= 0
-
-        // Time remaining accounts for how much of the episode has already been played.
-        private val BaseEpisode.timeRemaining get() = duration - playedUpTo
+        private val BaseEpisode.timeRemaining get() = if (hasNoDuration) 0.0 else (duration - playedUpTo).coerceAtLeast(0.0)
     }
 }
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/type/UpNextSortType.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/type/UpNextSortType.kt
@@ -37,7 +37,7 @@ enum class UpNextSortType(
     ) {
         // Episodes without a known duration always sort to the bottom.
         private val comparatorDelegate = compareBy<BaseEpisode> { it.hasNoDuration }
-            .thenBy { it.duration }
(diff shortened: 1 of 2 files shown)


modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/type/UpNextSortType.kt
```diff
@@ -60,6 +60,6 @@ enum class UpNextSortType(
     ;
 
     private companion object {
-        private val BaseEpisode.hasNoDuration get() = duration <= 0
+        private val BaseEpisode.timeRemaining get() = if (hasNoDuration) 0.0 else (duration - playedUpTo).coerceAtLeast(0.0)
     }
```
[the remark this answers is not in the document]
geekygecko: This AI comment doesn't seem correct as we still use it, and the app compiles and runs.

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/type/UpNextSortType.kt
```diff
@@ -37,7 +37,7 @@ enum class UpNextSortType(
     ) {
         // Episodes without a known duration always sort to the bottom.
         private val comparatorDelegate = compareBy<BaseEpisode> { it.hasNoDuration }
-            .thenBy { it.duration }
+            .thenBy { it.timeRemaining }
```
sztomek, line 40: ⛏️ / 💡 you can pass method refs like `.thenBy(BaseEpisode::timeRemaining)`

sztomek (approved):
worked as described!

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/type/UpNextSortType.kt
geekygecko: Thanks for the tip!

PR #5465, opened 2026-06-23
[HLS] Persist alternate enclosures as server sends them

sztomek:
## Description
This PR makes changes to our database schemas as we're now storing the alternate enclosures array as the server sends it for future purposes -- it's quite probable that we'll want to have an option on the ui to let users select which stream to be played.

Fixes PCDROID-617 https://linear.app/a8c/issue/PCDROID-617/persist-alternate-enclosures

## Testing Instructions
Just make sure the CI is green

## Checklist

### DIFF
--- a/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/AlternateEnclosureDaoTest.kt
+++ b/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/AlternateEnclosureDaoTest.kt
@@ -1,0 +1,89 @@
+package au.com.shiftyjelly.pocketcasts.models.db
+
+import androidx.room.Room
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.test.platform.app.InstrumentationRegistry
+import au.com.shiftyjelly.pocketcasts.models.db.dao.AlternateEnclosureDao
+import au.com.shiftyjelly.pocketcasts.models.db.dao.EpisodeDao
+import au.com.shiftyjelly.pocketcasts.models.di.ModelModule
+import au.com.shiftyjelly.pocketcasts.models.di.addTypeConverters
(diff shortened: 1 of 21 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

geekygecko:
@claude review this

geekygecko:
@claude review this

geekygecko:
@claude review this

sztomek:
Thanks for the review @claude — went through these critically:

**Addressed** (in \`Address PR comments\`):
- **#1 refresh perf** — good catch, this is the hot path. The persist loop now compares the incoming enclosures against what's already stored and only rewrites when they changed, so an unchanged feed no longer churns delete+insert every refresh.
- **compileOnly nit** — added a one-line note explaining why \`compileOnly(media3.common)\` is runtime-safe (the const inlines).

**Skipping for now:**
- **#2 subscribe churn** — that's a one-time cold path; batching would mean adding a speculative bulk delete-by-uuids + insertAll DAO API for marginal gain. Not worth it yet.
- **#3 stale rows when server drops enclosures** — intentional asymmetry; the rare case is cleaned up by the cascading FK when the episode is deleted anyway.
- **#4 \`default\` flag** — out of scope here. This PR only persists; we store \`default\` precisely so the future stream-selection UI can consume it. Selection behaviour stays unchanged on purpose.
- **duplicated parsing** — the two mappers take genuinely different inputs (raw \`org.json\` vs typed Moshi), so a shared mapper needs a common intermediate. Premature until it grows.
- **persistence-path tests** — driving \`refreshPodcast\` end-to-end needs disproportionate \`Response\`/\`PodcastResponse\` mocking. The new #1 gate is small/self-evident; can add DAO or instrumented coverage as a follow-up.

PR #5475, opened 2026-06-24
[Hyundai] Attempt to fix overlapping text on Link page

sztomek:
## Description
Although I never been able to reproduce this issue with the current main branch, I took claude's advice and changed how we create the page content -- now we're using `contentWithoutConsumedInsets` wrapper instead of directly creating a `ComposeView`.

Fixes PCDROID-619 https://linear.app/a8c/issue/PCDROID-619/hyundai-prt-554-text-overlap-on-qr-screen-after-navigation

## Testing Instructions
1. Install Pleos emulator and configure it
2. Disable chromium `adb shell pm disable-user --user 10 org.chromium.chrome.stable
3. Install our app
4. Launch it `adb shell am start --user 10 -n au.com.shiftyjelly.pocketcasts.debug/au.com.shiftyjelly.pocketcasts.AutomotiveSettingsActivity`
5. Navigate to Privacy Policy
6. Notice no overlap
7. Launch another app
8. Return to our app
9. Notice no overlap

## Screenshots or Screencast 
See: https://linear.app/a8c/issue/PCDROID-619/hyundai-prt-554-text-overlap-on-qr-screen-after-navigation#comment-6290defd

## Checklist

### DIFF
--- a/automotive/src/main/java/au/com/shiftyjelly/pocketcasts/AutomotiveLinkFragment.kt
+++ b/automotive/src/main/java/au/com/shiftyjelly/pocketcasts/AutomotiveLinkFragment.kt
@@ -2,7 +2,6 @@ package au.com.shiftyjelly.pocketcasts
 
 import android.os.Bundle
 import android.view.LayoutInflater
-import android.view.View
 import android.view.ViewGroup
 import androidx.compose.foundation.Image
 import androidx.compose.foundation.layout.Column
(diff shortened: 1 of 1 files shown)


geekygecko (approved):
I couldn't repeat this either. 

I'm not sure it's related, but it might be good to give the link some line height. 

AutomotiveLinkFragment.kt#L62
```
                lineHeight = 40.sp,
```

Before
Image

After
Image

PR #5478, opened 2026-06-25
Prevent endless transcript meta fetch/cancel loop in SummaryViewModel

geekygecko:
## Description

While an episode is playing, the AI summary feature was firing the transcript `…-meta.json` request in an endless fetch/cancel loop until the summary was loaded. In the logs there is a stream of `HTTP FAILED: java.io.IOException: Canceled` for `shownotes.pocketcasts.com/generated_transcripts/…-meta.json`.

Root cause: `PlayerContainerFragment` observes `listDataLive`, which is derived from `playbackStateObservable` and therefore re-emits on every playback position update (~once per second). Each emission calls `SummaryViewModel.loadSummary(episodeUuid)` with the same episode UUID. `loadSummary` only returned early for the `Loaded` and `Upsell` terminal states — while a load was still in flight (`Loading`), it cancelled the in-flight `loadJob` (and with it the OkHttp request) and restarted it. Because the next position update arrived before the request could settle into a terminal state, the request was cancelled and refetched indefinitely.

The fix adds an early return when a load for the same episode is already active, so position updates no longer interrupt an in-flight request. A genuine episode change still cancels and reloads as before, since the UUID won't match.

## Testing Instructions

1. Ensure the AI Summaries feature flag is enabled.
2. Play an episode of a podcast that has a generated transcript/summary, such as The Daily.
3. Open the full-screen player and watch Logcat (filter on `meta.json`).
4. Confirm the request is made once and completes, rather than repeatedly logging `HTTP FAILED: … Canceled` on every playback tick.
5. Skip to a different episode and confirm the summary reloads for the new episode.

## Screenshots 

Screenshot 2026-06-25 at 10 02 24 am

## Checklist

- [x] with different themes

### DIFF
--- a/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/viewmodel/SummaryViewModel.kt
+++ b/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/viewmodel/SummaryViewModel.kt
@@ -75,6 +75,7 @@ class SummaryViewModel @Inject constructor(
     fun loadSummary(episodeUuid: String) {
         if (currentEpisodeUuid == episodeUuid && _state.value is SummaryState.Loaded) return
         if (currentEpisodeUuid == episodeUuid && _state.value is SummaryState.Upsell) return
+        if (currentEpisodeUuid == episodeUuid && loadJob?.isActive == true) return
         currentEpisodeUuid = episodeUuid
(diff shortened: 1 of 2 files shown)


geekygecko:
@claude review this

sztomek (approved):
wow, nice catch! thanks for fixing it 🙏

PR #5480, opened 2026-06-25
[Hyundai] Keep automotive Now Playing card alive and replayable at end of playback

sztomek:
## Description
Our automotive partner at Hyundai has reported that the Now playing card gets cleared when the playback ends and the Autoplay next setting is Off on automotive.
This was a real issue that I had to address. Now I won't clear and clean up the media session, so the Now Playing card will show the last played media and give an option to replay it.

Fixes PCDROID-620 https://linear.app/a8c/issue/PCDROID-620/hyundai-prt-556-media-information-not-retained-after-playback-ends

## Testing Instructions
1. Disable Autoplay next setting
2. Play an episode, seek to close to end
3. Wait until playback completes
4. Press home
5. Verify you're still seeing the Now Playing 

## Screenshots or Screencast 

https://github.com/user-attachments/assets/6687112f-ae2e-44d8-849c-452023b86812

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 8.16
 -----
-
+*   Bug Fixes
+    *   Keep the Automotive Now Playing card alive and replayable when playback ends with Autoplay next off
+        ([#5480](https://github.com/Automattic/pocket-casts-android/pull/5480))
 
 8.15
 -----
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
@@ -1518,6 +1518,17 @@ open class PlaybackManager @Inject constructor(
                 LogBuffer.e(LogBuffer.TAG_PLAYBACK, "OnCompletion uuid does not match playback state current episode, ignoring onComplete event.")
                 return
             }
+
+            // Automotive: keep the Now Playing card alive and replayable at the end of playback.
+            if (Util.isAutomotive(application) &&
+                !hadSleepAfterEpisode &&
+                upNextQueue.queueEpisodes.isEmpty() &&
+                !settings.autoPlayNextEpisodeOnEmpty.value
+            ) {
+                pauseFinishedEpisodeForAutomotive(episode)
+                return
+            }
+
             eventHorizon.track(
                 PlayerEpisodeCompletedEvent(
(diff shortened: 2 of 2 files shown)


geekygecko:
@claude review this

sztomek:
Thanks for the thorough review @claude. I went through each point — here's how I addressed them (commit 775e0c7):

**Addressed**
- 🟠 **Completion analytics dropped** — good catch. The player genuinely finishes the episode, so `PlayerEpisodeCompletedEvent` should still fire. I now track it on the automotive path (inside the `isAutomotive` branch, so other platforms are unaffected). Sync/archive/status are still intentionally skipped (see below).
- 🟡 **CHANGELOG** — added a Bug Fixes entry under `8.16`.

**Intentional / out of scope**
- 🔴 **Episode left "unplayed"/unsynced/rewound** — this is deliberate and scoped to Automotive. The partner requirement is that the Now Playing card stays alive and the episode is instantly replayable from the start, so we keep it in the queue and rewind to 0 rather than completing/archiving it. Reworking cross-device completion sync is a separate product decision that I'm intentionally keeping out of this fix. Phone/Wear behavior is unchanged (the branch is gated by `Util.isAutomotive`).

**Not a defect**
- 🟠 **`player?.stop()` re-entrancy** — this matches the existing `stopPlayer()` pattern, which is also a bare `player?.stop()` (used by `loadCurrentEpisode(play = false)`). Calling `stop()` transitions ExoPlayer to `IDLE`, not `ENDED`, so it doesn't re-enter `onCompletion`. The `resettingPlayer` guard is specific to the player *rebuild* in `resetPlayer()` (the `play = true` path), not the stop.
- 🟠 **Pause/hibernate timer clearing the card** — the card lives exactly as long as any paused episode's does; `setupPauseTimer`/`PAUSE_TIMER_DELAY` hibernate applies equally to a normally-paused episode, so this isn't a new regression. Verified in the screencast.

**Skipped**
- 🟡 **Unit test** — there's no existing harness for `PlaybackManager`'s playback logic (26-arg constructor; `onCompletion` is private and fired through a private player callback gated on app metadata via `Util.isAutomotive`). A meaningful test would need the full dependency graph + a fake player + Robolectric metadata — disproportionate and brittle for this fix. Behavior is covered by the recorded screencast and the manual test steps. Happy to add one if the team feels the scaffolding is worth it.
- 🟢 **`getAppPlatform == Automotive` nit** — `Util.isAutomotive(application)` is already the idiom used elsewhere in this file (e.g. line 1814), so I kept it for local consistency.

sztomek:
@geekygecko closing this PR as we've agreed on we won't do it this way, but disable the controls when the playback ends.

PR #5488, opened 2026-06-29
[Hyundai] Reset player to empty state when playback completes

sztomek:
## Description
We have agreed to reset the player state on automotive after playback finishes and the Autoplay next setting is turned off.
This PR does that.

Fixes PCDROID-620 https://linear.app/a8c/issue/PCDROID-620/hyundai-prt-556-media-information-not-retained-after-playback-ends

## Testing Instructions
1. Disable Autoplay next
2. Start playing an episode
3. Seek close the its end
4. Wait until playback completes
5. Observe 

## Screenshots or Screencast 

https://github.com/user-attachments/assets/aa12d79e-5367-4e20-aa78-f40bede454f4

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,13 @@
         ([#5451](https://github.com/Automattic/pocket-casts-android/pull/5451))
     *   Wear OS performance improvements
         ([#5446](https://github.com/Automattic/pocket-casts-android/pull/5446))
+    *   Fix overlapping text issue on Automotive after orientation change
+        ([#5477](https://github.com/Automattic/pocket-casts-android/pull/5477))
+    *   Fix link screen's text overlap on Automotive
+        ([#5475](https://github.com/Automattic/pocket-casts-android/pull/5475))
+*   Updates
+    *   Reset player after playback finished on Automotive
+        ([#5488](https://github.com/Automattic/pocket-casts-android/pull/5488))
 
 8.14
 -----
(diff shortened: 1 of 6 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

sztomek:
@claude please review this

sztomek:
Thanks for the review! Addressed the actionable items in `303aa41fc0`.

### Addressed
- **Misleading `inactive` variable name (point 4)** — renamed the lambda param to `hasError` now that the mapping is just `it.isError`, so `setMedia3SessionAttached(!hasError)` reads correctly.
- **`resetToEmptyState()` untested (point 1)** — a full `MediaSessionManager` Robolectric test is disproportionate (it mostly orchestrates Media3 framework objects — player release/reassignment, which stays integration-level). Instead I added a unit test for the load-bearing, newly-added property reset depends on: that `swapPlayer()` propagates `isAutomotive`, so the swapped-in empty `SeedStatePlayer` actually renders the empty state (null `currentMediaItem` + stripped transport commands, `SET_MEDIA_ITEM` kept). That's the part that would silently break the feature.

### Not changing
- **Cold-start guard `if (wrappedPlayer is SeedStatePlayer) return` (point 2)** — this is intentional idempotency to avoid re-creating a `SeedStatePlayer` on repeated `data == null` emissions once we're already on the placeholder. The "seeded placeholder keeps showing controls after data==null" case can't actually arise: seeding only happens when there's a current episode, and `data == null` only fires for an empty playback state — so seeded ⟹ episode present ⟹ not empty ⟹ never null-data. An unseeded placeholder is already the empty state, so the early return is correct.
- **Transient command flicker at playback start (point 3)** — speculative (noted as "likely imperceptible") with no observed flicker. `installPlayer()` swaps in the real ExoPlayer with its timeline already populated, so the empty-state window doesn't materialise in practice. Not worth extra guarding.
- **CHANGELOG placement (point 5)** — this PR targets `release/8.15`, so the entry belongs in the 8.15 section where it is, and `* Updates` is an established category used throughout the changelog. The milestone due-date note is a separate Danger warning, unrelated to grouping.

geekygecko (approved):
That worked well in my tests.

Can we target `main` and cherry pick to the release branch in future.

PR #5059, opened 2026-03-02
[Player] Optimize battery usage 

sztomek:
## Description
This PR aims to achieve small wins related to battery usage.
It tries to offload decoding from the CPU to a low-energy DSP (Digital Signal Processor) when possible. Since DSPs are much more simpler than a CPU, it can only decode simple audio in its simplest form -- no effects. That means no trim silence, 1x playback speed, no volume boost. 
In sync with this, we enabled longer wakes when playing downloaded episodes, so the CPU won't wake up that often.
Also passing proper audio attributes to help the system decide the best course of action.

Fixes PCDROID-472

## Testing Instructions
Since it touches the player component which is part of our critical infrastructure, it would be best to test all platforms (wear, auto, phone)
Make sure that playback and effects are working properly
Make sure downloaded episodes play fine while the screen is locked
Make sure audio focus settings are respected (you can use Phil's p1770262733422059/1770262585.382239-slack-C028JAG44VD). When Settings>Notifications>Duck is set, steal focus (transparent, and exclusive) should steal focus until you abandon it)
You can check the logs (Profile -> Help -> Logs) to see "Audio offload preference" lines. When there are no effects enabled, it offloading should be ENABLED, otherwise DISABLED.

## Checklist
- [x] If this is a user-facing change, I have added an entry in CHANGELOG.md
- [x] Ensure the linter passes (`./gradlew spotlessApply` to automatically apply formatting/linting)
- [x] I have considered whether it makes sense to add tests for my changes
- [x] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml`
- [x] Any jetpack compose components I added or changed are covered by compose previews
- [x] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

PR #5491, opened 2026-06-29
Hold a wake lock during playback to reduce background playback stops

joashrajin:
## Description

Background playback of a long local **Files** episode can silently stop on its own (Pixel 6 / Android 16). The logs in PCDROID-555 show the system **SIGKILLing the Pocket Casts foreground-service process** mid-playback (`reason=2 SIGNALED status=9`, `importance=125`), with frequent `FREEZER`/cached-app activity on surrounding days — not an internal pause or audio-focus loss.

Root cause (closed gap): the background-audio `ExoPlayer` is built with **no wake mode set**, so it defaults to `WAKE_MODE_NONE` and holds **no wake lock during playback**. That's the missing standard configuration for background audio — without a wake lock the system is free to suspend the CPU / drop Wi-Fi and aggressively freeze/reclaim the process during long uninterrupted playback.

This sets a **source-aware wake mode** on the one background-audio player (`SimplePlayer`, shared by both the Media3 and legacy service paths): `WAKE_MODE_LOCAL` (partial CPU wake lock) for downloaded/local episodes, and `WAKE_MODE_NETWORK` (partial wake lock **+** Wi-Fi lock) for streams. `isStreaming` is known at build time, the `WAKE_LOCK` permission is already declared, and Media3 ties the lock to the playing state (it's released automatically on pause/stop), so there's no paused-but-locked battery drain.

> **Honest scope:** this is **best-effort hardening that closes a real, standard-config gap**, not a guaranteed cure. Because the process was still at foreground-service importance (`importance=125`) when killed, a wake lock **reduces** the system's freeze/kill eligibility and prevents screen-off CPU suspend, but cannot *prove* it eliminates every memory-pressure SIGKILL. It should be **verified on-device before PCDROID-555 is closed** (see Testing). Related to PCDROID-555 (intentionally not auto-closing).

> **History note for reviewers:** wake mode was previously added in #5059 and reverted in #5105 — but there it was **bundled with audio-offload** (`setAudioAttributes()` + `AudioOffloadPreferences` track selection) behind `Feature.AUDIO_OFFLOAD`, which defaulted to debug-only and so never shipped to production. Audio offload is the risky, device-specific part and the likely reason for that revert. **This change re-introduces only the `setWakeMode` line, decoupled from offload and audio attributes** — a strict, benign subset of what was reverted.

### Changes

- `SimplePlayer.prepare()` — add `.setWakeMode(if (isStreaming) C.WAKE_MODE_NETWORK else C.WAKE_MODE_LOCAL)` to the `ExoPlayer.Builder` chain, with an explanatory comment. No manifest, import, or permission changes needed (`C` already imported; `WAKE_LOCK` already declared).

## Testing Instructions

- Install this build on a device (preferably Pixel 6 / Android 16).
- Add a **long local Files episode** (~45 min), start playback, turn the screen off, and leave the app in the background.
- [ ] Confirm playback continues to the end without silently stopping (the original symptom). Re-test a few times, since the kill was intermittent.
- [ ] While playing with the screen off, confirm the app holds a partial wake lock — e.g. `adb shell dumpsys power | grep -i wake` shows a lock held by the package during playback.
- [ ] Pause for a while and confirm the wake lock is **released** (no lock held while paused) — i.e. no idle battery drain.
- [ ] Play a **streaming** (non-downloaded) episode and confirm a Wi-Fi lock is held during playback (`adb shell dumpsys wifi | grep -i lock`).
- [ ] After a long background session, check `adb shell dumpsys activity exit-info <pkg>` and confirm the `SIGKILL` during playback no longer recurs (or recurs less often). If it still recurs, the residual cause is likely memory pressure — see follow-ups below.
- [ ] Regression sanity: normal play/pause/skip/seek, lock screen controls, and Bluetooth/Android Auto playback behave as before.

_Not easily unit-testable: this is `ExoPlayer.Builder` configuration that the Media3 API doesn't expose for assertion. Behaviour is validated on-device. Follow-ups if the kill persists after verification: `onTrimMemory` handling in the playback service and moving episode-metadata decode off the player thread (tracked separately, lower confidence)._

## Screenshots or Screencast

n/a — no UI changes.

## Checklist
> Considered: this is non-asserting `ExoPlayer.Builder` config (Media3 doesn't expose wake mode for verification), so it's covered by on-device testing rather than a unit test.
- [ ] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml` — n/a (no strings)
- [ ] Any jetpack compose components I added or changed are covered by compose previews — n/a (no Compose changes)
- [ ] I have updated (or requested that someone edit) [the Event Horizon schema](https://github.com/Automattic/EventHorizonSchemas/blob/trunk/schema/pocket-casts.yml) to reflect any new or changed analytics — n/a (no analytics changes)

n/a — this PR contains no UI changes.


joashrajin:
## ✅ On-device verification

Built this branch as a debug APK and drove it through the full wake-lock matrix on hardware.

**Device: Samsung SM-G990E (Galaxy S21 FE), Android 16 / SDK 36** — chosen deliberately because it hits both risk surfaces at once: PCDROID-555 reproduced on Android 16, and the reverted #5105 regression was Samsung-specific.

Evidence captured via `dumpsys power` (wake locks), `dumpsys wifi` (Wi-Fi locks), and `dumpsys media_session` (playback state).

| # | Test | Expected | Result |
|---|------|----------|--------|
| 1 | Fresh install → onboarding → play | opens, no crash | ✅ launched clean |
| 2a | **Stream** episode → `dumpsys power` | `ExoPlayer:WakeLockManager` held | ✅ held |
| 2b | **Stream** → `dumpsys wifi` | `ExoPlayer:WifiLockManager` held | ✅ held (`type=4`, low-latency) |
| 2c | Stream reaches `PLAYING` | playback starts | ✅ **#5105 fresh-install regression does not recur** |
| 3 | Pause stream | both locks released | ✅ both gone |
| 4a | **Downloaded** episode → play | partial wake lock held | ✅ held |
| 4b | **Downloaded** → `dumpsys wifi` | **no** Wi-Fi lock | ✅ absent |
| 5 | Screen **off** during local playback | lock survives, audio continues | ✅ held, position advancing, `mWakefulness=Dozing` |
| 6 | Pause local | wake lock released | ✅ released |

**Source-aware branch confirmed live:** streaming took a partial wake lock **+** a Wi-Fi lock (`WAKE_MODE_NETWORK`); the downloaded file took **only** the partial wake lock (`WAKE_MODE_LOCAL`) — exactly the `if (isStreaming)` branch in `SimplePlayer.prepare()`.

Clean acquire/release lifecycle, no leaked locks:
```
ACQ ExoPlayer:WakeLockManager (stream play)  → REL (pause)
ACQ ExoPlayer:WakeLockManager (local play)   → REL (pause)
```

**Screen-off (Test 5)** is the direct PCDROID-555 scenario: with the screen off and the process backgrounded, the partial wake lock stays held and playback keeps advancing — the CPU-suspend path is closed.

### Not covered here (needs a long soak)
Two items from the Testing list require a multi-hour real-world session and were **not** exercised (I ran short episodes):
- Full ~45-min uninterrupted background playthrough without a silent stop.
- `SIGKILL`-during-playback no longer recurring in `dumpsys activity exit-info`. As the description notes, the kill happened at `importance=125`, so the wake lock reduces but cannot *prove* elimination of memory-pressure kills — that still wants on-device soak before PCDROID-555 is closed.

Everything mechanically assertable — lock acquisition, source-aware mode selection, release-on-pause, screen-off persistence, and absence of the #5105 regression on a Samsung — checks out.

PR #5492, opened 2026-06-29
[WearOS] Fix broken login with phone flow

sztomek:
## Description
We have received user reports on the WearOS up hanging up on the Log in with phone screen. The problem was that the watch only waited about 30 seconds for the phone to hand over the login info, but in real life it usually takes longer than that — you have to open the app on your phone, let it generate the login, and send it across — so the watch kept giving up too early. On top of that, the watch was checking whether a phone was nearby in a way that only worked over Bluetooth, so it would wrongly think no phone was connected even when one was reachable over Wi-Fi. 
To fix it, we made the watch simply wait patiently and keep listening for the phone's login info for as long as the screen is open, with no countdown to give up on. We also changed the starting screen so it shows the helpful "log in with your phone" instructions right away instead of an endless spinner. And we swapped in a smarter way of detecting whether a phone is actually within reach, so the "no phone connected" warning now only shows up when it's genuinely true and clears itself the moment a phone comes back. The result is that logging in from the phone now works reliably, even when the phone takes its time.

Fixes PCDROID-570 https://linear.app/a8c/issue/PCDROID-570/wear-os-login-resets-and-sync-with-phone-stalls

## Testing Instructions
1. Install the wear app on a watch
2. Make sure the phone app is also installed on your phone and you're logged in with a Plus account
3. Launch the wearOS app
4. If not signed in automatically, press Log in > Log in with phone
5. Verify you're being signed in

## Checklist


sztomek:
@claude review this please

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

sztomek:
Thanks for the review! Addressed the feedback in `ee0a7b66cf`.

### Addressed
- **Use `toUri()` instead of `Uri.parse`** (lint) — done in `PhoneConnectionMonitor`.
- **Cold-start flash of "no phone connected"** — `isPhoneConnected` is now `StateFlow<Boolean?>` seeded with `null` (unknown). The ViewModel only shows `NoPhoneConnection` once a real `false` arrives, so a reachable phone no longer flashes the error before the capability query resolves.
- **Edge-triggered reset hiding a real disconnection** — `restartSyncIfNeeded()`/`retrySync()` now re-check `isPhoneConnected.value` on reset, so the warning stays visible if the phone is genuinely still unreachable.
- **Dead `Timeout` / `Unknown` error states** — removed the unreachable error types, their UI branches/previews, and the now-unused strings.
- **Missing cold-start test** — added a test covering the `null` (unknown) → `false` path; mock flow updated to `Boolean?`.

### Not changing
- **`CapabilityClient` listener never removed** — intentional: it's an `@Singleton` on `ApplicationScope` that lives for the whole process and mirrors the existing `WatchPhoneCommunication.Watch` pattern, so there's nothing to clean up.
- **Throwaway `Failed(NoPhoneConnection)` allocation** — negligible; the collector only emits on connectivity changes, not in a hot path, so not worth the extra `is`-check noise.

sztomek:
@geekygecko I asked claude to fix the issue you experienced -- being stuck on the "Logged in frome phone!" screen.
Pushed a commit that is supposed to fix that case.  
Explanation: The watch's entire UI is gated behind signInState.isSignedInAsPlusOrPatron, but the auth token the phone hands to the watch contains no subscription tier — only a refresh token and identity. So after a token login the watch had to make its own network call to the server to discover it was Plus before it could leave the login screen. That single fetch was the weak point: when it stalled (common on a watch's proxied connection) the sign-in state never updated and the user was frozen on the green "Logged in from phone!" screen forever, and when it failed the error was silently turned into "no subscription," which the watch interpreted as a free account and signed the user out, bouncing them back to the landing screen on every launch. We fixed it by giving the fetch a bounded timeout plus a few retries and making it report failure distinctly from "confirmed free," so a transient network blip no longer demotes a paying user.  On the watch side we added a short watchdog that surfaces a retryable error instead of leaving the success screen hanging, and we stopped the activity from auto–signing-out while a phone login is still resolving. The net result: a Plus account now reliably reaches the watch list, and genuine failures show a retry rather than a freeze or a surprise logout.

Please take another look when you can 🙏

geekygecko:
> Please take another look when you can 🙏

I can still repeat the issue on my Pixel Watch 3, but it might be a different case. If the phone automatically picks up my Google account, then I sign out, and then tap "Log in on phone" it goes on the success page and it never disappears.

geekygecko (approved):

PR #5493, opened 2026-06-29
Fix conditions to fetch summary that includes chapters

sztomek:
## Description
The summary fetching logic (that actually pulls chapters as well) had a fatal flaw introduced by this commit: 1a70a0f85607414b793ae15d5543ad8e4e8b6d0a. If the `ai_summaries` feature flag is OFF, we won't fetch chapters at all. 
This PR addresses that and runs the fetch job either when `ai_summaries` or `generated_chapters` are on.
It worth noting that non-generated chapters were not affected by this plague.

Fixes PCDROID-637 https://linear.app/a8c/issue/PCDROID-637/bug-generated-chapters-are-not-fetched-when-ai-summaries-flag-is-off

## Testing Instructions
1. Build debugProd
2. Open developer settings
3. Turn off AI summaries (`ai_summaries` flag)
4. Open a podcast with generated chapters (like The Daily)
5. Open an episode, wait until data loads
6. Notice that Chapters tab is available on top
7. Verify chapter operations (tap to seek, preselect, etc) work
8. Now disable `generated_summaries`) as well
9. Find another episode with generated summaries and open it
10. Notice no Chapters tab
11. Find another podcast, this time with curated chapters (e.g. Changelog Master Feed)
12. Open an episode
13. Notice regular chapters are still dispalyed
14. Verify chapter operations work

## Screenshots or Screencast 

https://github.com/user-attachments/assets/c21d48fd-c32b-4f39-a2cc-a1078c3e130b

## Checklist

### DIFF
--- a/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/PlayerContainerFragment.kt
+++ b/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/PlayerContainerFragment.kt
@@ -253,7 +253,9 @@ class PlayerContainerFragment :
 
         viewModel.listDataLive.observe(viewLifecycleOwner) {
             adapter.updateNotes(addNotes = !it.podcastHeader.isUserEpisode)
-            if (FeatureFlag.isEnabled(Feature.AI_SUMMARIES) && !it.podcastHeader.isUserEpisode) {
+            val isSummaryOrChaptersEnabled =
+                FeatureFlag.isEnabled(Feature.AI_SUMMARIES) || FeatureFlag.isEnabled(Feature.GENERATED_CHAPTERS)
+            if (isSummaryOrChaptersEnabled && !it.podcastHeader.isUserEpisode) {
                 summaryViewModel.loadSummary(it.podcastHeader.episodeUuid)
             } else {
                 summaryViewModel.clearSummary()
--- a/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/episode/EpisodeFragmentViewModel.kt
+++ b/modules/features/podcasts/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/view/episode/EpisodeFragmentViewModel.kt
@@ -325,7 +325,7 @@ class EpisodeFragmentViewModel @Inject constructor(
                 _pageState.update { state ->
                     state.withSummary(result)
                 }
-                if (result != null) {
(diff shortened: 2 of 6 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

sztomek:
@claude please review this

sztomek:
Thanks for the review! Addressed the one substantive item in `87ce701543`.

### Addressed
- **#1 — chapters-only re-fetches `meta.json` on every emission** — the summary block sits downstream of the `combineLatest` that also emits on progress updates, so with `GENERATED_CHAPTERS` on / `AI_SUMMARIES` off the dedup key was never set and the meta fetch fired on every tick. Changed the guard to `if (result != null || !isSummaryEnabled)` so chapters-only episodes are marked processed. Kept it surgical — summary-mode semantics are unchanged (a null/blank summary with `AI_SUMMARIES` on still re-fetches, which is pre-existing and not this PR's regression), so no `TranscriptManager`/`SummaryViewModel` interface ripple.
  - No new VM test: `EpisodeFragmentViewModelTest` only covers tab-selection via `_pageState`; there's no harness for the `loadEpisode` Rx flow and counting fetches across emissions would be disproportionate/brittle. The flag matrix stays covered by `TranscriptManagerTest`.

### Not changing
- **#2 — rename `loadSummaryText` / `lastSummaryEpisodeUuid`** — a rename ripples through the `TranscriptManager` interface, `SummaryViewModel`, and ~30 test call sites for a cosmetic gain. Better as a standalone follow-up than bundled into a release-branch bugfix (you flagged it as non-blocking too).
- **#3 — `PlayerContainerFragment` has no per-episode dedup guard** — pre-existing and out of scope for this fetch-condition fix; that path goes through its own `SummaryViewModel`.

### Noted (housekeeping)
- **Branch target** — agreed it should also land in `main`; I'll make sure the commit is cherry-picked there per `AGENTS.md`.
- **Milestone past due** — leaving that to release management.

geekygecko (approved):
That worked but if the generated chapter is already downloaded then toggling the flag off doesn't seem to.

sztomek:
@geekygecko 
>That worked but if the generated chapter is already downloaded then toggling the flag off doesn't seem to.

Great finding, I'm going to address it.

PR #5494, opened 2026-06-30
Fix synced transcript auto-scroll for non-Latin transcripts (CJK, Thai, Hindi, Arabic, and more)

joashrajin:
## Description

The synced ("follow along") transcript never kept the highlighted line anchored for Japanese (CJK) transcripts: each time playback advanced to the next line, the transcript scrolled back to the **top**, so the user had to manually scroll down to find where playback was. It reproduced only with CJK text, not with Latin-text transcripts.

**Root cause:** the synced-transcript post-processing (`TranscripSanitization.joinSplitSentences()`) splits a transcript into per-sentence `TranscriptEntry` blocks using `endsAsSentence()` / `findMidSentence()`, whose punctuation lists were Latin-only (`. ! ? … - ) ] > } " ” ' ’`). Japanese sentences end with `。` `！` `？`, so no cue was ever treated as a sentence boundary and the whole transcript collapsed into a single `TranscriptEntry`. With one entry, `AutoScrollEffect` (which anchors the scroll on `highlightState.entryIndex`) had no per-line target: the index stayed at `0`, so the list stayed pinned to the top while only the highlighted word advanced.

**Changes:**
- Recognise CJK sentence terminators `。` `！` `？` (and half/full-width full stops `｡` `．`) so CJK transcripts split into per-sentence entries again, restoring per-line auto-scroll anchoring.
- Recognise CJK closing marks `」` `』` `）` as sentence-ending **only as terminal combos** (`。」`, `！」`, …), never bare — a bare closing quote also wraps a mid-sentence quoted term (`これは「AI」について話します。`), which must keep accumulating rather than split.
- Don't inject an ASCII space when joining adjacent CJK fragments, so a Japanese sentence spanning multiple word-timed cues renders without spurious gaps between characters (`これはペンです。`, not `これは ペンです。`).
- Single-source the CJK terminator/closer sets, and add unit tests for sentence/mid-sentence detection, the no-space join, terminal-combo splitting, the bare-closer case, mixed-script quoting, and CJK timing/word-offset preservation.

**Extended to other non-Latin scripts** (follow-up commits in this PR — the same root cause affected every script the punctuation list didn't know about):
- Recognise sentence terminators for other scripts with their own sentence-final punctuation: Arabic (`؟`), Urdu (`۔`), Devanagari/Hindi (`।` `॥`), Burmese (`။`), Khmer (`។` `៕`), Ethiopic (`።`), and Armenian (`։`).
- **Cap the phrase accumulator** for transcripts no terminator list can ever segment — Thai has *no* sentence-final punctuation, and unpunctuated auto-generated captions are common. Once the accumulated phrase outgrows a sentence-sized length (160 chars), it flushes at the next cue boundary, so every transcript yields multiple scrollable entries and the auto-scroll always has per-line anchors, regardless of script.
- Extend the no-space join rule from CJK to all space-less scripts (Thai, Lao, Khmer, Burmese), so their rejoined fragments render without injected gaps.

Pure-Latin transcripts are unaffected — the new punctuation is `endsWith`-matched and can't appear at the end of Latin text, and the accumulator cap only triggers where no sentence boundary was found for 160+ characters (punctuated text flushes on terminators long before that).

Fixes PCDROID-639

<details><summary>Out of scope / follow-ups (from review)</summary>

- **AI context window for CJK** — `TranscriptWindowExtractor.windowText()` gates on `split("\s+").size >= MIN_WORDS`; CJK has no whitespace, so AI bookmark/summary enrichment silently returns nothing for Japanese podcasts. Pre-existing, separate feature; not addressed here.
- **Nested quoted speech** — `findMidSentence()` prefers a terminator+quote combo over a later bare terminator, so `彼は「はい。」と答えた。それから帰った` splits at the inner `。」` rather than the true outer `。`. Leftover is preserved (degraded, not lost). Mirrors the existing Latin quotation-first heuristic.
- **More CJK closers** — only `」 』 ）` are covered; lenticular/angle brackets (`】 》 〉` etc.) used in some Japanese headings are not yet recognised as terminal-combo closers.
- **Generalisation** — sentence detection is a hand-maintained character list; `java.text.BreakIterator.getSentenceInstance(Locale)` could replace it locale-aware, but isn't a clean drop-in given the surrounding split/timing/offset logic.
- **U+3000** — `compactWhiteSpace()` doesn't normalise the ideographic space `　` (cosmetic, pre-existing).
</details>

## Testing Instructions

Automated:
- Run the transcript sanitization unit tests on **JDK 21**: `./gradlew :modules:services:repositories:testDebugUnitTest --tests "*TranscriptSanitizationTest*"` — 114 tests, 0 failures (includes the CJK cases, the new-script terminators, the Thai/unpunctuated cue-boundary flush, and the Thai/Devanagari spacing cases).

On device (see the before/after recordings below):
1. With `synced_transcripts` enabled and a Plus account, play an episode of a Japanese-language podcast that has a synced transcript — e.g. **EASY JAPANESE PODCAST** (podcast UUID `08b59600-4245-0139-3351-0acc26574db2`).
2. Open the transcript / "follow along" view and let it play without touching the screen.
3. Confirm:
   - [x] Each newly highlighted line auto-scrolls to stay in view (anchored ~30% from the top) instead of jumping to the top.
   - [x] The Japanese text renders without stray spaces between characters.

Latin transcripts are unaffected (covered by the existing Latin unit tests; the new punctuation can't match Latin text).

## Screenshots or Screencast

**Before:**

https://github.com/user-attachments/assets/1ab63aba-91cc-47f8-bab9-8fcbaa263063

**After:**

https://github.com/user-attachments/assets/a6cb449d-08f7-4c22-bd1a-058013317270

## Checklist

- [ ] with different themes


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

PR #5497, opened 2026-06-30
[HLS] Rework HLS stream url selection logic

sztomek:
## Description
Previously we had `BaseEpisode.hlsUrl` that was assigned after data > domain > database conversion. This way we exposed only the first alternate enclosure, preventing us to read the entire list. 
This PR gets rid of `hls_url` column from the database so we can work with the entire list as the server sends it.

Fixes PCDROID-608 https://linear.app/a8c/issue/PCDROID-608/update-api-parsing-logic

## Testing Instructions
1. Just make sure the databse migration tests run fine

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/AppDatabase.kt
```diff
@@ -1487,6 +1487,22 @@ abstract class AppDatabase : RoomDatabase() {
...
+            )
+            database.execSQL("DROP TABLE `podcast_episodes`")
+            database.execSQL("ALTER TABLE `podcast_episodes_new` RENAME TO `podcast_episodes`")
+            database.execSQL("CREATE INDEX IF NOT EXISTS `episode_last_download_attempt_date` ON `podcast_episodes` (`last_download_attempt_date`)")
+            database.execSQL("CREATE INDEX IF NOT EXISTS `episode_podcast_id` ON `podcast_episodes` (`podcast_id`)")
+            database.execSQL("CREATE INDEX IF NOT EXISTS `episode_published_date` ON `podcast_episodes` (`published_date`)")
+        }
```
geekygecko: This feels like a really dangerous migration to do as this table could be very large. How about we remove adding the `hls_url` column before it goes to production?
```
val MIGRATION_131_132 = addMigration(131, 132) { database ->
    database.execSQL("ALTER TABLE podcast_episodes ADD COLUMN hls_url TEXT") // delete this
}
```

sztomek: hmmm fair point, but we've already shipped database version 132 in our betas, so beta builds would break... as an alternative, we could keep the `hls_url` column untouched in the table but would not use its value in our domain layer.
what do you reckon? @geekygecko

geekygecko: > we've already shipped database version 132 in our betas, so beta builds would break...

If we leave the column for the beta users but remove the ALTER TABLE statement before we ship to production would there be an issue for either of them? I'm happy to do whatever you think but don't reckon we should do a full table copy.

sztomek: i think this would leave the app crashing for our beta users who installed the previous betas. claude has confirmed that as well, so i won't take this path.
as an alternative to full table migration, we should try dropping a column, but SQLite's `ALTER TABLE … DROP COLUMN` only exists since 3.35 and isn't available on older Android API levels, so the table-rebuild dance is the only portable way to actually remove a column.

i think the best course of action is to leave the column dormant and just get rid of the latest migration.
what are your thoughts, @geekygecko ?

geekygecko (approved):

PR #5515, opened 2026-07-01
[HLS] Implement analytics

sztomek:
## Description
This PR integrates the latest version of EventHorizon that contains the necessary analytics events and proprerties.
New events:
-  `playback_source_resolved` - fires on every episode open
- `playback_hls_toggled` - fires when show/hide video shelf action is tapped while playing a hls stream
Enriched properties of existing events:
- `playback_play` - `hls_available`, `audio_only_mode` are new ones
- `playback_episode_completed` - same as above
- `playback_failed` - podcast_uuid (if the episode is known),  source (where playback was triggered, content_type, playback_protocol (hls if the failing stream was HLS, else progressive), hls_error_detail = the Media3 error code (e.g. ERROR_CODE_IO_NETWORK_CONNECTION_FAILED) only for HLS failures; omitted otherise

Fixes PCDROId-640 https://linear.app/a8c/issue/PCDROID-640/analytics

## Testing Instructions
1. Build and install app
2. Make sure `hls_streaming` is enabled
3. Select a podcast that has hls streams like We Have A Meeting
4. Open one of their last 2 episodes and hit play
5. check logcat for related events, filter for `Event: playback_`

## Screenshots or Screencast 
SCR-20260701-qtlc

## Checklist
 
- [ ] with different themes


geekygecko (approved):

sztomek:
@copilot resolve the merge conflicts in this pull request

sztomek:
Thanks for the review! Went through the points in `9204b5dfc3` ("Address PR comments"):

**Addressed**
- **#1 (stale `source` on `playback_failed`)** — good catch. `lastPlaybackSource` is now also set in `loadCurrentEpisode()` when the episode is opened, so a failure that fires before `play()` (early buffering error, restore after process death) attributes to the episode actually being loaded rather than a stale previous one. `play()` still updates it so a later explicit play tap wins.
- **#4 (magic string `"exoplayer"`)** — extracted to a `HLS_ENGINE_EXOPLAYER` const.

**Not changing**
- **#2 (`findForEpisode()` now unconditional)** — intentional. `hls_available` must be reported regardless of the `HLS_STREAMING` flag, so the lookup has to run on every open. It's a single cheap suspend DB query.
- **#3 (`isFallback = false`)** — correct that it carries no signal yet; progressive fallback resolution isn't wired up. It'll become meaningful when that lands; leaving it as the primary-source default until then.
- **#5 (thread-safety of the new `var`s)** — acceptable for best-effort analytics; adding synchronization isn't worth the complexity here.
- **Tests** — skipping for now: these are private helpers on a manager with a large DI surface and no existing construction harness (current tests are narrow), and the logic is simple branching. Not worth standing up a harness for this change.

PR #5533, opened 2026-07-06
fix: Halt playback for the data warning and respect it on stream retry

joashrajin:
## Description

With **Warn before using data** ON and on a metered connection, two playback paths stream undownloaded episodes with no warning:

1. **Autoplay / Up Next auto-advance** — when the playing episode ends and the queue advances to a not-downloaded episode, it streams over mobile immediately.
2. **Missing-download fallback** — when a "downloaded" episode's file is gone (ENOENT), the player clears the download status and retries via stream with `forceStream = true`, bypassing the warning entirely.

**Root cause of path 1:** #5017 removed the `return` at the end of the data-warning branch in `PlaybackManager.loadCurrentEpisode()` (the deletion sits inside the `episodeObservable` if/else → `when` refactor hunk; the notification, its "Yes, keep playing" resume action, and the `EMPTY` state push all survived, so the halt was clearly still intended). The branch still posts the *"This episode is not downloaded, do you want to stream it?"* notification, but execution now falls through and `play()` starts anyway — since 8.10 the warning has been decorative for any non-forced play. This became much more visible after the setting became ON by default (PCDROID-534).

Changes (all in `PlaybackManager.kt`):

- Restore the `return` in the data-warning branch of `loadCurrentEpisode()`, so playback halts until the user taps **Yes, keep playing** on the notification (which resumes via the existing `playNow(forceStream = true)` receiver path) or **Play next downloaded**.
- The missing-download stream retry in `onPlayerError()` no longer passes `forceStream = true`, so the retry goes through the same warning check. Behavior on unmetered networks, or with the warning off, is unchanged.
- Skip the warning while **Android Auto is connected** (`Util.isAndroidAutoConnectedFlow`), because the confirmation notification is only visible on the phone — restoring the `return` without this would bring back the "playback silently stops in the car" behavior reported in PCDROID-397. The setting's description already says the warning doesn't apply to Android Auto.

Fixes PCDROID-647
Addresses PCDROID-397 (the halt-in-car symptom; keeping it open pending on-device Android Auto verification)

## Testing Instructions

Preconditions: metered/mobile connection, **Settings → Storage & data use → Warn before using data** = ON, Autoplay ON.

- Play a downloaded episode with a **not-downloaded** episode next in Up Next, and let it finish
- [ ] Verify playback stops when the queue advances and a "do you want to stream it?" notification appears (no mobile-data streaming starts)
- Tap **Yes, keep playing** on the notification
- [ ] Verify streaming starts
- Manually tap play on a not-downloaded episode
- [ ] Verify the streaming-warning dialog still appears and confirming plays as before
- For the missing-file path: download an episode, delete the audio file from `Android/data/au.com.shiftyjelly.pocketcasts/files/PocketCasts/podcasts/` (adb or file manager), then tap play on it while on mobile data
- [ ] Verify the log shows "Downloaded file missing … retrying via stream" and the data warning is shown instead of silent streaming
- On WiFi, repeat auto-advance and the missing-file retry
- [ ] Verify both stream immediately with no warning (unchanged behavior)
- Connect to Android Auto (head unit or DHU) on mobile data with the same queue setup
- [ ] Verify auto-advance keeps playing in the car with no halt (PCDROID-397)

## Screenshots or Screencast

> n/a

## Checklist
- [ ] If this is a user-facing change, I have added an entry in CHANGELOG.md *(added after PR creation, references the PR number)*
- [x] Ensure the linter passes (`./gradlew spotlessApply` to automatically apply formatting/linting) *(ran `spotlessCheck` — passed)*
- [x] I have considered whether it makes sense to add tests for my changes *(no unit test added: `loadCurrentEpisode`'s warning branch has no existing test harness and depends on player/network/notification state; behavior is covered by the manual steps above)*
- [ ] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml` — n/a, no new strings
- [ ] Any jetpack compose components I added or changed are covered by compose previews — n/a, no UI components changed
- [ ] I have updated (or requested that someone edit) [the Event Horizon schema](https://github.com/Automattic/EventHorizonSchemas/blob/trunk/schema/pocket-casts.yml) to reflect any new or changed analytics. — n/a, no analytics changes

n/a — no UI changes (existing notification/dialog behavior only)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
```diff
@@ -2049,7 +2053,9 @@ open class PlaybackManager @Inject constructor(
                 episode.uuid != lastWarnedPlayedEpisodeUuid &&
                 !Network.isUnmeteredConnection(application) &&
                 !forceStream &&
-                play
+                play &&
+                // Don't block playback while driving as the warning is only shown on the phone
+                withTimeoutOrNull(CAR_CONNECTION_TIMEOUT) { Util.isAndroidAutoConnectedFlow(application).first() } != true
```
geekygecko: The app crashed when auto playing an episode. 
```
java.lang.RuntimeException: Can't create handler inside thread Thread[DefaultDispatcher-worker-14,5,main] that has not called Looper.prepare()
	at android.os.Handler.<init>(Handler.java:253)
	at android.os.Handler.<init>(Handler.java:155)
	at android.content.AsyncQueryHandler.<init>(AsyncQueryHandler.java:125)
	at androidx.car.app.connection.CarConnectionTypeLiveData$CarConnectionQueryHandler.<init>(CarConnectionTypeLiveData.java:92)
	at androidx.car.app.connection.CarConnectionTypeLiveData.<init>(CarConnectionTypeLiveData.java:59)
	at androidx.car.app.connection.CarConnection.<init>(CarConnection.java:94)
	at au.com.shiftyjelly.pocketcasts.utils.Util.isAndroidAutoConnectedFlow(Util.kt:21)
	at au.com.shiftyjelly.pocketcasts.repositories.playback.PlaybackManager$loadCurrentEpisode$2.invokeSuspend(PlaybackManager.kt:2058)
```
I have merged some minor changes to fix this.

There's also a warning "Legacy Long overload can be converted to Duration" so I have changed `CAR_CONNECTION_TIMEOUT` to a Duration.

geekygecko (approved):
Thanks for making this fix.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
joashrajin: Thank you so much Phil 🙇

Issue #3543, opened 2025-02-04
Wireless Android Auto: Keeps pausing when bluetooth is enabled


joashrajin:
### Description

We have numerous users reporting that after some recent updates, Pocket Casts keeps pausing every few seconds when using wireless Android Auto. The issues do not occur when they switch Bluetooth off, so it sounds like the Bluetooth connection is sending some sort of command to pause playback. 

In our internal discussion at p1738325401040009/1738324786.080269-slack-C02A333D8LQ, it was mentioned that the issue is likely due to https://github.com/Automattic/pocket-casts-android/pull/3297

### Step-by-step reproduction instructions

1. Test using Pocket Casts on a car with Wireless Android Auto (unclear how to emulate)
2. Try with Bluetooth on vs off

### Did you search for existing bug reports?

- [x] I have searched for existing bug reports.

erricgunawan:
9374445-zen (1-star Android App review):

> I used to love this app. But for the last two weeks I have been unable to use it on Android Auto. It will play for a few seconds and then pause on its own. Press play and it plays for a few seconds and pauses again. It is unusable. I didn't change anything to cause it to do this. I tried everything "Google" said to no avail. None of my other audio apps have this problem. It sucks because I really like the interface. 🫤

Device: Samsung Galaxy S23 Ultra
App version code: 9306
App version name: 7.81
Android version: Android 14 (SDK 34)

danjjohnson:
9326880-zd-a8c

Android auto: 13.4.645054
App version : 7.80 (9303)
Phone: Sony - XQ-EC54 - XQ-EC54 - XQ-EC54
Android version: 15 SDK 35

Full logs and screen recording in ticket.

danjjohnson:
Another report in 9378713-zd-a8c

> I have just experienced the issue when listening to the latest Short Stuff episode at around 6.20am local time. Log attached as requested.

Wulfspade:
To clarify with this bug report, if you turn Bluetooth off while using wireless android auto you will have no audio, as such, it does not resolve the issue. This bug does not appear when using wired android auto where Bluetooth is utilized.

mmarcus:
[...]

PR #3297, opened 2024-11-27
Fix skip forward/back on some remote devices such as wired and wireless headsets

splendidbits:
## Description

Fixes #3292 (I'm not sure if this is the exact same issue that I experience, but this is more apt https://forums.pocketcasts.com/forums/topic/skip-and-rewind-with-headphone-controls/).

I have always had issues seeking forward and back using my USB C Apple headphones. It seems that some keycodes are different from those devices, and the extended path for these edge-cases do not support seeking forward and back. This also fixes sporadic play-pause issues for me when the app is in a background state.

## Testing Instructions
1. Start playing a podcast with wired USB C headphones (and possibly other remote devices)
2. Tap hardware play-pause button to play or pause 
3. Double-tap hardware play-pause button to skip forward
4. Triple-tap hardware play-pause button to skip backwards

## Checklist
- [x] If this is a user-facing change, I have added an entry in CHANGELOG.md
- [x] Ensure the linter passes (`./gradlew spotlessApply` to automatically apply formatting/linting)
- [x] I have considered whether it makes sense to add tests for my changes
- [x] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml`
- [x] Any jetpack compose components I added or changed are covered by compose previews
- [ ] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

PR #5535, opened 2026-07-07
Treat KEYCODE_MEDIA_PLAY as play-only to stop repeated Android Auto pauses

joashrajin:
## Description

Wireless Android Auto users report that playback repeatedly pauses a few seconds after pressing play (PCDROID-516, #3543). Logs show the head unit (`com.google.android.projection.gearhead`) sending redundant `KEYCODE_MEDIA_PLAY` key events while playback is already running. Since #3297 routed `KEYCODE_MEDIA_PLAY` into the same single-tap play/pause **toggle** as `KEYCODE_MEDIA_PLAY_PAUSE`, each redundant play event toggled playback into a pause — producing the play/pause loop.

This PR gives `KEYCODE_MEDIA_PLAY` its explicit play-only semantics: a single tap from it is ignored when playback is already running, in both `Media3SessionCallback` and the legacy `MediaSessionManager`. The key stays in the `MediaEventQueue` multi-tap system, so the headset double/triple-tap skip behavior introduced by #3297 (and the Pixel Buds spurious-event suppression) is unchanged, and `KEYCODE_MEDIA_PLAY_PAUSE` / `KEYCODE_HEADSETHOOK` still toggle as before.

This should fix the repeated-pause loop; it needs on-device verification with a wireless Android Auto head unit (see TODOs below).

Fixes PCDROID-516

## Testing Instructions

Unit tests:
- Run `./gradlew :modules:services:repositories:testDebugUnitTest --tests "*Media3SessionCallbackTest*"`
- [ ] Verify the new tests pass (`KEYCODE_MEDIA_PLAY is ignored while already playing`, `KEYCODE_MEDIA_PLAY_PAUSE still toggles while playing`)

On device:

> ⚠️ **TODO (manual):** Needs a car/head unit with wireless Android Auto to reproduce.

- Connect to a car with wireless Android Auto (Bluetooth enabled) and start playing an episode
- [ ] Verify playback no longer pauses on its own every few seconds
- Pause and resume via the steering wheel / head unit controls and the Android Auto screen
- [ ] Verify play, pause, and resume all still work
- With regular Bluetooth headphones, single/double/triple-tap the play button
- [ ] Verify play/pause, skip forward, and skip back still work per the headphone controls settings

## Screenshots or Screencast

n/a — no UI changes.

## Checklist
- [x] I have considered whether it makes sense to add tests for my changes — added 3 unit tests in `Media3SessionCallbackTest`
- [x] I have updated (or requested that someone edit) [the Event Horizon schema](https://github.com/Automattic/EventHorizonSchemas/blob/trunk/schema/pocket-casts.yml) to reflect any new or changed analytics — n/a, no analytics changes

n/a — no UI changes.


joashrajin:
Addressed the review feedback in 13a6806e4: extracted `PlaybackManager.playIfNotPlaying()` so the play-only guard is shared by both the Media3 and legacy callbacks (no more duplicated logic or double `isPlaying()` read), and added `PlaybackManagerPlayIfNotPlayingTest` covering the helper directly — so the legacy path now relies on tested shared logic rather than untested copy-pasted code. The CHANGELOG target (8.17) is confirmed correct: 8.16 is already at rc. Remaining gate before marking ready: on-device verification with a wireless Android Auto head unit.

joashrajin:
Superseded by #5645, which includes this PR’s explicit play-only handling and adds the immediate-resume fix for #5631 while preserving multi-tap handling and spurious-event suppression. Closing this PR so review can continue on the consolidated fix.

PR #2742, opened 2024-08-28
Improved full screen player Talkback

geekygecko:
## Description

After reading an [issue on the player accessibility](https://github.com/Automattic/pocket-casts-android/issues/2264), I checked our TalkBack. I found some issues that I have fixed in this PR.

The following items were missing a label
- The Up Next button
- The time remaining label
- The rearrange button

## Testing Instructions
1. Enable Talkback
2. Open the full screen player
3. Navigate through the items above

✅ Verify the labels are now spoken out

## Screenshots

| Before | After |
| --- | --- |
| Screenshot 2024-08-28 at 10 07 18 AM | Screenshot 2024-08-28 at 10 20 54 AM | 
| Screenshot 2024-08-28 at 10 07 02 AM | Screenshot 2024-08-28 at 10 21 07 AM | 
| Screenshot 2024-08-28 at 10 08 03 AM | Screenshot 2024-08-28 at 10 21 23 AM | 

## Checklist
- [x] If this is a user-facing change, I have added an entry in CHANGELOG.md
- [x] Ensure the linter passes (`./gradlew spotlessApply` to automatically apply formatting/linting)
- [x] I have considered whether it makes sense to add tests for my changes
- [x] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml`
- [x] for accessibility with TalkBack

Issue #2264, opened 2024-05-21
Accessibility: Android Talkback issues


joashrajin:
### Description

> The full player progress bar does not work with Android talkback. Swiping across the progress bar should move through the podcast but with each swipe progression leaps  back to the zero timepoint. So you can't quickly move through podcasts. Also on that page every time I touch the background I'm told  "icon view" and it gets really annoying. 
Another problem is the four buttons along the bottom of the screen which I think includes search and profile are not detected by the screen reader. So unless you have some vision you wouldn't know these buttons  were there.

### Step-by-step reproduction instructions

1. Enable Android Talkback as explained here: https://support.google.com/accessibility/android/answer/6007100?hl=en
2. Open Pocket Casts
3. Try to check if the progress bar works with talkback
4. Also check the buttons at the bottom of the page (Podcasts, Up Next, Filters, Discover, Profile)

### Did you search for existing bug reports?

- [X] I have searched for existing bug reports.

### Device, Operating system, and Pocket Casts app version

App version : 7.63 (9232)
Phone: Google - Pixel 3a - Pixel 3a - sargo
Android version: 12 SDK 32

PR #5547, opened 2026-07-09
Fix TalkBack seeking on the full player progress bar

joashrajin:
## Description

TalkBack users could move the full-player progress bar, but the position immediately snapped back and the seek was never committed (reported in #2264 as “progression leaps back to the zero timepoint”).

### Root cause

TalkBack `ACTION_SET_PROGRESS` actions and keyboard arrows update a native `SeekBar` through `onProgressChanged(fromUser = true)` without the touch-only `onStartTrackingTouch` / `onStopTrackingTouch` callbacks. The player seek bar therefore did not enter its seeking state or send the requested position to playback.

A deeper review also found that a simple delayed commit was not sufficient: an older async completion could clear a newer interaction, a delayed request could target the next episode, video reported completion before playback actually moved, detach could silently discard an accepted accessibility action, and seek analytics could race the playback-state mutation.

### What changed

- Model touch, TalkBack, and keyboard seeks as explicit, generation-scoped sessions.
- Debounce consecutive non-touch changes while capturing the target, duration bound, listener, and episode UUID for that session.
- Keep touch takeover paired as one start/stop interaction, and prevent stale gesture events after an episode change from targeting the replacement episode.
- Flush accepted non-touch actions on detach and explicitly cancel lifecycle-invalidated touch/in-flight sessions.
- Guard playback seeks against the current queue episode, playback state, and player instance; serialize UI seeks and make the latest request win.
- Deliver UI seek completions on the main thread and ignore stale completions without unlocking newer interactions.
- Wait for the real playback completion in full-screen video and prevent stale callbacks from restarting its hide-controls timer.
- Capture seek analytics from the validated episode before mutating its playback position.
- Preserve the legacy `seekIfPlayingToTimeMs` overload’s synchronous guard, callback ABI, and sync-caller behavior.
- Add focused Robolectric coverage for native accessibility actions, real touch events, detach/reattach, episode changes, stale completions, and video timer generations.

This benefits both consumers of `PlayerSeekBar`: the Compose full-screen player and the full-screen video player.

References #2264 — this resolves the **progress bar** item. The missing-labels item was already fixed in #2742, and the bottom-navigation item needs separate on-device confirmation, so this PR intentionally does **not** auto-close the issue.

## Testing Instructions

Automated validation:

```bash
./gradlew :modules:features:player:testDebugUnitTest
./gradlew :modules:features:player:lintRelease
./gradlew spotlessCheck
```

The new tests cover:

- repeated `ACTION_SET_PROGRESS` changes coalescing to the latest target
- stale async completion versus a newer seek session
- non-touch-to-touch takeover without duplicate callbacks
- episode changes during pending and active gestures
- detach flush plus detach/reattach accessibility input
- playback updates remaining blocked only until the matching completion
- stale video completion/cancellation timer behavior
- forwarding the expected episode UUID, position, and analytics source

The original accessibility path was also verified on-device (Samsung, Android SDK 34) by dispatching `AccessibilityAction.ACTION_SET_PROGRESS` at the live full-player seek bar: a seek from 695s to 995s held at the target and playback continued from there instead of snapping back.

Optional manual smoke test:

1. Enable TalkBack.
2. Play an episode and open the full-screen player.
3. Focus the progress bar and swipe up/down repeatedly.
   - [ ] The position accumulates and does not jump back.
   - [ ] Playback resumes from the final position after the debounce.
4. Repeat with a video episode in the full-screen video player.
   - [ ] The same behavior holds.
5. Disable TalkBack and drag the progress bar normally.
   - [ ] Touch dragging remains unchanged.
6. Optionally rotate or move to the next episode during a seek.
   - [ ] No stale seek is applied to the replacement episode.

## Screenshots or Screencast

n/a — no visual changes.

## Checklist

- [x] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml` — n/a, no new strings
- [x] Any Jetpack Compose components I added or changed are covered by Compose previews — n/a, no visual Compose change
- [x] I have updated (or requested that someone edit) the Event Horizon schema to reflect new or changed analytics — n/a, the existing seek event/schema is unchanged

- [ ] with different themes — n/a, no visual changes
- [ ] with a landscape orientation — covered structurally by detach/reattach tests; optional manual smoke above
- [ ] with the device set to have a large display and font size — n/a, no visual changes
- [x] for accessibility with TalkBack — live accessibility action verified on-device; native action path covered by Robolectric


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

joashrajin:
Thanks for the review — addressed both correctness points in da7fccfc7:

**1. `seeking` stuck `true` after detach.** `onDetachedFromWindow()` now resets `seeking` and `touchSeeking` (in addition to cancelling the pending commit), so a reused instance can't get stuck ignoring `setCurrentTime`/`setChapters` updates.

**2. Double `onSeekPositionChangeStart` when a touch drag interrupts an in-flight non-touch seek.** Start is now dispatched exactly once per seek session via a shared `beginSeekIfNeeded()` used by both `onStartTrackingTouch` and the non-touch path. If a drag begins while a TalkBack/keyboard seek is pending, the drag cancels the pending commit and `beginSeekIfNeeded()` no-ops (session already active), so the single `onStopTrackingTouch` keeps start/stop paired. This also orders start before the changing callback for non-touch seeks, addressing the ordering nit.

Left the state machine in the `View` (not extracted for unit testing) for this PR; compiles + `spotlessApply` clean. The primary fix was verified on-device via `ACTION_SET_PROGRESS`; these are defensive edge-case hardening not exercised by that path.

PR #5549, opened 2026-07-09
Improve podcast artwork caching

joashrajin:
## Description

Podcast covers disappear when the artwork CDN is unavailable, because nothing guarantees a subscribed podcast's artwork is on disk. This gives the phone image cache a usable floor, keeps subscribed artwork topped up, and stops failed loads from poisoning the cache.

Fixes [PCDROID-318](https://linear.app/a8c/issue/PCDROID-318/improve-podcast-artwork-caching)

## What changed

**Cache sizing** — `UiModule` keeps Coil's free-space-relative sizing but raises the share to 5% and the floor to 64MB, phone only. Coil's default is 2% clamped to `[10MB, 250MB]`; on a full phone that collapses to the 10MB floor. Note `maxSizeBytes()` is deliberately *not* used — it sets `maxSizePercent = 0`, pinning an absolute size regardless of free space, which on a device with 500MB free would let the cache claim half of it.

**`ArtworkCacheStrategy`** — a custom Coil `CacheStrategy` on the shared `ImageLoader`. Coil's `DefaultCacheStrategy` caches `{300, 301, 404, 405, 410, 414, 501}` and replays them from disk forever, so a cover that 404s once stays broken even after the server recovers. Error responses are no longer written, and any an earlier version already wrote are re-requested rather than replayed — otherwise upgrading users would keep their existing poisoned entries indefinitely.

**`PrefetchArtworkWorker`** — daily work that walks subscribed podcasts and fetches only the artwork missing from the disk cache. It never returns `Result.retry()`: the daily cadence *is* the retry, and because errors are no longer cached the UI itself heals covers as the user scrolls. Retrying within a run would re-walk the library every day for any podcast whose artwork is simply gone.

**`RefreshArtworkWorker`** — the Appearance settings button. It no longer wipes the disk cache before refetching. Each request uses `diskCachePolicy(WRITE_ONLY)`, so it always re-downloads and replaces a cover only once the new bytes arrive; a failed refresh leaves the old artwork in place. Previously, tapping it during a CDN outage cleared every cached image and then failed to refill any of them — the exact scenario this ticket is about. Enqueued as unique work so repeated taps can't stack concurrent passes.

Requests in all three paths decode at `1x1`. The original bytes still reach the disk cache — only the throwaway decode is sampled — which avoids decoding a full 960x960 bitmap per podcast.

## Trade-offs worth knowing

- **Three sizes per podcast.** The UI can request 200/480/960px, so all three are prefetched to be fully offline-safe. That is ~52KB per podcast across a cache shared with every other image in the app; a very large library on a nearly-full phone can thrash. The run summary now logs `Cache XMB of YMB` so that is visible in the support log.
- **Prefetch respects "Sync on metered networks".** With it off, a mobile-data-only user never gets a prefetch pass. Reusing a sync preference as consent for bulk image download is arguable — flagging it for a product call rather than deciding it here.
- **Permanently missing images are now re-requested rather than cached.** Correct, but it applies to every image the app loads, and `CleanAndRetryInterceptor` doubles those requests.
- **Scope is subscribed-podcast covers.** Episode-level and User Files artwork are not prefetched.
- **The grid already degrades gracefully.** Since #4973/#4976 a failed cover shows the podcast title on a UUID-hashed colour card, so this branch improves how often real artwork is present, not what happens when it is missing.

## Testing Instructions

1. Install `debugProd` and subscribe to several podcasts.
2. Clear only the image cache: `adb shell run-as au.com.shiftyjelly.pocketcasts.debug rm -rf cache/ImageCache`
3. Relaunch online and let `PrefetchArtworkWorkerPeriodic` run; the log should show `Prefetched N missing podcast artwork images (0 failed). Cache XMB of YMB.` and a second run should report `Prefetched 0`.
4. Force-stop, enable airplane mode, cold launch. Subscribed artwork should still be visible in the Podcasts grid and on podcast pages.
5. With the network up but the CDN unreachable, tap **Settings → Appearance → Refresh artwork**. Existing covers must remain visible — this is the case that previously wiped them.
6. Toggle **Settings → Advanced → Sync on Metered Networks** and confirm the existing worker's constraint switches between `CONNECTED` and `UNMETERED` without resetting its schedule.

## Checklist
- [x] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml` — no strings added
- [ ] Any jetpack compose components I added or changed are covered by compose previews — n/a
- [ ] I have updated (or requested that someone edit) [the Event Horizon schema](https://github.com/Automattic/EventHorizonSchemas/blob/trunk/schema/pocket-casts.yml) to reflect any new or changed analytics. — n/a

n/a — no layout or screen changes.
- [ ] with different themes


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

sztomek (approved):

PR #5550, opened 2026-07-10
Catch seek exception in SimplePlayer to prevent playback crash

geekygecko:
## Description

`SimplePlayer.handleSeekToTimeMs` calls `player.seekTo(...)` directly. In production this can throw an `IllegalArgumentException` originating deep inside Media3's `Player.PositionInfo` constructor (via `ExoPlayerImpl.getPositionInfo` → `updatePlaybackInfo` → `seekTo`) while ExoPlayer updates its internal playback state. This surfaced as a production crash.

This change wraps the seek in a `try/catch` and logs the failure to the playback log buffer instead of crashing, matching the existing defensive pattern already used elsewhere in this class (e.g. `handleStop`).

`onSeekComplete` is intentionally kept inside the `try`, so it is skipped when the seek throws. This is safe: nothing awaits that event, and `PlaybackManager` already updates the playback state position optimistically (`LastChangeFrom.OnUserSeeking`) before the seek runs, so the UI position still reflects the target.

[Sentry](https://a8c.sentry.io/issues/7601424614/?alert_rule_id=16493820&alert_timestamp=1783577967963&alert_type=email&environment=release&notification_uuid=97a93ac9-aa8b-4f2e-ab62-9d8c56b83340&project=4510465631911937) 

## Testing Instructions

1. Play any episode.
2. Scrub/seek the playback position multiple times, including near the very start and very end of the episode.
3. Verify playback continues to work and seeking behaves as expected.
4. Confirm no crash occurs. If a seek does fail internally, a `Failed to seek to <n> ms.` entry is written to the playback log instead of crashing the app.

## Checklist


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SimplePlayer.kt
```diff
@@ -135,8 +135,12 @@ class SimplePlayer(
...
             Toast.makeText(context, "Unable to seek. File headers appear to be invalid.", Toast.LENGTH_SHORT).show()
         } else {
-            player?.seekTo(positionMs.toLong())
-            super.onSeekComplete(positionMs)
+            try {
+                player?.seekTo(positionMs.toLong())
+                super.onSeekComplete(positionMs)
+            } catch (e: IllegalArgumentException) {
+                LogBuffer.e(LogBuffer.TAG_PLAYBACK, e, "Failed to seek to $positionMs ms.")
+            }
```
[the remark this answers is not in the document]
geekygecko: I'm going to leave 1. as it's not related to this change. I think leaving 2. is fine as well.

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SimplePlayer.kt
```diff
@@ -135,8 +135,13 @@ class SimplePlayer(
         if (player?.isCurrentMediaItemSeekable == false && player?.isPlaying == true) {
             Toast.makeText(context, "Unable to seek. File headers appear to be invalid.", Toast.LENGTH_SHORT).show()
         } else {
-            player?.seekTo(positionMs.toLong())
-            super.onSeekComplete(positionMs)
+            try {
+                player?.seekTo(positionMs.toLong())
+                super.onSeekComplete(positionMs)
+            } catch (e: IllegalArgumentException) {
```
sztomek: ❓ don't we want to catch anything here ?

geekygecko: Thanks for the feedback. This is a good suggestion and I have changed it.

geekygecko:
I have address the PR feedback and tested the change so I'm going to self approve this PR. I hope that's ok.

PR #5552, opened 2026-07-10
Reduce the chance of episode playback progress reverting during listening history sync

geekygecko:
## Description

When the listening history syncs from the server, the app updates each affected episode with its latest "last played" info. It was doing this by rewriting the episode's entire database record from a copy it had loaded a moment earlier.

The problem: if something else changed that same episode at the same time, most importantly the live playback position while the user was listening, or its download status, the history sync could overwrite those newer changes with its older copy. The visible symptom would be an episode's playback progress occasionally jumping back to an earlier point after a sync.

This PR makes the history sync update only the "last played" fields it actually needs to change, instead of rewriting the whole record. Everything else about the episode (your current position, download state, and so on) is left untouched, so concurrent changes are no longer clobbered.

## Testing Instructions

This change is a little hard to test and as the change isn't very big the unit test added might be enough. 

## Checklist
- [x] Any jetpack compose components I added or changed are covered by compose previews — n/a, no Compose changes
- [x] I have updated (or requested that someone edit) the Event Horizon schema to reflect any new or changed analytics — n/a, no analytics changes


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

sztomek (approved):

PR #5555, opened 2026-07-12
Set up a8c-secrets

mokagio:
Won't be used just yet. Just getting ahead with the work. See https://linear.app/a8c/issue/AINFRA-2685

### DIFF
--- a/.a8c-secrets/keys.pub
+++ b/.a8c-secrets/keys.pub
@@ -1,0 +1,4 @@
+# dev
+age1yxwzlhtnnlxx0djwdwjmy5j5j8th6ugljypcqymz28r0edhlmflqqhptjp
+# ci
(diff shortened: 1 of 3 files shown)


geekygecko (approved):

.a8c-secrets/keys.pub
```diff
@@ -0,0 +1,4 @@
+# dev
+age1yxwzlhtnnlxx0djwdwjmy5j5j8th6ugljypcqymz28r0edhlmflqqhptjp
+# ci
+age12uzrcyqlt8usvdzkyssyx66d2al9mfhmezpqct9ncxvu7lx0jv6selemnq
```
AliSoftware, line 4: Double-checked using `age-keygen -y` that this is indeed the public key corresponding to the private key we have in our CI secrets for the repo/pipeline of `pocket-casts-android` 👍

AliSoftware (approved):

PR #5572, opened 2026-07-14
[HLS] Integrate audio-only setting

sztomek:
## Description
This PR follows up with recent developments on other platforms -- the audio-only setting. It lives under General > Player section and controls whether we enable  the video track (when available) or not. Opting in for audio-only has a positive impact on data consumption as we won't bother with video tracks of the HLS stream.
This is a global setting, affecting all podcast episodes.
Fullscreen player's shelf also respects this and won't render the Show/Hide video action in audio-only mode.

Fixes PCDROID-669 https://linear.app/a8c/issue/PCDROID-669/add-audio-only-setting

## Testing Instructions
1. Open settings > general and notice new audio only setting
2. Turn it on
3. Start playing an HLS episode (E.g. The confidence chromicles)
4. Notice it's audio-only
5. Open shelf on fullscreen player
6. Notice no option to show/hide video
7. Turn off audio-only
8. Resume to fullscreen player
9. Notice video starts playing
10. Open shelf
11. Notice hide video option available

## Screenshots or Screencast 

https://github.com/user-attachments/assets/dfdb493c-c9f5-45da-bb33-c1d494bea50b

## Checklist


sztomek:
Thanks for the review 🙏 Went through each point:

**✅ Addressed in `Address PR comments`:**
- **Android Lint / `UnsafeOptInUsageError`** — the new `trackSelector` field and `applyAudioOnly()` touch media3 `@UnstableApi`, which was failing the Lint check. Added the opt-in (field marker + `@OptIn` on the method, matching `prepare()`). Lint is green locally now.
- **CHANGELOG** — added an entry under `8.17 → New Features`. Good catch, the checklist was ticked prematurely.

**⏭️ Intentionally not addressing here (justified):**
- **Analytics event** — agreed it *should* be tracked for consistency with the other toggles. It can't land in this PR though: these events are code-generated from the EventHorizon schema and filtered by `includedPlatforms`, so the tracked event has to exist in the schema first (that's exactly why the schema checklist item is unticked). Will follow up with a schema PR + the `track(...)` call once it's available, rather than block this behavioural change on it.
- **Thread consistency for `updateAudioOnly()`** — `DefaultTrackSelector.setParameters` is documented as safe to call from any thread (you noted this too), and `updateAudioOnly()` is a non-`suspend` interface method by design. Routing it through `withContext(Dispatchers.Main)` would mean making the interface method suspending purely for symmetry, with no correctness gain — not worth the churn.
- **Summary string scope** — intentional. HLS adaptive streams are the only video source the app renders today, so "episodes with video" is accurate for the current surface; the copy is deliberately source-agnostic so it still reads correctly if/when other video sources are added.
- **Unit tests** — `SimplePlayer` wraps `ExoPlayer` directly and isn't unit-testable without heavy fakes, and the `PlaybackManager` branch is a trivial 3-way `when`. The behaviour is covered by the manual test steps + screencast in the description. Happy to add targeted tests if you feel strongly.

geekygecko (commented):
Did this get past design approval?

sztomek:
@geekygecko asked for design approval here: p1784280532991529-slack-C0B9J1M3DQX, in the meantime would you mind reviewing the PR?

geekygecko (approved):

PR #5587, opened 2026-07-15
[Synced Transcripts] Fix broken seek failed event tracking

sztomek:
## Description
It's been reported that the `synced_transcripts_seek_failed` event was never reported due to a bug in the code.
After further investifagation, it turned out that both the UI (`TranscriptPage`) and `seekToTranscriptEntry` gated on `isSyncedActive` -- a state that already requires fingerprint syncing to be working, so the failure path was unreachable. The fix introduced a looser `isTapToSeekAvailable` gate (generated text transcript, no paywall, matching playing episode) for enabling the tap handler, and only attempts the fingerprint time mapping when sync is actually active,so when sync isn't active or the mapping returns null, the seek-failed event now fires.
Also added tests for proper coverage

## Testing Instructions
 1. Build and install the debug app
  2. Confirm the SYNCED_TRANSCRIPTS feature flag is on
  3. Start streaming (not downloaded) an episode that has a Pocket Casts-generated transcript, using a Plus account.
  4. Open the full-screen player and tap the Transcript tab.
  5. Immediately (within the first seconds, while fingerprint sync is still preparing) tap any transcript line.
  6. Verify a "streaming unavailable" toast appears and no seek happens.
  7. In logcat, filter for synced_transcripts_seek_failed and verify the event fired with synced_state=preparing (or idle).

## Screenshots or Screencast 

SCR-20260715-kkae

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/features/transcripts/src/main/kotlin/au/com/shiftyjelly/pocketcasts/transcripts/TranscriptViewModel.kt
+++ b/modules/features/transcripts/src/main/kotlin/au/com/shiftyjelly/pocketcasts/transcripts/TranscriptViewModel.kt
@@ -178,18 +178,22 @@ class TranscriptViewModel @AssistedInject constructor(
 
     /**
      * Seeks playback to the tapped transcript [entry]. Returns the playback position (ms) sought
-     * to, or `null` if the entry is untimed, syncing is inactive, or no mapping is available. The
-     * caller uses the returned position to drive the highlight directly rather than re-deriving it
-     * from the (lossy) playback position.
+     * to, or `null` if the entry is untimed, tap-to-seek is unavailable, or no mapping is
+     * available (which tracks a seek failure). The caller uses the returned position to drive the
+     * highlight directly rather than re-deriving it from the (lossy) playback position.
      */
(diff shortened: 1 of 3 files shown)


geekygecko (approved):
The event and toast worked well. I'm not sure about the toast as it does work once it has loaded. Does iOS do this?

sztomek:
@geekygecko 

iOS does exactly this. In `TranscriptViewController.transcriptTapped`, when the fingerprint mapping isn't ready it fires `syncedTranscriptsSeekFailed` and then shows the same `transcriptTapToSeekStreamingUnavailable` toast for non-downloaded episodes. It only stays silent in two cases: when the synced state is `.unavailable`, or when the episode is already downloaded (so the mapping will never help). So the "works once it's loaded, toast while it's still preparing" behavior is intentional and matches iOS
Toast + seek-failed logic: [link](https://github.com/Automattic/pocket-casts-ios/blob/2b644cdd2b5731f7b9deac38e4cc5d12db57f3aa/podcasts/TranscriptViewController.swift#L1020-L1032)

PR #5589, opened 2026-07-15
Reduce network traffic from streaming fingerprint decoding

sztomek:
## Description
  Fingerprinting a streamed episode decodes the enclosure as a second full download. This PR makes that spend deliberate on metered networks:

  - Preparation now carries a trigger (`PLAYBACK` / `TRANSCRIPT_VIEW` / `BOOKMARK`). On metered data, streamed episodes only fingerprint when the user is looking at the transcript and hasn't enabled "warn about data usage"; otherwise preparation ends `Unavailable(metered_network)` and a `NetworkCallback` retries automatically once an unmetered network appears.
  - That consent expires when the transcript is dismissed and is re-elevated whenever the transcript surface re-enters composition. Without this, a rotation would leave consent downgraded while the user keeps reading.
  - A running remote decode re-checks the network every 5 s and stops itself on a Wi-Fi→cellular handoff; restart paths re-evaluate the policy per attempt, so the handoff also stops future restarts.
  - A seek landing well inside the mapped range cancels a running streaming decode instead of letting it keep pulling enclosure bytes it no longer needs.
  - Before the first playback-progress tick (e.g. preparing while paused) the decode throttles against its own start position, so it can no longer free-run through the whole enclosure.
  - Reference fingerprints for streamed episodes are cached in `cacheDir` (LRU, 20 entries), so re-preparing an episode doesn't re-download them.
  - Fingerprinting is now phone-only, and ChapterManager logs once per episode when generated chapters run unaligned without a mapping their auto-skip boundaries may be offset by dynamically inserted ads.
  

  ## Testing Instructions
  1. Enable `SYNCED_TRANSCRIPTS`, mark the Wi-Fi network as metered (or use cellular), and stream an episode without opening the transcript: no fingerprint enclosure download should start.
  2. Open the transcript,  fingerprinting should start (with "warn about data usage" off) or stay unavailable (with it on).
  3. Close the transcript, seek around, no fingerprint stream restarts on metered data.
  4. Rotate the device while reading the transcript,  sync should survive the rotation.
  5. Start a decode on Wi-Fi, then switch to cellular mid-stream, the decode should stop within ~5 s; switch back to Wi-Fi and it should recover.
  6. Stream the same episode twice,  the second preparation should reuse the cached reference (no reference re-download).

  ## Checklist
  - [ ] I have updated (or requested that someone edit) [the Event Horizon schema](https://github.com/Automattic/EventHorizonSchemas/blob/trunk/schema/pocket-casts.yml) to reflect any new or changed
  analytics.

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/fingerprint/FingerprintReferenceRetriever.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/fingerprint/FingerprintReferenceRetriever.kt
@@ -1,6 +1,8 @@
 package au.com.shiftyjelly.pocketcasts.repositories.fingerprint
 
+import android.content.Context
 import au.com.shiftyjelly.pocketcasts.servers.di.NoCache
+import dagger.hilt.android.qualifiers.ApplicationContext
 import java.io.ByteArrayOutputStream
 import java.io.File
 import java.io.IOException
@@ -26,6 +28,7 @@ import timber.log.Timber
 @Singleton
 class FingerprintReferenceRetriever @Inject constructor(
     @NoCache private val okHttpClient: OkHttpClient,
+    @ApplicationContext private val context: Context,
 ) {
     private val inFlightRequests = mutableMapOf<String, Deferred<ByteArray?>>()
     private val requestMutex = Mutex()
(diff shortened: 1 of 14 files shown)


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/fingerprint/FingerprintReferenceRetriever.kt
```diff
@@ -122,8 +125,38 @@ class FingerprintReferenceRetriever @Inject constructor(
...
+    }
+
+    private fun cachedReferenceFile(episodeUuid: String) = File(File(context.cacheDir, CACHE_DIR_NAME), "$episodeUuid.ref.fp.json")
+
+    private fun pruneCachedReferences() {
+        val files = File(context.cacheDir, CACHE_DIR_NAME).listFiles() ?: return
+        files.sortedByDescending { it.lastModified() }
+            .drop(REFERENCE_CACHE_MAX_FILES)
+            .forEach { it.delete() }
+    }
```
geekygecko, line 154: It would be good to not roll our file cache and instead use an approach like ShowNotesServiceManager.kt with the OkHttp cache if possible.

sztomek:
Thanks for the thorough pass 🙏 Quick rundown of what I did / didn't action:

- **#6 (test gap)** — good catch, added a `decideOnProgress` case for a jump landing in a mapped run that *isn't* far ahead while streaming + decode-active (expects `None`). Covers the `mappedFarAhead == false` boundary.
- **#1 (consent lifecycle)** — left as-is; you traced it as correct and the "confirming test" would need to drive the `ConnectivityManager`/`NetworkCallback` integration end-to-end, which is more machinery than the already-verified behavior warrants.
- **#3 (cache prune not synchronized)** — skipped. Prepare is single-flighted per episode upstream, and concurrent cross-episode prunes are benign (idempotent deletes, self-healing cache), so a mutex here is contention for a race that can't really bite.
- **#5 (recheck holds the mutex over the `ConnectivityManager` call)** — leaving it; it's once per 5s as you noted.
- **Nits** — the trailing `Unit` in `saveCachedReference` is load-bearing (keeps the return type `Unit` rather than `Result<Unit>`), and fully-qualifying `android.net.Network` in the one callback override reads clearer to me than adding an alias that collides with the `Network` util. Kept both.

#2 and #4 were already flagged as fine. 👍

geekygecko (approved):

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/fingerprint/FingerprintReferenceRetriever.kt
sztomek: Looked into this and I'm going to keep the manual cache for now. The OkHttp cache only stores a response when the origin sends caching signals (`Cache-Control`/`Expires`, or `ETag`/`Last-Modified`). The references come from `shownotes.pocketcasts.com/generated_transcripts/…`, and that host returns no caching headers — Cloudflare reports `cf-cache-status: DYNAMIC`/`BYPASS`, so nothing would get cached and every prepare would re-download the reference (exactly the metered traffic this PR is removing).

The `ShowNotesServiceManager` path works because it rides `SERVER_CACHE_URL` (the pcache server, built to return cacheable responses); the fingerprint host doesn't. To make OkHttp cache these I'd have to add a response-rewriting interceptor to inject a synthetic `Cache-Control` plus a dedicated `Cache` instance — more moving parts than the current ~30-line unconditional LRU, and it'd need re-testing. The manual cache is also a better fit here since references are immutable, so keeping them offline indefinitely is correct. Happy to revisit if the origin starts sending cache headers.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>CacheBackedMediaDataSource</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>

</p>

PR #5600, opened 2026-07-17
Homogenize milestone error handling across the release lanes

AliSoftware:
Closes AINFRA-2707

- 🧵 Thread in Slack: p1784280315812929/1784270734.514819-slack-CC7L49W13
- 🔗 Companion PR in pocket-casts-ios: https://github.com/Automattic/pocket-casts-ios/pull/4778

## Why?

Milestone manipulation across the release lanes was handled inconsistently:
 - `finalize_release` called `set_milestone_frozen_marker` + `close_milestone` with no error handling at all, so re-running the lane (when the milestone is already closed or missing) would crash it
 - `finalize_hotfix_release` only logged via `UI.important`; and `code_freeze` only logged via `UI.error` without a Buildkite annotation.

This PR makes all of them behave the same way.

## How?

It extracts a `report_milestone_error(error_title:)` helper — mirroring the convention already used in the repos for our other apps — which logs a descriptive warning and, on CI, adds a `warning` Buildkite annotation explaining that the error is expected on a re-run.

Every milestone rescue path (`code_freeze`, `finalize_release`, and `finalize_hotfix_release`) now routes through that helper, and the previously-missing `begin/rescue` around `finalize_release` is added.

Note: I also took the occasion to standardize our use of `is_ci?` to `is_ci` (both the one with or without `?` are valid methods in fastlane and alias to one another, so the idea was just to be consistent across our `Fastfile`)

## Testing Instructions

This only affects release automation, so it can't be exercised from a normal build. To sanity-check the behavior, on a release branch run `bundle exec fastlane finalize_release` (or `finalize_hotfix_release`) a second time, or against a version whose milestone is already closed, and confirm the lane now logs the "Error closing the milestone" warning (plus a `warning` Buildkite annotation on CI) and completes instead of aborting.

## Checklist
- [x] If this is a user-facing change, I have added an entry in CHANGELOG.md (Not user-facing — release automation only.)
- [x] Ensure the linter passes (`./gradlew spotlessApply` to automatically apply formatting/linting) (Ruby `Fastfile` change; `ruby -c` passes and the `Fastfile` is excluded from the RuboCop length cops.)
- [x] I have considered whether it makes sense to add tests for my changes (Fastlane lanes; not unit-testable in this repo.)
- [x] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml` (N/A)
- [x] Any jetpack compose components I added or changed are covered by compose previews (N/A)
- [x] I have updated (or requested that someone edit) [the Event Horizon schema](https://github.com/Automattic/EventHorizonSchemas/blob/trunk/schema/pocket-casts.yml) to reflect any new or changed analytics. (No analytics changes.)

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -400,10 +400,10 @@ platform :android do
         api_token: ENV.fetch('GITHUB_TOKEN')
       )
 
-      buildkite_annotate(context: 'translations-pr-url', message: "Translations PR: #{pr_url || 'Error creating PR'}") if is_ci?
+      buildkite_annotate(context: 'translations-pr-url', message: "Translations PR: #{pr_url || 'Error creating PR'}") if is_ci
     else
       UI.important('No translations were downloaded, so no PR was created')
-      buildkite_annotate(context: 'translations-pr-url', message: 'No translations were downloaded, so no PR was created') if is_ci?
+      buildkite_annotate(context: 'translations-pr-url', message: 'No translations were downloaded, so no PR was created') if is_ci
       Fastlane::Helper::GitHelper.checkout_and_pull(base_branch)
       Fastlane::Helper::GitHelper.delete_local_branch_if_exists!(pr_branch)
     end
@@ -508,7 +508,7 @@ platform :android do
         RELEASE_VERSION: release_version_current
       }
     }
-    if is_ci?
+    if is_ci
       buildkite_pipeline_upload(**pipeline_args)
     else
       buildkite_trigger_build(
@@ -990,6 +990,6 @@ platform :android do
     MESSAGE
 
     UI.error(error_message)
-    buildkite_annotate(style: 'warning', context: 'error-with-milestone', message: error_message) if is_ci?
+    buildkite_annotate(style: 'warning', context: 'error-with-milestone', message: error_message) if is_ci
   end
 end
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -404,10 +400,10 @@ platform :android do
         api_token: ENV.fetch('GITHUB_TOKEN')
       )
 
-      buildkite_annotate(context: 'translations-pr-url', message: "Translations PR: #{pr_url || 'Error creating PR'}") if is_ci?
+      buildkite_annotate(context: 'translations-pr-url', message: "Translations PR: #{pr_url || 'Error creating PR'}") if is_ci
```
mokagio, line 403: To be honest, I like when methods that return a boolean have the `?` suffix, in the same way as methods that raise have the `!` suffix. But consistencies across repos is key (and we can always change in a dedicated set of PRs across repos in the future if the feeling is shared)

mokagio (approved):

mokagio:
Taking the liberty of merging this so that it's available for the upcoming 8.17 release, due to code freeze in a few hours.

fastlane/Fastfile
AliSoftware: Yeah, I like that convention a lot too, and tbh I initially asked Claude to use `is_ci?` everywhere. And it's only when it showed me the summary stats at the of the task that I realized that there were initially more `is_ci` instances than `is_ci?` on this repo, and apparently same for other repos, so why I ended up asking it (a bit relunctantly ^^) to switch back the `is_ci?` standardization we just did on this repo to still do a standardization, but using `is_ci` instead.

PR #5605, opened 2026-07-18
Use Ruby 3.4.9 and Bundler 4.0.16

mokagio:
Just a routine, and long overdue, tooling upgrade. If CI is green, we should be good.

See https://linear.app/a8c/issue/AINFRA-2705/use-ruby-349-in-pocket-casts-android


geekygecko (approved):
Thanks for doing this.

PR #5609, opened 2026-07-18
Fix duplicate custom app icons

### LINKED ISSUE #1252: Change the custom App Icon implementation
### Description

At the moment, our custom icon implementation results in 2 app icons being shown. This has resulted in confusion and users uninstalling the app by mistake. 

It would be great if we could investigate how other apps which support custom app icons like DuckDuckGo and ToDoList have implemented this. 

Note: The JP Mobile team looked into this too; you can view their comments here: https://github.com/wordpress-mobile/WordPress-Android/issues/14476#issuecomment-1451884262

1. Go to Appearance and set a custom App Icon
2. Notice you will now see two app icons for Pocket Casts 

image

Samsung Z Fold 4, Android 13, Pocket Casts v7.43

joashrajin:
## Description

Pocket Casts currently exposes both MainActivity and the selected activity alias as launcher entries. On affected launchers, especially Samsung One UI, selecting a custom icon therefore creates a second app-drawer icon.

This changes the launcher architecture so MainActivity remains enabled for notifications, sharing, deep links, media intents, shortcuts, and explicit launches, while every launcher icon, including the default icon, is represented by an activity alias. Exactly one alias is enabled at a time.

The change also:
- migrates existing installations to a new default alias after an app update
- switches aliases atomically on Android 13 and newer
- enables the replacement before disabling the previous alias on older Android versions
- keeps static shortcuts and Samsung multi-window launcher categories on every alias
- updates the icon confirmation copy and adds regression tests

Fixes #1252

## Testing Instructions

1. Install an existing Pocket Casts build and select a custom app icon.
2. Install this branch over that build without clearing app data.
3. Confirm only the selected icon appears in the all-apps screen.
4. Launch Pocket Casts from the selected icon.
5. Switch from the custom icon to another custom icon, then back to the default icon.
6. After each switch, confirm exactly one Pocket Casts entry appears in the all-apps screen.
7. Verify a notification tap, app shortcut, shared OPML file, Pocket Casts deep link, and media search intent still open the app.
8. Verify launching from recents and Samsung multi-window still works.

Automated checks completed:
- focused AppIcon unit tests on API 32 and API 33
- debug, debugProd, prototype, and release manifest merging
- debug and debugProd APK assembly
- app lint
- Spotless

> [!NOTE]
> **One UI launcher cache:** after switching icons, Samsung's launcher may keep showing app-drawer entries for previously enabled aliases for a while, even though PackageManager reports exactly one enabled launcher component (`adb shell cmd package query-activities -a android.intent.action.MAIN -c android.intent.category.LAUNCHER --brief | grep pocketcasts`). These stale entries disappear once the launcher refreshes (restarting the launcher or rebooting forces it). Testers should use the `query-activities` output as ground truth rather than the drawer, so launcher cache isn't mistaken for a regression.

Verified on a Galaxy S21 FE (Android 16 / One UI 8): duplicate icons reproduced on the pre-fix build, then a dirty upgrade to this branch migrated to exactly one enabled alias (custom selection preserved), with notification launches, launch-intent resolution, static shortcuts, and icon switching all confirmed working. Android Auto is unaffected: it surfaces the app via the `MediaBrowserService` declarations and `com.google.android.gms.car.application` meta-data, which this change doesn't touch.

## Screenshots or Screencast

Captured on a Galaxy S21 FE (Android 16 / One UI 8), same app-drawer page, debug build. Before: pre-fix build with the Dark icon selected shows two entries (default + Dark). After: dirty upgrade to this branch, the migration receiver leaves only the selected Dark icon.

**Before (pre-fix build, Dark icon selected, two entries):**

before-duplicate-icons

**After (this branch installed over it, one entry, selection preserved):**

after-single-icon

## Checklist
- [x] I have considered whether it makes sense to add tests for my changes — added `AppIconTest` covering the atomic API 33+ switch, the pre-13 enable-before-disable ordering, and the default-icon case
- [x] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml` — the reworded dialog message uses a new key (`settings_app_icon_updated_message_v2`) so stale translations of the old copy don't shadow it
- [ ] Any jetpack compose components I added or changed are covered by compose previews — n/a, no Compose components changed
- [ ] I have updated (or requested that someone edit) [the Event Horizon schema](https://github.com/Automattic/EventHorizonSchemas/blob/trunk/schema/pocket-casts.yml) to reflect any new or changed analytics. — n/a, no analytics added or changed (the existing `SettingsAppearanceAppIconChangedEvent` is untouched)

n/a — this PR changes launcher/alias behavior and one dialog message; there are no layout or screen changes. Launcher behavior was verified on-device (Galaxy S21 FE, One UI 8): duplicate reproduced pre-fix, dirty upgrade migrated to a single entry with the user's selection preserved, and notification launches, launch-intent resolution, static shortcuts, and icon switching all confirmed working.
- [ ] with different themes


sztomek:
Reviewed this and verified it end-to-end on a Pixel 6 (Android 16 / API 36). The approach is right and it does what it says — I've rebased it onto `main` and pushed a few small additions.

## Verified on device

**Before** (pre-fix build), after selecting Dark:

```
au.com.shiftyjelly.pocketcasts.debug/…ui.MainActivity
au.com.shiftyjelly.pocketcasts.debug/…ui.MainActivity_1
```

Two drawer entries — #1252 reproduced.

**After** (this branch): exactly one entry, and installing over a pre-fix install that had Dark selected left only `MainActivity_1` — the user's choice survived the migration.

The detail I want to call out because it's invisible and easy to "tidy" later: renaming `MainActivity_0` → `MainActivityDefault` is load-bearing. Users on the old build carry a persisted `COMPONENT_ENABLED_STATE_DISABLED` override for `MainActivity_0` (the old code disabled it even for the default icon). Reusing that name would leave them with the default alias disabled and, since `MainActivity` no longer carries `LAUNCHER`, **zero** launcher icons. A fresh component name has no override and falls back to `android:enabled="true"`. Worth a comment in the code so nobody reverts the name.

## Pushed on top

- `Move the changelog entry to 8.22` — it was under 8.19, which shipped a while ago. This was also the only merge conflict.
- `Check every app icon has an activity alias` — an `androidTest` walking `AppIcon.AppIconType` against the merged manifest. The 19 aliases have to stay in lockstep with an enum in a *different* module, nothing failed the build on drift, and `setComponentEnabledSetting` on a missing component throws at runtime. I confirmed the test isn't vacuous: pointed at a non-existent alias it fails with `Missing activity alias au.com.shiftyjelly.pocketcasts.ui.MainActivityDefault_MISSING`.
- `Test the default icon on the atomic path` — DEFAULT was only covered on API 32, and API 33+ is what most devices run.
- `Remove a stray blank line in the manifest`.

Branch is rebased on current `main`; `spotlessApply`, unit tests and the new instrumentation test are green (Pixel 6 and a Google TV Streamer).

## On the review bot's suggestions

**Keep `AppIconUpdateReceiver` — I'd reject the suggestion to delete it.** The argument was that `setupApp()` already syncs before `onReceive`, making the receiver a redundant second PackageManager batch. That inverts cause and effect: the broadcast is what starts the process at all. Without it nothing runs until the user next opens the app, and an upgrading user with a custom icon sees exactly the two icons this PR removes. I saw the migration work on a dirty upgrade even though the app then crashed for an unrelated reason, because `enableSelectedAlias` runs first inside `runBlocking`. The KDoc is accurate as written.

The "unreachable `one` plural forms" style notes elsewhere are also safe to ignore — those forms are reached for 21, 101, … in Russian and Polish.

The enum/manifest drift comment was a good one, hence the test above.

## Home screen and shortcuts: checked, no problem

I expected pinned/home-screen shortcuts to break, since they bind to a component that now gets disabled. They don't, at least on Pixel Launcher:

- A home-screen icon survived both the upgrade and two icon switches, re-skinning itself each time (radioactive → Dark → Round Light).
- All four shortcuts (3 static + the dynamic `top_filter`) re-bound to the newly enabled alias with `disabledReason=[Not disabled]`.

So no shortcut re-publishing is needed. One caveat: I could only test Pixel Launcher, and #1252 was reported on One UI — your S21 FE testing covers that side. Given the shortcut doesn't actually disappear here, it may be worth double-checking whether the new copy ("If your Home screen shortcut disappears, add Pocket Casts again…") describes something you actually saw on One UI.

## One pre-existing bug found along the way — not for this PR

Changing the app icon closes the app and leaks the confirmation dialog:

```
E WindowManager: android.view.WindowLeaked: Activity …ui.MainActivity has leaked window
    …[MainActivity_1] that was originally added here
    at androidx.appcompat.app.AlertDialog$Builder.show(AlertDialog.java:1007)
    at …AppearanceSettingsFragment.onViewCreated$lambda$2$1(AppearanceSettingsFragment.kt:108)
```

`AppearanceSettingsFragment.kt:102-108` flips the alias and *then* shows the dialog. The running task was launched through the alias being disabled, so the system tears it down — the dialog never becomes readable and the user lands on the home screen.

**This is not caused by this PR.** I reproduced it identically on `main`: launch via `am start -n …/MainActivity_1`, switch Dark → Round Light, same bounce, same leak at the same line. What changes here is coverage — today a default-icon user is exempt because their task root is `MainActivity`, which is never disabled; after this PR every launch goes through an alias, so it applies to everyone rather than only to people already on a custom icon.

Worth its own ticket. Two options when someone picks it up:

1. Persist the choice, show the dialog, apply `enableSelectedAlias` from the OK button — small, fixes the leak, makes the close deliberate rather than crash-looking.
2. Defer the switch until the app backgrounds (a `ProcessLifecycleOwner` observer in `PocketCastsApplication`, which already injects `AppIcon`; `setupApp()` already re-applies on cold start). Needs checking whether the backgrounded task survives the alias being disabled, plus a product call on when the icon visibly updates.

sztomek (approved):

PR #5613, opened 2026-07-20
Resolve generated chapter seeks through on-demand audio fingerprinting

sztomek:
## Description

Generated chapter timestamps live on the reference (ad-free) timeline, so seeking to `chapter.startTime` lands early whenever dynamically inserted ads shift the stream. This PR routes generated-chapter seeks through a bounded on-demand fingerprint resolve, adapting the approach that landed on iOS (pocket-casts-ios#4728) while keeping Android's existing strengths.

How a seek resolves, in order:
1. **Session cache** — a chapter already resolved this session seeks instantly.
2. **Dense mapping fast path** — if the continuous fingerprint map already brackets the chapter start with anchors ≤8s apart, interpolate directly (no decode).
3. **Bounded on-demand resolve** — decode and fingerprint only a small search window: cold `[t, t+240s]`, or warm `[max(t, est−180s), est+120s]` when the continuous map provides an estimate. Requires ≥2 anchors, hard 5s timeout, last-tap-wins supersession, and the scratch mapping never touches the continuous map.
4. **Raw-seek fallback** — anything unresolved falls back to the chapter's own start time (current behavior).

While a resolve is in flight, the chapter row shows a small spinner and the player's next/previous chapter button shows the same indicator.

As a quick recap, these are the areas where the android implementation is stronger than the iOS:
| Strength | What it buys |
  |---|---|
  | Eager whole-file pass for **downloaded** episodes | Accurate chapter-list timestamps + instant, correct seeks anywhere — even past the 240s cold-window limit. Costs only CPU on downloads |
  | Reactive whole-list chapter realignment | The chapter list shows true playback times as the map grows; iOS shows raw reference times until a chapter is tapped |
  | `densePlaybackTime` fast path | Skips the bounded decode entirely when the continuous map already brackets the chapter — instant seek, no spinner |
  | Proactive prep on playback start (+ bookmarks) | Mapping is often warm before the user ever opens a transcript or taps a chapter; iOS only starts when the transcript opens |
  | Metered-network gating + unmetered retry | Data-saving safeguard iOS has none of |
  | Cache-following decode source | Resolves can succeed in regions the player hasn't buffered; iOS falls back to a raw seek there |
  | Richer restart heuristics | Debounced restarts, bootstrap cooldown, run-end continuation — fewer wasted decode restarts on scrubbing |

Fixes PCDROID-671 https://linear.app/a8c/issue/PCDROID-671/check-on-the-latest-ios-implementation-changes

## Testing Instructions

1. Debug build with `SYNCED_TRANSCRIPTS` and `GENERATED_CHAPTERS` enabled.
2. Stream (don't download) an episode with generated chapters and dynamic ads.
3. Open the chapters list and tap a chapter well ahead of the playhead — the row shows a spinner, then playback lands on the chapter content (not early into an ad); if resolution fails, the seek falls back to the raw chapter time within 5s.
4. Use the next/previous chapter buttons in the full player — the tapped button shows the spinner during the resolve.
5. Download the same episode and reopen it — the chapter list re-aligns via the eager pass and taps seek instantly (unchanged behavior).
6. Verify transcript highlight and tap-to-seek still behave as before.

## Screenshots or Screencast

https://github.com/user-attachments/assets/085c2f3c-3107-42a4-849f-8c174e53dca1

## Checklist

- [ ] with different themes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
         ([#5536](https://github.com/Automattic/pocket-casts-android/pull/5536))
     *   HLS support
         ([#5602](https://github.com/Automattic/pocket-casts-android/pull/5602))
+    *   Resolve generated chapter seeks through on-demand audio fingerprinting
+        ([#5577](https://github.com/Automattic/pocket-casts-android/pull/5577))
 *   Bug Fixes
     *   Fix episodes being cached over mobile data when Warn before using data is enabled
         ([#5567](https://github.com/Automattic/pocket-casts-android/pull/5567))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
     *   HLS support
         ([#5602](https://github.com/Automattic/pocket-casts-android/pull/5602))
     *   Resolve generated chapter seeks through on-demand audio fingerprinting
-        ([#5577](https://github.com/Automattic/pocket-casts-android/pull/5577))
+        ([#5613](https://github.com/Automattic/pocket-casts-android/pull/5613))
 *   Bug Fixes
     *   Fix episodes being cached over mobile data when Warn before using data is enabled
         ([#5567](https://github.com/Automattic/pocket-casts-android/pull/5567))
--- a/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/chapters/ChapterRow.kt
+++ b/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/chapters/ChapterRow.kt
@@ -19,6 +19,7 @@ import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.layout.widthIn
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material.Checkbox
+import androidx.compose.material.CircularProgressIndicator
 import androidx.compose.material.Icon
 import androidx.compose.material.IconButton
 import androidx.compose.runtime.Composable
@@ -56,6 +57,7 @@ fun ChapterRow(
     onClick: () -> Unit,
     onUrlClick: () -> Unit,
     modifier: Modifier = Modifier,
+    isResolving: Boolean = false,
 ) {
     val chapter = state.chapter
     var selectedState by remember(chapter.index) { mutableStateOf(state.chapter.selected) }
@@ -100,12 +102,22 @@ fun ChapterRow(
                 )
                 Spacer(Modifier.width(8.dp))
             }
-            TextH50(
-                text = (chapter.uiIndex).toString(),
-                color = textColor,
-                modifier = Modifier
-                    .padding(horizontal = 12.dp),
-            )
+            if (isResolving) {
+                CircularProgressIndicator(
+                    color = textColor,
+                    strokeWidth = 2.dp,
+                    modifier = Modifier
+                        .padding(horizontal = 12.dp)
+                        .size(16.dp),
+                )
+            } else {
+                TextH50(
+                    text = (chapter.uiIndex).toString(),
+                    color = textColor,
+                    modifier = Modifier
+                        .padding(horizontal = 12.dp),
+                )
+            }
             Spacer(Modifier.width(8.dp))
             TextH50(
                 text = chapter.title,
--- a/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/chapters/ChaptersFragment.kt
+++ b/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/chapters/ChaptersFragment.kt
@@ -56,6 +56,7 @@ class ChaptersFragment : BaseFragment() {
         savedInstanceState: Bundle?,
     ) = contentWithoutConsumedInsets {
         val state by viewModel.uiState.collectAsState()
+        val resolvingChapterIndex by viewModel.resolvingChapterIndex.collectAsState()
 
         AppThemeWithBackground(mode.themeType) {
             ChaptersTheme(state.podcast) {
@@ -73,6 +74,7 @@ class ChaptersFragment : BaseFragment() {
                     onSkipChaptersClick = viewModel::enableTogglingOrUpsell,
                     onSelectionChange = viewModel::selectChapter,
                     onUrlClick = ::openChapterUrl,
+                    resolvingChapterIndex = resolvingChapterIndex,
                     modifier = Modifier.nestedScroll(rememberNestedScrollInteropConnection()),
                 )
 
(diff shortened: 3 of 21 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

sztomek:
Addressed the review in bb090ee2ef:

- **Metered-network gap (🟠)**: added `shouldBlockOnDemandResolve` as an upfront gate in `performResolve` — a streaming resolve on a metered network now returns `Unresolved(metered_network)` (raw-seek fallback) when `warnOnMeteredNetwork` is on, mirroring the transcript-view policy for user-initiated actions. Covered by 4 new unit tests.
- **`_resolvingChapter` race (🟡)**: the publish now happens inside the same `withLock` as the `activeCaller` swap, and cleanup is ownership-based under `NonCancellable` — a superseded tap can no longer wipe or flicker the new tap's spinner (including the same-chapter double-tap case the old `compareAndSet` missed). New regression test.
- **Streaming reference cache (🟡)**: `loadOrFetchReferenceData` now checks `loadCachedReference` and saves via `saveCachedReference` for streaming episodes, mirroring the prepare path.
- **Stale `computeEager` KDoc (🔵)**: removed the leftover `[isUnmetered]` sentence.

Not changed, deliberately:
- **Two concurrent codecs**: inherent to the isolated on-demand path (iOS has the same shape with its dedicated `onDemandQueue`); the resolve is bounded to ≤5s.
- **Indicator scoping to the playing episode**: resolves can only fire through `PlaybackManager.seekToChapterStart` for the current episode, so the scoping matches reality.
- **`referenceStartTime` idempotency**: chapters are always re-derived from raw DB rows before aligning, so a double-align can't occur.

geekygecko (approved):
It's great to show the users the loading state.

> Session cache — a chapter already resolved this session seeks instantly.

Is this caching working correctly? Here's an example video of me using a downloaded episode and jumping between the same chapters. It still seems to load each time. Is that the expected behaviour?

https://github.com/user-attachments/assets/86c07a60-f4d9-409e-98b1-18a1dc75f1ba

sztomek:
@geekygecko 

> Is this caching working correctly? Here's an example video of me using a downloaded episode and jumping between the same chapters. It still seems to load each time. Is that the expected behaviour?

Good catch! I went through the video frame by frame and I'm pretty sure the cache itself is fine, it just never got anything to store. Every revisit lands exactly on the raw chapter start with a full spinner, which means the resolve was returning `Unresolved `every single time on your device. And since we only cached **successful** resolves, every tap re-ran the whole thing from scratch. That's the loading you kept seeing.
I've pushed a fix that remembers failures for the session too: missing reference = no more resolve attempts for that episode (taps just fall back instantly, no spinner), no-match is remembered per chapter. Transient stuff (timeout, offline, metered) still retries.
Can you please verify that you were testing on `prodDebug` builds?

sztomek:
>One thing worth a second look: REASON_NO_REFERENCE conflates permanent and transient failures 🟡
The new memoization treats a no_reference result as a permanent, session-long block for the whole episode (GeneratedChapterSeeker.kt:104 sets referenceUnavailable = true, and :79 then short-circuits every future chapter tap). That's correct for a genuinely-absent reference, but performResolve returns REASON_NO_REFERENCE whenever loadOrFetchReferenceData yields null (FingerprintTimingManager.kt:359) — and that null collapses two very different cases:
Permanent: the server returns 404/403 (FingerprintReferenceRetriever.kt:80-83) — no reference exists. Remembering it is right.
Transient: IOException/offline or 5xx after exhausting MAX_RETRIES (:102-109, :86-88) — also returns null.
So if a user taps a chapter while briefly offline (or the CDN 5xxes through the retry budget), the on-demand resolver is disabled for that episode for the rest of the session, even after connectivity returns — which is the opposite of the "transient stuff still retries" intent stated in the thread. REASON_TIMEOUT/REASON_AUDIO_UNAVAILABLE/REASON_METERED_NETWORK are correctly not remembered; it's only this no_reference path that leaks a transient failure into a permanent block.
Mitigations that make this narrow (so this is a 🟡, not a blocker):
fetchReferenceData already retries 3× with backoff internally, so a momentary blip is usually absorbed before it ever surfaces as null.
The referenceUnavailable short-circuit sits after the dense-map fast path (:76 before :79), so the continuous/eager map can still serve these chapters as it grows — nicely locked in by the new dense mapping still serves a chapter that previously failed test.
It resets on episode change.
The clean fix would be to distinguish "reference absent (404/403)" from "reference fetch failed (transient)" — e.g. a separate reason like REASON_REFERENCE_UNAVAILABLE for the transient branch that the seeker treats like REASON_TIMEOUT (retryable), reserving REASON_NO_REFERENCE for the 404/403 case. Not required for merge, but it would make the "still retries" guarantee actually hold when the failure was network-caused. [Fix this →](https://claude.ai/code?q=In%20FingerprintReferenceRetriever.performFetch%20and%20FingerprintTimingManager.performResolve%20on%20branch%20feat%2Fchapters-fingerprint-improvements%2C%20distinguish%20a%20genuinely-absent%20reference%20%28HTTP%20404%2F403%29%20from%20a%20transient%20fetch%20failure%20%28IOException%2Foffline%20or%205xx%20after%20MAX_RETRIES%29.%20Return%20REASON_NO_REFERENCE%20only%20for%20404%2F403%20and%20a%20new%20retryable%20reason%20%28e.g.%20REASON_REFERENCE_UNAVAILABLE%29%20otherwise%2C%20and%20in%20GeneratedChapterSeeker%20do%20NOT%20set%20referenceUnavailable%20for%20the%20transient%20reason%20so%20it%20retries%20like%20REASON_TIMEOUT.%20Add%20a%20unit%20test.&repo=Automattic/pocket-casts-android)
A weaker version of the same point applies to REASON_NO_MATCH on a streaming episode: if the search window only partially decoded because the region wasn't buffered yet, "too few anchors" can be a transient state rather than a true no-match. It's better mitigated than the reference case (the dense path still recovers it), so I'd only flag it as something to keep in mind.

This is already fixed in the follow-up PR (#5614). There the fetch result distinguishes a real 404 from a network error, and transient failures get their own reference_fetch_failed reason which the seeker retries. I'd rather not backport that split here since #5614 rewrites this exact code and it would just create rebase conflicts. Until that lands, the risk is small: the fetch already retries 3 times with backoff before giving up, the dense map can still serve those chapters as it grows, and the block resets when the episode changes.

geekygecko:
> Can you please verify that you were testing on prodDebug builds?

Yes I only use the `debugProd`  build variant for testing. Especially as in staging, only The Daily has transcripts.

PR #5614, opened 2026-07-20
Refine fingerprinting mechanism

sztomek:
## Description
This branch is a hardening pass on the on-demand chapter resolver compared to the parent, and the gains are pretty concrete. 
**Data**: a streaming chapter tap used to decode the full search window (~4MB at 128kbps) every time — now it stops as soon as it finds the target, so a typical tap costs whatever the ad offset is (usually well under 1MB), and repeat taps near a resolved region cost nothing because the anchors get merged back into the shared map. 
**Latency**: warm taps skip re-downloading and re-parsing the multi-MB reference file entirely, and with "cache entire episode" on, far-ahead chapter taps went from a guaranteed 5-second timeout + wrong landing spot to actually working. 
**Reliability**: timeouts now salvage partial results instead of discarding them, a flaky reference fetch is retryable instead of killing sync for the whole episode, and a resolve can never yank playback after you've seeked elsewhere. 
**UX**: skip buttons stay tappable so double-tapping advances two chapters, and sparse transcript taps land on real content instead of extrapolated guesses (or mid-ad).
In summary, performance and reliability should improve thanks to these optimisations.

## Testing Instructions
Smoke test the following areas:
- seeking in audio while playing
- chapters: skip to chapter to start playback, skip to chapter during playback
- transcripts: monitor transcripts highlighting stays in sync as you're seeking in the audio, scroll on transcripts and hit a paragraph

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/nowplaying/EpisodeTitles.kt
+++ b/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/nowplaying/EpisodeTitles.kt
@@ -126,7 +126,7 @@ private fun Content(
         ) {
             if (state.isChaptersPresent) {
                 ChapterPreviousButton(
-                    enabled = !state.isFirstChapter && !isResolvingPrevious,
+                    enabled = !state.isFirstChapter,
                     alpha = if (state.isFirstChapter) 0.5f else 1f,
                     isResolving = isResolvingPrevious,
                     iconTint = playerColors.contrast01,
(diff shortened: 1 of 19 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

geekygecko (approved):
I'm not sure if it's related to this pull request, but there are occasions where the next chapter button doesn't work correctly and it keeps jumping to the same chapter. I was using this episode https://pca.st/dyp8hzul

https://github.com/user-attachments/assets/6ae25b4a-bfe8-472b-a199-0a9dd6a822e8

sztomek:
@geekygecko 

 Great find, it was related to this PR. From the video: next picks the target chapter by its displayed (aligned) start, but the resolved seek time for that chapter can land a few seconds before that boundary (the boundary keeps shifting as the continuous mapping refines, while the session-cached resolve is frozen). Landing short leaves you in the previous chapter, so the next button re-targets the same chapter indefinitely. Fixed by clamping chapter-seek resolves to the chapter's own displayed start, so a skip can never land short of the boundary it was picked from — pushed in the latest commit.

PR #2380, opened 2024-06-19
Cache playing episode 

ashiagr:
## Description

This is a bare-bone implementation of caching the entire playing episode to minimize streaming while playing and utilizing the LRU cache eviction to save disk space.

I'm raising this PR early to gather feedback on whether we should proceed in this direction given [this ExoPlayer limitation](https://rb.gy/uf5w2z) of using a single location for downloading and caching. To be safe, I'm enabling it only from advanced settings as an experimental feature.

P.S. Cache size depends on `settings.getExoPlayerCacheEntirePlayingEpisodeSizeInMB()`  which can be configured remotely.

#### Test.1 Caching playing episode - happy path

1. Install the app
2. Go to `Profile` -> `Settings` -> `Advanced`
3. Enable `Cache playing episode` setting
4. Play an episode which is not downloaded earlier while being connected to wifi
5. Look for `CacheWorker: Caching complete for`  in the logs
6. ✅  Notice that playback continues without any noticeable glitch in switching from streaming to cached location
7. ✅  Notice that episode is cached in device explorer at `/data/data/au.com.shiftyjelly.pocketcasts.debug/cache/pocketcasts-exoplayer-cache`
8. Disable internet
9. Seek to another position
10. ✅ Notice that playback continues from the new seek position
11. Restart the app while being disconnected from internet
12. ✅ Play the episode in the player
13. ✅ Notice that playback continues from cache

#### Test.2 Caching not initiated if on mobile network
1. Continue from above
2. Enable internet and switch to mobile network
3. Clear this directory in device-explorer: `/data/data/au.com.shiftyjelly.pocketcasts.debug/cache/pocketcasts-exoplayer-cache`
4. Play an episode which is not downloaded or cached earlier
5. ✅  Notice that the episode is not cached in device-explorer at `/data/data/au.com.shiftyjelly.pocketcasts.debug/cache/pocketcasts-exoplayer-cache`

#### Test.3 Previous caching cancelled on playing another episode
1. Continue from above
2. Switch to wifi
3. Clear this directory in device-explorer: `/data/data/au.com.shiftyjelly.pocketcasts.debug/cache/pocketcasts-exoplayer-cache`
4. Play an episode which is not downloaded or cached earlier
5. Before the caching complete log appears, play another episode 
6. ✅ Notice in logs that cache worker is cancelled for the previous episode

#### Test.4 Disabled advanced setting
1. Continue from above
2. Go to `Profile` -> `Settings` -> `Advanced`
[...]

PR #5618, opened 2026-07-21
Tap into ExoPlayer's cache instead of a standalone MediaSource for fingerprinting

sztomek:
## Description
Instead of running a second, parallel decode of the episode just to fingerprint it, we now tap the PCM that ExoPlayer is already decoding for playback by a passthrough processor at the head of the audio chain copies the samples, and a thin sink wrapper stamps them with accurate media-time positions. 
That means fingerprinting is literally free on the network side: the old approach re-downloaded the audio it was fingerprinting (up to a full second copy of the episode for streamed playback), while the tap adds zero bytes and zero extra codec work — we only pay for the fingerprint math itself, a few percent of one core. It also killed a whole pile of machinery: the playhead-following throttle, restart debouncing, metered-network gating and retry callbacks all just disappeared, because there's nothing to gate when you're not downloading anything. 
The positions are also more trustworthy — we're fingerprinting the exact audio the user hears, anchored to the player's own timeline, and since the tap sits upstream of trim-silence and Sonic, speed changes and silence trimming can't skew the map. The one trade-off is that the tap only sees audio in realtime, so seeks would leave the map ~8 seconds behind —we cover that with a tiny bounded catch-up decode that reads through the player's own cache and has anchors at the new position in about a second. 
Chapter jumps and taps into far-off transcript regions go through the same bounded resolver, decoding a ~12-second window instead of anything close to a full pass.
**Net result**: same sync quality on screen, roughly half the data usage of the old streamed path gone entirely, less battery, and way less code to babysit.

## Testing Instructions
Make sure to rigorously test anything that might be affected by fingerprinting: seeking while playing an episode with transcripts open, jumping to chapters, jumping to transcript rows.

## Checklist
 
- [ ] with different themes


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-android/milestone/131">8.17 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

geekygecko (approved):
I just tested it, and it does seem to use more data than the episode file itself. I think that might be because ExoPlayer's cache isn't storing the entire episode. As we jump between different parts, it appears to stream each section again instead of reading it from the cache.

I wonder if there's a way to improve the caching so the whole episode is available locally. What do you reckon?

sztomek:
> I just tested it, and it does seem to use more data than the episode file itself. I think that might be because ExoPlayer's cache isn't storing the entire episode. As we jump between different parts, it appears to stream each section again instead of reading it from the cache.
> 
> I wonder if there's a way to improve the caching so the whole episode is available locally. What do you reckon?

@geekygecko 

I dug in properly and benchmarked it. You're right about the symptom, but it turns out to be pre-existing behaviour rather than something this PR added.
  
  **What I measured.** Two emulators (same AVD, wifi, cold caches), streaming this exact episode (The Daily, served file 29,460,794 bytes ≈ 28.1 MB) and jumping between chapters via the player skip buttons (11 jumps), measured via per-UID netstats:

  | Run | Base branch (#5614) | This branch (pcm-tap) |
  |---|---|---|
  | 1 | 41.89 MB (1.42×) | 41.48 MB (1.41×) |
  | 2 | 38.67 MB (1.31×) | 40.51 MB (1.38×) |
  | Final ExoPlayer cache | 32.4 MB (full episode) | 32.4 MB (full episode) |

  Both branches sit at ~1.3–1.4× the file size, and the gap between them is run-to-run noise (every request gets a slightly different ad-stitched file, so single runs vary by a couple of MB anyway).

  **Where the extra ~40% comes from.** The player reads through ExoPlayer's cache but has writes disabled (`setCacheWriteDataSinkFactory(null)` in `createMediaSource`, since the original cache-entire-episode work in #2380). The whole episode is cached by the background `CacheWriter` sweeping from byte 0. Jump
  ahead of the sweep and the player streams that section without storing it — then the worker downloads the same bytes again later. That's the "streams each section again" you saw, and it predates this chain.

  **What this PR changes on data: nothing — by design.** The fingerprinting side stopped being a network problem back in the cache read-through work; what it still cost was a *second decode* of the same audio. That's what the PCM tap removes: the mapping is now built from the samples the player already decoded,
  so continuous fingerprinting costs zero extra network *and* zero extra codec/CPU, follows the live edge without lag, and matches exactly what's rendered. The bounded chapter/transcript resolves are the only remaining fetches, and they *do* write into the cache, so the worker skips those bytes.

  **Can we fix the double-fetch? Yes, but not by just re-enabling player writes.** That was disabled deliberately: with dynamic ad stitching, every request can return a differently-stitched file (this exact episode varies in size per request). `SimpleCache` keys spans by episode UUID and assumes one underlying
  file — letting the player's per-seek requests write spans would interleave different stitch variants under one key: audio glitches at span joins, and cached content-length mismatches, which is the old 416 story (#3183). The worker-as-sole-writer guarantees the cached copy comes from a single request, i.e. a
  single consistent variant.

  The safe version is gating write-through on response identity — only let a connection write when its strong ETag / exact content-length matches what's already cached, read-only otherwise. Static-file feeds (most of them) would get full write-through and ~1× traffic; stitched feeds keep today's safe behaviour.
  That's a self-contained change with its own testing burden, so I'd rather land it as a follow-up PR than grow this one. Happy to open it next.

  One note on the table: I labeled the samples "Run 1 / Run 2" per branch for readability, but they weren't all simultaneous pairs — each branch had one solo-valid run and one paired run (the other side's run failed to launch twice due to emulator cold-start, not app behavior). The numbers are exactly as
  measured; only the pairing is loose. If you'd rather be strictly precise in the comment, swap the table rows for: base 41.89 / 38.67 MB, pcm-tap 41.48 / 40.51 MB, "two valid samples each".

PR #5619, opened 2026-07-21
[TV] Show dynamic content on the Home tab

sztomek:
## Description
This PR connects the Home tab with out backend systems and the app now fetches and displays actual data.
We can distinguish 2 cases - when the user is signed in (in this case the Up next, Keep listening sections also appear) and when they are not.
This PR only focuses on the data bits, you should not pay too much attention to the designs as they will be amended later to match the latest figma designs.

Fixes PCDROID-603 https://linear.app/a8c/issue/PCDROID-603/populate-home-tab-with-real-data

## Testing Instructions
A - Browse without sign in
1. Clean install TV app
2. Hit Browse without sign in
3. Observe Home tab content

B - Sign in
1. Clean install app
2. Sign in with QR 
3. Observe Home tab content.

## Screenshots or Screencast 
| Signed in | Not signed in |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/7294fbdd-9d3f-4abd-86e8-bd16a8b27924" /> | <video src="https://github.com/user-attachments/assets/d96c19af-00f4-4cf3-aa87-341595b46ece" /> |

## Checklist
 
- [ ] with different themes

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/lists/ListRepository.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/lists/ListRepository.kt
@@ -17,7 +17,7 @@ class ListRepository(
 ) {
 
     suspend fun getDiscoverFeed(): Discover {
-        val version = if (FeatureFlag.isEnabled(Feature.RECOMMENDATIONS)) 3 else 2
+        val version = if (platform == PLATFORM_TV || FeatureFlag.isEnabled(Feature.RECOMMENDATIONS)) 3 else 2
         return listWebService.getDiscoverFeed(platform = platform, version = version)
     }
 
@@ -45,5 +45,11 @@ class ListRepository(
 
     suspend fun getPodcastRecommendations(podcastUuid: String, countryCode: String?): ListFeed? {
         return getListFeed(url = "${Settings.SERVER_API_URL}/recommendations/podcast/$podcastUuid?country=${countryCode ?: "global"}")
+    }
+
+    companion object {
+        const val PLATFORM_ANDROID = "android"
+        const val PLATFORM_AUTOMOTIVE = "automotive"
+        const val PLATFORM_TV = "tv"
     }
 }
--- a/tv/src/main/java/au/com/shiftyjelly/pocketcasts/component/TvRow.kt
+++ b/tv/src/main/java/au/com/shiftyjelly/pocketcasts/component/TvRow.kt
@@ -13,14 +13,16 @@ import androidx.compose.foundation.lazy.LazyRow
 import androidx.compose.foundation.lazy.itemsIndexed
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableIntStateOf
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
+import androidx.compose.runtime.saveable.rememberSaveable
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.focus.FocusRequester
+import androidx.compose.ui.focus.focusProperties
 import androidx.compose.ui.focus.focusRequester
-import androidx.compose.ui.focus.focusRestorer
 import androidx.compose.ui.focus.onFocusChanged
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.text.PlatformTextStyle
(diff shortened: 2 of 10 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/lists/ListRepository.kt
```diff
@@ -17,7 +17,7 @@ class ListRepository(
 ) {
 
     suspend fun getDiscoverFeed(): Discover {
-        val version = if (FeatureFlag.isEnabled(Feature.RECOMMENDATIONS)) 3 else 2
+        val version = if (platform == PLATFORM_TV || FeatureFlag.isEnabled(Feature.RECOMMENDATIONS)) 3 else 2
```
geekygecko, line 20: This is probably a separate PR, but I think we can remove the recommendations feature flag, it has been available for ages.

tv/src/main/java/au/com/shiftyjelly/pocketcasts/home/TvHomeScreen.kt
```diff
@@ -0,0 +1,218 @@
...
+            }
+        }
+    }
+}
+
+@Composable
+private fun TvHomeRows(
+    rows: List<TvHomeRow>,
+    modifier: Modifier = Modifier,
+) {
```
geekygecko, line 94: It seems to be showing all the category sponsors at the top of the page. 

Image

tv/src/main/java/au/com/shiftyjelly/pocketcasts/home/TvHomeViewModel.kt
```diff
@@ -0,0 +1,273 @@
...
+
+    private suspend fun loadRow(row: DiscoverRow): TvHomeRow? {
+        return when (row.type) {
+            is ListType.PodcastList -> loadPodcastsRow(row)
+            is ListType.EpisodeList -> loadEpisodesRow(row)
+            is ListType.Categories, is ListType.Unknown -> null
+        }
+    }
+
+    private suspend fun loadPodcastsRow(row: DiscoverRow): TvHomeRow? {
```
geekygecko, line 187: A lot of this seems like common logic for loading the Discover section. I wonder whether it should be pulled out into a module which isn't related to the TV so that we can reuse it for mobile Discover when we switch away from Rx or if we have another platform we want to use it on.

geekygecko (approved):

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/lists/ListRepository.kt
sztomek: Good idea, I'll raise a dedicated one to remove the flag!

tv/src/main/java/au/com/shiftyjelly/pocketcasts/home/TvHomeScreen.kt
sztomek: Great spotting, addressed it in [3086681](https://github.com/Automattic/pocket-casts-android/pull/5619/commits/30866810215e4d8196b7b348004a873c90854170)

tv/src/main/java/au/com/shiftyjelly/pocketcasts/home/TvHomeViewModel.kt
sztomek: Hmm this one is a bit trickier, I'll try to get it done in a new PR

tv/src/main/java/au/com/shiftyjelly/pocketcasts/component/TvRow.kt
```diff
@@ -74,21 +76,32 @@ fun <T> TvRow(
...
 
         LazyRow(
             contentPadding = contentPadding,
             horizontalArrangement = Arrangement.spacedBy(itemSpacing),
             modifier = Modifier
                 .focusGroup()
-                .focusRestorer(firstItemFocusRequester),
+                .focusProperties {
+                    onEnter = {
+                        focusRequesters.getOrNull(lastFocusedIndex)?.requestFocus()
```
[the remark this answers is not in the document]
sztomek: Re the TvRow focus-restore concern (requestFocus() on a disposed item's FocusRequester throwing IllegalStateException):

This doesn't throw on the Compose version we're on. Since 1.11 (we ship 1.11.2 via the BOM), requestFocus() on a requester with no attached nodes logs a FocusWarning and returns false instead of throwing — see findFocusTarget in FocusRequester.kt:

internal inline fun findFocusTarget(onFound: (FocusTargetNode) -> Boolean): Boolean {
    check(this !== Default) { InvalidFocusRequesterInvocation }
    check(this !== Cancel) { InvalidFocusRequesterInvocation }
    if (focusRequesterNodes.isEmpty()) {
        println("$FocusWarning: $FocusRequesterNotInitialized")
        return false
    }
    ...
The IllegalStateException("FocusRequester is not initialized...") quoted in the review only exists for the Default/Cancel sentinel requesters, which can't apply to our remembered per-item requesters. So when the last-focused item has been disposed, the call safely no-ops and focus enter falls back to the default target (first visible item), which is the intended behavior.

PR #5628, opened 2026-07-22
Copy description and show notes links with a long press

joashrajin:
## Description

Copies the exact target URL when a link is long-pressed in:

- podcast descriptions
- Now Playing > Details
- Episode > Details

Compose previously supported link taps only, while the two WebViews kept Android's default URL-preview popup because no link long-click listener was installed. This adds reusable Compose and WebView handling plus one shared, version-aware clipboard helper. Regular taps and non-link interactions remain unchanged.

Linear: [PCDROID-681](https://linear.app/a8c/issue/PCDROID-681/android-copy-description-and-show-notes-links-with-a-long-press)

## Testing Instructions

1. Open a podcast with a linked podcast description, expand it, then long-press the link.
2. Confirm the URL is copied once and a regular tap still opens it.
3. Open Now Playing > Details and long-press an external show-notes link.
4. Open Episode > Details and repeat.
5. Long-press non-link text and confirm it is not copied.

Automated checks:

- `:modules:services:views:testDebugUnitTest --tests '*WebViewTest'`
- `:modules:services:compose:testDebugUnitTest --tests '*ExpandableTextTest'`
- player and podcasts Kotlin compilation
- `spotlessCheck`
- views, player, and podcasts lint
- `:app:assembleDebugProd`

Physical validation on a Samsung SM-G990E running Android 16 covered all three surfaces, regular link taps, non-link long presses, and fatal-crash monitoring.

## Screenshots or Screencast

Not included. The system clipboard confirmation was verified on the physical test device.

## Checklist

- [x] If user-facing, added an entry to `CHANGELOG.md`
- [x] All checks and linter rules pass locally
- [x] Added unit tests for the shared Compose and WebView handlers
- [x] No new strings required; the existing localized clipboard confirmation is reused where Android does not provide one
- [x] Existing Compose previews continue to cover the changed component
- [x] No analytics changes required

#### I have tested the UI changes on:

- [x] Light and dark themes
- [ ] Landscape orientation
- [ ] Large display size and font
- [ ] TalkBack

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 8.18
 -----
-
+*   Bug Fixes
+    *   Allow copying description and show notes links with a long press
+        ([#5628](https://github.com/Automattic/pocket-casts-android/pull/5628))
 
 8.17
 -----
--- a/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/NotesFragment.kt
+++ b/modules/features/player/src/main/java/au/com/shiftyjelly/pocketcasts/player/view/NotesFragment.kt
@@ -24,6 +24,7 @@ import au.com.shiftyjelly.pocketcasts.ui.theme.ThemeColor
 import au.com.shiftyjelly.pocketcasts.utils.extensions.toSecondsFromColonFormattedString
 import au.com.shiftyjelly.pocketcasts.utils.log.LogBuffer
 import au.com.shiftyjelly.pocketcasts.views.extensions.cleanup
+import au.com.shiftyjelly.pocketcasts.views.extensions.copyLinkOnLongPress
 import au.com.shiftyjelly.pocketcasts.views.extensions.show
 import au.com.shiftyjelly.pocketcasts.views.extensions.showIf
 import au.com.shiftyjelly.pocketcasts.views.fragments.BaseFragment
@@ -124,6 +125,7 @@ class NotesFragment : BaseFragment() {
                 // stop the web view jumping after loading
                 isFocusable = false
                 setBackgroundColor(Color.TRANSPARENT)
+                copyLinkOnLongPress()
                 webViewClient = object : WebViewClient() {
                     override fun shouldOverrideUrlLoading(view: WebView, request: WebResourceRequest): Boolean {
                         val url = request.url.toString()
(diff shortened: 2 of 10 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">View files have been modified, but no screenshot or video is included in the pull request. Consider adding some for clarity.</td>
    </tr>
  </tbody>
</table>

</p>

joashrajin:
I did a deeper check of the two remaining notes, and neither requires a code change here.

### `pointerInput` callback identity

This project uses Kotlin and the Compose compiler plugin `2.3.21`, with no configuration disabling Strong Skipping. Strong Skipping has been enabled by default since Kotlin 2.0.20 and automatically remembers lambdas created inside composables, keyed by their captures: [Android documentation](https://developer.android.com/develop/ui/compose/performance/stability/strongskipping).

I also checked the generated `debugProd` bytecode for `PodcastDetails` on this commit. Around the callback passed to `ExpandableText`, the compiler emits:

- `changedInstance(context)`
- `rememberedValue()`
- allocation of the `Function1` only when that remembered slot is empty or the captured `Context` instance changes
- `updateRememberedValue()`

So the lambda is not freshly allocated on ordinary recompositions and does not restart `pointerInput` each time. If the captured `Context` genuinely changes, rebuilding the callback and gesture detector is appropriate. Adding an explicit `remember(context)` would duplicate compiler-generated behavior.

### Link annotations at the truncation boundary

The exact dependency resolved by this branch is `ui-text-android 1.11.2`. Its `AnnotatedString.Builder.append(text, start, end)` copies every intersecting annotation through `getLocalAnnotations()`. That function clamps only the annotation's character range and retains the original annotation item with `item = it.item`: [AndroidX source](https://android.googlesource.com/platform/frameworks/support/+/refs/heads/androidx-main/compose/ui/ui-text/src/commonMain/kotlin/androidx/compose/ui/text/AnnotatedString.kt).

`ExpandableText.urlRanges()` then reads the target from `LinkAnnotation.Url.url`, not from the displayed substring. Therefore:

- a link crossing the cutoff keeps its complete href while its visible range is shortened;
- a link entirely beyond the cutoff has no visible glyph and no gesture target;
- the appended ellipsis and “See more” text are outside the copied annotation.

There is no path where a visible clipped link copies a truncated URL or loses its URL annotation.

For the image-anchor notes, consuming the gesture before the asynchronous href callback is intentional. Falling back to the default popup could expose the image `src`, which is precisely the incorrect value this fix avoids. The one-shot `Handler` has no retained lifetime after its message is dispatched.

geekygecko (approved):
Thanks for making this change. I’ve seen other users request this as well. 

The only minor thing I noticed is that long pressing an email includes the prefix. We could strip that out. For example with https://pca.st/tz2btaj3 when you copy the email from the header you get `mailto:lovetrappedpod@gmail.com` in your clipboard.

geekygecko:
I have merged in `main` as there was a change log conflict.

joashrajin:
Thanks @geekygecko 

I made some tweaks to address it :)

PR #5634, opened 2026-07-24
Switch default artwork placeholder to vector drawables to fix main-thread decode ANRs

geekygecko:
## Description

This change is to try help fix the recurring ANRs caused by a synchronous bitmap decode on the main thread. 

`defaultartwork.png` shipped at 1038x1038, which decodes to roughly 4MB of `ARGB_8888`, and only had `xhdpi` and`xxhdpi` buckets. On `xxxhdpi` devices Android decodes the `xxhdpi` asset and bilinearly upscales it, which is why the traces show Skia resampling symbols (`neon bilerp`, `swizzle_index_to_n32`, `store_8888`) rather than pure decode.

Changes:
- Replaced the placeholder with a vector drawable, so there is no decode and no density buckets.
- Deleted the 8 PNGs (~185KB of source assets).
- Dropped the separate `_small` variants. A vector scales freely, so `PlaceholderType.Small` and `?attr/defaultArtworkSmall` now resolve to the same drawable. Both the enum case and the theme attribute are kept so the existing call sites are untouched.

Fixes https://linear.app/a8c/issue/PCDROID-685/anrs-from-synchronous-bitmap-decode-on-the-main-thread-player-bottom

## Testing Instructions

The placeholder shows while artwork loads, so it is easiest to see on the Discover page and when a podcast is missing artwork.

1. Open Discover and scroll the podcast rows before images load
2. Open the podcast page https://pca.st/T5AzSO as it is missing artwork
3. Play an episode from the podcast to see the artwork in the mini player, full screen player, and Up Next list
4. Confirm the placeholder still renders correctly, and that no artwork slot is blank or wrongly sized

## Screenshots

| Before | After |
| --- | --- |
| Screenshot_20260724_193707 | Screenshot_20260724_223103 | 
| Screenshot_20260724_193742 | new | 
| Screenshot_20260724_194620 | Screenshot_20260724_223020 | 

## Checklist
 
- [x] with different themes


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/images/PocketCastsImageRequestFactory.kt
```diff
@@ -82,8 +82,7 @@ data class PocketCastsImageRequestFactory(
     @get:DrawableRes private val placeholderId
         get() = when (placeholderType) {
             PlaceholderType.None -> 0
-            PlaceholderType.Small -> if (isDarkTheme) IR.drawable.defaultartwork_small_dark else IR.drawable.defaultartwork_small
-            PlaceholderType.Large -> if (isDarkTheme) IR.drawable.defaultartwork_dark else IR.drawable.defaultartwork
+            PlaceholderType.Large, PlaceholderType.Small -> if (isDarkTheme) IR.drawable.defaultartwork_dark else IR.drawable.defaultartwork
```
geekygecko, line 85: We might be able to merge these PlayholderTypes together but I thought I would keep the PR simple for now.

sztomek (approved):
Seems to work fine across different themes. Good job 👍

PR #5647, opened 2026-07-27
Release fingerprint codec buffers before suspending

geekygecko:
## Description

Investigating a native crash reported mainly on Huawei devices:

```
#00  /system/lib64/libstagefright.so (android::ACodec::BaseState::onOMXFillBufferDone(...)+2064)
#01  /system/lib64/libstagefright.so (android::ACodec::BaseState::onOMXMessage(...)+764)
...
#11  /system/lib64/libstagefright_foundation.so (android::ALooper::loop()+556)
```

No Java frames: this is a SIGSEGV on the codec's own looper thread, inside the handler that runs when an OMX component returns a filled output buffer. That crashes when the buffer's bookkeeping entry has been freed while the component still had it in flight, i.e. the classic ACodec teardown race. `ACodec` is the legacy OMX path, which is why it clusters on vendor decoders like HiSilicon's rather than on Codec2 devices.

The only hand-rolled `MediaCodec` loop we have is the fingerprint decoder, and it had exactly that race. In `decodeAndFingerprint`, the dequeued output buffer was held across the `onWindows` callback, which suspends on `mutex.withLock` for the continuous path and runs under `withTimeoutOrNull` for both the on-demand and catch-up resolve paths. Cancelling at that point skipped `releaseOutputBuffer`, and `AudioStream.close()` then tore the codec down while we still owned a buffer. Cancellation there is routine, not an edge case: every generation change and every resolve timeout hits it.

### What changed

- **`decodeAndFingerprint`**: copy the samples out and call `releaseOutputBuffer` *before* any suspending work. `extractFloatSamples` already copies into a `FloatArray`, so the `ByteBuffer` is not referenced afterwards. The codec now owns all of its buffers at every cancellation point.
- **`AudioStream.close()`**: drop `stop()`. `release()` is valid from any state, while `stop()` forces the Executing to Idle transition that requires a full buffer census and is where these decoders fail. Each teardown step is individually guarded so one failure cannot leak the others.
- **`openAudioStream`**: a malformed track format (missing sample rate or channel count) or a failing `createDecoderByType` used to leak the extractor and cache source and escape the `AudioOpenException` classification. Both are now cleaned up and rethrown as `AudioOpenException`, which callers already degrade gracefully on.

### Notes for reviewers

- The native crash cannot be caught from Kotlin, so the only available handling is to stop provoking it. I checked the Java-level paths separately: every entry into the codec loop already catches `Exception` and degrades gracefully (eager decode marks failed, chapter resolve returns `Unresolved(REASON_AUDIO_UNAVAILABLE)`, catch-up resolve is logged and dropped), so `CodecException`/`IllegalStateException` were never app-killers.
- One residual case left alone deliberately: a cancelled network read can abandon a dequeued *input* buffer. There is no API to hand an unqueued input buffer back, `release()` is documented to cope with it, and the crash is on the output (`FillBufferDone`) side.
- This branch's PCM-tap rework already removed the largest exposure by dropping the always-on parallel decoder during playback. On top of that, the codec loop now only runs for eager passes on downloaded episodes and short bounded resolves.
- If Huawei crash volume does not drop after this ships, we should decide if we disable this feature for these users. It seems to only be happening for users on Android 9.0.

## Testing Instructions

This is a race on a cancellation path, so the useful test is to exercise cancellation hard while the eager decoder runs:

1. Make sure the Synced Transcripts feature flag is on in Developer settings.
2. Play a **downloaded** episode of a podcast with a generated transcript (the eager local decode only runs for downloaded episodes).
3. Open the transcript and seek repeatedly and rapidly, including tapping transcript rows to seek. Each seek cancels and restarts the decode generation, which is the path that used to abandon a buffer.
4. Confirm playback and transcript highlighting stay correct and the app does not crash.
5. Ideally repeat on a Huawei or other HiSilicon device, or any Android 9 or below device, where the legacy ACodec path is in use.

## Checklist
- [x] If this is a user-facing change, I have added an entry in CHANGELOG.md — n/a, this fixes an unreleased feature behind the `synced_transcripts` flag, on its feature branch
- [x] I have considered whether it makes sense to add tests for my changes — the existing fingerprint suite passes; the race is in real `MediaCodec` buffer ownership and is not reachable from JVM unit tests
- [x] I have updated (or requested that someone edit) the Event Horizon schema — n/a, no analytics changes


modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/fingerprint/FingerprintTimingManager.kt
```diff
@@ -1100,10 +1101,10 @@ class FingerprintTimingManager @Inject constructor(
         }
 
         override fun close() {
-            runCatching { codec.stop() }
-            codec.release()
-            extractor.release()
-            cacheSource?.close()
+            // Skip stop(): release() works from any state and the Executing→Idle transition crashes some OMX decoders.
```
sztomek, line 1104: ❓ don't we want to log something when we catch an exception on any of these operations? maybe that could help later?

sztomek (approved):
I don't have Huawei, but tested it on my phone and fingerprinting worked as before.

geekygecko:
Closing in favour of https://github.com/Automattic/pocket-casts-android/pull/5677

PR #5656, opened 2026-07-27
Prevent stale Cast resumes from breaking local playback

### LINKED ISSUE #5248: Android: stale Cast session auto-resume (error 2152) tears down the media notification and stops local playback
> **Updated 2026-07-27** with root cause confirmed from a full debug log on 8.16. The original report is preserved under "Original report" below. Retitled from "Intermittent broken playback with 8.10-rc-2": this is neither intermittent nor rc-specific, it reproduces on demand and is still present in 8.16.

## Summary

Pocket Casts persists a Chromecast session and the Cast SDK auto-resumes it on app foreground. When the receiver is unreachable the resume fails with error `2152`, and **that failure is routed into the local playback error path**. The result is that the media notification and foreground service are torn down and local playback stops, even though the user is not casting and never asked to.

The user sees a "Could not connect to the Chromecast device" toast, loses lock screen / headphone / Android Auto transport controls, and playback halts.

## Log signature

```
E Cast session resume failed with error code 2152
E Cast session failed with error code 2152
I stopForeground state: 7 removing notification: true
E Playback state error: 1 Could not connect to the Chromecast device. Please check the device and try again.
```

Repeats roughly 10 seconds after **every** app foreground event (10+ occurrences in a single log), each preceded by the standard foreground sequence (`Loading subscription plans` / `Billing client connected` / `Set up periodic refresh`).

Sequence:

1. Cast SDK attempts to resume a previously saved session.
2. Resume fails with `2152`.
3. The failure surfaces as a generic playback error instead of being scoped to the cast path. **This is the defect.**
4. `stopForeground(... removeNotification = true)` tears down the media notification and foreground service.
5. Local playback stops; transport controls go dead.

## Reproduction

1. Cast to a Chromecast or Nest receiver from Pocket Casts.
2. Leave without explicitly tapping **Stop Casting** (walk out of range, or power the receiver off).
3. Later, with the receiver unreachable, foreground Pocket Casts and play an episode locally.
4. Within ~10 seconds the toast appears, the media notification disappears, and playback stops.

Step 2 is the trigger. The saved session persists indefinitely until a session is explicitly ended.

## What this is not

* **Not a network or proximity problem.** Confirmed firing with `Current connection: Type: mobile, Metered? yes` and a Bluetooth headset as the active audio route. Cellular data, no receiver reachable anywhere, and the resume is still attempted and still kills playback. Toggling Wi-Fi does nothing.
* **Not cleared by a process restart.** Four separate `App started. 8.16 (9441)` events within a ~5 minute window, error recurring after each. The saved session outlives the process, so force stop and device reboot do not help.
* **Not user-initiated.** Affected users are not casting. One had never knowingly cast from the app.
* **Not PCDROID-555.** The same log contains `reason=3 (LOW_MEMORY)` process exits, but those are a separate issue. This failure is directly attributable to the cast error tearing down the foreground service and is reproducible on demand.

The user-visible toast is gated behind the `display_errors_on_player: true` flag, enabled in current builds.

## Workarounds

**Works:** rejoin the network with the receiver powered on, connect from Pocket Casts, then explicitly tap **Stop Casting**. This ends the session cleanly and clears the persisted state. Independently discovered by the reporter on PCDROID-549.

**Does not work,** all confirmed by affected users: force stop, device restart, Wi-Fi toggle, and "disconnect via the Cast icon" (there is no visible session to disconnect from, which is precisely the problem). Support should not be recommending these.

## Suggested direction

Two separable fixes:

1. Do not propagate a Cast **session resume** failure into `PlaybackState` or the local player error path when no cast session is actively in use. Failing silently back to local playback resolves the user-facing breakage on its own, and is the smaller change.
2. Reconsider `CastOptions.setResumeSavedSession`, or bound the resume attempt: stop retrying once the route is undiscoverable, and clear the persisted session when the app is not on a network where that receiver exists.

## Impact

* App versions **8.9 through 8.16**.
* Android 16 (SDK 36) and Android 17 (SDK 37).
* Multiple OEMs: Nothing A024, Pixel 10 Pro XL.
* **10 support tickets** carrying this error string since April 2026. Two are attached to this issue; at least one affected customer has stated they are switching apps over it.

Likely the same underlying defect as PCDROID-549, which can probably be closed as a duplicate of this once confirmed.

---

### Description

This issue occurs intermittently. During podcast playback (over headphones so no casting), I'll get a toast notification that says

> Could not connect to the Chromecast device. Please check that the device and try again.

Sometimes playback will continue and sometimes it'll pause (and only resume when I re-enter the app). What will consistently happen is that the playback controls will disappear from the notification area and I can no longer play/pause with the controls on my headphones. If playback continues, it will usually be able to finish the podcast but then it won't go to the next one in the queue (it'll just pause at the end of the current podcast). When I open the app it displays a banner underneath the bottom bar which says

> This episode can't be played. Learn more

where the "Learn more" link takes me to an unhelpful webpage. Note that this banner appears consistently even though the podcast is playing.

Going into the playback screen, hitting play/pause twice will usually reset the app to working normally temporarily.

I do wonder if this is related to casting that I don't exit properly. I do sometimes cast to speakers, pause playback and then leave home without explicitly cancelling casting. This has never previously been a problem.

*(Confirmed 2026-07-27: this hunch was correct.)*

Unfortunately this issue occurs intermittently so I don't have good reproduction instructions. When it does occur, it's often something like the following:

1. Open Pocket Casts and start playing a podcast.
2. Leave the app (not always required)
3. Wait a short period of time
4. Observe the toast notification which says

   > Could not connect to the Chromecast device. Please check that the device and try again.<br>   as well as the playback controls disappearing from the notification area.
5. Try play/pausing using my headphone controls. They don't work.
6. Open Pocket Casts and observe the bottom banner which says

   > This episode can't be played. Learn more
7. The app can be reset to normal by going to the "Playing" screen and tapping Play/pause twice.

Toast notification:

Image

Bottom banner:

Image

* Google Pixel 10 Pro XL
* Android 16
* Pocket Casts 8.10-rc-2 (9423)

joashrajin:
## Description

A failed Cast SDK saved-session resume was routed through the same error path as a user-initiated Cast start. When the saved receiver was unavailable, the callback changed healthy local playback to an error state, removing the media notification and breaking lock-screen, headphone, and Android Auto controls.

This change preserves whether a Cast session failure came from an explicit start or an automatic resume. Resume failures are ignored only while local playback is active, while explicit Cast start failures and failures affecting an installed Cast player remain visible.

Fixes #5248
Fixes [PCDROID-554](https://linear.app/a8c/issue/PCDROID-554/android-stale-cast-session-auto-resume-error-2152-tears-down-the-media)

## Testing Instructions

1. Cast an episode to a receiver.
2. Leave the Cast session without tapping **Stop Casting**, then make the receiver unavailable.
3. Start local playback and bring Pocket Casts to the foreground.
4. Wait for the saved Cast session resume attempt to time out.
5. Verify local playback continues without an error toast or banner.
6. Verify the media notification remains visible and lock-screen, headphone, and Android Auto controls still work.
7. Start a new Cast connection to an unavailable receiver and verify the explicit connection failure is still shown.

Automated validation:

`JAVA_HOME='/Applications/Android Studio.app/Contents/jbr/Contents/Home' ./gradlew spotlessCheck :modules:services:repositories:testDebugUnitTest`

## Screenshots or Screencast 

Not applicable — no UI changes.

## Checklist
 
- [ ] with different themes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
 *   Bug Fixes
     *   Allow copying description and show notes links with a long press
         ([#5628](https://github.com/Automattic/pocket-casts-android/pull/5628))
+    *   Prevent a failed Cast session resume from interrupting local playback or removing playback controls
+        ([#5656](https://github.com/Automattic/pocket-casts-android/pull/5656))
 
 8.17
 -----
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/chromecast/CastManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/chromecast/CastManager.kt
@@ -6,6 +6,11 @@ interface CastManager {
 
     val isConnectedFlow: StateFlow<Boolean>
 
+    enum class SessionFailureType {
+        START,
+        RESUME,
+    }
+
     suspend fun isAvailable(): Boolean
     suspend fun isConnected(): Boolean
     suspend fun endSession()
@@ -17,6 +22,6 @@ interface CastManager {
         fun sessionStarted()
         fun sessionEnded()
         fun sessionReconnected()
-        fun sessionFailed(errorCode: Int) {}
+        fun sessionFailed(errorCode: Int, failureType: SessionFailureType) {}
     }
 }
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/chromecast/CastManagerImpl.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/chromecast/CastManagerImpl.kt
@@ -124,7 +124,7 @@ class CastManagerImpl @Inject constructor(
 
         override fun onSessionStartFailed(session: Session, i: Int) {
             LogBuffer.e(LogBuffer.TAG_PLAYBACK, "Cast session start failed with error code $i")
-            sessionListener?.sessionFailed(i)
+            sessionListener?.sessionFailed(i, CastManager.SessionFailureType.START)
         }
 
         override fun onSessionEnding(session: Session) {
@@ -150,7 +150,7 @@ class CastManagerImpl @Inject constructor(
 
         override fun onSessionResumeFailed(session: Session, i: Int) {
             LogBuffer.e(LogBuffer.TAG_PLAYBACK, "Cast session resume failed with error code $i")
-            sessionListener?.sessionFailed(i)
+            sessionListener?.sessionFailed(i, CastManager.SessionFailureType.RESUME)
         }
 
         override fun onSessionSuspended(session: Session, i: Int) {
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackManager.kt
@@ -308,7 +308,16 @@ open class PlaybackManager @Inject constructor(
                     castReconnected()
                 }
 
-                override fun sessionFailed(errorCode: Int) {
+                override fun sessionFailed(errorCode: Int, failureType: CastManager.SessionFailureType) {
+                    // The Cast SDK can resume a saved session without user interaction. A failed cold resume must not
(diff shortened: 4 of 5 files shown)


paulybeebe:
I am really suffering with this issue. Any chance we can get it resolved please? Been a little while now.

sztomek (approved):

PR #5676, opened 2026-07-29
[TV] Add an episode actions menu to the TV playlist details screen

sztomek:
## Description

Adds the per-episode actions menu to the TV playlist details screen, mirroring the Apple TV implementation: a circular `…` button appears next to the focused episode row and opens a modal action sheet titled with the episode. The menu lists the same actions in the same order as tvOS's playlist context (`.other(showGoToPodcast: true)`):

- Episode details
- Go to podcast
- Play next in Up Next
- Play last in Up Next
- Mark as played / Mark as unplayed (toggles on the episode's played state)
- Archive / Unarchive (toggles on the episode's archived state)

- New reusable `TvMoreButton` (circular ellipsis icon button, focus-scaled, styled like the existing TV icon buttons) and `TvEpisodeActionsModal` (built on `TvModal`, first action self-focuses), plus a horizontal-ellipsis drawable to match the tvOS glyph.
- The row and its `…` button are focus siblings: unfocused rows span the full width to the right margin, and when a row gains focus its container shrinks while the `…` button slides in from the right (and reverses on focus loss). LEFT from a row still focuses `Play all episodes`; RIGHT reaches the `…` button.
- **Scope is the menu UI only.** The action buttons are intentionally no-ops for now — wiring each action to the ViewModel/`EpisodeManager` comes in a follow-up, as does analytics (`episode_actions_shown` and per-action events).

Fixes PCDROID-693 https://linear.app/a8c/issue/PCDROID-693/playlist-details.
Designs: Ftk3KwnfqaK4g57yCN63p0-fi-3258_12136.

Stacked on `feat/tv-playlist-details-archived` (#5669) — merge bottom-up.

## Testing Instructions

1. Install the TV app (`./gradlew :tv:installDebugProd`) and open a playlist's details.
2. Move through the episode list with the d-pad — verify unfocused rows reach the right margin, and the focused row shrinks as the `…` button slides in from the right (and reverses when focus leaves).
3. Focus a row, press RIGHT to reach the `…` button, press select — a modal action sheet opens titled with the episode, first action focused.
4. Verify the actions and their order: Episode details, Go to podcast, Play next in Up Next, Play last in Up Next, Mark as played/unplayed, Archive/unarchive.
5. Verify the played/archive labels reflect the episode's state (e.g. an in-progress, unarchived episode reads "Mark as played" / "Archive").
6. Press back — the modal dismisses and focus returns to the `…` button.
7. Confirm the action buttons currently do nothing (no-op by design).

## Screenshots or Screencast

https://github.com/user-attachments/assets/c6d24dff-5314-42da-ac1a-21e4a9e2f2cc

## Checklist
- [ ] I have updated (or requested that someone edit) [the Event Horizon schema](https://github.com/Automattic/EventHorizonSchemas/blob/trunk/schema/pocket-casts.yml) to reflect any new or changed analytics. _(analytics deliberately excluded — follow-up)_

### DIFF
--- a/modules/services/images/src/main/res/drawable/ic_ellipsis_horizontal.xml
+++ b/modules/services/images/src/main/res/drawable/ic_ellipsis_horizontal.xml
@@ -1,8 +1,10 @@
 <vector xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
     android:width="24dp"
     android:height="24dp"
     android:viewportWidth="24"
-    android:viewportHeight="24">
+    android:viewportHeight="24"
+    tools:ignore="UnusedResources">
 
     <path
         android:fillColor="#FF000000"
--- a/modules/services/images/src/main/res/drawable/ic_ellipsis_horizontal.xml
+++ b/modules/services/images/src/main/res/drawable/ic_ellipsis_horizontal.xml
@@ -1,0 +1,11 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
+
+    <path
+        android:fillColor="#FF000000"
+        android:pathData="M4,12a2,2 0 1,0 4,0a2,2 0 1,0 -4,0M10,12a2,2 0 1,0 4,0a2,2 0 1,0 -4,0M16,12a2,2 0 1,0 4,0a2,2 0 1,0 -4,0" />
+
+</vector>
--- a/modules/services/localization/src/main/res/values/strings.xml
+++ b/modules/services/localization/src/main/res/values/strings.xml
@@ -268,6 +268,7 @@
     <string name="tv_playlist_empty_title">No episodes</string>
     <string name="tv_playlist_empty_subtitle">Either it’s time to celebrate completing this list or edit your rules in the mobile app to get some more.</string>
     <string name="tv_playlist_empty_subtitle_manual">Add episodes to this playlist from the mobile app and they’ll show up here.</string>
+    <string name="tv_episode_details">Episode details</string>
     <string name="unavailable">Unavailable</string>
     <string name="browse_podcasts">Browse podcasts</string>
     <string name="pocket_casts_plus_badge">Pocket Casts Plus badge</string>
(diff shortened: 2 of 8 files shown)


geekygecko (approved):
That worked really well. 

Can we make sure we are creating reusable components.

There were a few slight differences in the Apple TV version, such as not including the title in the popup and one version having a Cancel button. I'm not sure it really matters, but I thought I should mention it.

Image

sztomek:
Thanks @geekygecko, great catch on the Apple TV differences 🙏 Figma is my source of truth, though -- as far as i know that's common ground

I've brought this dialog in line with the tvOS version:
- **Removed the title** from the popup
- **Trimmed the padding** to match the Figma spec (36/40 px → 24/27 dp)
- **Added a Cancel button** to dismiss alongside the hardware back key

On reusable components — agreed. The modal container (`TvModal`) is already shared, and the action-button styling will be pulled out into a reusable `TvModalButton` in one of the upstream/child PRs in this stack rather than here, so the dedupe lands in one place across all the TV modals.

sztomek:
Thanks for the condensed pass — here's how I triaged the still-open items:

**Done in the latest push:**
- **Duplicated icon-button colours** — added `TvButtonDefaults.iconButtonColors(containerColor = TvColors.BgActive20)` alongside the existing `filledButtonColors()` / `prominentButtonColors()` / `borderlessButtonColors()`, and routed all three sites through it: `TvMoreButton`, `SortDropdownButton`, and `TvProfileButton` (passing `TvColors.Gray`).
- **`modifier.size(48.dp)` ordering** → `Modifier.size(48.dp).then(modifier)` so a caller can override the size.
- **`MORE_BUTTON_ANIMATION_MILLIS`** → `MORE_BUTTON_ANIMATION_DURATION_MS` to match the module's `_MS` convention.

**Deferring to the actions-wiring follow-up (deliberately, not overlooked):**
- **Mutable-entity snapshot (`actionsEpisode`)** — agreed it's the one item that becomes a real bug, but only once the played/archive toggles actually mutate the entity. The fix (hold the uuid, resolve against the live `episodes` list, dismiss if it drops out) is untestable while every action is a no-op, so it belongs with the code that makes the labels change rather than here.
- **Stringly-typed actions → `(label, onClick)`/enum** — same PR: the pairs only earn their keep once each action has a real handler + per-action analytics, and doing it there keeps this change purely UI.

**Not doing:**
- **TODO markers on the no-ops** — the no-op scope is already called out in the PR description, house style favours terse code over inline notes, and I'd rather not commit a placeholder ticket ref; the enum refactor above will make the no-ops self-evident anyway.
- **Focused-row `1.02` scale gap / `DOWN` landing on the next card** — both are intentional tvOS-parity behaviours of a focus-scoped accessory; worth a final eyeball against the spec on device, but not defects.

Separately, per @geekygecko's note I also removed the dialog title, trimmed the padding to the Figma spec, and added a Cancel button.

PR #5678, opened 2026-07-29
Open HLS videos in the player like mp4, and fix their content_type analytics

sztomek:
## Description

When you play a video episode, the player opens on its own (fullscreen in landscape). Until now this only worked for mp4 videos, not HLS ones. That's because an HLS episode isn't known to be video until its stream starts. This makes HLS videos behave the same: once the stream reports video, the player opens like it does for mp4.

Two related things came up while testing and are fixed here too:
- mp4 videos no longer open when Audio-only mode is on.
- Toggling Audio-only off mid-playback no longer pops the player open.

It also fixes the analytics. `playback_play` and `playback_source_resolved` were reporting the wrong `content_type` for HLS (`audio`, then `unknown`) because the type isn't known yet when playback starts. These events now wait for the stream to resolve and report the real value.

**Known limitation (casting):** `CastPlayer` doesn't report its stream tracks, so while casting we can't resolve an HLS stream to video. To avoid `content_type=unknown` (and events that never fire until stop), a Cast session resolves the type up front from the episode metadata. For the usual alternate-enclosure shape (audio enclosure + HLS video alternate) this means casting an HLS video reports `content_type=audio`. This restores the pre-existing behaviour and only affects the Cast path. Follow-up to make it exact: have `CastPlayer` emit `VideoTrackChanged` from the media metadata it already computes, then drop the remote short-circuit in `StreamVideoState.initialFor`.

## Testing Instructions

Best tested on a real device — video rendering is flaky on emulators. You'll need a build with HLS enabled, a Plus account, and a podcast with an HLS video episode.

**Auto-open**
1. Play an HLS video episode. The player should open by itself once the video starts (fullscreen if you're in landscape).
2. Play an mp4 video episode. Same thing should happen (existing behavior — just check it still works).
3. Play an audio-only episode. The player should not auto-open into video.

**Audio-only setting**
4. Turn Audio-only on (Settings → General, or the toolbar toggle).
5. Play an mp4 video. It should not open fullscreen and should not show video.
6. Play an HLS video. Same — no fullscreen, audio only.
7. With that HLS video playing in Audio-only, turn Audio-only off. The player should not pop open on its own. Video can start rendering in place, but nothing should jump to fullscreen.
8. Toggle Audio-only on and off a few more times while it plays. No unexpected fullscreen.

**Orientation**
9. Rotate to landscape while an HLS video plays → it should go fullscreen.
10. Send the app to the background in landscape during an HLS video, then reopen it → it should return to fullscreen.

**Analytics** (watch the Tracks debug log)
11. Play an HLS video → `playback_play` and `playback_source_resolved` should both report `content_type=video` (not `audio` or `unknown`).
12. Play an HLS audio-only stream → `content_type=audio`.
13. Play an mp4 video → `content_type=video`, fired right away.
14. Tap play then quickly skip to another episode before the first one loads → the first play should still report an event (falls back to `unknown`), nothing lost.

## Screenshots or Screencast

https://github.com/user-attachments/assets/e7556630-ae6e-4b4c-9146-7a05e70e03ea

SCR-20260729-nosy

## Checklist

- [ ] with different themes

### DIFF
--- a/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
+++ b/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
@@ -163,6 +163,7 @@ import au.com.shiftyjelly.pocketcasts.repositories.opml.OpmlImportTask
 import au.com.shiftyjelly.pocketcasts.repositories.playback.PlaybackManager
 import au.com.shiftyjelly.pocketcasts.repositories.playback.PlaybackNoticeType
 import au.com.shiftyjelly.pocketcasts.repositories.playback.PlaybackState
+import au.com.shiftyjelly.pocketcasts.repositories.playback.StreamVideoState
 import au.com.shiftyjelly.pocketcasts.repositories.playback.UpNextSource
 import au.com.shiftyjelly.pocketcasts.repositories.playlist.Playlist
 import au.com.shiftyjelly.pocketcasts.repositories.podcast.EpisodeManager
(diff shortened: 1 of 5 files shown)


geekygecko (approved):

sztomek:
SCR-20260731-jvur


`audio_only_mode` is now being reported on other playback events

PR #5680, opened 2026-07-29
[TV] Add the Your Podcasts tab

sztomek:
## Description

Adds the **Your Podcasts** tab to the TV app, mirroring the Apple TV `PodcastsView`. This PR covers the two states in scope — the empty view and the populated podcasts grid. Folders are handled as a follow-up (stacked on this branch).

- **Grid:** subscribed podcasts are shown as a 6-column grid of square artwork tiles, matching tvOS's `LazyVGrid` of 6 fixed columns. Focus scales the tile and the last-focused position is restored when re-entering the grid (the same focus pattern as the Playlists tab).
- **Data:** observes `PodcastManager.findSubscribedFlow()` and sorts with `PodcastsSortType.NAME_A_TO_Z` — the article-aware A→Z order tvOS uses (`HomeGridDataHelper.gridItems(orderedBy: .titleAtoZ)`). The list is kept fresh reactively from the DB (synced at sign-in); see the note below on why the tab doesn't force its own network refresh.
- **Empty state:** "Time to fill this up" with a **Discover podcasts** button that jumps to the Home tab, matching tvOS's `ContentUnavailableView` action (`tabRouter.selectedTab = .home`).
- **Reuse:** generalized the existing `TvPodcastTile` with an `imageModifier` (defaulting to the current fixed 123.dp used by the Home rows) so the same component fills a grid cell here — this also gives each tile a proper artwork content description.

**Scope / deliberate deviations** (consistent with the rest of the TV module):
- The screen title is pinned above the grid rather than scrolling with it as on tvOS — this matches the sibling Playlists / Up Next tabs, favouring platform-internal consistency.
- Tapping a podcast is a no-op — the podcast detail screen isn't wired anywhere on TV yet, and it's out of scope for this ticket (grid + empty view only). The tiles remain focusable, matching the Home tab tiles.
- No analytics — the TV module has none; adding the tvOS `podcastsListShown` events would need an EventHorizon schema change and would be inconsistent with the sibling tabs.

Fixes PCDROID-696 https://linear.app/a8c/issue/PCDROID-696/podcasts-grid-and-empty-view.
Designs: Ftk3KwnfqaK4g57yCN63p0-fi-2052_15345.

Stacked on `feat/tv-up-next` (#5679) — merge bottom-up.

## Testing Instructions

1. Install the TV app (`./gradlew :tv:installDebugProd`) and sign in with an account that follows several podcasts.
2. Open the **Your Podcasts** tab — verify the followed podcasts appear as a 6-column grid, sorted A→Z (ignoring a leading "The"), under a "Your Podcasts" title.
3. Navigate the grid with the d-pad — the focused tile scales up; leave and re-enter the grid and verify focus returns to the last-focused tile.
4. Empty state: with an account that follows no podcasts, verify "Time to fill this up" with a focused **Discover podcasts** button; selecting it switches to the Home tab.
5. Navigate in and out of the tab repeatedly — verify it never crashes (regression check for the token-refresh crash described above).

## Screenshots or Screencast

https://github.com/user-attachments/assets/91b45bd3-2d7b-4118-b0af-b53cff9428ff

## Checklist
- [x] I have considered whether it makes sense to add tests for my changes _(added `TvYourPodcastsViewModelTest`)_
- [ ] I have updated (or requested that someone edit) [the Event Horizon schema](https://github.com/Automattic/EventHorizonSchemas/blob/trunk/schema/pocket-casts.yml) to reflect any new or changed analytics. _(analytics deliberately excluded — matches the rest of the TV module)_

### DIFF
--- a/modules/services/localization/src/main/res/values/strings.xml
+++ b/modules/services/localization/src/main/res/values/strings.xml
@@ -272,9 +272,9 @@
     <string name="tv_up_next_empty_title">Nothing queued up</string>
     <string name="tv_up_next_empty_subtitle">Add episodes to Up Next and they’ll play in order, back to back</string>
     <string name="tv_up_next_empty_action_title">Add episodes</string>
-    <string name="tv_podcasts_empty_title">Time to fill this up</string>
-    <string name="tv_podcasts_empty_subtitle">Followed podcasts show up here, ready to play.</string>
-    <string name="tv_podcasts_empty_action_title">Discover podcasts</string>
+    <string name="tv_your_podcasts_empty_title">Time to fill this up</string>
+    <string name="tv_your_podcasts_empty_subtitle">Followed podcasts show up here, ready to play.</string>
+    <string name="tv_your_podcasts_empty_action_title">Discover podcasts</string>
     <string name="unavailable">Unavailable</string>
     <string name="browse_podcasts">Browse podcasts</string>
     <string name="pocket_casts_plus_badge">Pocket Casts Plus badge</string>
--- a/modules/services/localization/src/main/res/values/strings.xml
+++ b/modules/services/localization/src/main/res/values/strings.xml
@@ -272,6 +272,9 @@
     <string name="tv_up_next_empty_title">Nothing queued up</string>
     <string name="tv_up_next_empty_subtitle">Add episodes to Up Next and they’ll play in order, back to back</string>
     <string name="tv_up_next_empty_action_title">Add episodes</string>
+    <string name="tv_podcasts_empty_title">Time to fill this up</string>
+    <string name="tv_podcasts_empty_subtitle">Followed podcasts show up here, ready to play.</string>
+    <string name="tv_podcasts_empty_action_title">Discover podcasts</string>
     <string name="unavailable">Unavailable</string>
     <string name="browse_podcasts">Browse podcasts</string>
     <string name="pocket_casts_plus_badge">Pocket Casts Plus badge</string>
--- a/tv/src/main/java/au/com/shiftyjelly/pocketcasts/component/TvPodcastTile.kt
+++ b/tv/src/main/java/au/com/shiftyjelly/pocketcasts/component/TvPodcastTile.kt
@@ -22,6 +22,7 @@ fun TvPodcastTile(
     podcastTitle: String,
     onClick: () -> Unit,
     modifier: Modifier = Modifier,
+    imageModifier: Modifier = Modifier.width(123.dp),
 ) {
     TvTile(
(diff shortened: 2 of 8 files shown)


sztomek:
@claude please review this

sztomek:
@claude review again pls

geekygecko (approved):
Looks great and worked well.

sztomek:
Rebased onto `main` now that #5679 has merged — the 8 Your Podcasts commits replayed cleanly with no conflicts.

Treating the latest review (which had no blockers, all nits):

**Fixed:**
- **Nit A — `FocusRequester` attach race.** Wrapped the `onEnter` `requestFocus()` in `runCatching`, matching `TvPlaylistsScreen.kt:207`. Agreed the `visibleItemsInfo` guard is the primary fix, but `layoutInfo` reflects the last completed pass, so a list-size change can still leave an index 'visible' with its requester unattached — the backstop closes that crash window for free.
- **Nit C — repeated `123.dp`.** Added `TvPodcastTileDefaults.RowImageWidth` and routed all four row call sites (`TvHomeScreen`, `TvTabPlaceholder` ×2, the tile preview) through it. Since this PR moved the width out of the tile's default, the named constant is the right completion — it kills the magic number and makes the zero-width footgun discoverable.

**Deferring, with reasons:**
- **Nit D — pass a UI model instead of `Podcast`.** Real, but a moderate refactor of the shared tile's data contract, and low practical impact here: the tile receives derived `artworkUrl` (from the stable `uuid`) + `title` as plain strings, and the grid is `key`-ed by `uuid`, so sync churn doesn't meaningfully re-render tiles. Better done deliberately with the `TvHomePodcast` owners than folded in here.
- **Nit B — `Modifier.focusRestorer()`.** Would delete the manual focus block, but you flagged it needs a device check and it ideally updates Playlists too — a focused follow-up.
- **Nit H / the `PodcastDao.findSubscribedFlow` ordering test** — both land outside this PR: H is a `TvPlaylistsScreen` change, and the DAO test is a model-module `androidTest` (worth doing to pin the contract the ViewModel now leans on, but as its own change).

**Rejecting:**
- **E** (`GRID_COLUMNS` placement) — stylistic; consistent with how the other TV screens place their private constants.
- **F** (no `Loading` preview) — `Loading` is a plain `LoadingView` spinner; both content-bearing states are already previewed.
- **G** (`UpNextEmpty` still hand-rolled) — that's on the already-merged Up Next branch, not this PR.

Validation: `spotlessApply` clean, `:tv:compileDebugKotlin` + `:tv:testDebugUnitTest` both green.

PR #5681, opened 2026-07-29
[TV] Add folder support to the Your Podcasts tab

sztomek:
## Description

Adds folder support to the TV **Your Podcasts** tab, mirroring the Apple TV `FolderCardView` / `FolderDetailView`. Follow-up to the podcasts grid (#5680).

- **Grid:** the top-level grid now shows folders alongside loose podcasts. Folders render as a square card in the folder's colour with a 2×2 grid of the folder's top-4 podcast covers and the folder name at the bottom — the same layout tvOS's `FolderCardView` draws (covers top, name bottom, 12dp corners).
- **Folder detail:** selecting a folder opens a detail screen — the folder name over a 6-column grid of that folder's podcasts, matching tvOS `FolderDetailView`. Back (or the empty-state button) returns to the grid.
- **Data:** the view model reuses `FolderManager.getHomeFolder()` (folders + podcasts-not-in-a-folder) and re-queries it reactively whenever folders or subscriptions change, re-sorting the items A→Z with `PodcastsSortType.NAME_A_TO_Z` to match tvOS's `titleAtoZ`. Folder covers and detail podcasts come from `FolderManager.findFolderPodcastsSorted()`.
- **Colour:** the card resolves the folder colour through the Compose theme (`LocalColors.current.colors.getFolderColor(...)`), so it uses the TV app's Extra Dark palette — the analog of tvOS `AppTheme.folderColor(colorInt:)`.
- **Reuse:** extracted the title + focus-restoring 6-column grid into a shared `TvPodcastGridScaffold`, now used by both the top-level grid and the folder detail. The new `TvFolderCard` reuses the existing `TvTile` + `TvArtworkImage` (whose placeholder already matches tvOS's empty-cover fill).
- **Empty folder:** "Your folder is empty" with an **OK** button that returns to the grid, matching tvOS's `ContentUnavailableView`.

Fixes PCDROID-697 https://linear.app/a8c/issue/PCDROID-697/folder-support.
Designs: Ftk3KwnfqaK4g57yCN63p0-fi-3246_3336.

Stacked on `feat/tv-your-podcasts` (#5680) — merge bottom-up.

## Testing Instructions

1. Install the TV app (`./gradlew :tv:installDebugProd`) and sign in with an account that has at least one folder with podcasts and some podcasts outside folders.
2. Open the **Your Podcasts** tab — verify folders and loose podcasts are interleaved A→Z; each folder card shows the folder colour, up to four covers in a 2×2 grid, and the folder name.
3. Select a folder — verify the detail screen opens with the folder name and a grid of that folder's podcasts, and the first tile is focused. Press Back — verify it returns to the grid.
4. Open a folder that has no podcasts (create one in the mobile app) — verify "Your folder is empty" with a focused **OK** button that returns to the grid.
5. Add/remove a folder or move a podcast in/out of a folder on another device — re-open the tab and verify the grid reflects the change.

## Screenshots or Screencast

https://github.com/user-attachments/assets/13d1f239-5c5f-4f0b-bf7c-23077b0f8d55

## Checklist
- [x] I have considered whether it makes sense to add tests for my changes _(extended `TvYourPodcastsViewModelTest`)_
- [ ] I have updated (or requested that someone edit) [the Event Horizon schema](https://github.com/Automattic/EventHorizonSchemas/blob/trunk/schema/pocket-casts.yml) to reflect any new or changed analytics. _(analytics deliberately excluded — matches the rest of the TV module)_

### DIFF
--- a/tv/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/TvFolderDetailScreen.kt
+++ b/tv/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/TvFolderDetailScreen.kt
@@ -7,39 +7,29 @@ import androidx.compose.animation.fadeOut
 import androidx.compose.animation.togetherWith
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Box
-import androidx.compose.foundation.layout.Column
-import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
-import androidx.compose.foundation.layout.height
-import androidx.compose.foundation.layout.widthIn
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
+import androidx.compose.runtime.rememberUpdatedState
 import androidx.compose.runtime.setValue
-import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.focus.FocusRequester
-import androidx.compose.ui.focus.focusRequester
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.res.stringResource
-import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.tooling.preview.Devices
 import androidx.compose.ui.tooling.preview.Preview
-import androidx.compose.ui.unit.dp
-import androidx.tv.material3.Button
 import androidx.tv.material3.MaterialTheme
-import androidx.tv.material3.Text
+import au.com.shiftyjelly.pocketcasts.component.TvEmptyState
+import au.com.shiftyjelly.pocketcasts.component.TvPodcastGridScaffold
 import au.com.shiftyjelly.pocketcasts.component.TvPodcastTile
 import au.com.shiftyjelly.pocketcasts.compose.AppTheme
 import au.com.shiftyjelly.pocketcasts.compose.loading.LoadingView
 import au.com.shiftyjelly.pocketcasts.models.entity.Podcast
 import au.com.shiftyjelly.pocketcasts.repositories.images.PodcastImage
-import au.com.shiftyjelly.pocketcasts.theme.TvButtonDefaults
 import au.com.shiftyjelly.pocketcasts.theme.TvColors
-import au.com.shiftyjelly.pocketcasts.theme.TvTextStyles
 import au.com.shiftyjelly.pocketcasts.ui.theme.Theme
 import au.com.shiftyjelly.pocketcasts.localization.R as LR
 
@@ -52,8 +42,9 @@ fun TvFolderDetailScreen(
     modifier: Modifier = Modifier,
 ) {
     var uiState by remember(folderUuid) { mutableStateOf<TvFolderDetailUiState>(TvFolderDetailUiState.Loading) }
-    LaunchedEffect(folderUuid, getFolderPodcasts) {
-        val podcasts = getFolderPodcasts(folderUuid)
+    val currentGetFolderPodcasts by rememberUpdatedState(getFolderPodcasts)
+    LaunchedEffect(folderUuid) {
+        val podcasts = currentGetFolderPodcasts(folderUuid)
         uiState = if (podcasts.isEmpty()) TvFolderDetailUiState.Empty else TvFolderDetailUiState.Loaded(podcasts)
     }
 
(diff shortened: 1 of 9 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

sztomek:
@claude please review this

sztomek:
@claude review again pls

tv/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/TvFolderDetailScreen.kt
```diff
@@ -0,0 +1,182 @@
...
+            Button(
+                onClick = onClose,
+                colors = TvButtonDefaults.filledButtonColors(),
+                modifier = Modifier.focusRequester(focusRequester),
+            ) {
+                Text(stringResource(LR.string.ok))
+            }
+        }
+    }
+}
```
geekygecko: Can this use a common component?

geekygecko (approved):
Looks great.

This has similar behaviour to what I mentioned on another screen. The scrollable content doesn't include the header and title. It might be nice to address that in a future PR.

sztomek:
Rebased onto `main` now that #5680 has merged. The rebase surfaced one conflict in `TvYourPodcastsScreen.kt`: this PR's `e064` extracts the grid into `component/TvPodcastGridScaffold.kt` and deletes the local copy, while main had absorbed my #5680 `runCatching` focus-crash backstop. Resolved by taking the extraction **and** re-applying the `runCatching` guard to the extracted scaffold's `onEnter` so that fix isn't lost.

Treating round 6:

**Fixed:**
- **The `TvEmptyState` consolidation (the 'one leftover').** Added an `autoFocusAction` parameter to `TvEmptyState` and deleted the ~38-line hand-rolled `TvFolderDetailEmpty`, which was line-for-line the shared component minus `ModalButtonLabel` on the button. The folder-detail empty now goes through `TvEmptyState(autoFocusAction = true)` — so it keeps its focused OK button *and* picks up the correct button typography it was missing. I defaulted `autoFocusAction = false` to avoid changing the already-merged Your Podcasts / Playlists empty states; flipping their default to auto-focus is a reasonable separate UX call if you want the D-pad to land on the action there too.
- **Grid auto-focus crash guard.** Wrapped the `autoFocusFirstItem` `requestFocus()` in `TvPodcastGridScaffold` in `runCatching` — same class as your 'worth a device check' note, and the same guard I added to `onEnter`, since a lazy grid child may not be attached when the effect runs.
- **`compose:lambda-param-in-effect` on `getFolderPodcasts`.** ktlint (locally) flags the `LaunchedEffect(folderUuid)` that references the `getFolderPodcasts` lambda directly — the very change praised as a fix last round drops it as a key. Wrapped it in `rememberUpdatedState` so the effect stays keyed on `folderUuid` only while satisfying the rule. (Moot once **C** lands and the lambda goes away, but it was blocking spotless.)

**Deferring as follow-ups — taking you up on the offer to stop re-raising A–E:**
- **A** (folder contents use the mobile sort, not A→Z) needs a product decision, so not mine to force here.
- **B** (rebuild items from `findSubscribedFlow()` grouped by `folderUuid` — kills the N+1 and resolves A for free) is the strategic fix, but it decides A and is a real ViewModel rewrite; better as its own change.
- **C/D** (pass the folder's podcast list through instead of the re-query lambda; retain focus/scroll returning from a folder) are the interrelated data-flow cleanups — C in particular deletes the whole `Loading/Empty/Loaded` machine here, worth doing deliberately with the `TvPlaylistsScreen` focus-restore work.
- **E** and the smaller nits (`FOLDER_COVER_COUNT` placement, `TvColors.getFolderColor`, the title-rename re-sort) — filing alongside the above.

Validation: `spotlessApply` clean, `:tv:compileDebugKotlin` + `:tv:testDebugUnitTest` both green.

sztomek:
Both caught correctly — thank you, and you were right on the mechanism.

**1. The 13 orphaned imports.** Confirmed: all 13 had zero body references and were still in the file, so `spotlessCheck` would have failed on Buildkite. You nailed the cause — my `spotlessApply` gave a false clean: the first run aborted on the `lambda-param-in-effect` error *before* writing the import auto-fixes, and the re-run served a stale up-to-date result for the file. Removed all 13, and this time I verified with `./gradlew spotlessApply spotlessCheck --rerun-tasks` (observed green, no stale caching) rather than trusting `apply`.

**2. `runCatching` changed the failure mode on the two auto-focus sites.** Agreed — on `onEnter` it's correct (user event, grid already placed), but on the two `LaunchedEffect(Unit)` sites it was swallowing the normal first-frame not-yet-attached case with no retry, so `autoFocusFirstItem`/`autoFocusAction` could silently no-op exactly where it matters (large folder, cold cache, the `AnimatedContent` fade). Switched both to wait for placement:
- **Grid:** `snapshotFlow { gridState.layoutInfo.visibleItemsInfo.isNotEmpty() }.first { it }` before the `requestFocus()` (kept `runCatching` as a pure backstop behind the wait).
- **`TvEmptyState`:** no lazy container, so a one-shot `withFrameNanos {}` before the `requestFocus()`.

Validation this round: `spotlessApply` + `spotlessCheck --rerun-tasks` green, `:tv:compileDebugKotlin` + `:tv:testDebugUnitTest` green.

Issue #10, opened 2022-06-27
Starred Podcasts Menu does not load anything


joashrajin:
When on the Beta (7.20), if you go to Profile > Starred, then click the ellipsis to open the menu, there is no menu that opens up. 

Screen Recording: 

https://user-images.githubusercontent.com/16253818/175949537-1cbf03b9-e9e1-42ff-9c75-c1257a52820b.mp4

Reported at https://pocketcastsbeta.slack.com/archives/C013F9S9C2W/p1656327446320359

Replicated on:

```
App version : 7.20.0 (8085)
Phone: samsung - R9s - SM-G990E - r9s
Kernel: 5.4.129-23694594-abG990EXXS2CVC8 - 21DJF604
Android version: 12 SDK 31
```

PR #5683, opened 2026-07-30
[TV] Add the podcast details screen

sztomek:
## Description

Adds a **podcast details screen** to the Android TV app, ported from the tvOS Apple TV app: **podcast info on the left pane, episode list on the right pane**. Previously, tapping a podcast tile anywhere in the TV app was a no-op.

This PR is scoped to **UI + data integration**. Left for follow-ups: wiring the CTA actions and analytics, and the show/hide-archived + sort controls. Archived episodes are **hidden by default**, matching the tvOS default state.

What's included:
- `TvPodcastDetailsViewModel` — resolves the podcast via `PodcastManager.findOrDownloadPodcastRxSingle(uuid, waitForSubscribe = false)` (finds it locally, or fetches + persists it unsubscribed from the server, mirroring tvOS `loadPodcast`), then observes `EpisodeManager.findEpisodesByPodcastOrderedFlow` (ordered by the podcast's own sort type) and filters out archived episodes. Exposes a `Loading` / `NotFound` / `Loaded` `StateFlow`.
- `TvPodcastDetailsScreen` — a two-pane layout mirroring the sibling `TvPlaylistDetailsScreen`:
  - **Left pane**: hero artwork, author, title, a 3-line description, and two focusable action buttons — **Follow this podcast / Following** (reflects `Podcast.isSubscribed`, with plus/check icon like tvOS) and **More info**. The button click handlers are no-ops for now (deferred CTAs).
  - **Right pane**: an **All episodes** title above a `LazyColumn` of the reused `TvEpisodeListItem`. Each focused row shows the **⋯** menu (opens the shared `TvEpisodeActionsModal`, same as playlist/up-next; its actions are no-op stubs for now).
- Wired **every** podcast tile to open the screen: the Your Podcasts grid, folder detail, and the Home rows. Because it resolves unsubscribed podcasts from the server, discover podcasts from Home open correctly too.
- Shrunk the details hero artwork and made it **uniform across the podcast and playlist details screens** via a shared `TvDetailsArtworkSize` token.

Because the details screen is opened by uuid from several entry points, it is nested per-tab (like the existing playlist/folder detail screens) so that Back returns to wherever you came from (e.g. a folder), and switching tabs behaves correctly.

Fixes PCDROID-699 https://linear.app/a8c/issue/PCDROID-699/episode-list
Figma: Ftk3KwnfqaK4g57yCN63p0-fi-2052_16040

## Testing Instructions

1. Sign in with an account that is subscribed to some podcasts
2. Open the TV app and go to **Your Podcasts**.
3. Move focus into the grid and select any podcast → the details screen opens: podcast info + Follow/More info buttons on the left, "All episodes" list on the right. Archived episodes are not shown.
4. From the episode list, press **D-pad left** → focus moves to the Follow button; **right** → More info. Both are focusable.
5. Focus an episode and open its **⋯** menu → the actions modal appears (actions are no-op for now).
6. Press **Back** → you return to the grid.
7. Open a **folder**, select a podcast inside it, then press **Back** → you return to the folder (not the top-level grid).
8. On the **Home** tab, select a podcast in a "podcasts" row → its details open (fetched from the server if you don't follow it).

## Screenshots or Screencast

Screenshot_20260730_145001

## Checklist

- [x] with different themes 

### DIFF
--- a/tv/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/TvPodcastDetailsScreen.kt
+++ b/tv/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/TvPodcastDetailsScreen.kt
@@ -160,7 +160,7 @@ private fun PodcastInfo(
                 )
             }
         }
-        Column(verticalArrangement = Arrangement.spacedBy(8.dp)) {
+        Row(horizontalArrangement = Arrangement.spacedBy(8.dp)) {
             Button(
                 onClick = {},
                 colors = TvButtonDefaults.filledButtonColors(),
@@ -258,7 +258,7 @@ private fun NoEpisodes(modifier: Modifier = Modifier) {
     }
 }
 
-private val InfoPaneWidth = 280.dp
+private val InfoPaneWidth = 380.dp
 
 @Preview(device = Devices.TV_1080p)
 @Composable
--- a/tv/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/TvPodcastDetailsScreen.kt
+++ b/tv/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/TvPodcastDetailsScreen.kt
@@ -132,7 +132,7 @@ private fun PodcastInfo(
         modifier = modifier,
     ) {
         TvArtworkImage(
-            model = PodcastImage.getArtworkUrl(size = null, uuid = podcast.uuid, isWearOS = false),
+            model = PodcastImage.getMediumArtworkUrl(podcast.uuid),
             modifier = Modifier
                 .size(TvDetailsArtworkSize)
                 .clip(RoundedCornerShape(8.dp)),
@@ -172,7 +172,7 @@ private fun PodcastInfo(
                     modifier = Modifier.size(18.dp),
                 )
                 Spacer(Modifier.width(8.dp))
-                Text(stringResource(if (podcast.isSubscribed) LR.string.tv_podcast_following else LR.string.tv_podcast_follow))
+                Text(stringResource(if (podcast.isSubscribed) LR.string.podcast_subscribed else LR.string.tv_podcast_follow))
             }
             Button(
                 onClick = {},
--- a/tv/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/TvPodcastDetailsScreen.kt
+++ b/tv/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/TvPodcastDetailsScreen.kt
@@ -171,7 +171,6 @@ private fun EpisodeList(
                 episode = episode,
                 dateFormatter = dateFormatter,
                 onClick = {},
-                onOpenActions = {},
                 episodeFocusRequester = firstEpisodeFocusRequester.takeIf { index == 0 },
             )
         }
--- a/tv/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/TvPodcastDetailsScreen.kt
+++ b/tv/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/TvPodcastDetailsScreen.kt
@@ -1,0 +1,230 @@
+package au.com.shiftyjelly.pocketcasts.podcasts
+
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.fillMaxHeight
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.layout.width
+import androidx.compose.foundation.layout.widthIn
+import androidx.compose.foundation.lazy.LazyColumn
+import androidx.compose.foundation.lazy.itemsIndexed
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.remember
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.focus.FocusRequester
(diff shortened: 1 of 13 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

sztomek:
@claude please review this

sztomek:
@claude please review again

tv/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/TvPodcastDetailsScreen.kt
```diff
@@ -0,0 +1,291 @@
...
+            Text(
+                text = stringResource(LR.string.podcast_no_episodes),
+                style = MaterialTheme.typography.bodyLarge,
+                color = TvColors.TextSecondary,
+                textAlign = TextAlign.Center,
+                modifier = Modifier.widthIn(max = 400.dp),
+            )
+        }
+    }
+}
```
geekygecko, line 256: Can this be a reusable component?

tv/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/TvPodcastDetailsScreen.kt
```diff
@@ -0,0 +1,291 @@
...
+            model = PodcastImage.getMediumArtworkUrl(podcast.uuid),
+            modifier = Modifier
+                .size(TvDetailsArtworkSize)
+                .clip(RoundedCornerShape(8.dp)),
+        )
+        Column(verticalArrangement = Arrangement.spacedBy(6.dp)) {
+            if (podcast.author.isNotBlank()) {
+                Text(
+                    text = podcast.author,
+                    style = TvTextStyles.PlaylistCardCaption,
```
geekygecko: It doesn't seem right using a Playlist text style here. How about Caption like the design?

Image

tv/src/main/java/au/com/shiftyjelly/pocketcasts/podcasts/TvPodcastDetailsScreen.kt
```diff
@@ -0,0 +1,291 @@
...
+            }
+            Text(
+                text = podcast.title,
+                style = TvTextStyles.ScreenTitle,
+                color = Color.White,
+            )
+            if (podcast.podcastDescription.isNotBlank()) {
+                Text(
+                    text = podcast.podcastDescription,
+                    style = TvTextStyles.PlaylistCardCaption,
```
geekygecko: Same as above

geekygecko (approved):
Looks good, I only had some minor comments.

sztomek:
Thanks @geekygecko — all three addressed (rebased onto `main` first now that #5681 merged):

**Caption text style (the two `PlaylistCardCaption` usages).** Agreed the Playlist-scoped style was the wrong fit. Added a proper `TvTextStyles.Caption` from the Figma spec (25px / weight 510 / 32px line-height → **17sp / FontWeight(510) / 21sp** at the 1.5 TV ratio) and used it for both the author and the description.

**Reusable component (the `NoEpisodes` block).** Rather than add another one-off, I generalised the existing shared `TvEmptyState` to make its action button optional, and the podcast details empty state now reuses it — so there's one component for all the centered title+subtitle(+optional CTA) empty states. (The playlist screen still has its own near-identical `NoEpisodes`; I left that swap out of this PR to keep the diff off a shared screen, but it's the obvious next dedupe.)

One thing worth your call: the Figma values you'd expect also specify a larger details **title** (57px → 38sp) and a matching info row. I only touched the captions here since that's what was flagged and the title currently uses the shared `ScreenTitle`; happy to align the full header across both detail screens as a follow-up if you want it.

sztomek:
Round-6 bot review — treatment (rebased onto `main`; the base's duplicated focus effect you flagged is already gone there via #5679's fix, and the merge kept main's `iconButtonColors`/`_MS`/`RowImageWidth` cleanups):

**Fixed:**
- **⋯ button in the overscan margin.** Right call — the trailing more button sat flush at `end = 32.dp`, inside the ~5% safe area. Bumped to `end = 56.dp` to match the sibling playlist screen.
- **`NoEpisodes` duplication (your #5, and geekygecko's ask).** Folded into the generalised `TvEmptyState`.
- **Previews (#9).** Added the `isSubscribed = true` "Following" + check-icon variant and the empty-episodes variant.

**Deferring, with reasons:**
- **#1 ViewModel one-shot snapshot / #6 no refresh-on-open** — both are the follow-up wiring (the CTAs are still `onClick = {}`); `podcastByUuidFlow` + `flatMapLatest` + a `refreshPodcast` call land together when Follow/More-info get wired.
- **#2 all-archived → generic empty / #3 network-error state / #11 empty-branch focus** — need the VM to expose an archived/error signal; distinct states with retry are their own change.
- **#4 `rememberSaveableStateHolder`** — this is the meaningful UX one (Back from details drops grid/feed focus+scroll), but the fix spans the `TvHomeScreen`/`TvYourPodcastsScreen` hosts, so it's better as a focused change copying `TvPlaylistsScreen`'s `SaveableStateProvider` than buried here.
- **#5 `EpisodeList` dedup** — did `NoEpisodes`; the full `EpisodeList`/nested-screen extraction is bigger and pairs naturally with the play/actions wiring.
- **#7 HTML description** — real (entities render literally), but swapping to `podcastHtmlDescription` + `AnnotatedString.fromHtml` in a 3-line clamp wants a device check; follow-up.
- **#8 `TvDimensions` bare val → object** — trivial, but the rename touches the shared playlist screen too; folding into a cleanup rather than churning the stack for a naming nit.
- **#10 test gaps** — coupled to #1 (the query hand-off changes when the flow lands).

On the `hasRequestedInitialFocus` note: this screen's effect keys on `Unit`, not `episodes.isNotEmpty()`, and `EpisodeList` only composes in the non-empty branch — so it already runs once per composition and can't re-fire the way the base's `isNotEmpty()`-keyed one could. I read the guard as unnecessary here; happy to add it for symmetry if you'd rather.

Validation: `spotlessCheck --rerun-tasks` green (verified, not `apply`), `:tv:compileDebugKotlin` + `:tv:testDebugUnitTest` green.

PR #5684, opened 2026-07-30
[TV] Show archived and sort options on podcast details

sztomek:
## Description

Brings the Apple TV podcast-details controls to Android TV: the **All episodes** header now hosts an **episode sort order** menu and a **show / hide archived** filter, mirroring `PodcastDetailView` on tvOS.

Persistence matches Apple TV exactly:
- **Sort order** persists to the podcast's **synced** `episodesSortType` (same as tvOS and mobile Android) — reflects across devices.
- **Show / hide archived** is stored **locally** in a TV-only preference, exactly like Apple TV's `UserDefaults` flag. It does not touch the podcast's synced `showArchived`, so toggling it on the TV doesn't change the phone. The preference is wiped on sign-out.

This PR also **consolidates the TV local preferences**: the previous per-feature `TvPlaylistPreferences` / `TvPodcastPreferences` are merged into one `TvPreferences` manager (single backing store, one `clearAll()`), used by both the playlist and podcast detail view models.

The controls are the same components the TV **playlist details** screen already uses. Those were private to that screen, so this PR extracts them into shared components (`TvArchivedFilterButton`, and a generic `TvSortButton<T>`) and points the playlist screen at them — no behaviour change there.

### What changed
- **`TvEpisodeListControls.kt`** (new) — shared `TvArchivedFilterButton` + generic `TvSortButton<T>`, extracted from the playlist screen.
- **`EpisodesSortTypeLabel.kt`** (new, `compose`) — `EpisodesSortType.displayLabel()`, mirroring the existing `PlaylistEpisodeSortType` helper, reusing the existing `episode_sort_*` strings.
- **`TvPreferences.kt`** (new) — unified TV-local preference manager (replaces `TvPlaylistPreferences` + `TvPodcastPreferences`); holds the per-playlist and per-podcast archived flags; cleared on sign-out via `TvSignOutManager`.
- **`TvPodcastDetailsViewModel`** — episodes come from `combine(podcastByUuidFlow(uuid).filterNotNull().flatMapLatest { … }, isShowingArchivedFlow)`, so a sort change (synced podcast) or an archived toggle (local flow) re-filters reactively. `changeSortType` persists to the podcast; `toggleArchiveFilter` flips the local preference. `Loaded` exposes `archivedEpisodeCount` and `isShowingArchived`.
- **`TvPodcastDetailsScreen`** — the “All episodes” title becomes a header row with the archived + sort controls (left-focus wired to the Follow button). When every episode is archived and the filter hides them, the controls stay reachable and an “all archived” message is shown instead of a dead-end empty state.

### Notes / out of scope
- **No analytics events** were added — the existing TV view models (podcast + playlist details) don’t track anything, so adding events here would be inconsistent. Easy to add if wanted (iOS fires `podcast_screen_toggle_archived` / `podcasts_screen_sort_order_changed`).
- Sort options are shown in the same order the mobile app uses: Title A–Z, Title Z–A, Newest→Oldest, Oldest→Newest, Shortest→Longest, Longest→Shortest.

## Testing Instructions

1. Open the Android TV app and go to a followed podcast → **podcast details**.
2. In the **All episodes** header, open the **sort** button (icon on the right): pick a different order → the list reorders and the choice is checked/persisted (reopen to confirm; it also reflects on the phone app).
3. Open the **show / hide archived** dropdown: toggle it → archived episodes appear / disappear; the choice persists across reopen.
4. From either control, press **left** on the D-pad → focus returns to the **Follow** button.
5. On a podcast where every episode is archived: with archived hidden you get an “all N episodes archived” message (not an empty dead-end) and can still open the archived dropdown to show them again.

## Checklist
- [x] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml` (reused existing strings; none added)
- [ ] I have updated (or requested that someone edit) the Event Horizon schema to reflect any new or changed analytics. 

- [ ] with different themes 


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

sztomek:
@claude review this pls

tv/src/main/java/au/com/shiftyjelly/pocketcasts/preferences/TvPreferences.kt
```diff
@@ -0,0 +1,40 @@
...
+    fun setPlaylistShowingArchived(playlistUuid: String, isShowingArchived: Boolean) {
+        prefs.edit().putBoolean(playlistArchivedKey(playlistUuid), isShowingArchived).apply()
+    }
+
+    fun isPodcastShowingArchived(podcastUuid: String): Boolean {
+        return prefs.getBoolean(podcastArchivedKey(podcastUuid), false)
+    }
+
+    fun setPodcastShowingArchived(podcastUuid: String, isShowingArchived: Boolean) {
+        prefs.edit().putBoolean(podcastArchivedKey(podcastUuid), isShowingArchived).apply()
```
geekygecko, line 29: Do we need this preference class? We have this on the podcast entity.

geekygecko (approved):

sztomek:
Thanks @geekygecko — on the `TvPreferences` question: it's a deliberate split rather than a duplicate of the entity.

- **Sort order** *does* go on the synced `Podcast` entity (`changeSortType` → `updateEpisodesSortTypeBlocking`), exactly as you'd expect — that's shared with mobile.
- **The archived filter** is the one that's local (`TvPreferences.isPodcastShowingArchived`). The entity's `showArchived` is a *synced* field, so writing the TV toggle there would flip "show archived" on the user's phone too. The commit that introduced this ("Store the TV podcast archived filter locally instead of on the synced podcast") was specifically to keep the TV's view-filter from syncing back — it's a per-device UI preference, not user data.

So `TvPreferences` earns its place for that one setting; happy to reconsider if you'd rather the TV filter *did* sync to the entity (that's a product call). One fair follow-up the review bot also raised: the class currently lives in the `…pocketcasts.preferences` package shared with `:services:preferences`, so `…tv.preferences` would read cleaner — I can move it if we keep the class.

sztomek:
Round-6 bot review — treatment (rebased onto `main` first now that #5683 merged; the conflicts were the sort/archived-control extraction vs. main's `5683` treatment — resolved to keep both: the extracted `TvEpisodeListControls` **and** main's `Caption`/`TvEmptyState`/`RowImageWidth` changes, and the all-archived branch now uses the shared `TvEmptyState` for the truly-empty case while keeping `AllEpisodesArchived` for the filtered case).

**Fixed — E (the one you'd push for):** added `a non-sort podcast change does not re-run the episodes query`. It emits a podcast differing only in `title` (same `episodesSortType`) and asserts `findEpisodesByPodcastOrderedFlow` is called **once** — so `distinctUntilChangedBy { it.episodesSortType }` is now pinned; the test goes red (times 2) the moment that key regresses. The existing `times(2)` sort-change test stays as the positive case.

**Deferring, with reasons (all the optional items):**
- **B — `podcastFlow` collected twice.** Real (two Room executions + two invalidation observers), but `shareIn` inside the `flow { }` builder has enough lifecycle subtlety (scope, replay, WhileSubscribed interplay with the outer `stateIn`) that I'd want a device check rather than land it blind; it's a contained follow-up.
- **C — intermediate `(new podcast, old episodes)` emission.** You confirmed it self-corrects in a frame and doesn't defeat the scroll reset; cosmetic, and it collapses together with B.
- **4 — `podcast_no_episodes_all_archived` is `%d`, not a plural** ("All 1 episodes"). Genuine grammar bug, but the fix adds a new plural string (GlotPress) against the PR's no-new-strings premise — flagging for your call rather than sneaking a string in.
- **5 — `AllEpisodesArchived` duplicated across both screens.** Right that it belongs in `TvEpisodeListControls` with a `text` param; deferring only to keep this rebase off a third shared screen, and it pairs naturally with #4 (the shared version can take a plural).
- **G — unguarded scroll reset.** You noted it's arguably better here (it makes the initial-focus request reachable after a restore); I read it the same way, so leaving it as the intentional choice.
- **H — `hasRequestedInitialFocus = episodes.isEmpty()` reads oddly.** It mirrors the sibling screen and is covered by tests; skipping the clarifying comment to stay consistent with it (and the house preference for terse code).
- **8 — `tv_playlists` → `tv_preferences` file rename + orphaned file.** Non-issue pre-release (nothing shipped to wipe); noted for whenever storage compat matters.

Validation: `spotlessCheck --rerun-tasks` green (observed), `:tv:compileDebugKotlin` + `:tv:testDebugUnitTest` green.

sztomek:
Fixed **#4** (the `All 1 episodes` plural) after all.

`podcast_no_episodes_all_archived` turned out to be shared with mobile (`PodcastAdapter.kt:541`) and already translated into ~20 locales, so converting it in place would have broken the mobile call site and reset its translations. Instead I added a **new TV-specific plural** `tv_podcast_all_archived` mirroring the sibling `tv_playlist_all_archived`:

```xml
<plurals name="tv_podcast_all_archived">
    <item quantity="one">The only episode of this podcast has been archived.</item>
    <item quantity="other">All %1$d episodes of this podcast have been archived.</item>
</plurals>
```

and `AllEpisodesArchived` now uses `pluralStringResource(...)`, so a 1-archived podcast reads "The only episode of this podcast has been archived." English-only per the localization guidelines; GlotPress will pull the translations.

Validation: `:tv:compileDebugKotlin` + `:tv:testDebugUnitTest` green, `spotlessCheck --rerun-tasks` green.

sztomek:
Both of these are right — fixed.

**Inlined icon-button colours in the extraction.** Good catch — `TvSortButton` had copy-pasted the four `IconButtonDefaults.colors(...)` values that are exactly `TvButtonDefaults.iconButtonColors()`. Swapped it back to the shared helper; `IconButtonDefaults`, `TvColors` **and** `Color` all became unused in this file (no other references), so removed all three imports. `spotlessCheck --rerun-tasks` is green.

**`changeSortType` gate.** Confirmed `PodcastManagerImpl.updateEpisodesSortTypeBlocking` only forwards `podcast.uuid` to the DAO's targeted `UPDATE … WHERE uuid`, so the `(uiState.value as? Loaded)?.podcast ?: return` read was dead weight that could silently drop a tap (state reverts to `Loading` under `WhileSubscribed(replayExpiration = ZERO)`). Dropped it and pass `Podcast(uuid = podcastUuid)` straight through using the constructor arg. Kept it contained in the TV VM rather than adding a uuid overload to the shared `PodcastManager`; happy to switch to the overload if you'd prefer not to fabricate the entity. Updated the persist test to verify against `Podcast(uuid = "podcast-uuid")` accordingly.

`:tv:compileDebugKotlin` + `:tv:testDebugUnitTest` green.

PR #5685, opened 2026-07-30
[TV] More info modal on the podcast details screen

sztomek:
## Description
Adds the **More info** modal to the Android TV podcast details screen, porting the Apple TV design. Tapping **More info** opens a glassy blur-behind dialog with:

- A **fixed left panel**: podcast artwork plus a metadata list — **Network** (podcast author), **Website** (short URL), **Schedule** (release frequency), and **Next episode** (upcoming date). Each row is hidden when its value is unavailable.
- A **scrollable right panel**: the network/author caption, the podcast title, and the full podcast description, which scrolls with the D-pad while the left panel stays put.

The modal reuses the existing glassy `TvModal` surface (as `TvEpisodeActionsModal` / `TvProfileModal` do), opened at a wider fixed width. The previously no-op **More info** button now opens it.

Fixes PCDROID-700 https://linear.app/a8c/issue/PCDROID-700/more-info
Design: Ftk3KwnfqaK4g57yCN63p0-fi-3279_2008

## Testing Instructions
1. On the Android TV app, open **Your Podcasts** and select a followed podcast (e.g. *BBC Inside Science* or *The Daily*).
2. Focus the **More info** button (next to Follow) and press select.
3. Confirm the modal shows artwork + Network / Website / Schedule / Next episode on the left, and the network caption + title + description on the right.
4. With the modal open, press D-pad **up/down** — a long description scrolls while the left metadata panel stays fixed.
5. Press **Back** to dismiss.

## Screenshots or Screencast

https://github.com/user-attachments/assets/d8837178-86ef-4c97-bfc0-b1bdb31e0b6c

## Checklist
- [ ] I have updated (or requested that someone edit) the Event Horizon schema to reflect any new or changed analytics.

### DIFF
--- a/tv/src/main/java/au/com/shiftyjelly/pocketcasts/component/TvPodcastInfoModal.kt
+++ b/tv/src/main/java/au/com/shiftyjelly/pocketcasts/component/TvPodcastInfoModal.kt
@@ -1,7 +1,6 @@
 package au.com.shiftyjelly.pocketcasts.component
 
 import android.content.Context
-import android.text.format.DateFormat
 import android.text.format.DateUtils
 import androidx.compose.foundation.focusable
 import androidx.compose.foundation.gestures.animateScrollBy
@@ -34,20 +33,18 @@ import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import androidx.tv.material3.MaterialTheme
 import androidx.tv.material3.Text
 import au.com.shiftyjelly.pocketcasts.compose.AppTheme
+import au.com.shiftyjelly.pocketcasts.localization.helper.RelativeDateFormatter
 import au.com.shiftyjelly.pocketcasts.models.entity.Podcast
 import au.com.shiftyjelly.pocketcasts.repositories.images.PodcastImage
 import au.com.shiftyjelly.pocketcasts.theme.TvColors
 import au.com.shiftyjelly.pocketcasts.theme.TvTextStyles
 import au.com.shiftyjelly.pocketcasts.ui.theme.Theme
-import java.text.SimpleDateFormat
-import java.util.Calendar
-import java.util.Date
-import java.util.Locale
 import kotlinx.coroutines.launch
 import au.com.shiftyjelly.pocketcasts.localization.R as LR
 
@@ -86,6 +83,7 @@ private fun PodcastInfoPane(
     modifier: Modifier = Modifier,
 ) {
     val context = LocalContext.current
+    val dateFormatter = remember(context) { RelativeDateFormatter(context) }
     Column(
         verticalArrangement = Arrangement.spacedBy(24.dp),
         modifier = modifier,
@@ -103,7 +101,7 @@ private fun PodcastInfoPane(
             )
             InfoRow(
                 label = stringResource(LR.string.tv_podcast_info_website),
-                value = podcast.podcastUrl?.takeIf { it.isNotBlank() },
+                value = podcast.getShortUrl().takeIf { it.isNotBlank() },
             )
             InfoRow(
                 label = stringResource(LR.string.tv_podcast_info_schedule),
@@ -111,7 +109,7 @@ private fun PodcastInfoPane(
             )
             InfoRow(
                 label = stringResource(LR.string.tv_podcast_info_next_episode),
-                value = podcast.displayableNextEpisode(context),
+                value = podcast.displayableNextEpisode(context, dateFormatter),
             )
         }
     }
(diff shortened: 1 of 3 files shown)


sztomek:
@claude please review this

tv/src/main/java/au/com/shiftyjelly/pocketcasts/component/TvPodcastInfoModal.kt
```diff
@@ -0,0 +1,277 @@
...
+        return null
+    }
+    val now = System.currentTimeMillis()
+    return when {
+        expectedDate.time < now - 7 * DateUtils.DAY_IN_MILLIS -> null
+        DateUtils.isToday(expectedDate.time) -> context.getString(LR.string.today)
+        DateUtils.isToday(expectedDate.time - DateUtils.DAY_IN_MILLIS) -> context.getString(LR.string.tv_podcast_next_episode_tomorrow)
+        expectedDate.time < now -> context.getString(LR.string.tv_podcast_next_episode_soon)
+        expectedDate.time < now + 6 * DateUtils.DAY_IN_MILLIS -> expectedDate.formatSkeleton("EEEE")
+        else -> expectedDate.formatSkeleton(if (expectedDate.isSameYearAsNow()) "dMMMM" else "yMMMMd")
```
geekygecko: It would be good if we didn't rewrite this logic again. Is it possible to use the phone version? https://github.com/Automattic/pocket-casts-android/blob/615dbf4ea0180eb09f265b945a6a3a96ebf09ec3/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/entity/Podcast.kt#L308

sztomek:
Round-6 bot review — treatment (rebased onto `main` first now that #5684 merged; clean rebase, no conflicts).

**Fixed — #3 (the duplication half, which is also @geekygecko's point):** `displayableNextEpisode`'s custom weekday/skeleton-date formatting now delegates to the shared `RelativeDateFormatter` instead of reimplementing it (see the reply above). The remaining half of #3 — lifting the whole thing into `TvPodcastDetailsUiState` so composition is idempotent (no `System.currentTimeMillis()` in the composable) and unit-testable — is a bigger refactor (inject `Context`/resources into the VM, add fields to `Loaded`, wire the modal to take strings); deferring that as a focused follow-up now that the duplication itself is gone.

**Deferring the rest, with reasons:**
- **#1 held-D-pad scroll crawls** (`animateScrollBy` cancelled by the next repeat). Real, and viewer-noticeable, but the fix (accumulate a target + `animateScrollTo(coerceIn)`, derive the step from `viewportSize`) changes scroll feel and wants tuning on a device rather than landing blind.
- **#2 left pane can overflow `InfoModalHeight`** when a value wraps to two lines. It's a genuine tension, but the PR deliberately *reduced* the modal to 336dp — bumping it back to ~392dp or forcing `maxLines = 1` on the values is a design call (which value is allowed to wrap?), so flagging for a design decision rather than picking one.
- **#4 preview re-implements the Row layout** — maintainability; the `…ModalContent` extraction pairs naturally with the #3 lift.
- **Minors** (`tomorrow` generic word behind a TV key, raw-HTML description vs `HtmlCompat`, `InfoArtworkSize` vs `TvDetailsArtworkSize`, author rendered in both panes) — mostly design/pre-existing; happy to pick any up if you want them.

Note the `context.getString` at the helper call sites is correct — those are non-`@Composable` `Podcast` extensions, so they can't use `stringResource`.

Validation: `spotlessCheck --rerun-tasks` green, `:tv:compileDebugKotlin` + `:tv:testDebugUnitTest` green.

sztomek:
Thanks for the branch-by-branch verification of the formatter swap — and good catch on the allocation.

**Fixed — formatter allocated per recomposition.** `RelativeDateFormatter(context)` was being constructed inside `displayableNextEpisode` (called straight from `PodcastInfoPane`'s body), so a fresh instance + its 4 `Calendar`s built on every recomposition. Now `remember(context) { RelativeDateFormatter(context) }` in the pane and passed into the helper, matching the `TvPodcastDetailsScreen.kt:233` convention. (Went with `remember` over threading a parameter through `TvPodcastInfoModal` to keep the modal's signature to just `podcast` — the details screen's instance isn't in scope here anyway.)

**Still deferring, with reasons:**
- **#1 held-D-pad scroll crawl** — genuine and viewer-noticeable, but the fix (accumulate target + `animateScrollTo(coerceIn)`, step from `viewportSize`) changes scroll feel and I'd rather tune it on a device than land it blind; also relates to the `maxValue == 0` "dead D-pad" minor, so worth doing together on a device pass.
- **#2 left-pane overflow at 2 lines** — the PR deliberately shrank the modal to 336dp, so "bump to ~392dp" vs "cap values at `maxLines = 1`" is a design call on which value is allowed to wrap; flagging for design rather than picking one.
- **#3 lift into `TvPodcastDetailsUiState`** — the duplication half is done (shared formatter); the idempotent-composition + testability half is a real refactor that touches phone code (your `NextEpisodeEstimate` sealed-type sketch is the right shape). Happy for that to be a separate issue as you suggested.
- **#4 preview extraction** — maintainability; pairs with #3.
- Minor `context.getString` at the helper sites is correct — those are non-`@Composable` `Podcast` extensions, so `stringResource` isn't available.

geekygecko (approved):

PR #5688, opened 2026-07-31
[TV] Add episode details modal

sztomek:
## Description

Adds a TV **episode details modal** and wires it to the "Episode details" action in the episode actions menu. The modal mirrors the existing `TvPodcastInfoModal`: it's built on the shared `TvModal`, shows a header (artwork + podcast name + episode title + "date · duration"), an "Episode details" heading, and a **D-pad-scrollable** description (per Figma, the description scrolls within the modal).

The description is the episode's **show notes**, loaded on demand via the shared `ShowNotesManager` (the stored `episodeDescription` field is frequently empty) and rendered with the shared `HtmlText` component. While the notes load a spinner is shown; if they can't be found the modal shows an "Unable to find show notes" message. The podcast name is resolved from the episode's `podcastUuid` via a small `TvEpisodeInfoViewModel`, so nothing extra has to be threaded through the list screens.

Selecting "Episode details" **replaces** the actions menu (the menu closes and the details modal opens); Back returns to the episode list. It is reachable from every current entry point that surfaces the episode actions menu:
- Podcast details episode list
- Up Next
- Playlist details

Fixes PCDROID-703 https://linear.app/a8c/issue/PCDROID-703/episode-details-modal

## Testing Instructions

1. Sign in on an Android TV device/emulator with some subscribed podcasts and a non-empty Up Next.
2. Go to **Your Podcasts** → open a podcast → focus an episode → open the **…** actions menu → select **Episode details**.
3. Confirm the modal shows artwork, podcast name, episode title, "date · duration", a brief loading spinner, then the show notes, and that they **scroll** with D-pad up/down.
4. Press **Back** — the modal closes and focus returns to the episode list.
5. Repeat from **Up Next** and from a **Playlist**'s details.
6. Open details for episodes from two different podcasts in a row and confirm the podcast name and notes are never stale.

## Screenshots or Screencast

https://github.com/user-attachments/assets/d945727a-eddf-45ae-aa0f-da9b6efc354c

## Checklist
- [x] I have updated (or requested that someone edit) the Event Horizon schema to reflect any new or changed analytics. (No analytics in this change.)

- [ ] with different themes

### DIFF
--- a/tv/src/main/java/au/com/shiftyjelly/pocketcasts/component/TvEpisodeInfoModal.kt
+++ b/tv/src/main/java/au/com/shiftyjelly/pocketcasts/component/TvEpisodeInfoModal.kt
@@ -63,7 +63,8 @@ fun TvEpisodeInfoModal(
     LaunchedEffect(episode.podcastUuid) {
         viewModel.load(episode.podcastUuid)
     }
-    val podcastTitle by viewModel.podcastTitle.collectAsStateWithLifecycle()
+    val podcastTitleState by viewModel.podcastTitle.collectAsStateWithLifecycle()
+    val podcastTitle = podcastTitleState?.takeIf { it.podcastUuid == episode.podcastUuid }?.title
     TvModal(
         onDismissRequest = onDismissRequest,
         width = EpisodeInfoModalWidth,
@@ -165,10 +166,11 @@ private fun EpisodeDescriptionPane(
             .focusable()
             .verticalScroll(scrollState),
     ) {
+        val hasDescription = description.isNotBlank()
         Text(
-            text = description,
+            text = if (hasDescription) description else stringResource(LR.string.error_loading_show_notes),
             style = TvTextStyles.FeaturedTileDescription,
-            color = Color.White,
+            color = if (hasDescription) Color.White else TvColors.TextSecondary,
         )
     }
 }
--- a/tv/src/main/java/au/com/shiftyjelly/pocketcasts/component/TvEpisodeInfoModal.kt
+++ b/tv/src/main/java/au/com/shiftyjelly/pocketcasts/component/TvEpisodeInfoModal.kt
@@ -33,7 +33,9 @@ import androidx.compose.ui.input.key.type
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.AnnotatedString
 import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.text.fromHtml
 import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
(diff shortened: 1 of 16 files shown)


tv/src/main/java/au/com/shiftyjelly/pocketcasts/component/TvEpisodeInfoModal.kt
```diff
@@ -0,0 +1,237 @@
...
+            model = episode.imageUrl ?: PodcastImage.getMediumArtworkUrl(episode.podcastUuid),
+            modifier = Modifier
+                .size(ArtworkSize)
+                .clip(RoundedCornerShape(8.dp)),
+        )
+        Column(verticalArrangement = Arrangement.spacedBy(6.dp)) {
+            if (!podcastTitle.isNullOrBlank()) {
+                Text(
+                    text = podcastTitle,
+                    style = TvTextStyles.PlaylistCardCaption,
```
geekygecko: It doesn't seem right using the playlist card style here. Should it be a generic name from the design.

geekygecko (approved):
Worked well. Thanks for including videos and screenshots in the PRs.

sztomek:
Thanks @geekygecko 🙏 — same fix as the podcast-details screen: swapped `TvTextStyles.PlaylistCardCaption` for the generic `TvTextStyles.Caption` (the style added for exactly this in #5683) on the podcast-title caption you flagged, and also on the metadata line below it since it's the same secondary-caption element. Left the episode title on `PlaylistCardTitle` since you didn't flag it and the modal's a smaller surface — shout if you'd like that generalised too.

sztomek:
Re-review treatment (rebased onto the updated #5687 branch first — the one conflict was #5676's Cancel-button `onDismissRequest` vs. this PR's `onShowEpisodeDetails` in `TvEpisodeActionsModalContent`; resolved to keep both params since the body uses both).

**Fixed — the two you'd want before merge:**
- **#2 sticky failed load.** `loadedEpisodeUuid` was set before the load and never cleared on failure, so a load that failed offline showed "Unable to find show notes" forever. Now cleared when the result is `Unavailable`, so reopening the same episode retries. Went with this over the assisted-injection restructure to keep it contained.
- **#4 artwork bypassed the episode-artwork setting.** `TvEpisodeInfoModal` now uses `PodcastImage.getMediumArtworkUrl(episode.podcastUuid)` unconditionally, matching `TvEpisodeRow` — respects `useEpisodeArtwork` and sidesteps the blank-`imageUrl` failure.

**Also done:**
- **Item 1 follow-up test** — added `blank show notes map to Unavailable and can be retried`, which pins both the `Loaded("")` → `Unavailable` mapping *and* the #2 retry (asserts `loadShowNotes` runs `times(2)` on reopen). That's the branch that regressed once before.
- **#7 `onShowEpisodeDetails` default** — dropped the `= {}` on the private content composable and passed `onShowEpisodeDetails = {}` at the preview, so a new call site can't silently no-op.
- **geekygecko** — playlist-card caption → generic `Caption`.

**Deferring, with reasons:**
- **#5 wiring duplicated 3×** (`TvEpisodeModals` extraction) and **#6 duplicated `formatSkeleton` / D-pad-scroll helpers** — both real, but they're shared-component extractions spanning three screens + the sibling info modal; better as a focused cleanup than folded into this rebase (and the D-pad one pairs with #5685's deferred scroll work). The `Action(label, onClick)` data-class nit rides along with that.
- The markup-only-notes emptiness check, the exhaustive `when`, and the test/description nits — noted; low-value or need a Figma/design call.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #5692, opened 2026-07-31
[TV] Add action toasts

sztomek:
## Description

Adds a reusable **toast** (transient notification) component to the TV app. These are the short confirmations that appear after an episode action — "Marked as played", "Will play next", "Archived", etc.

The component is wired into the episode actions modal so that selecting an option gives immediate visual feedback.

> [!IMPORTANT]
> **The underlying actions are not implemented yet — this PR is visual feedback only.** Selecting e.g. "Mark as played" shows the toast and dismisses the modal, but does not yet change the episode's played/archived state or Up Next. Wiring the real actions is a follow-up. "Go to podcast" simply dismisses the modal (no toast, matching iOS, which treats it as navigation).

### What's added
- **`TvToast.kt`** — the reusable pieces:
  - `TvToastHostState` — a small state holder with `show(message)`; keeps a single toast at a time (a new toast replaces the current one and restarts its timer).
  - `TvToastHost` — the overlay composable. Small centered text on the same glassy surface as `TvModal` (reuses `TvOverlayContainerColor`/`TvOverlayBorderColor`), pinned top-right, slides in from the right + fades in (300ms), holds 3s, then fades out with a slight downward drift — matching the iOS animation.
  - `LocalTvToastHostState` — a `CompositionLocal` so any screen (including modal `Dialog` content) can enqueue a message.
- **`TvOnboardingNavHost`** — hosts `TvToastHost` above the whole navigation graph and provides the `CompositionLocal`.
- **`TvEpisodeActionsModal`** — the action buttons now show the matching toast and dismiss the modal.
- 6 new `tv_episode_*` strings mirroring the iOS toast copy.
- Unit tests for `TvToastHostState` (show / replace / dismiss / stale-dismiss).

Stacked on #5688 (`feat/tv-episode-details-modal`).

Fixes PCDROID-701 https://linear.app/a8c/issue/PCDROID-701/wire-up-actions
Figma: Ftk3KwnfqaK4g57yCN63p0-fi-3280_3432

## Testing Instructions

1. Run the TV app on an Android TV device/emulator and sign in (or continue without an account).
2. Go to **Your Podcasts** → open a podcast → focus an episode row and open the **⋯** actions menu.
3. Select **Play next**, **Play last**, **Mark as played/unplayed**, or **Archive/Unarchive**.
4. The modal closes and a toast slides in from the top-right ("Will play next", "Marked as played", "Archived", …), then fades out after ~3s.
5. Select **Go to podcast** → the modal just closes, no toast.
6. Trigger two actions quickly → only the latest toast is shown, and its timer restarts.

## Screenshots or Screencast
Screenshot_20260731_165236

## Checklist

- [ ] with different themes

### DIFF
--- a/modules/services/localization/src/main/res/values/strings.xml
+++ b/modules/services/localization/src/main/res/values/strings.xml
@@ -278,6 +278,12 @@
     <string name="tv_playlist_empty_subtitle">Either it’s time to celebrate completing this list or edit your rules in the mobile app to get some more.</string>
     <string name="tv_playlist_empty_subtitle_manual">Add episodes to this playlist from the mobile app and they’ll show up here.</string>
     <string name="tv_episode_details">Episode details</string>
+    <string name="tv_episode_marked_as_played">Marked as played</string>
+    <string name="tv_episode_marked_as_unplayed">Marked as unplayed</string>
+    <string name="tv_episode_archived">Archived</string>
+    <string name="tv_episode_unarchived">Unarchived</string>
+    <string name="tv_episode_will_play_next">Will play next</string>
+    <string name="tv_episode_will_play_last">Will play last</string>
     <string name="tv_up_next_empty_title">Nothing queued up</string>
     <string name="tv_up_next_empty_subtitle">Add episodes to Up Next and they’ll play in order, back to back</string>
     <string name="tv_up_next_empty_action_title">Add episodes</string>
--- a/tv/src/main/java/au/com/shiftyjelly/pocketcasts/component/TvEpisodeActionsModal.kt
+++ b/tv/src/main/java/au/com/shiftyjelly/pocketcasts/component/TvEpisodeActionsModal.kt
@@ -4,6 +4,7 @@ import androidx.compose.foundation.layout.ColumnScope
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Modifier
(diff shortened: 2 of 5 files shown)


geekygecko (approved):

sztomek:
Review treatment (rebased onto the updated #5688 branch first — the conflict was the toast wiring vs. #5676's title-removal/Cancel-button + #5688's `onShowEpisodeDetails` in `TvEpisodeActionsModalContent`; resolved to keep the title removed, keep both no-default params, and take the toast wiring + the preview's `CompositionLocalProvider`).

**Done:**
- **Preview fix (nit):** `TvToastPreview` now has `device = Devices.TV_1080p` and the `AppTheme(EXTRA_DARK)` facade so it renders at TV scale in the dark theme. Left `MaterialTheme` out on purpose — `TvToastContent` uses explicit `Color.White` + standalone `TvTextStyles.ModalBody` + the top-level `TvOverlay*` colours, so it consumes nothing from `MaterialTheme`.
- **The `onDismissRequest` default nit** is already gone as a side-effect of the rebase — the private `TvEpisodeActionsModalContent` now has no defaults and the preview passes `{}` explicitly.

**Rejected — the `const val` nit:** making `ToastDurationMillis`/`ToastAnimationMillis` `const` trips ktlint's `property-naming` (const ⇒ `SCREAMING_SNAKE_CASE`), which would leave them inconsistent with the sibling `ToastMaxWidth`/`ToastShape` that *can't* be const (`dp`/`RoundedCornerShape`). Kept all four as PascalCase `val` for consistency.

**Deferring, with reasons (all non-blocking per your review):**
- **#1 `AnimatedContent`** — the backwards `lastMessage` write is a real smell and `AnimatedContent` would also give the missing replace-transition, but it changes the animation + TalkBack behaviour you flagged, so it wants a device/TalkBack pass rather than a blind swap.
- **#2 overlay owned by the onboarding graph** — agreed it should hoist into `TvActivity.setContent`/a `TvAppRoot`; that's a structural move best done deliberately (and would undo the `Box`/`fillMaxSize` indentation).
- **#3 toast behind the modal dim** — latent only (every action currently dismisses first); worth documenting on `LocalTvToastHostState` or rendering a host inside the dialog when real failure-toasts land.
- **#4 RTL slide** and the remaining nits (`internal TvToast` — cascades to the public `currentToast`; `TvSafeArea` const; `remember` around `actions`) — follow-ups.

PR #5700, opened 2026-08-05
[TV] Hide the top bar on nested detail screens

sztomek:
## Description

On Android TV the top bar (profile + tab row + logo) was shown on every screen, including nested detail screens. On the Apple TV app the tab bar is hidden on pushed detail screens. This brings Android TV to parity: the top bar is hidden on the nested detail screens so they use the full height, and it reappears when you back out.

Detail screens that now hide the top bar:
- Podcast details (opened from Home, Your Podcasts, or a folder)
- Folder details
- Playlist details

### How it works
- A small `TvTopBarVisibility` state holder (counter-based) is provided through a `CompositionLocal` (`LocalTvTopBarVisibility`) in `TvScaffold`, mirroring the existing `TvToast`/`LocalTvToastHostState` pattern in the module.
- Each nested detail is rendered through `TvDetailOverlay`: the parent grid/list **stays composed** and the detail **fades in over the top** of it as an opaque, full-height layer. The overlay drives `HideTvTopBar()` (so the bar fades out under it) and owns the Back handler, so the detail composables stay unchanged and reusable — analogous to iOS driving this from the navigation stack rather than from inside the detail views.
- The layer behind the overlay is deactivated for focus (`tvFocusInactiveWhen`) so the D-pad can't reach the hidden grid. In Your Podcasts a folder and a podcast detail stack as two overlays (the folder stays composed underneath the podcast).

### Addressing review feedback (@geekygecko)
1. **"It should open as a completely new page… faded in over the top, rather than transitioning the previous page."** The detail is now a separate full-height layer that fades in over the still-composed parent. The parent no longer reflows: the animated top-bar/content padding was removed in favour of a static top inset, so opening a detail doesn't shove the content around — it just fades the new page in on top.
2. **"The previous page scrolls to the top when you return — preserve the scroll position."** Because the parent is never removed from composition, its `LazyGrid`/`LazyList` state survives, so backing out lands you exactly where you left off. On close the overlay also fires an `onHide` callback that re-requests the previously focused tile (reusing each grid/row's existing `onEnter`/`lastFocusedKey` restoration), so **focus returns to the tile you opened** instead of jumping to the top bar.

### Layout tweak
While in here, evened out the horizontal padding on the two-pane detail screens (podcast + playlist details): they were `start = 32.dp, end = 56.dp`; both are now `32.dp`, matching the app's horizontal rhythm.

Fixes POC-792 https://linear.app/a8c/issue/POC-792/hide-top-bar-on-nested-destinations-like-playlist-details-and-podcast

## Testing Instructions

1. Launch the TV app and sign in (so Your Podcasts / Playlists have content).
2. On **Home**, **Your Podcasts**, and **Playlists** confirm the top bar is visible.
3. Scroll a tab **down** a few rows, then open a **podcast** (from Home or Your Podcasts), a **folder**, and a **playlist**. In each detail screen the top bar should be hidden, the content should fill the height (with a small top margin), and the detail should **fade in over the top** without the underlying page shifting.
4. Press **Back** — the top bar fades back in and the top-level grid/list is shown **at the same scroll position you left it**, not reset to the top.
5. Switch between tabs and confirm the content **crossfades**.
6. Confirm the **Up Next** tab (which has no detail screen) always keeps the top bar.

## Screenshots or Screencast
Screenshot_20260805_140909

## Checklist

- [ ] with different themes

### DIFF
--- a/tv/src/main/java/au/com/shiftyjelly/pocketcasts/component/TvDetailOverlay.kt
+++ b/tv/src/main/java/au/com/shiftyjelly/pocketcasts/component/TvDetailOverlay.kt
@@ -1,0 +1,73 @@
+package au.com.shiftyjelly.pocketcasts.component
+
+import androidx.activity.compose.BackHandler
+import androidx.compose.animation.AnimatedVisibility
+import androidx.compose.animation.core.FastOutSlowInEasing
+import androidx.compose.animation.core.tween
+import androidx.compose.animation.fadeIn
+import androidx.compose.animation.fadeOut
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.padding
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.focus.focusProperties
+import au.com.shiftyjelly.pocketcasts.theme.TvDetailTopInset
+import au.com.shiftyjelly.pocketcasts.theme.TvScreenBackgroundBrush
+
+/**
+ * Fades a detail screen in over the top of the still-composed parent it was opened from, so the
+ * parent keeps its scroll position and focus and never reflows. The overlay is opaque (it carries
+ * the screen background), hides the top bar while shown, and handles Back.
+ *
+ * [target] drives visibility: non-null shows the overlay, null fades it out. The last non-null
+ * value is retained so [content] still renders during the exit animation. Render this as the last
+ * child of the tab's root [Box] and mark the layers behind it inactive with [tvFocusInactiveWhen]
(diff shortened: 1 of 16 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

geekygecko (approved):
Going to the new page looks great. My only comment is that it feels like it should open as a completely new page, rather than transitioning the previous page, if that's possible.

I think it would be fine if the new page simply faded in over the top. I wonder whether changing the navigation on Apple TV is more related to having to rather than polish. 

Image

One other comment is that the previous page scrolls to the top when you return, it would be good if the scroll position is preserved where you left it. 

https://github.com/user-attachments/assets/7ef4a629-b469-446e-98e1-19c0f387d0f4

sztomek:
Thanks for the review @geekygecko.
Both of your concerns have been addressed in the recent commits, here's a recording of the current behavior.

https://github.com/user-attachments/assets/1a5158c8-5708-41ea-8551-f6e6524d7ca7

PR #5375, opened 2026-06-02
[TV] Add tv module with placeholder screen

sztomek:
## Description
This is the first step in the TV app development.
Created a new module `tv` that builds for the tv platform that has similar build configurations to wear and automotive: `debug, debugProd and release`.
Added basic dependencies to build upon them later.
Placeholder screen shows when the app launches.

Fixes PCDROID-572 https://linear.app/a8c/issue/PCDROID-572/create-new-tv-module

## Testing Instructions
1. Create an AVD for TV
2. Select tv build target
3. Build and install the tv app

## Screenshots or Screencast 
Screenshot_20260602_173805
Screenshot_20260602_173800

## Checklist
- [x] If this is a user-facing change, I have added an entry in CHANGELOG.md
- [x] Ensure the linter passes (`./gradlew spotlessApply` to automatically apply formatting/linting)
- [x] I have considered whether it makes sense to add tests for my changes
- [x] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml`
- [x] Any jetpack compose components I added or changed are covered by compose previews
- [x] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.
 
#### I have tested any UI changes...

- [x] with different themes
- [x] with a landscape orientation
- [x] with the device set to have a large display and font size
- [x] for accessibility with TalkBack

PR #5702, opened 2026-08-05
[CI] Fix `D8` Heap OOM Failures in the `Instrumented tests` Job

ParaskP7:
Closes: [AINFRA-2706](https://linear.app/a8c/issue/AINFRA-2706)

---

## Description

The `Instrumented tests` job has been failing intermittently for about two months with `ERROR: D8: java.lang.OutOfMemoryError: Java heap space`. A retry usually passed, so it read as flakiness, but it became frequent enough to be a real drag.

It was not flaky. The job was running out of heap because it was doing a large amount of work that it then threw away.

**The cause.** Both task names in the `build_and_instrumented_test` lane were unscoped:

```ruby
gradle(tasks: %w[assembleDebug assembleDebugAndroidTest])
```

An unscoped task name runs in every project, so this built debug APKs for `app`, `automotive`, `tv` and `wear`, plus an `androidTest` variant for all ~47 modules. Firebase Test Lab only ever consumes two artifacts, `app-debug.apk` and `app-debug-androidTest.apk`. Everything else was built and discarded. Only `app` (76 files) and `automotive` (1 file, unused by CI) have `androidTest` sources at all, so most of those variants were empty yet still dex-merged their external dependencies.

**Why that exhausted the heap.** Build scans show the failing job peaking at 7.4 GiB of a 7.8 GiB daemon heap with five D8 dex merges running concurrently, all sharing that one heap. For contrast, `Assemble app release APK` on the same agent spec peaked at 2.7 GiB while running full R8 minification. The constraint was never the size of the heap, it was how many dex merges ran at once.

**Why it started recently.** The most likely trigger is the `tv` module added in #5375 (8e1fb3e, 5 June), which took the repository from three application modules to four and so gave `assembleDebug` a fourth concurrent app-level dex merge, with the first report followed on [22 June](p1784247427512889-slack-CC7L49W13). This means that the timing matches.

---

## Changes

1. Scope both tasks to the app module so the job builds only what Firebase Test Lab consumes b2eee32e2
2. Replace `-Xmx8048M` with `-Xmx8g`, which states the intent rather than resolving to 7.859 GiB 1f53c1557
3. Temporarily bump the [Compose BOM](https://developer.android.com/develop/ui/compose/bom/bom-mapping) to force a dexing cache miss, so the fix could be measured on a build where D8 actually ran 51d8c9cde
4. Revert that temporary bump bd492a29d

Commits 3 and 4 are kept deliberately rather than squashed away, so the verification build stays reproducible from this branch.

FYI: This also made the `Instrumented tests` job faster as a side-effect, by close to a minute on a normal run and by several minutes whenever it saves a failure and a retry.

---

## Results

Three builds tell the story. The middle column is why the temporary commit above exists: the first scoped run served every dex task from cache, so it proved nothing about heap under real load.

| | [18061](https://buildkite.com/automattic/pocket-casts-android/builds/18061/list?sid=019fc7d9-2c59-46fa-9f7d-81dbedfc7a48&tab=output#019fc7d9-2cc0-4ec2-977a-b694c957a24e/L2533) unscoped, FAILED | [18142](https://buildkite.com/automattic/pocket-casts-android/builds/18142/list?sid=019fd294-e248-4efc-ab0c-31d394b283d9&tab=output#019fd294-e2e3-4d34-b27f-674c24a6b226/L1535) scoped, all cached | [18161](https://buildkite.com/automattic/pocket-casts-android/builds/18161/list?sid=019fd61a-f800-4bfd-a8d9-64881f4e6c29&tab=output#019fd61a-f87b-44ce-a211-6901630df238/L1535) scoped, dexing ran |
|---|---|---|---|
| `:app:mergeExtDexDebug` | executed | `FROM-CACHE` | executed |
| Peak heap | 7.4 GiB / 7.8 (95.0%) | 1001.5 MiB / 8 (12.2%) | 1.7 GiB / 8 (22.2%) |
| GC time | 24.825s | 1.332s | 4.030s |
| Execution time | 4m 8s | 22.7s | 2m 38s |
| Tasks executed | 1139 of 2066 (55.1%) | 416 of 1217 (34.2%) | 525 of 1217 (43.1%) |
| Serial execution factor | 10.88x | 6.97x | 2.78x |
| Outcome | D8 OOM | Passed, 542 tests | Passed, 542 tests |

---

## Testing Instructions

This is a CI change, so it is verified by reading builds rather than by tapping through the app.

1. Confirm the tests still run and pass. Open [build 18161](https://buildkite.com/automattic/pocket-casts-android/builds/18161) and check the `Instrumented tests` job reports `542 test cases passed`, and that both APKs were uploaded to Firebase Test Lab from `app/build/outputs/apk/`.
2. Confirm the heap is no longer the constraint. Open the [build scan for that run](https://buildkite.com/automattic/pocket-casts-android/builds/18161/list?sid=019fd61a-f800-4bfd-a8d9-64881f4e6c29&tab=output#019fd61a-f87b-44ce-a211-6901630df238/L1535) and check **Peak heap memory usage** reads 1.7 GiB of 8 GiB.
3. Confirm nothing else lost coverage. `Unit tests` runs `testDebugUnitTest` unqualified, so every module including `tv` is still compiled there, and `automotive` and `wear` keep their own release APK jobs.

---

## Screenshots or Screencast

Not applicable, no UI changes.

---

## Checklist

Not applicable, no UI changes.


AliSoftware (approved):
❤️ 

I'm guessing that not only will this help with CI build time for that job (having to build way less modules than before)… but also with the "Agent lost" occurrences we've been seeing in our EC2 infra recently due to OOM killing the buildkite agent and leaving the EC2 machine a zombie instance?

ParaskP7:
FYI: On Claude's [review](https://github.com/Automattic/pocket-casts-android/pull/5702#issuecomment-5203006246) and this below:

> Verdict: the scoping change is a good, correct fix. I'd resolve https://github.com/Automattic/pocket-casts-android/pull/1 (or explicitly decide the automotive test is dead) and adjust the framing of the heap change; https://github.com/Automattic/pocket-casts-android/pull/2 is optional follow-up.

---

> 1. automotive's instrumented test is now compiled by nothing in CI ⚠️ (unresolved)

I wouldn't go for `gradle(tasks: %w[:app:assembleDebug :app:assembleDebugAndroidTest :automotive:compileDebugAndroidTestSources])` as we don't do much with UI testing on `automotive` atm. 

1. Maybe, actually, this is test is stale and should be first reviewed, then deleted. 🤷 
2. Else, if we really need this test to run on Firebase Test Lab, we should update our infra on that. 💭

> 2. tv loses its only packaging check in CI (minor, follow-up)

If we want the `tv` module to be compiled as part of CI, then we should be explicit about that and have it be part of the the `Assemble release APKs` group, just like the rest of them (`app`, `automotive` and `wear`. I wouldn't have `Instrumented tests` do the heavy lifting here.

---

FYI: No matter, I consider both out-of-scope of this PR and CI fix and thus I wouldn't block merging it because of any of them.

AliSoftware:
One good point from Claude's review is that this means that now no job in our CI validates that targets for `automotive` and `tv` are even just _compiling_ without error

> The PR description characterises `automotive/src/androidTest/.../AutomotiveSettingsActivityRecreateTest.kt` as "1 file, unused by CI". That's true of execution — FTL only ever received the app test APK — but it was not true of compilation. The unscoped `assembleDebugAndroidTest` compiled it on every PR, and that's what kept it from bit-rotting

I wonder if we want to address that somehow (all while avoiding to re-introduce the OOM)?

[EDIT: Jinx — seems me and [Petros](https://github.com/Automattic/pocket-casts-android/pull/5702#issuecomment-5203096241) posted our comments at exactly the same time 😅 ]

ParaskP7:
👋 @AliSoftware and thanks for taking a look at this! 💯 

1. On [this](https://github.com/Automattic/pocket-casts-android/pull/5702#pullrequestreview-4873297850) comment of yours and the zombie instances problem, I think PCAndroid isn't causing any. This is because even if this job OOMs, I still see it `Exited with status 1` (see [here](https://buildkite.com/automattic/pocket-casts-android/builds/18061/list?sid=019fc7d9-2c59-46fa-9f7d-81dbedfc7a48&tab=output#019fc7d9-2cc0-4ec2-977a-b694c957a24e/L2533)) instead of `Exited with status -1 (agent lost)`, which might be due to Fastlane helping us here... 🤷 

2. On [this](https://github.com/Automattic/pocket-casts-android/pull/5702#issuecomment-5203103264) comment of yours and `automotive` & `tv` module, as per your edited update, indeed jinx, I consider those out-of-scope of this PR and CI fix personally. But, we can definitely add new tasks in our backlog if need be for both, after me discussing that further with, at least, @geekygecko ! 😄

ParaskP7:
👋 @geekygecko and @wzieba , just a friendly reminder of this! 🙏

FYI: I know @AliSoftware already approved this, but I'll be waiting on either of you to give me another ✅ on this before merging.

geekygecko (approved):

geekygecko:
Thanks for making this change. I just had another "Caused by: java.lang.OutOfMemoryError: Java heap space" so this will help speed us up.

ParaskP7:
Thanks for reviewing and merging this @geekygecko ! 🙇 ❤️ 🚀

PR #5708, opened 2026-08-07
Use public Sentry org/project, env-injected token

mokagio:
See https://linear.app/a8c/issue/AINFRA-2790. Also, https://github.com/wordpress-mobile/WordPress-Android/pull/23189 for a similar implementation.

- Inlines the Sentry organization and project names, which do not need to be secret (and cannot be anyway because they can be accessed in the published binary), instead of reading from encrypted `sentry.properties`.
- Leaves the DSN as a secret, however, not because it _is_ secret, but just to avoid open source contributors accidentally spamming our production logs
- Moves reading the API token from `.properties` to env var, so that it can be kept off any machine that doesn't need it (i.e., ideally CI only)

The pinned `mobile-secrets` checkout now points at `789d0d6`, which drops the properties this PR stops reading — [per review](https://github.com/Automattic/pocket-casts-android/pull/5708#discussion_r3926832452), so CI exercises this change against the cleaned-up `secret.properties` rather than the one that still carried them.

<details>
<summary>AI-generated details</summary>

## Rationale

`sentryOrg`, the per-variant `sentry*Project` slugs, and `sentryAuthToken` aren't secrets: the org/project slugs are Sentry-public identifiers, and the auth token is only needed by CI to upload ProGuard mappings. Moving them out of `secret.properties` shrinks what `mobile-secrets` needs to hold for this repo, matching the pattern already shipped in [simplenote-android#1850](https://github.com/Automattic/simplenote-android/pull/1850) and in review in dayone-android.

## Intentional tradeoff

Unlike those sibling repos, the Sentry **DSN stays in `secret.properties` unchanged**. Pocket Casts Android is open source and hasn't otherwise adopted `a8c-secrets` yet, so per the paaHJt-ahO-p2 RFC the DSN can't become a public repo constant the way it did there.

## Verification

No ordinary CI job uploads to Sentry — `assemble-release-apk.sh` and the prototype lane both pass `-PskipSentryProguardMappingUpload=true`, and the only path that registers the upload tasks is `build_bundle`, reachable solely from the API-triggered `release-builds.yml`. So without a deliberate check, the first exercise of this wiring would have been the next code freeze, after the AAB was built.

Throwaway PR #5750 ran it for real against this branch, at the current pin. Now closed; the evidence:

- [Build 18798](https://buildkite.com/automattic/pocket-casts-android/builds/18798) — R8 ran for each module (106s / 104s / 83s) and sentry-cli uploaded three real mappings with the token taken from the environment: `UPLOADED df18268e…` (automotive), `6fde9224…` (app), `76eb31ac…` (wear). Confirms the `a8c` org and all three inlined project slugs authenticate.
- [Build 18799](https://buildkite.com/automattic/pocket-casts-android/builds/18799) — with the token blanked, `:app:uploadSentryProguardMappingsRelease FAILED` with `SENTRY_AUTH_TOKEN is not set (or is blank)`. The guard fails the build loudly rather than letting a release ship unsymbolicated.

Both runs also asserted that `secret.properties` no longer defines `sentryAuthToken`, `sentryOrg`, `sentryAndroidProject`, `sentryAutomotiveProject` or `sentryWearProject`, and that the upload tasks genuinely executed rather than coming back cached.

## Gotchas

- `SENTRY_AUTH_TOKEN` is exported on the `android` queue agents, not by anything in this repo — grep finds it only in `build.gradle.kts`. Build 18798 confirms it arrives; nothing here would tell you where from.
- The `mobile-secrets` counterpart is applied and pinned. `sentryTvProject` and the DSN keys deliberately stay — the TV module's Sentry setup is #5840 / #5841.

## How to test

`./gradlew spotlessCheck` passes, and any task evaluates all Gradle scripts cleanly. The upload path itself can't be exercised from an ordinary build — see Verification above for the CI runs that did.

</details>


build.gradle.kts
```diff
@@ -490,6 +492,19 @@ fun Project.applyCommonSentryConfiguration() {
         includeDependenciesReport = false
         ignoredBuildTypes = setOf("debug", "debugProd", "prototype")
     }
+
+    // The upload task only exists when includeProguardMapping is on, so this fires exactly when a
+    // mapping upload is expected — fail there rather than ship a release we can't deobfuscate.
+    tasks.matching { it.name.startsWith("uploadSentryProguardMappings") }.configureEach {
+        doFirst {
+            if (!sentryAuthToken.isPresent) {
```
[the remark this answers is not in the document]
mokagio: Good catch, adopted in 9d74335bd — with one adjustment.

Rather than blank-checking at both sites, I normalise once at the source so the bound value and the guard can't drift:

```kotlin
val sentryAuthToken = providers.environmentVariable("SENTRY_AUTH_TOKEN").orNull?.takeIf { it.isNotBlank() }
```

`authToken = sentryAuthToken` and `if (sentryAuthToken == null)` then read the same thing by construction.

One correction to the premise: the code being replaced here didn't use `isNullOrBlank()` for the token — it was `project.findProperty("sentryAuthToken")?.toString()`, unchecked. The `isNullOrBlank()` you're thinking of was in `app/build.gradle.kts`, guarding a warning `println` about `sentryAndroidProject`, and this PR removes it. Doesn't change the conclusion — an empty env var reading as `""` rather than absent is exactly the failure mode worth defending against.

build.gradle.kts
```diff
@@ -490,6 +492,19 @@ fun Project.applyCommonSentryConfiguration() {
         includeDependenciesReport = false
         ignoredBuildTypes = setOf("debug", "debugProd", "prototype")
     }
+
+    // The upload task only exists when includeProguardMapping is on, so this fires exactly when a
+    // mapping upload is expected — fail there rather than ship a release we can't deobfuscate.
+    tasks.matching { it.name.startsWith("uploadSentryProguardMappings") }.configureEach {
```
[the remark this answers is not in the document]
mokagio: Both points valid — fixed in 9d74335bd by going one level up the type hierarchy instead of broadening the string match:

```kotlin
tasks.withType<SentryCliExecTask>().configureEach { ... }
```

Checking the 6.16.0 plugin jar: `SentryCliExecTask` is the public base of every task that shells out to `sentry-cli` with the auth token — `SentryUploadProguardMappingsTask`, `UploadSourceBundleTask`, and the native-symbols/app-artifact tasks. So one type covers both cases you raised, and a future rename is a compile error rather than a silently vanishing guard.

Your first point checks out: `sentryUploadSourceBundle*` is registered by `UploadSourceBundleTask` and gated by the same `includeSourceContext` flag, so the old prefix genuinely missed it.

Two consequences worth noting:

- The message is now "upload debug files to Sentry" rather than "upload ProGuard mappings", since the net is wider. The `-PskipSentryProguardMappingUpload` escape hatch stays accurate — in this repo that one flag gates both `includeProguardMapping` and `includeSourceContext`.
- The extra tasks this now catches are inert here: nothing sets `uploadNativeSymbols` or app-artifact upload, so those tasks are never registered.

Note that CI cannot exercise this guard: every PR-triggered build passes `-PskipSentryProguardMappingUpload=true` (`assemble-release-apk.sh`, and the prototype lane), so the upload tasks are never registered and `doFirst` never runs. The first real exercise is `build_bundle` on a beta/release build, which passes no skip property. What CI does confirm is that the build scripts compile, so the typed match resolves against the plugin.

build.gradle.kts
```diff
@@ -474,9 +475,14 @@ subprojects {
 }
 
 fun Project.applyCommonSentryConfiguration() {
+    // A secret-injection step that resolves to nothing exports an empty SENTRY_AUTH_TOKEN rather
+    // than leaving it unset, so treat blank as missing — otherwise the guard below passes and
+    // sentry-cli fails mid-upload with an opaque auth error.
```
mokagio:
```suggestion
```

build.gradle.kts
```diff
@@ -490,6 +496,21 @@ fun Project.applyCommonSentryConfiguration() {
         includeDependenciesReport = false
         ignoredBuildTypes = setOf("debug", "debugProd", "prototype")
     }
+
+    // SentryCliExecTask is the base of every task that shells out to sentry-cli with the auth token
+    // (ProGuard mappings and source bundles here). Those tasks only exist when the upload flags are
+    // on, so this fires exactly when an upload is expected — fail there rather than ship a release
+    // we can't deobfuscate.
```
mokagio:
```suggestion
```

build.gradle.kts
```diff
@@ -474,9 +475,11 @@ subprojects {
 }
 
 fun Project.applyCommonSentryConfiguration() {
+    val sentryAuthToken = providers.environmentVariable("SENTRY_AUTH_TOKEN").orNull?.takeIf { it.isNotBlank() }
+
     extensions.getByType(SentryPluginExtension::class.java).apply {
-        authToken = project.findProperty("sentryAuthToken")?.toString()
-        org = project.findProperty("sentryOrg")?.toString()
+        authToken = sentryAuthToken
```
iangmaia, line 468: Not sure why this assignment is needed if you already have it above?

You could also keep the fact that it exists or not like `val hasSentryAuthToken = providers.environmentVariable("SENTRY_AUTH_TOKEN").map { it.isNotBlank() }.orElse(false)` or similar.

mokagio: Adopted the Provider form in 5dad76003, but kept the assignment — it's load-bearing.

Checking the 6.16.0 plugin jar: `SentryPluginExtension.authToken` is `objects.property(String::class.java).convention(null as String?)`. It has no environment default, so without `authToken = …` the extension property stays absent. `SentryCliExecTask` then goes the other way — it *exports* `SENTRY_AUTH_TOKEN` into the `sentry-cli` process **from** that property. And the extension value isn't only used for that: `SentryOrgValueSource`, `SentryTelemetryService`, and the source-bundle/native-symbol registration read it too.

So `sentry-cli` picking the token up "for free" would rely on it happening to be in the Gradle daemon's inherited environment, which is stale whenever the daemon outlives the shell that started it. The explicit assignment makes it deterministic.

Your second point stands on its own though, so I took it:

```kotlin
val sentryAuthToken = providers.environmentVariable("SENTRY_AUTH_TOKEN").filter { it.isNotBlank() }

authToken = sentryAuthToken
// …
if (!sentryAuthToken.isPresent) { throw GradleException(…) }
```

Same single source for the bound value and the guard, but lazy now — a build that never configures Sentry never resolves the variable. I used `filter { it.isNotBlank() }` rather than `map { … }.orElse(false)` so the one provider serves both roles; a blank value stays absent instead of reaching the plugin as `""`.

Verified against the three env states (unset / set-but-empty / set), with configuration cache on: the extension property reads `<absent>`, `<absent>`, `dummy-token`, and the guard throws, throws, passes.

mokagio: Correcting myself — I've reverted the provider refactor in 3ad15eeea, so this is back to the `String?` val you commented on.

The refactor was a local improvement that made this repo the odd one out. The same migration already merged elsewhere with the eager read:

- [WordPress-Android#23189](https://github.com/wordpress-mobile/WordPress-Android/pull/23189) — `WordPress/build.gradle:71`, `def sentryAuthToken = providers.environmentVariable("SENTRY_AUTH_TOKEN").orNull?.trim() ?: null`, bound with `authToken = sentryAuthToken` and guarded with `if (sentryAuthToken == null)` on `SentryCliExecTask`. Same shape as here.
- [simplenote-android#1850](https://github.com/Automattic/simplenote-android/pull/1850) — `Simplenote/build.gradle:77`, `authToken = providers.environmentVariable("SENTRY_AUTH_TOKEN").orNull` (no guard).

The point of this migration is one pattern across the fleet, so consistency wins over the marginally nicer form here.

On the assignment itself: it is needed. In the 6.16.0 plugin jar, `SentryPluginExtension.authToken` is `objects.property(String::class.java).convention(null as String?)` — no environment default. `SentryCliExecTask` goes the other way, *exporting* `SENTRY_AUTH_TOKEN` into the `sentry-cli` process from that property, and `SentryOrgValueSource` / `SentryTelemetryService` read it too. Without the assignment we'd be relying on the token happening to sit in the Gradle daemon's inherited environment, which is stale whenever the daemon outlives the shell that started it.

Why `takeIf { it.isNotBlank() }` here rather than WordPress-Android's `?.trim() ?: null`: that form relies on Groovy truth, where `"" ?: null` is `null`. In Kotlin the elvis only fires on `null`, so a set-but-empty variable would sail through as `""`. Verified both against the four env states — unset, empty, whitespace, real — and they agree.

mokagio: @iangmaia _real Gio here, not a meat proxy_.

> You could also keep the fact that it exists or not like val hasSentryAuthToken = providers.environmentVariable("SENTRY_AUTH_TOKEN").map { it.isNotBlank() }.orElse(false) or similar.

I updated the implementation in https://github.com/Automattic/pocket-casts-android/pull/5708/changes/829a341ed694bec4d551c60feaa8ebcbd77f08b5 and aligned it with the other apps that already do this.

> Not sure why this assignment is needed if you already have it above?

My understanding is that the `val sentryAuthToken =` assigns the token from the env if any making it available to the other steps. The `authToken = sentryAuthToken` is what passes it to the Sentry plugin itself.

AliSoftware: > My understanding is that the val sentryAuthToken = assigns the token from the env if any making it available to the other steps. The authToken = sentryAuthToken is what passes it to the Sentry plugin itself.

That's my understanding as well.

app/build.gradle.kts
```diff
@@ -9,7 +9,7 @@ plugins {
 }
 
 sentry {
-    projectName = project.findProperty("sentryAndroidProject")?.toString()
+    projectName = "pocket-casts-android"
```
AliSoftware, line 12: Not critical (it won't break anything if we don't do it), but technically, since you also already removed those now-unnecessary properties (`sentryAuthToken`, `sentryOrg`, `sentry*Project`) from `secrets.properties` in `mobile-secrets`'s commit `789d0d6`, it would make sense to also update the `.configure` config file to update its `pinned_hash ` to `789d0d6ec3d5e23886ca25d767fb63452082412c` so it points to this cleaned-up `secrets.properties`, also so that we can validate that things still work as expected with this PR's changes _even after_ said `secrets.properties`' cleanup

AliSoftware:
_This PR has been through **9 rounds** of the CI Claude Review bot. Since the branch was rebased, the commit SHAs those comments reference (`e6a053a`, `9d74335`, `5dad760`, `358eddc`, `3ad15ee`, `829a341`, `f981a32`) no longer exist in the branch — commit *subjects* are the reliable join key._

Here's what's still worth reading vs. what's been superseded, plus my own pass.

---

<details><summary><h3>1. Summary of the different rounds of review</h3></summary>

| # | Commit (subject) | What triggered it | Outcome |
|---|---|---|---|
| R1 | `Use public Sentry org/project…` | initial | **2 real findings** |
| R2 | `Guard all sentry-cli tasks…` | fixes R1 | both resolved |
| R3 | `Remove verbose AI comments` | comment cleanup | no findings |
| R4 | `Delete unused sentry.properties.enc` | ready-for-review | no findings |
| R5 | `Keep the Sentry auth token lazy` | answering @iangmaia | no findings |
| R6 | `Revert "Keep the…lazy"` | fleet consistency | no findings |
| R7 | `Trim the Sentry auth token…` | newline defence | no findings |
| R8 | `Say "or is blank"…` | message wording | no findings |
| R9 | *(rebase)* | rebase | no findings |

**Only R1 produced findings; R2 confirmed them fixed.** R3–R9 are seven rounds of re-verification with zero new findings — they exist because the bot re-runs on every push, and each round re-pasted the same "still open" list. That repetition is what makes the thread look noisier than it is.
</details>
<details><summary><h3>2. Obsolete — safe to skip</h3></summary>

- **R1 #1 — blank token slips past the guard.** Fixed, then improved twice. Current code is `?.trim()?.ifEmpty { null }`.
- **R1 #2 — guard misses the source-bundle upload / prefix matching is fragile.** Fixed by `tasks.withType<SentryCliExecTask>()`. Verified against plugin **6.19.0**: `SentryCliExecTask` is the base of *all six* CLI tasks (proguard mappings, source bundle, bundle sources, native symbols, snapshots, app artifact). The typed match is genuinely complete.
- **R2's `System.getenv()["CI"]` nit.** The bot withdrew it itself in R5.
- **R5's "Gradle input validation may pre-empt your friendly message."** ❌ Not correct. `SentryCliExecTask.sentryAuthToken` is `@get:Input @get:Optional`, so Gradle will never emit *"no value has been specified for property"*.
- **R2's "assumption: those tasks only exist when the upload flags are on."** ✅ Verified, closed. `SentryUploadProguardMappingsTask` registers only under `if (isMinifyEnabled && includeProguardMapping.get())`; source-bundle tasks only under `if (includeSourceContext.get())` *and* carry `onlyIf` guards; snapshots default to `enabled = false`; app-artifact needs `sizeAnalysis`/`distribution` (unset here); native symbols needs `uploadNativeSymbols` (unset here). **The suggested `--dry-run` is unnecessary.**
- **R7's "does the plugin actually override the child process env from the property?"** ✅ Verified. `SentryCliExecTask.setSentryAuthTokenEnv()` calls `environment("SENTRY_AUTH_TOKEN", …)` when the property is set, so the `trim()` is effective. This also confirms @mokagio's answer to @iangmaia: `authToken` is still `convention(null as String?)` in 6.19.0, so the explicit assignment **is** load-bearing.
- **The 6.16 → 6.18 → 6.19 plugin version drift discussion.** I re-verified every claim against 6.19.0; all hold.

</details>

<details><summary><h3>3. Closed by verification</h3></summary>

**Hardcoded org and project slugs — all four are correct.** Checked against the Sentry API:

| Value in PR | Status |
|---|---|
| `org = "a8c"` | ✅ Automattic |
| `pocket-casts-android` | ✅ exists |
| `pocket-casts-automotive` | ✅ exists |
| `pocket-casts-wear` | ✅ exists |

This was the bot's #1 "still open" item in all 9 rounds, and it no longer matters that the originals were already removed from `mobile-secrets` in `789d0d6`.

**Orphaned `android/pocket-casts/sentry.properties` in `mobile-secrets` — nothing left to sweep.** My human checked `mobile-secrets` for me and assured me that the file is no longer there. Combined with `.configure` having no `files_to_copy` entry targeting it (so it was never decrypted in the first place), this follow-up item is already handled — it can come off the list rather than being carried into the cleanup PR.

</details>

<details open><summary><h3>4. Still valid</h3></summary>

**A. `SENTRY_AUTH_TOKEN` is an untested, load-bearing assumption.** Confirmed: **zero references anywhere in the repo.** `.buildkite/commands/assemble-release-apk.sh:19` and the prototype lane both pass `-PskipSentryProguardMappingUpload=true`; `build_bundle` (`fastlane/Fastfile:583`, `:app:bundle` Release) passes nothing. So the guard is **unreachable on every PR build**, and first fires mid-release, *after* the AAB is built.

**B. Config-cache caveats** (token resolved at configuration time and serialised into `.gradle/configuration-cache/`; the env read is a tracked input on every app-module configuration, so rotating the token invalidates CC for *all* builds, not just upload builds). Real but low impact — and note `release-build.sh` doesn't call `restore-cache.sh` at all, and the cached artifact is the *dependency* cache, so the exposure is narrower than the bot implied.

</details>

<details open><summary><h3>5. My Claude's own review</h3></summary>

The diff is correct. Three things not raised in any round:

**1. A shadowing trap that turns out to be fine.** `SentryCliExecTask` has its own `sentryAuthToken: Property<String>` member, and the guard reads a bare `sentryAuthToken` from inside `configureEach { doFirst { … } }` — two nested implicit receivers deep. If the task member won name resolution, `Property<String> == null` would be *always false* and **the guard would silently never fire**. I built a minimal Gradle repro to check: the outer local `String?` wins, and the guard fires correctly. **No bug** — but it's a hair's breadth from one, and renaming the local to something like `sentryAuthTokenValue` would make it robust against a future reader or refactor.

**2. The guard over-reaches onto `BundleSourcesTask`** (`build.gradle.kts:496`). That task runs `sentry-cli debug-files bundle-jvm` — purely local, **needs no auth token** — but it extends `SentryCliExecTask`, so it throws too. Harmless *today* because `includeProguardMapping` and `includeSourceContext` move together off the same flag. It would misfire if anyone ever sets `autoUploadSourceContext = false`. Arguably a feature (fails earlier, before doing work); just worth knowing it's unintentional.

**3. On the `.configure` `pinned_hash`** (my inline comment). The description's rebuttal — "they don't need new ones, so I kept the diff smaller" — answers a different question. The point isn't *need*, it's **evidence**: bumping to `789d0d6` is the only way CI proves this works against the cleaned-up `secret.properties` before someone discovers it during a release. Currently `.configure` still pins `c65d2bd`.

**Bonus, out of scope:** `pocket-casts-tv` exists as a Sentry project and the TV app already ships a DSN (`dependencies.gradle.kts:113`), so it reports crashes today — but `tv/build.gradle.kts` doesn't apply the Sentry Gradle plugin, so those crashes will never be deobfuscated. Pre-existing and not caused by this PR; the `sentryTvProject` removal stays correct. Just noting that whoever wires the plugin into `tv` later will need `projectName = "pocket-casts-tv"`.

</details>

## Key points to focus on

1. **Don't read rounds of review R3–R9.** Seven rounds, zero findings. R1 + R2 are the whole substantive history, and both findings are fixed.
2. **The code is fine — the remaining risk is operational.** This PR converts *"release silently ships without mappings"* into *"release build hard-fails after the AAB is built."* That's the right trade, but it's unexercised by any PR job. Confirm `SENTRY_AUTH_TOKEN` is present on the agents before the next beta, and keep `-PskipSentryProguardMappingUpload=true` within reach for that first run.
3. ~~Verify the hardcoded slugs~~ — **done, all four correct.**
4. **The `pinned_hash` bump** [suggested here](https://github.com/Automattic/pocket-casts-android/pull/5708#discussion_r3926832452) is the one change that would buy actual CI evidence.
5. **Optional nit:** rename the local `sentryAuthToken` so the guard isn't one Kotlin resolution rule away from being silently dead.

---

*Reviewed with Claude Code (Opus 5), posted by @AliSoftware.*

AliSoftware:
Regarding the point mentioned by my Claude above about the `SENTRY_AUTH_TOKEN`, I noticed something surprising when I looked in `mobile-secrets`:

- Before this PR's change, the Sentry Auth Token was provided by `secrets.properties`, and had a value looking like `sentryAuthToken=e473[…]3b7c`
- Now it's provided by `~/.mobile-secrets/CI/secrets/pocket-casts-android/env` which points to `export SENTRY_AUTH_TOKEN="${SENTRY_AUTH_TOKEN}"`… and the value for that (from `shared-secrets.yml`) is something like `360d[…]664a`

So this means that technically this PR will now switch to using a different Sentry Auth Token as before 🤔 

In practice, that shouldn't be a problem, because that `SENTRY_AUTH_TOKEN` from `shared-secrets.yml` has been in use by all our other repos already (and in particular, `pocket-casts-ios` is on the same Sentry team (`hogwarts`) as `pocket-casts-android`, uses that shared token, and uploaded debug files as recently as 2026-08-31). It's just that I found it surprising that we'd switch tokens as a side-effect, but I don't think that should prevent this PR from being merged.

**This effectively closes item 4A** that my Claude review/summary above initially marked as still to be checked.

---

I did some archeology to try to find where those 2 tokens were defined in Sentry. It happens that they are not [Org Auth tokens](https://a8c.sentry.io/settings/auth-tokens/), but instead are custom integration + personal tokens:
 - The `…664a` token (the shared one we'll start using after this PR) is actually the [Internal Custom Integration](https://a8c.sentry.io/settings/developer-settings/) token called [CI Symbol Upload](https://a8c.sentry.io/settings/developer-settings/circleci-df81e8/). It has scopes `project:read`, `project:releases` and `project:write` at org level — which is why we can be confident that it'll work for `pocket-casts-android` too
 - The `…3b7c` one (that we had in `secrets.properties` so far) is a personal token belonging to @geekygecko (as confirmed by using `SENTRY_AUTH_TOKEN=...3b7c sentry-cli info`

So **it's actually a good thing that we don't rely on that old user token anymore** and that this PR will switch to the shared one. (And once everything is in place and landed, we should probably ultimately ask Philip to revoke that user token to clean things up).

In the long term it'd probably be better to replace the `SENTRY_AUTH_TOKEN` in `shared-secrets.yml` with a modern (`sntrys_*`) org token — like [this one](https://a8c.sentry.io/settings/auth-tokens/197930/), which is already used by `CI/secrets/beeper-ios/env` and `CI/secrets/day-one-apple/env` (or even better security-wise, switch to using a different org auth token per repo?), instead of using an integration token. But that's for a different time and PR.

app/build.gradle.kts
mokagio: Fair. I relied on the fact that all the removals from the file upstream matched properties no longer read here, but one can only be sure when everything is up to date.

Addressed in 19e282212914f68c6066ac5a7dc9d3beccc91705

mokagio:
> SENTRY_AUTH_TOKEN is an untested, load-bearing assumption. Confirmed: zero references anywhere in the repo. .buildkite/commands/assemble-release-apk.sh:19 and the prototype lane both pass -PskipSentryProguardMappingUpload=true; build_bundle (fastlane/Fastfile:583, :app:bundle Release) passes nothing. So the guard is unreachable on every PR build, and first fires mid-release, after the AAB is built.

That's not 100% correct. I had a validation PR open from the start of this work https://github.com/Automattic/pocket-casts-android/pull/5750.

But it's true that I haven't updated it with the latest changes (on my defense, they didn't really touch the token ingestion iteslf). Regardless, that's validated now.

- https://buildkite.com/automattic/pocket-casts-android/builds/18799 - happy path
- https://buildkite.com/automattic/pocket-casts-android/builds/18798 - invalid token

mokagio:
@AliSoftware @iangmaia it's hard to tell if everything is addressed given the amount of words in the PR, but between me and Claude, there's fair confidence.

All points addressed, and the token wiring is now verified rather than assumed.

<details>
<summary>Details</summary>

**Pin bump** — `.configure` moved to `789d0d6` in 19e28221, so CI builds against the
cleaned-up `secret.properties` rather than the one that still carried the removed keys.

**The upload path is now actually exercised.** Nothing in ordinary CI uploads to Sentry:
`assemble-release-apk.sh` and the prototype lane both pass
`-PskipSentryProguardMappingUpload=true`, and the only path that registers the upload
tasks is `build_bundle`, reachable solely from the API-triggered `release-builds.yml`.
So the first real test of this change would otherwise have been the next code freeze,
after the AAB was built. Throwaway PR #5750 ran it for real against this branch:

- [Build 18798](https://buildkite.com/automattic/pocket-casts-android/builds/18798) —
  R8 ran for each module (106s / 104s / 83s) and sentry-cli uploaded three real mappings
  with the token taken from the environment: `UPLOADED df18268e…` (automotive),
  `6fde9224…` (app), `76eb31ac…` (wear). Confirms the `a8c` org and all three inlined
  project slugs authenticate.
- [Build 18799](https://buildkite.com/automattic/pocket-casts-android/builds/18799) —
  with the token blanked, `:app:uploadSentryProguardMappingsRelease FAILED` with
  `SENTRY_AUTH_TOKEN is not set (or is blank)`. The guard fails the build loudly rather
  than letting a release ship unsymbolicated.

Each build is decisive for one half: on 18799 the remote build cache made R8 a no-op, so
sentry-cli reported `Nothing to upload, all files are on the server` instead of new UUIDs.

</details>

Thanks!

build.gradle.kts
iangmaia: > The authToken = sentryAuthToken is what passes it to the Sentry plugin itself.

Ah indeed, I guess I missed `authToken` is a field in the extension we're changing with `apply`  👍

.configure
```diff
@@ -1,7 +1,7 @@
 {
   "project_name": "pocketcasts-android",
   "branch": "trunk",
-  "pinned_hash": "c65d2bd5456a8d0d33d675376b182cd9b529e397",
+  "pinned_hash": "789d0d6ec3d5e23886ca25d767fb63452082412c",
```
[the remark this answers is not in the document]
AliSoftware: - I verified that `.mobile-secrets`'s `789d0d6` contains `sentryAndroidDsn`, `sentryAutomotiveDsn` and `sentryWearDsn`.
 - That commit `789d0d6` does _not_ contain `sentryTvDsn` though, which was added (alongside `sentryTvProject`) in `mobile-secrets`' commit `1e17440` stacked on top of the `789d0d6` one.

For consistency I think it could make sense to keep this PR point to `789d0d6`, and only update the `pinned_hash` to `1e17440` in https://github.com/Automattic/pocket-casts-android/pull/5841 (the PR stacked on top of this one and which adds the inlining of `projectName = "pocket-casts-tv"` and enables Sentry for the TV module). At least that's what I suggested in https://github.com/Automattic/pocket-casts-android/pull/5841#pullrequestreview-5132115070

@mokagio You might also as well want to enable GitHub's Stacked PRs feature on this PR + #5841, so that we end up merging them both in sync (inlining public values + enabling TV with sentryTvProject value inlined too)

AliSoftware (approved):
I think at that stage it's ready to merge.

I'd also suggest to update the `pinned_hash` in https://github.com/Automattic/pocket-casts-android/pull/5841 to `1e17440`, un-draft that PR, enable GitHub's Stacked PR feature on this #5708 + #5841, and merge them both together.

PR #5715, opened 2026-08-10
[TV] Wire up Play all episodes on the playlist details screen

sztomek:
## Description

Wires up the **Play all episodes** action on the Android TV playlist details screen, mirroring the Apple TV (tvOS) implementation. The button already existed on the screen but its `onClick` was an empty stub.

Behaviour is ported from tvOS `PlaylistDetailsViewModel.playAll()` → `PlaybackManager.playIfSafe(playlist:episodeIDs:)`, reusing the existing phone-app `PlayAllHandler` so the Android phone and TV share identical semantics:

- **Up Next is empty, or is already this playlist** → replace/append and start playing immediately, then open Now Playing (no dialog).
- **Up Next is non-empty and different** → show a confirmation modal, matching the three tvOS options:
  - **Play without saving** — clear Up Next and play the playlist.
  - **Save and play** — snapshot the current Up Next into new manual playlists ("Up Next – <date>"), then play. A toast confirms the save once it completes.
  - **Cancel**.

The visible, sort/archive-filtered episode list is what gets played (matching what the user sees).

Fixes PCDROID-711 https://linear.app/a8c/issue/PCDROID-711/play-all-episodes

### Notable implementation points
- New `PlayAllHandler.handlePlayAllEpisodes(List<BaseEpisode>)` overload reuses the existing generic core so the TV screen can pass its `List<PodcastEpisode>` directly — no new play-all logic was written.
- One-shot navigation/toast signals go through a `Channel(BUFFERED).receiveAsFlow()` so an event can't be dropped across an activity-recreation window.
- `playAll()` and `replaceUpNextAndPlay()` are both re-entrancy guarded (a TV remote double-press on "Save and play" would otherwise create duplicate synced playlists), and the destructive queue-replace runs under `NonCancellable`.
- The 3-button confirmation is a `TvModal` (consistent with `TvEpisodeActionsModal`/`TvConfirmationContent`), not the phone's `Switch`-based dialog, because discrete buttons are far more D-pad friendly — the strings and logic remain the iOS ones.
- Analytics are intentionally omitted, consistent with the rest of the TV episode-action work.

## Testing Instructions

1. On Android TV, sign in and open a playlist that has episodes (Playlists tab → a playlist).
2. Focus the left pane and select **Play all episodes**.
   - With an **empty Up Next**: playback starts on the first episode and Now Playing opens.
3. Add some unrelated episodes to Up Next, then select **Play all episodes** again on a different playlist.
   - The **"This will clear your Up Next"** confirmation appears.
   - **Play without saving** → Up Next is replaced, playback starts, Now Playing opens.
   - **Save and play** → your previous Up Next is saved as new manual playlist(s), a toast confirms, then the playlist plays.
   - **Cancel** / Back → nothing changes.
4. Select **Play all episodes** when the playlist is already playing as Up Next → resumes / no-op, no dialog.

## Screenshots or Screencast
Screenshot_20260810_145033

## Checklist
- [ ] I have updated (or requested that someone edit) the Event Horizon schema to reflect any new or changed analytics

- [ ] with different themes

### DIFF
--- a/modules/services/localization/src/main/res/values/strings.xml
+++ b/modules/services/localization/src/main/res/values/strings.xml
@@ -266,8 +266,8 @@
     <string name="tv_playlists_download_title">Create playlists on your phone</string>
     <string name="tv_playlists_download_subtitle">They’ll be waiting here when you’re done. Scan to get the app</string>
     <string name="tv_playlist_play_all">Play all episodes</string>
-    <string name="tv_playlist_play_all_clear_up_next_title">This will clear your Up Next</string>
-    <string name="tv_playlist_play_all_clear_up_next_message">We can save your current Up Next as a playlist</string>
+    <string name="tv_playlist_play_all_clear_up_next_title">This will clear your Up\u00A0Next</string>
+    <string name="tv_playlist_play_all_clear_up_next_message">We can save your current Up Next as a\u00A0playlist</string>
     <string name="tv_playlist_play_all_save_and_play">Save and play</string>
     <string name="tv_playlist_play_all_play_without_saving">Play without saving</string>
     <plurals name="tv_playlist_all_archived">
--- a/modules/services/localization/src/main/res/values/strings.xml
+++ b/modules/services/localization/src/main/res/values/strings.xml
@@ -266,6 +266,10 @@
     <string name="tv_playlists_download_title">Create playlists on your phone</string>
     <string name="tv_playlists_download_subtitle">They’ll be waiting here when you’re done. Scan to get the app</string>
     <string name="tv_playlist_play_all">Play all episodes</string>
+    <string name="tv_playlist_play_all_clear_up_next_title">This will clear your Up Next</string>
+    <string name="tv_playlist_play_all_clear_up_next_message">We can save your current Up Next as a playlist</string>
+    <string name="tv_playlist_play_all_save_and_play">Save and play</string>
+    <string name="tv_playlist_play_all_play_without_saving">Play without saving</string>
     <plurals name="tv_playlist_all_archived">
         <item quantity="one">The only episode of this playlist has been archived.</item>
         <item quantity="other">All %1$d episodes of this playlist have been archived.</item>
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlayAllHandler.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlayAllHandler.kt
@@ -30,6 +30,10 @@ class PlayAllHandler @AssistedInject constructor(
         return handlePlayAllAction(episodes, PlaylistEpisode::toPodcastEpisode)
     }
 
+    suspend fun handlePlayAllEpisodes(episodes: List<BaseEpisode>): PlayAllResponse {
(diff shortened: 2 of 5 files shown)


sztomek:
@claude please review this

modules/services/localization/src/main/res/values/strings.xml
```diff
@@ -266,6 +266,10 @@
     <string name="tv_playlists_download_title">Create playlists on your phone</string>
     <string name="tv_playlists_download_subtitle">They’ll be waiting here when you’re done. Scan to get the app</string>
     <string name="tv_playlist_play_all">Play all episodes</string>
+    <string name="tv_playlist_play_all_clear_up_next_title">This will clear your Up Next</string>
```
geekygecko:
```suggestion
    <string name="tv_playlist_play_all_clear_up_next_title">This will clear your Up\u00A0Next</string>
```
It would be good to remove the widow word here so that Next isn't left on a line on its own.

modules/services/localization/src/main/res/values/strings.xml
```diff
@@ -266,6 +266,10 @@
...
+    <string name="tv_playlist_play_all_clear_up_next_message">We can save your current Up Next as a playlist</string>
```
geekygecko:
```suggestion
    <string name="tv_playlist_play_all_clear_up_next_message">We can save your current Up Next as a\u00A0playlist</string>
```

geekygecko (approved):

geekygecko:
Thanks for always including an image or video with your pull requests.

sztomek:
Thanks — addressed the latest round in `d707e00d23`:

- **#1 `OpenNowPlaying` firing with nothing played** — `PlayAllHandler.playAllPendingEpisodes()` now returns a `Boolean`, and `replaceUpNextAndPlay` only emits `OpenNowPlaying` when it actually played something. Closes the `rememberSaveable`-restore path where the VM (and its `pendingEpisodes`) is gone.
- **#2 silent failures / `playAll()` crash** — replaced `runCatching` with explicit `try/catch` that rethrows `CancellationException` and logs via `Timber.e`, matching `TvHomeViewModel`; wrapped the save, the play, and `playAll()` itself so an exception can no longer take the process down. I deliberately log rather than add a new failure-toast event/string, to avoid the string proliferation flagged in the nits — happy to add one if you'd prefer the user-visible surface.
- **#3 cross-action re-entrancy window** — both `playAll()` and `replaceUpNextAndPlay()` now bail on a shared `isBusy` check (either job active), so a second press during a slow save can't overwrite `pendingEpisodes` mid-flight.
- **#4 `pendingEpisodes` never cleared** — nulled out at the end of `playAllPendingEpisodes()`.
- **#5 test gaps** — added `advanceUntilIdle()` to the rapid-replace test; new tests for the `playAll()` re-entrancy guard, the cross-action guard (replace ignored while play-all runs), and `OpenNowPlaying` not emitting when there's nothing to play.

Not changing (nits): the `extraBufferCapacity = 2` comment, `rememberUpdatedState`, and the `@JvmName` rename — all called out as convention/justified rather than defects; and the CHANGELOG checkbox, which matches the surrounding TV PRs.

`:tv:testDebugUnitTest` (TvPlaylistDetailsViewModelTest) and `:modules:services:repositories` + `:modules:features:filters` compile all pass locally; spotless clean.

PR #5721, opened 2026-08-11
Migrate Files Settings screen to Jetpack Compose

clementiano9:
## Description
- Migrates `CloudSettingsFragment` (the **Files Settings** screen) to Jetpack Compose, adding `CloudSettingsPage` and deleting `fragment_cloud_settings.xml`
- Follows the pattern already used by `HeadphoneControlsSettingsFragment`, reusing the shared `SettingSection` / `SettingRow` / `SettingRowToggle` / `Banner` components
- Reworks `CloudSettingsViewModel` to expose a single `StateFlow<UiState>` instead of a `LiveData` sign-in state plus direct `settings.*.value` reads in the fragment. This also fixes a latent issue: the old screen read each switch value **once** at bind time, so a preference changed elsewhere (e.g.by sync) while the screen was open would not be reflected
- Adds `CloudSettingsViewModelTest` 

Fixes # [753](https://github.com/Automattic/pocket-casts-android/issues/753)

## Testing Instructions
1. Sign out (or use a free account) so the Plus rows render locked
4. Go to **Profile → Files → ⋮ → Files settings**
5. Toggle **Auto Add To Up Next** and **Delete local file**; reopen the screen and confirm the values persisted
6. Confirm **Delete cloud file** is *hidden* (not merely disabled) while not Plus
7. Tap any of the three locked Plus rows, the upgrade screen should open
8. Tap **Learn More** in the banner, the upgrade screen should open
9. Dismiss the banner with **✕**; reopen the screen and confirm it stays dismissed
10. Sign in as Plus/Patron and confirm all six toggles are enabled, **Delete cloud file** appears, and the banner is gone
11. Confirm the analytics still fire: `settings_files_shown` on open, and the per-toggle events on each change

## Screenshots 

new_banner
Screenshot_20260811-183017
Screenshot_20260811-183042

## Checklist
- [x] I have updated (or requested that someone edit) [the Event Horizon schema](https://github.com/
Automattic/EventHorizonSchemas/blob/trunk/schema/pocket-casts.yml) to reflect any new or changed analytics.

- [x] with different themes


modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/cloud/CloudSettingsPage.kt
```diff
@@ -0,0 +1,371 @@
...
+                role = Role.Switch,
+            ) {
+                if (isPlus) onCloudOnlyOnWifiChange(!uiState.cloudDownloadOnlyOnWifi) else onUpgradeClick()
+            },
+        )
+    }
+}
+
+/**
+ * Mirrors the layout of the `upgradeLayout` block from the deleted `fragment_cloud_settings.xml`:
```
geekygecko, line 228: It might not be necessary to mention the removed XML view file.

geekygecko (approved):
Looks great, thanks for the change.

modules/features/profile/src/main/java/au/com/shiftyjelly/pocketcasts/profile/cloud/CloudSettingsPage.kt
clementiano9: Removed the comment

PR #5706, opened 2026-08-06
[TV] Add a Now Playing tab with basic playback

sztomek:
## Description

Adds a **Now Playing** tab to the Android TV app's top bar and wires up real playback for the first time, mirroring the Apple TV app's behavior. Scope agreed for this PR: basic playback screen (play/pause, seek, skip), video + audio (audio shows static artwork), and Up Next queue interactions — analytics and everything else the tvOS player has is deliberately left for follow-ups.

** IMPORTANT **
Please note that the UI is not final, figma (Ftk3KwnfqaK4g57yCN63p0-fi-2639_2612) shows apple tv's built-in player interface. Screenshots attached to the linear ticket is different, so this needs to be clarified with design team and make amends when necessary.

**What's in it:**
- The TV app now calls `PlaybackManager.setup()` (it previously only seeded the Up Next queue and every Play button was a no-op).
- A dynamic **Now Playing** tab between Up Next and Search, mirroring tvOS: it only exists while there is a current episode, and auto-redirects to Home if the queue empties while selected. Tab selection is now keyed on the tab (not its index) so the list can change shape safely.
- A **playback screen**: static artwork + titles for audio, an aspect-fitted video surface for video episodes, a focusable seek bar (DPAD left/right seeks by the user's skip amounts), skip back/forward and play/pause buttons with a buffering spinner, and error text. During video playback the controls and top bar auto-hide after 4s; any navigation press reveals them (media keys keep their function).
- **Current-episode actions** on the player (`…` button): details, go to podcast, mark played/unplayed, archive/unarchive — the tvOS player ellipsis set, reusing `TvEpisodeActionsModal` with a new `NowPlaying` context.
- **Entry points that start playback**: Up Next row click, a new *Play* action in the episode actions modal (all contexts), and the Home screen video-tile play buttons (discover episodes are fetched via `findOrDownloadPodcastRxSingle` first; failures show a toast). Starting playback switches to the Now Playing tab via a `LocalOpenNowPlaying` composition local.
- **Shared-module fixes** needed to make playback work on TV: the metered-network stream warning is skipped on TV (it has no warning UI — same exclusion as Auto/Wear), and `getLaunchActivityPendingIntent` falls back to the leanback launch intent (the compat media session path would otherwise NPE on release TV builds, which only declare a `LEANBACK_LAUNCHER` activity).

**Notable decisions:**
[...]

PR #5737, opened 2026-08-13
Fix Android TV app being unable to play audio; add Now Playing screen

nolengreenspan:
## Description

The Android TV app can browse podcasts and manage the Up Next queue, but has no `MediaSession` and therefore **cannot play audio at all** — `tv/src/main/AndroidManifest.xml` never declared the `PlaybackService`/`LegacyPlaybackService` that the phone module ships with, and `TvApplication.onCreate()` only ever called `upNextQueue.setupBlocking()`, never `playbackManager.setup()`.

This PR:
- Adds the missing playback services to the TV manifest (copied from `app/src/main/AndroidManifest.xml`), plus `FOREGROUND_SERVICE`, `FOREGROUND_SERVICE_MEDIA_PLAYBACK`, and `WAKE_LOCK`. Services are declared `enabled="false"` and switched on at runtime via `PlaybackServiceToggle.ensureCorrectServiceEnabled()`, matching the phone app's own pattern.
- Replaces the TV-only `upNextQueue.setupBlocking()` call with `playbackManager.setup()`, which already calls `setupBlocking()` internally (so this replaces rather than duplicates the queue sync) and also starts the media session.
- Adds a **Now Playing** screen: artwork or a 16:9 video surface for video episodes, progress/buffered track, seek-adjacent skip buttons (using the user's own skip-seconds settings), play/pause, playback speed, and a fullscreen toggle for video.
- Adds a **mini-player drawer**, reachable from any screen via the remote's Options/Menu key, so playback can be controlled without leaving whatever you're browsing.
- Adds a `LAUNCHER` category alongside `LEANBACK_LAUNCHER` so the app also shows outside the Leanback row on devices that support it.

Fixes # 

## Testing Instructions

1. Build and install `:tv:assembleDebugProd` on an Android TV / Google TV device.
2. From Your Podcasts or a podcast's episode list, select an episode and press Play.
   - **Before this PR:** nothing happens — no audio, no session, `Now Playing` isn't reachable.
   - **After this PR:** audio starts playing.
3. Navigate to the new **Now Playing** tab in the top bar — shows artwork, title, progress, and skip/play-pause/speed controls that all work.
4. Play a video episode and use "Fullscreen" — a 16:9 `SurfaceView` shows the picture (previously video episodes played audio-only with no surface at all).
5. From any other screen, press the remote's Options/Menu key — the mini-player drawer slides in with the same transport controls, and Menu (or Back) dismisses it.

Tested on a Philips 4K Google TV device (`Philips_4K_A1`, Android 11).

## Screenshots or Screencast

Happy to attach TV screenshots of the Now Playing screen and mini-player drawer if useful — let me know.

## Checklist
- [ ] I have updated (or requested that someone edit) the Event Horizon schema to reflect any new or changed analytics 

- [ ] with different themes

This is a draft while I finish the remaining checklist items (compose previews, tests) — happy to take direction on scope/approach before investing further.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 8.19
 -----
+*   Bug Fixes
+    *   Fix the Android TV app being unable to play audio (no MediaSession was
+        registered), and add a Now Playing screen
+        ([#0000](https://github.com/Automattic/pocket-casts-android/pull/0000))
 
 
 8.18
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 *   Bug Fixes
     *   Fix the Android TV app being unable to play audio (no MediaSession was
         registered), and add a Now Playing screen
-        ([#0000](https://github.com/Automattic/pocket-casts-android/pull/0000))
+        ([#5737](https://github.com/Automattic/pocket-casts-android/pull/5737))
 
 
 8.18
--- a/tv/src/main/AndroidManifest.xml
+++ b/tv/src/main/AndroidManifest.xml
@@ -3,6 +3,10 @@
     xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools">
 
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE_MEDIA_PLAYBACK" />
+    <uses-permission android:name="android.permission.WAKE_LOCK" />
+
     <uses-feature
         android:name="android.software.leanback"
         android:required="true" />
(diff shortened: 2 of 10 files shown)


sztomek:
Hey @nolengreenspan — first off, thank you for taking the time to dig into the TV module and put together this PR, genuinely appreciated. 🙏

It looks like your branch is based on a fork from before our recent TV work landed. On current main the Now Playing screen already exists (added in #5706 back on the 10th, with advanced controls following in #5710) — `TvNowPlayingScreen`, `TvNowPlayingViewModel`, video surface, seekbar, playback speed and all. `TvApplication` also already calls `playbackManager.setup()` rather than `upNextQueue.setupBlocking()`, and audio does play today. So the core of this PR unfortunately overlaps with what's already shipped, and the new files collide with existing ones at the same paths.
**The TV app is under active development on our side right now**, which is why things are moving fast and a fork from a few days ago can already be behind. Because of that I'm going to close this one out — but please don't let it put you off contributing!

One thing you spotted that is still worth doing: the TV manifest doesn't yet declare `PlaybackService/LegacyPlaybackService`, so system media-session and background integration aren't wired up on TV yet. We'll track that separately. Thanks again for the sharp eye, and sorry the timing didn't line up on this one. 🙏

PR #5741, opened 2026-08-13
[TV] Episode actions & details analytics

sztomek:
## Description

Fifth screen-area of the **Android TV → Apple TV analytics parity** program (EventHorizon only): the shared **episode options modal** and **episode details modal**. Stacked on the Up Next analytics PR. Verified event-by-event against the tvOS app.

| Event | EventHorizon class | Properties | Fires when |
|---|---|---|---|
| `episode_actions_shown` | `EpisodeActionsShownEvent` | `source` | the episode options modal (`TvEpisodeActionsModal`) appears |
| `episode_detail_shown` | `EpisodeDetailShownEvent` | `source` | the episode details / show-notes modal (`TvEpisodeInfoModal`) appears |

tvOS fires these bare (`EpisodeRow.swift:292` `.onAppear` and `EpisodeShowNotesView.swift:26` `.task`, no properties), but both Android EventHorizon classes **require** a `source: EpisodeViewSourceType`. We map it from the surface the modal is shown on:

| Surface | `source` |
|---|---|
| Podcast details | `podcast_screen` |
| Playlist | `filters` |
| Up Next | `up_next` |
| Now Playing | `now_playing` |
| Search | `search` |

Wired the established way: `EventHorizon` injected into `TvEpisodeActionsViewModel` (new `trackActionsShown(source)` on the `TvEpisodeActions` interface) and `TvEpisodeInfoViewModel` (new `trackDetailShown(source)`); each fires once from the modal's `LaunchedEffect(episode.uuid)` on appear. The `source` mapping lives in one place — the new `episodeViewSource` column on the `TvEpisodeActionContext` enum — and both modals derive it from the `TvEpisodeActionContext` passed at each of the 5 call sites, so the two events can't drift apart.

### Two decisions worth calling out

**1. `episode_actions_shown` from Search — intentionally fired (diverges from Apple TV).**
Apple TV's Search/Discover results use a *reduced* menu (`DiscoveryEpisodeMenuButtons`) that has no tracking, so Apple TV does **not** emit `episode_actions_shown` there. Android TV reuses the full `TvEpisodeActionsModal` for search results, and we chose to fire the event from Search too, stamped `source=search`. This is a deliberate divergence for completeness of Android's own analytics — noted here so it's not mistaken for an accidental over-fire.

**2. The `source` mapping lives only in the `TvEpisodeActionContext` enum.**
Each context already carried a `SourceView` (used for playback actions); this PR adds an `episodeViewSource: EpisodeViewSourceType` column alongside it for the two new shown-events. Both the actions modal and the details modal read their `source` from the same `TvEpisodeActionContext`, so a screen can't report `episode_actions_shown` and `episode_detail_shown` with mismatched sources, and a future surface only has to add one enum entry. No existing playback `source` changes here — Search already used `TvEpisodeActionContext.SearchResults` (`SourceView.SEARCH_RESULTS`) on the base branch.

### Not in scope: `up_next_queue_reordered`

Apple TV fires `up_next_queue_reordered` from the shared Play Next / Play Last helper, but **only** when the episode is already in Up Next (a move). Android's `PlaybackManager.playNext/playLast` instead unconditionally fire a *different* event (`episode_added_to_up_next`) with no move-detection, and the reorder event is fired nowhere in the Android codebase. Layering iOS's semantics on top would double-count and needs a `slots` value iOS never sends — so it's deliberately deferred as its own decision.

Fixes PCDROID-724 https://linear.app/a8c/issue/PCDROID-724/podcast-details-analytics

## Testing Instructions

1. `./gradlew :tv:installDebug`.
2. From Podcast details / Playlist / Up Next / Now Playing / Search, open an episode's options menu → confirm one `episode_actions_shown` with the matching `source` in `adb logcat`.
3. From the options menu tap **Episode details** → confirm one `episode_detail_shown` with the same `source`.

## Screenshots or Screencast
SCR-20260814-qgni

## Checklist
- [ ] If this is a user-facing change, I have added an entry in CHANGELOG.md — N/A (analytics only)
- [x] All strings that need to be localized are in localization — N/A
- [x] Any jetpack compose components I added or changed are covered by compose previews — existing previews updated (no-op tracking)
- [x] I have updated (or requested that someone edit) the Event Horizon schema to reflect any new or changed analytics — existing events, no schema change needed

### DIFF
--- a/tv/src/main/java/au/com/shiftyjelly/pocketcasts/component/TvEpisodeActionsModal.kt
+++ b/tv/src/main/java/au/com/shiftyjelly/pocketcasts/component/TvEpisodeActionsModal.kt
@@ -63,7 +63,7 @@ private fun ColumnScope.TvEpisodeActionsModalContent(
     var pendingConfirmation by remember { mutableStateOf<TvEpisodeActionConfirmation?>(null) }
     var returnFocusLabel by remember { mutableStateOf<String?>(null) }
     val focusRequester = remember { FocusRequester() }
-    LaunchedEffect(Unit) {
+    LaunchedEffect(episode.uuid) {
         actions.trackActionsShown(actionContext.episodeViewSource)
     }
     LaunchedEffect(pendingConfirmation) {
--- a/tv/src/main/java/au/com/shiftyjelly/pocketcasts/component/TvEpisodeActionsModal.kt
+++ b/tv/src/main/java/au/com/shiftyjelly/pocketcasts/component/TvEpisodeActionsModal.kt
@@ -21,6 +21,7 @@ import au.com.shiftyjelly.pocketcasts.analytics.SourceView
 import au.com.shiftyjelly.pocketcasts.models.entity.PodcastEpisode
 import au.com.shiftyjelly.pocketcasts.models.type.EpisodePlayingStatus
 import au.com.shiftyjelly.pocketcasts.theme.TvTheme
+import com.automattic.eventhorizon.EpisodeViewSourceType
 import java.util.Date
 import au.com.shiftyjelly.pocketcasts.localization.R as LR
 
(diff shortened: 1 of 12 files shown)


sztomek:
@claude please review this

geekygecko (approved):

PR #5746, opened 2026-08-14
[TV] Search suggestions and history

sztomek:
## Description

Brings the Android TV **Search** screen up to parity with the Apple TV app (`pocket-casts-ios/Pocket Casts TV App/UI/Search`), closing a set of gaps found in a behavioural audit. **Stacked on #5745** (the combined-search crash fix) — review/merge that first; this PR's diff excludes it.

### What changed

| Area | Before (Android TV) | After (Apple TV parity) |
|---|---|---|
| **Debounce** | 2000ms (shared phone remote-config default) | **300ms** TV constant — matches iOS, far snappier |
| **Empty results** | generic "No results" | **"No results for "term""** (shows the query) |
| **Loading** | bare spinner | spinner + **"Searching…"** label (string existed, was unused) |
| **Scope on clear** | reset to Top Results | **preserved** (matches iOS) |
| **Field prompt** | "Search" | **"Podcasts, shows, authors"** |
| **Search flow** | single `combinedSearch` (all-at-once) | **two-phase**: predictive (`autoCompleteSearch`) → fast podcasts + term suggestions, then full search fills in episodes |
| **Suggestions** | none | **term suggestions while typing**, rendered above the keyboard; tap to fill |
| **History** | never saved/shown (manager only used to *wipe* on sign-out) | **saved** on finishing a search / picking a suggestion + **"Recent searches"** row in the idle area |
| **Video episodes** | `has_video` dropped by the DTO | parsed through DTO → model, driving a **Featured video row** |
| **Top Results** | Podcasts → Episodes (vertical) | **Featured (video) → Episodes → Podcasts** horizontal carousels (matches `SearchTopResultsView`) |
| **Podcast ordering** | subscribed-first | **server-relevance first**, subscribed fill gaps (matches iOS) — see flag below |

Shared `servers`/`model` change: `CombinedResult.EpisodeResult` now parses `has_video`, and `ImprovedSearchResultItem.EpisodeItem` carries `hasVideo` (additive, defaults false — benefits phone too).

### ⚠️ Needs design approval
There is an ongoing conversation about the UX, see slack: p1786735532787109-slack-C0ATWH7BNH3

**UPDATE**
UX is approved!

Fixes POC-848 https://linear.app/a8c/issue/POC-848/search-suggestions-and-history

### Notes / follow-ups
- **Suggestions can't live inside the system keyboard** — Android exposes no API to inject candidates into a third-party IME (and a custom keyboard was previously rejected by design), so they render in-app just above the keyboard.
[...]

PR #5747, opened 2026-08-14
[TV] Local folder results in search 

sztomek:
## Description

Adds **local folder results** to Android TV Search. Stacked on the TV search parity PR (#5746).

### Behaviour
- When a **signed-in** user searches, their local folders whose **name** matches the term (case-insensitive) are surfaced, each with its podcasts' artwork.
- Signed-out users get nothing here — the folder lookup is skipped entirely (no DB hit).
- Folders appear as a **Folders** carousel in Top Results, ordered **Featured → Episodes → Folders → Podcasts**; tapping one opens the folder in `TvFolderDetailScreen` (reusing the exact overlay/nav pattern from the Your Podcasts tab, including focus-restore and podcast-opened-from-folder returning to the folder).
- When a search yields at least one folder, a dedicated **Folders** filter pill appears after **Episodes** and shows a folder-only grid. The pill is hidden when there are no folder results; if the user is on the Folders filter and a subsequent search has no folders, the view transparently falls back to **Top Results**.

### Implementation
- `TvSearchViewModel` injects `FolderManager`. `searchFolders()` gates on `syncManager.isLoggedIn()`, filters `folderManager.getAll()` by name, attaches each folder's podcasts. It runs in a `runCatching { … }` `async` **concurrently** with the network search, so a (rare) local-DB failure degrades to "no folders" rather than failing the whole search, and it never delays the remote results.
- `folders` added to `TvSearchState.Results` (+ the NoResults emptiness check). `folderPodcasts(uuid)` feeds the detail screen.
- New `TvSearchFilter.Folders` entry (after `Episodes`). The visible pill set is derived from the current results — `Folders` is only included when the results contain folders — and an `effectiveFilter` keeps the selected pill and the rendered content in sync without mutating ViewModel state when folders come and go. `TvSearchFilters` now takes the filter list to render.
- UI reuses the existing `TvFolderCard`, `TvFolderDetailScreen`, and `TvPodcastGridScaffold` components (the Folders filter grid mirrors the Podcasts filter layout).

Folders are inherently a paid feature (a free user can't create them, and they're wiped on downgrade), so the gate only needs to check that the user is signed in. This matches how the rest of the TV module treats folders — the Your Podcasts tab renders them straight from `FolderManager` with no subscription check. An earlier revision gated on `SignInState.isSignedInAsPlusOrPatron`, but the subscription tier is not reliably hydrated on TV (it's the only place in the module that read it), which silently suppressed every folder even for Plus users; gating on `isLoggedIn()` — the same guard this ViewModel already uses elsewhere — fixes that.

Reuses existing infra throughout — the only shared-module change is the new `search_filters_folders` string alias; otherwise tv-only.

**UPDATE**
got design approval: p1787763569156349/1787663272.838409-slack-C0ATWH7BNH3

## Testing Instructions

1. Sign in with at least one folder whose name matches a query.
2. `./gradlew :tv:installDebug`; open **Search**, type the folder's name → a **Folders** row appears in Top Results, and a **Folders** filter pill appears after Episodes; select it → a folder-only grid; open a folder → its podcasts; open a podcast → back returns to the folder.
3. Refine the query so no folder matches → the **Folders** pill disappears and the view falls back to Top Results.
4. As a **signed-out** user, neither the Folders row nor the Folders pill ever appears.
5. `./gradlew :tv:testDebugUnitTest` — covers signed-in surfaces folders, signed-out skips the lookup, folder-alone counts as a result.

## Screenshots or Screencast

https://github.com/user-attachments/assets/9358a318-19b8-4b17-ad5b-4004852cd14a

## Checklist
- [ ] If this is a user-facing change, I have added an entry in CHANGELOG.md — N/A (TV, pre-release)
- [x] All strings that need to be localized are in localization — reuses existing `folders` via a `search_filters_folders` alias
- [x] I have updated (or requested that someone edit) the Event Horizon schema — N/A (no analytics change)

### DIFF
--- a/modules/services/localization/src/main/res/values/strings.xml
+++ b/modules/services/localization/src/main/res/values/strings.xml
@@ -2883,6 +2883,7 @@
     <string name="search_filters_top_results">Top Results</string>
     <string name="search_filters_podcasts" translatable="false">@string/podcasts</string>
     <string name="search_filters_episodes" translatable="false">@string/episodes</string>
+    <string name="search_filters_folders" translatable="false">@string/folders</string>
     <string name="search_suggestions_view_all">View all for \"%1$s\"</string>
     <string name="search_suggestions_no_results_title">No Results</string>
     <string name="search_suggestions_no_results_message">Try a new search, or by adding a podcast feed URL</string>
--- a/tv/src/main/java/au/com/shiftyjelly/pocketcasts/search/TvSearchFilters.kt
+++ b/tv/src/main/java/au/com/shiftyjelly/pocketcasts/search/TvSearchFilters.kt
@@ -38,6 +38,7 @@ internal fun TvSearchFilters(
     selected: TvSearchFilter,
     onFilterSelect: (TvSearchFilter) -> Unit,
     modifier: Modifier = Modifier,
+    filters: List<TvSearchFilter> = TvSearchFilter.entries,
     upFocusRequester: FocusRequester? = null,
 ) {
     Row(
@@ -47,6 +48,7 @@ internal fun TvSearchFilters(
         FilterDivider(modifier = Modifier.weight(1f))
         Spacer(modifier = Modifier.width(20.dp))
         TvSearchFilterPills(
+            filters = filters,
             selected = selected,
             onFilterSelect = onFilterSelect,
             upFocusRequester = upFocusRequester,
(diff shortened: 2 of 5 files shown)


geekygecko (approved):
Looks great.

I don't see search by folder on mobile or Apple TV search but it seems like a good idea. Maybe we should ask the designers about the extra tab.

Image

Thanks again for including a video in your PR description.

sztomek:
@geekygecko good point, asked the designers here: p1787663272838409-slack-C0ATWH7BNH3

PR #5750, opened 2026-08-18
DO NOT MERGE: verify Sentry upload on CI

mokagio:
**Do not merge.** Stacked on #5708 purely to answer one question, then to be closed and deleted.

## Description

#5708 moves the Sentry upload auth token out of the `configure`-provisioned `secret.properties` and into the `SENTRY_AUTH_TOKEN` environment variable, and inlines the org and per-module project slugs. Nothing in normal CI exercises any of that: `.buildkite/commands/assemble-release-apk.sh` and the prototype lane both pass `-PskipSentryProguardMappingUpload=true`, so the upload tasks are never registered. The only path that registers them is `build_bundle`, reachable solely from the API-triggered `release-builds.yml`. Without this, the first real test of the new wiring is the next code freeze.

This is a re-run. The [first pass](https://buildkite.com/automattic/pocket-casts-android/builds/18420) passed, but against `.configure` pinned at `c65d2bd` — the `secret.properties` that *still carried* `sentryAuthToken`, `sentryOrg` and the per-module project slugs. Following [a review comment](https://github.com/Automattic/pocket-casts-android/pull/5708#discussion_r3926832452), #5708 advanced the pin to `789d0d6`, which drops them, so the tree under test is no longer the tree that was validated.

## What the step asserts

Read the **:sentry: Verify Sentry Upload** job. Every step is expected to pass.

- **The cleanup landed.** `sentryAuthToken`, `sentryOrg`, `sentryAndroidProject`, `sentryAutomotiveProject` and `sentryWearProject` are absent from the provisioned `secret.properties`. The DSNs and `sentryTvProject` are still there by design and are not checked.
- **The upload works for real**, for `app`, `automotive` and `wear` — each declares its own project slug, so a typo in one would otherwise surface only at release. Deliberately not `assembleRelease`: `uploadSentryProguardMappingsRelease` pulls in `minifyReleaseWithR8` to produce the mapping and leaves packaging and signing out of the graph.
- **The guard fails the build on a blank token**, and fails *through the guard* — the assertion greps for its message rather than trusting a non-zero exit. This is new; the first pass only proved the happy path.

A successful run uploads real ProGuard mappings to the three production Sentry projects. Mappings are addressed by debug ID, so a spare one is dead weight rather than corruption.

## What a failure would mean

- **On the `secret.properties` check** — `.configure` is not on the cleaned-up hash.
- **On the guard message** — the guard does not fire when the token is blank, and a release would fail somewhere less legible instead.
- **Inside `sentry-cli`** — the token is reaching Gradle, but an org or project slug, or the token's scope, is wrong.

Nothing here should be "fixed" — close the PR once it has answered the question.

## Checklist
- [x] If this is a user-facing change, I have added an entry in CHANGELOG.md — n/a, CI-only and throwaway
- [x] Ensure the linter passes
- [x] I have considered whether it makes sense to add tests for my changes — n/a, this PR *is* the test


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

mokagio:
Answered. Evidence moved to #5708's description.

- [Build 18798](https://buildkite.com/automattic/pocket-casts-android/builds/18798) — real mapping uploads for `app`, `automotive` and `wear` with the token from the environment, against the cleaned-up `secret.properties` at pin `789d0d6`.
- [Build 18799](https://buildkite.com/automattic/pocket-casts-android/builds/18799) — the blank-token guard fires and fails the build.

Closing without merging, as intended.

PR #5763, opened 2026-08-20
Show account-creation modal recurrently (every 60 days) for logged-out users

yaelirub:
## Description

Port the iOS recurring "Encourage Account Creation" (EAC) modal to Android.

**Background:** signup conversions have been trending down. The EAC modal — the *"We noticed you're not logged in"* prompt with the sync / backups / recommendations benefits — had a positive impact on signups when it ran in 2025, but it was a **one-shot** (shown once then disabled, and suppressed entirely for new installs) tied to app updates. This makes it **recurring**: shown to logged-out users on a **60-day cadence**, decoupled from app updates, matching the iOS change ([pocket-casts-ios#4985](https://github.com/Automattic/pocket-casts-ios/pull/4985) / PCIOS-900).

### What changed

- **`AccountEncouragement` (utils, new)** — pure, unit-testable cadence logic: `decide(isEligible, lastShown, now, interval) → Show / Wait`, with a 60-day interval. Shows on the first eligible launch (no anchor yet), once the interval elapses, or when the stored anchor is in the future (backwards device clock / restored skewed backup, which would otherwise suppress the modal indefinitely).
- **`Settings`** — replaced the one-shot `showFreeAccountEncouragement: Boolean` with `freeAccountEncouragementLastShown: Instant?` (the cadence anchor).
- **`MainActivity.encourageAccountCreation()`** — gates on the new `ENCOURAGE_ACCOUNT_CREATION` feature flag, computes eligibility (logged out **and** past onboarding via `hasCompletedOnboarding()`), and shows the modal on the **first eligible launch**, then every 60 days while the user stays logged out. The anchor is recorded when the modal is shown. The post-restore login prompt anchors the clock so it doesn't double-show.
- **`AppLifecycleObserver`** — fresh installs anchor the cadence at install time, so the modal waits a full 60-day interval before its first show (matching iOS, which anchors in `AppDelegate`'s `case .installed`). Existing users upgrading leave the anchor `null` and see it on the first eligible launch.
- **`Feature`** — added `ENCOURAGE_ACCOUNT_CREATION` (default on, Firebase remote flag + dev toggle) for parity with the iOS feature flag / a remote kill-switch.

Signed-in users never see it; existing inline login prompts are unchanged.

Part of DSGPOC-88.

Fixes PCDROID-728

## Testing Instructions

The real interval is 60 days; to exercise the cadence quickly, use a debug build and adjust the device clock, or clear app data between runs.

1. Ensure the `ENCOURAGE_ACCOUNT_CREATION` feature flag is enabled (Beta Features screen).
2. On a **logged-out** device that has completed initial onboarding, launch the app → confirm the EAC modal **appears on this first eligible launch**.
   - On a **fresh install**, the cadence is anchored at install time, so the modal does **not** appear on the next launch — it waits a full interval.
3. Dismiss the modal and relaunch within the interval → confirm it does **not** reappear.
4. Advance past the 60-day interval (device clock / data reset) and relaunch → confirm it appears again.
5. Log in and relaunch → confirm the modal never appears for a signed-in user.
6. Tablet: confirm the benefits **dialog** (`AccountBenefitsFragment`) shows instead of the full onboarding flow.

> Note: unit tests for the new `AccountEncouragement` decision logic pass (`:modules:services:utils:testDebugUnitTest --tests "*AccountEncouragementTest"`). The `:app` module compiles cleanly. I was unable to run `:modules:features:shared` unit tests locally due to a **pre-existing** `:modules:services:crashlogging` KSP/Dagger build failure (`cannot find symbol: class CrashLogging`) that reproduces on `main` without these changes — flagging separately.

## Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/featureflag/Feature.kt
```diff
@@ -170,6 +170,15 @@ enum class Feature(
         hasDevToggle = true,
         addedOn = LocalDate.parse("2025-11-05"),
     ),
+    ENCOURAGE_ACCOUNT_CREATION(
+        key = "encourage_account_creation",
```
geekygecko: Image

This appears to use an existing feature flag name in Firebase Remote Config. Can we include a different one?

app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
```diff
@@ -683,21 +698,49 @@ class MainActivity :
...
+                if (launchedInitialOnboarding) {
                     return@repeatOnLifecycle
                 }
 
-                if (Util.isTablet(this@MainActivity)) {
-                    AccountBenefitsFragment().show(supportFragmentManager, "account_benefits_fragment")
-                } else {
-                    openOnboardingFlow(OnboardingFlow.AccountEncouragement)
+                // Eligible = logged out and past initial onboarding. Shown on the first eligible
+                // launch, then every 60 days while the user stays logged out.
```
geekygecko: Would it be possible to add to the prompt for Opus to reduce the comments to one line? It seems to be overexplaining and producing comments that might not be necessary.

geekygecko (approved):
Looks good!

Is it showing the message too soon after they first install the app? They’ve already decided not to create an account, but then the next time they open the app, they’re presented with this page.

https://github.com/user-attachments/assets/18209393-9821-4a02-bfad-cbfb7c44c229

yaelirub:
Thanks for the review, @geekygecko . You're right about the modal appearing first launch after declining account creation in onboarding, and this turned out to be a divergence from iOS rather than intended behaviour.

Fixed in 99efa04 by anchoring on fresh install in `AppLifecycleObserver` (the direct analogue of the iOS hook) and removing the fragile completion-handler anchoring. Added coverage for both halves: new installs anchor, upgrades don't.

I've also corrected the PR description, which claimed dropping the new-install suppression "matched iOS" — that was backwards.

modules/services/utils/src/main/java/au/com/shiftyjelly/pocketcasts/utils/featureflag/Feature.kt
yaelirub: Update name key name in 5999dad4c1fe60d4f2688b090b053cd2b6bddcd2. Also updated on iOS https://github.com/Automattic/pocket-casts-ios/pull/5029

app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivity.kt
yaelirub: Yes! Updated here: 6fcf1cebb901eb903eb8103cf681c2c0c8792f4d. I wish we could use a shared Claude memory

yaelirub:
> * `app/.../ui/MainActivity.kt:717` — signing out still re-triggers the modal on the next foreground (see the existing thread on this block). The anchor is only written on show / restore / fresh install, so a user signed in since before this ships has a `null` anchor; sign out → background → return gives them _"We noticed you're not logged in"_ immediately. Either anchor on sign-out or confirm iOS does the same.

I think it's good that it shows up in the next foreground after sign-out as most sign outs are accidental

PR #5767, opened 2026-08-25
Remove obsolete `android_build_preflight`

mokagio:
## Description

The action runs checks that are mostly unnecessary:

- Decrypt secrets, which should already be decrypted
- Check gems are up to date, unrelated to app builds
- Ensure `bundletool` is available, but builds don't use it — its only
  client is `android_send_app_size_metrics`, which runs its own check

See https://github.com/wordpress-mobile/release-toolkit/blob/9322fff453fb4622d4d1add8040dc8d13cee2a1d/lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_build_preflight.rb#L6-L23

The `skip_prechecks` block stays — only the `android_build_preflight` line goes, and the `@param` doc now names the checks that remain.

Fixes https://linear.app/a8c/issue/AINFRA-2965

## Testing Instructions

Validated on CI by throwaway PR #5768, which ran `build_and_upload_to_play_store` **without** `skip_prechecks:true` — so the block this PR edits actually executed — and stopped before the Play Store upload. The build [passed](https://buildkite.com/automattic/pocket-casts-android/builds/18485).

## Checklist
- [x] If this is a user-facing change, I have added an entry in CHANGELOG.md
- [x] Ensure the linter passes (`./gradlew spotlessApply` to automatically apply formatting/linting)
- [x] I have considered whether it makes sense to add tests for my changes
- [x] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml`
- [x] Any jetpack compose components I added or changed are covered by compose previews

PR #5768, opened 2026-08-25
[DO NOT MERGE] Validate build lane for #5767

mokagio:
## Validation only — do not merge

Stacked on #5767 ([AINFRA-2965](https://linear.app/a8c/issue/AINFRA-2965)).
Close this PR once the job has answered; never merge it.

## Why the ordinary pipeline misses this

#5767 removes `android_build_preflight` from `build_and_upload_to_play_store`, and now also drops the `skip_prechecks` parameter that gated the block it sat in.

Two things keep that code away from PR CI. The lane is only reached from `.buildkite/commands/release-build.sh`, which the release pipeline triggers — PR jobs run `build_and_upload_prototype_build` instead. And until #5767 that script passed `skip_prechecks:true`, so the block never executed even on the release pipeline. It has effectively never run on CI.

## Which job to read

**Validate build lane with prechecks ON (no Play Store upload)** is expected to **pass**.
Red means #5767 actually broke the lane — do not "fix" this PR.

The step runs `build_and_upload_to_play_store skip_confirm:true`. The prechecks are unconditional now, so invoking the lane is enough to exercise them.

Assertions:

- The lane prints `VALIDATION: prechecks block completed` — the code the PR edited ran end to end.
- The lane prints `VALIDATION: bundle built` with a non-zero size — a signed Release AAB was produced, so nothing downstream depended on the `configure_apply` that `android_build_preflight` used to run.

The AAB is attached as a build artifact.

Note that every check now left in that block is guarded by `unless is_ci` or by `skip_confirm`, so on CI this job exercises the path rather than the checks themselves. Its value is proving the lane still runs end to end and still produces a bundle.

## What CI still cannot answer

Whether the Play Store would accept the bundle. On this branch the loop over `APPS` is replaced by a single `app` build, and `upload_to_play_store`, `download_universal_apk_from_google_play` and `create_gh_release` are deleted rather than exercised.

### DIFF
--- a/.buildkite/commands/validate-build-lane.sh
+++ b/.buildkite/commands/validate-build-lane.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+echo "--- :rubygems: Setting up Gems"
+install_gems
+
+echo "--- :closed_lock_with_key: Installing Secrets"
+bundle exec fastlane run configure_apply
+
+echo "--- :hammer_and_wrench: Running build_and_upload_to_play_store with prechecks ON (stop before Play Store upload)"
+LOG="$(mktemp)"
+# Deliberately no `skip_prechecks:true`: the prechecks block is the code this PR changed.
+bundle exec fastlane build_and_upload_to_play_store skip_confirm:true | tee "${LOG}"
+
+if ! grep -F 'VALIDATION: prechecks block completed' "${LOG}"; then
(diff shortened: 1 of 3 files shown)


mokagio:
Answered: prechecks block runs and a 52 MB Release AAB is produced. Evidence moved to #5767; closing as planned.

mokagio:
Validation answered: build #18509 passed, both `VALIDATION:` markers printed, 52,040,683-byte AAB, no Play Store upload, and the `ensure_git_*` actions short-circuited on CI as intended. Closing as intended — this PR must never be merged.

PR #5773, opened 2026-08-25
[TV] Add onboarding analytics

sztomek:
## Description

Ports the Apple TV onboarding analytics to Android TV. The Android TV onboarding flow (Welcome, Sign In, Create Account, Syncing) previously fired **no** analytics; this brings it to parity with tvOS so both platforms report the same onboarding funnel.

Events added:

| Screen | Event | Properties |
| --- | --- | --- |
| Welcome | `setup_account_shown` | `flow` |
| Welcome | `setup_account_button_tapped` | `flow`, `button = sign_in` |
| Welcome | `setup_account_button_tapped` | `flow`, `button = create_account` |
| Welcome | `browse_no_account_tapped` | — |
| Sign In | `sign_in_shown` | — |
| Create Account | `create_account_shown` | `flow` |
| Syncing | `sign_in_sync_shown` | — |

Implementation mirrors the existing merged TV analytics pattern (`EventHorizon` injected into a `@HiltViewModel` exposing `trackXxx()` methods, invoked from the screen via `CallOnce { }`).

The catalog bump also required adapting two unrelated call sites to a newly-required `flow` property (`user_account_created`, `create_account_next_button_tapped`); both pass `OnboardingFlowType.Unknown` as that layer has no onboarding flow in scope.

The pin is `pocket-casts-2026-08-27_07-40-16` rather than the `08-25` build so it also carries the new `SearchResultFilterType.Folders` value ([EventHorizonSchemas#125](https://github.com/Automattic/EventHorizonSchemas/pull/125)). Going `08-25` → `08-27` adds no further required-property breakage — only the enum value.

With #5769 (TV Search analytics) now merged, this branch also flips the TV Search folder filter chip from the placeholder `SearchResultFilterType.Unknown` to the real `SearchResultFilterType.Folders` (`TvSearchViewModel`), so focusing the Folders chip reports `filter = folders` instead of `unknown` — closing the blocking finding from [that thread](https://github.com/Automattic/pocket-casts-android/pull/5769#discussion_r3869815592). Covered by a new `TvSearchViewModelTest` case.

### Decisions / notes
- **`flow = OnboardingFlowType.Unknown`** on the onboarding screen events. tvOS sends no flow, but the Android generated events require a non-null `OnboardingFlowType`.
- **`sign_in_shown` fires when the Sign In screen is shown** (not on the Welcome "Sign In" tap), which covers both entry points into sign-in: the Welcome screen and the browse-without-account Home.
[...]

PR #5783, opened 2026-08-26
[TV] Show an alert when the server signs the user out

sztomek:
## Description

Adds a **server-forced sign-out alert** to the Android TV app, matching Apple TV. When a user is signed out by the **server** (token revoked / refresh failed / account removed) rather than by their own action, iOS shows a full-screen alert instead of silently dropping them; Android TV previously surfaced nothing (only a system notification, which is invisible on a TV). This closes functional gap #5 from the Android-TV-vs-iOS-tvOS onboarding parity audit and lands the deferred `signed_out_alert_shown` analytics event (one of the three events split out of the onboarding-analytics PR #5773).

### How it works
- **Signal at the source (shared):** `UserManager` gains `onServerSignOut: SharedFlow<Unit>`, emitted from `signOut()` only when `wasInitiatedByUser == false` (the Android equivalent of iOS's `user_initiated`). Transient (`replay = 0`) = in-session only, matching iOS.
- **TV observes it:** `TvOnboardingNavHost` collects the flow and navigates to a new `SIGNED_OUT` route from any screen.
- **Alert screen:** `TvSignedOutScreen` (Compose + tv-material3, styled like `TvWelcomeScreen`) shows "You've been logged out" + message + a **Log in** button; `signed_out_alert_shown` fires on appear. The button fully wipes local data (via `TvSignOutManager`) and returns to Welcome — matching iOS's `logout()`.
- **No double-count:** `TokenErrorNotification` (which already fires `SignedOutAlertShownEvent`) is suppressed on TV, so the in-app alert is the single event source.

A user-initiated Log Out (Profile) is unaffected — it passes `wasInitiatedByUser = true`, so no alert and no emit.

### Known limitation (conscious decision)
Because the TV token-error notification is suppressed, a forced sign-out that surfaces **only** via the OkHttp interceptor's swallowed `RefreshTokenExpiredException` (rather than `RefreshPodcastsThread` or the account-removed listener) won't show the alert until the next podcast refresh runs (`refreshOnLaunch` + periodic). The primary forced-sign-out paths are covered; this is a delay, not a permanent gap.

Fixes POC-867 https://linear.app/a8c/issue/POC-867/server-forced-signout-alerts

## Testing Instructions
1. Sign in on Android TV.
2. Trigger a server-forced sign-out (revoke the session server-side / let the refresh token expire so the next podcast refresh fails).
3. The **"You've been logged out"** alert appears over whatever screen you were on.
4. Press **Log in** → local data is wiped and you land on the **Welcome** screen.
5. Confirm a normal Profile → **Log Out** does **not** show the alert (it goes straight to Welcome as before).

## Screenshots or Screencast
Screenshot_20260826_125009

## Checklist
- [x] I have updated (or requested that someone edit) the Event Horizon schema to reflect any new or changed analytics. (`signed_out_alert_shown` already exists in the schema; no new events added.)

- [ ] with different themes (TV is single dark theme)
- [ ] with a landscape orientation (TV is landscape-only)

### DIFF
--- a/modules/services/localization/src/main/res/values/strings.xml
+++ b/modules/services/localization/src/main/res/values/strings.xml
@@ -273,6 +273,8 @@
     <string name="tv_sign_in_step_scan">Scan the QR code or go to %1$s</string>
     <string name="tv_sign_in_step_login">Log in to your account</string>
     <string name="tv_sign_in_step_confirm_code">If asked, confirm with the code below</string>
+    <string name="tv_account_signed_out_alert_title">You\'ve been logged out</string>
+    <string name="tv_account_signed_out_alert_message">There was a problem and you\'ve been logged out.\nTap the button below to log in to your Pocket Casts account again.</string>
     <string name="tv_create_account_modal_title">Your podcasts deserve a home</string>
     <string name="tv_create_account_modal_subtitle">Create a free account to follow podcasts, build your Up Next, and pick up right where you left off, on any device.</string>
     <string name="tv_create_account_modal_step_create">Create your account or log in</string>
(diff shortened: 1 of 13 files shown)


geekygecko (approved):
This might work well with the changes to the account authentication PR. https://github.com/Automattic/pocket-casts-android/pull/5775

PR #5786, opened 2026-08-26
[TV] Reveal the user's real podcast covers while syncing

sztomek:
## Description

On the Android TV **syncing** screen (shown after a successful sign-in / account creation), the artwork was a static, infinitely-scrolling row of bundled decorative covers. This replaces it with the user's **real podcast covers**, revealed with an iOS-style **pop-in stack** as the sync restores the library — matching Apple TV's `SigningInView`.

### How it works
- `TvSyncingViewModel` already collects `findSubscribedFlow()` into `uiState.podcastUuids` (which grows as sync brings podcasts in). The screen now maps those to artwork URLs via `PodcastImage.getMediumArtworkUrl(uuid)` and reveals them.
- **Reveal:** covers stack centered (272dp, 16dp corners); each springs in — opacity 0→1, scale 0.6→1, rotation 0→±2–10° (alternating lean); **max 2** stay on screen, the oldest fades out (0.5s) as a 3rd appears; ~0.8s between reveals; the pool recycles for small libraries via a pure `nextDistinctCoverIndex` helper (no duplicate covers stacked).
- **Empty/tiny library** (e.g. a brand-new account with 0 podcasts): falls back to the bundled artwork so the screen still animates.
- **Completion is unchanged** — the screen still advances on `uiState.syncComplete` (3s min + subscription-queue drain), and the reveal loop stops when the screen leaves composition. (iOS's fade-to-black-before-handoff is intentionally out of scope to keep the existing completion logic untouched.)

Scope is artwork/animation only — the copy ("Welcome back!" etc., vs iOS "Good to see you") is a separate audit item (the copy-discrepancies branch).

Fixes POC-868 https://linear.app/a8c/issue/POC-868/dynamic-assets-on-syncing-screen

## Testing Instructions
1. Sign in on Android TV as a **returning user** (an account that has podcasts).
2. On the syncing screen you should see your **own** covers pop in one at a time, stacked and gently rotated, cycling until sync completes and the app moves to Home.
3. New account with 0 podcasts → the bundled artwork animates as a fallback.

## Screenshots or Screencast

https://github.com/user-attachments/assets/a337cd1a-690b-4734-855a-3fff013b3a3c

## Checklist
- [x] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml` (no string changes)
- [x] I have updated (or requested that someone edit) the Event Horizon schema to reflect any new or changed analytics. (No analytics changes.)

- [ ] with different themes (TV is single dark theme)
- [ ] with a landscape orientation (TV is landscape-only)


geekygecko (approved):
It looks great seeing your own artwork being shown!

PR #5797, opened 2026-08-27
[TV] Play video previews on focused Made for TV tiles

sztomek:
## Description

Adds inline **video previews** to the **Made for TV** discover tiles (Apple TV parity): when a tile is focused, a muted-aware preview of the episode video fades in and plays; the extracted poster frame is shown while idle. This PR also aligns the tile chrome with the Apple TV app and makes the focus-loss transition smooth.

A `video_preview_list` discover display style + `DiscoverEpisode` video URL carry the preview source down to the tile; the media3 video-frame decoder and a `MediaMetadataRetriever`-based poster loader back the preview.

**Apple TV parity (tile chrome)**
- Removed the inline **Play this episode** / **Go to podcast** buttons that appeared on focus. On tvOS (`DiscoverVideoEpisodeCell`) the video cell has no inline buttons — the whole card is a single button.
- The podcast artwork + podcast title + episode title info row is now shown **at all times** (including while the preview is playing), matching tvOS's always-visible `infoContent`.
- Selecting the card plays the episode; **Go to podcast** moves to **long-press** — the Android-TV analog of the tvOS Menu-key context menu — so the action stays reachable without an inline button.

**Smoother preview transition on focus loss**
The transition from a playing preview back to the still poster was jarring. Root causes and fixes:
- `PlayerView` defaults to a `SurfaceView`, whose pixels live in a separate compositor layer and **don't blend with Compose `.alpha()`** — so the intended fade never actually rendered. Switched the surface to a **`TextureView`** (`view_tv_video_preview.xml`, `surface_type=texture_view`) with a transparent shutter so the fade composites correctly and no black frame can flash.
- The preview now **fades the video out (alpha → 0, revealing the poster) while the player is still rendering**, and only tears the player down once it is invisible — never releasing a visible surface.
- The `ExoPlayer` is created **lazily**, only for the focused tile and only after the 2s focus-settle delay (one instance at a time), and its **decoder is freed once the fade completes**. This avoids allocating a player for every composed tile and prevents hardware-decoder exhaustion when arrowing across a row of video tiles on low-end TV hardware.

Sound behavior was already faithful to tvOS and is unchanged: 2s play delay, 0.5 preview volume, muted when a podcast is already playing, 300 ms fade, 30 s cap, no looping.

Fixes POC-874 https://linear.app/a8c/issue/POC-874/discover-made-for-tv-should-play-inline-video

## Testing Instructions

1. Sign in (or browse without an account) and open the **Home** tab on the Android TV app.
2. Focus a tile in the **Made for TV** row (e.g. *Daily Tech News Show (VIDEO)*).
3. Confirm there are **no inline buttons** — only the podcast artwork + podcast/episode title row is shown, at all times.
4. Wait ~2s: the muted-aware video preview fades in and plays.
5. Arrow away to another tile: the preview should **cross-fade smoothly back to the still poster** (no black flash, no snap).
6. Press and hold **SELECT** on a tile: it navigates to the podcast (**Go to podcast**).
7. Press **SELECT**: the episode starts playing and opens Now Playing.
8. Arrow quickly back and forth across several video tiles and confirm playback stays healthy (no decoder errors / previews still appear).

## Screenshots or Screencast

https://github.com/user-attachments/assets/fc18009b-cba6-4670-818e-2d60be6374b8

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

tv/src/main/java/au/com/shiftyjelly/pocketcasts/component/TvVideoPreviewPlayer.kt
```diff
@@ -0,0 +1,136 @@
...
+    AndroidView(
+        factory = { viewContext ->
+            (LayoutInflater.from(viewContext).inflate(R.layout.view_tv_video_preview, null) as PlayerView).apply {
+                useController = false
+            }.also { playerViewHolder.view = it }
+        },
+        update = { view -> view.player = player },
+        onRelease = { view -> view.player = null },
+        modifier = modifier.alpha(videoAlpha),
+    )
```
geekygecko: I wonder if it's possible to do this with Compose instead. 

> While the media3-ui-compose library doesn't include out-of-the-box composables (such as buttons, indicators, images, or dialogs), you can find a [demo app written fully in Compose](https://github.com/androidx/media/tree/release/demos/compose) that avoids any interoperability solutions like wrapping [PlayerView](https://developer.android.com/reference/androidx/media3/ui/PlayerView) in [AndroidView](https://developer.android.com/reference/kotlin/androidx/compose/ui/viewinterop/AndroidView.composable#AndroidView(kotlin.Function1,androidx.compose.ui.Modifier,kotlin.Function1)).

https://developer.android.com/media/media3/ui/compose#compose-demo-app

geekygecko (approved):
Looks great!

tv/src/main/java/au/com/shiftyjelly/pocketcasts/component/TvVideoPreviewPlayer.kt
sztomek: thanks, TIL!
updated the code and now it uses the compose way for the preview! 🙏

PR #5805, opened 2026-08-28
[TV] Home/Discover robustness: per-row retry, region fallback, live local rows

sztomek:
## Description

**Android TV parity — Section B / PR 5: Home/Discover robustness.** Makes the TV Home feed degrade and update like tvOS: failed rows are visible and retryable, the feed never hard-fails on region resolution, and local rows reflect reality without relogging.

- **Per-row failure → inline retry.** `ListRepository.getListFeed` swallowed errors to `null` and the loader dropped null rows, so on a flaky connection rows silently vanished. Added `ListRepository.getListFeedResult(): Result<ListFeed?>` (getListFeed now delegates to it — phone behavior byte-identical); the TV row loaders `getOrThrow()` so a genuine fetch **failure** becomes a `TvDiscoverRow.Failed` (title + "Try again"), while an empty feed is still dropped. Pressing Try again reloads just that row (`reloadHomeRow`/`reloadSearchRow`) and replaces it in place. iOS renders the same per-row retry (`DiscoverRetryView`).
- **Region fallback.** `resolveRegion` fell back to `error(...)`, error-screening the whole feed; it now falls back to `"us"` then any available region (matches iOS `DiscoverManager`).
- **Live local rows.** `TvHomeViewModel` rebuilt rows only on login-state change; it now observes Up Next queue changes, playback state, and the new-releases flow, and rebuilds **only** the local rows (Keep Listening / Up Next / New Releases) with a ~1 s debounce, reading the latest state atomically so a refresh can't clobber a concurrent retry or full reload. Discover (network) rows keep their load-once behavior.
- **Keep Listening session rule.** Shown only until playback first starts in the session (iOS `playbackNotYetStarted`), instead of permanently duplicating the Now Playing tab after listening begins.
- **Up Next row threshold.** Requires ≥2 queued beyond the current episode (iOS), was ≥1.
- Unknown discover row types are now logged instead of silently dropped.

## Testing Instructions

1. `debugProd` TV build.
2. **Retry:** with a flaky/offline connection, load Home → a failed discover row shows its title + "Try again"; restore connectivity and press it → the row loads in place.
3. **Region:** with an unresolvable country code, Home still loads (no error screen).
4. **Live rows:** add/remove Up Next episodes or start playback elsewhere → Keep Listening / Up Next / New Releases update within ~1 s without leaving Home.
5. **Keep Listening:** it shows on launch, then disappears once you start playing something this session.
6. **Up Next:** the row appears only with ≥2 queued beyond the current episode.

Unit tests in `TvHomeViewModelTest`: failed row → `Failed` (not dropped) + successful re-fetch; unresolvable region → us-fallback; queue change → local rows rebuilt; Keep Listening hidden after playback starts; Up Next hidden at 1 queued.

## Screenshots or Screencast

Reproduced on the Android TV emulator by forcing the discover feed's network fetches to fail, then restoring connectivity and pressing **Try again**. Both columns are the same emulator under the same failing-connection condition.

**Failed discover rows — before vs after**

| Before (`main`) | After (this PR) |
| --- | --- |
| Failed rows are silently dropped; Home skips straight to *Browse categories*. | Failed rows stay visible as their title + **Try again**, each individually retryable. |

pr5805_discover_retry_before_after

**Per-row recovery**

| Pressing "Try again" reloads just that row in place (rows recover independently) |
| --- |
| <video src="https://github.com/user-attachments/assets/c6f67f50-6c07-49b0-a646-f4efe934ee4c" /> | 

## Checklist
- [ ] If this is a user-facing change, I have added an entry in CHANGELOG.md — n/a (TV-only, no phone-facing change)
- [x] I have considered whether it makes sense to add tests for my changes — `TvHomeViewModelTest` covers each item
- [x] All strings that need to be localized are in localization — reuses existing `try_again`
- [x] Any jetpack compose components I added or changed are covered by compose previews — the failed row is trivial; VM logic is unit-tested
- [x] I have updated the Event Horizon schema to reflect any new or changed analytics — n/a (no analytics change)

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/lists/ListRepository.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/lists/ListRepository.kt
@@ -32,6 +32,14 @@ class ListRepository(
     }
 
     suspend fun getListFeed(url: String, authenticated: Boolean? = false): ListFeed? {
+        return getListFeedResult(url, authenticated)
+            .onFailure { exception ->
+                Timber.e(exception, "Failed to fetch list feed $url")
+            }
+            .getOrNull()
+    }
+
+    suspend fun getListFeedResult(url: String, authenticated: Boolean? = false): Result<ListFeed?> {
         return runCatching {
             if (authenticated == true) {
                 checkNotNull(syncManager) { "Sync Manager is null" }
@@ -43,10 +51,6 @@ class ListRepository(
                 listWebService.getListFeed(url)
             }
         }
-            .onFailure { exception ->
-                Timber.e(exception, "Failed to fetch list feed $url")
-            }
-            .getOrNull()
     }
 
     suspend fun getCategoriesList(url: String): List<DiscoverCategory> {
--- a/tv/src/main/java/au/com/shiftyjelly/pocketcasts/discover/TvDiscoverFeedAnalytics.kt
+++ b/tv/src/main/java/au/com/shiftyjelly/pocketcasts/discover/TvDiscoverFeedAnalytics.kt
@@ -98,6 +98,7 @@ class TvDiscoverFeedAnalytics(
 
         is TvDiscoverRow.Banner,
         is TvDiscoverRow.Categories,
+        is TvDiscoverRow.Failed,
         -> null
     }?.takeUnless { it in localRowIds }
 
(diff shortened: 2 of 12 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

PR #5812, opened 2026-08-31
[TV] Reveal browse category covers only once their artwork has loaded

sztomek:
## Description

On Android TV, the Discover **Browse categories** cards reveal two peeking podcast covers (fade + scale + horizontal slide) when a card gains focus. Because a card's cover URLs are only fetched on first focus, the reveal spring — which was driven purely by focus progress — had already run to completion by the time the artwork arrived. The result was jarring: `TvArtworkImage`'s black placeholder ("shadow") slid in first, then Coil swapped the decoded bitmap in with no transition, so the image "popped" in abruptly.

This change gates each cover's reveal on its artwork actually being decoded. `CategoryCover` now loads via `rememberAsyncImagePainter` and observes `painter.state` (the same idiom used by `ProfileImage`/`CoilImage`); the reveal spring targets `1f` only when the card is focused **and** the image is `Success`. The cover stays fully transparent and off-edge until its bitmap is ready, then springs in already showing the real artwork — no placeholder shadow, no pop. The request also enables `crossfade(true)` as a belt-and-suspenders smoothing of the swap. The focus gradient behind the covers still fades on focus as before.

Loading is safely decoupled from the reveal: Coil starts the request in the painter's `onRemembered` and decodes at `SizeResolver.ORIGINAL`, so an alpha-0 / off-edge / not-yet-drawn cover still loads — there is no "reveal waits for image, image waits for reveal" deadlock.

Only `tv/src/main/java/au/com/shiftyjelly/pocketcasts/component/TvCategoryTile.kt` changes.

## Testing Instructions

1. Launch the TV app and choose **Browse without an account** (or sign in).
2. On Home, scroll down to the **Browse categories** row.
3. Move focus onto a category card (e.g. Arts) for the first time.
4. Observe: the two peeking covers now fade/scale/slide in **already showing real artwork** — the black placeholder no longer slides in ahead of the image, and the image no longer pops in abruptly.
5. Move focus across other cards (Business, Comedy, …) and confirm the same clean reveal. The same row also appears on the Search screen.

## Screenshots or Screencast

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/dcdebc50-3690-4859-9c57-a955fa75c89f" /> | <video src="https://github.com/user-attachments/assets/8ba543ba-0543-4835-b23c-7aa641575c2b" /> |

## Checklist

- [ ] with different themes

### DIFF
--- a/tv/src/main/java/au/com/shiftyjelly/pocketcasts/component/TvCategoryTile.kt
+++ b/tv/src/main/java/au/com/shiftyjelly/pocketcasts/component/TvCategoryTile.kt
@@ -3,6 +3,7 @@ package au.com.shiftyjelly.pocketcasts.component
 import androidx.compose.animation.core.Spring
 import androidx.compose.animation.core.animateFloatAsState
 import androidx.compose.animation.core.spring
+import androidx.compose.foundation.Image
 import androidx.compose.foundation.background
 import androidx.compose.foundation.interaction.MutableInteractionSource
 import androidx.compose.foundation.interaction.collectIsFocusedAsState
@@ -19,6 +20,7 @@ import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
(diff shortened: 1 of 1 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

geekygecko (approved):
This looks great. 

Don't worry if it's too hard, but I wonder if they should be gated together, otherwise it could cause one to pop in quite a bit later. 

https://github.com/user-attachments/assets/1639d8c1-a138-484b-a914-d210f6a65c6c

sztomek:
Good call @geekygecko — gated them together now. Both covers share a single reveal animation that only fires once *every* cover in the tile has loaded (`covers.all { it.isReady }`), so a slow second image no longer pops in after the first; they slide in as a pair. Painter creation + `Success` tracking is hoisted to the parent tile and the two `CategoryCover`s just render the shared reveal progress.

sztomek:
@geekygecko great idea, updated the animation and now the cover artworks sync, so the anim only plays when both images are loaded!

https://github.com/user-attachments/assets/d08107dc-e57d-4685-a1b6-594c828be570

PR #5814, opened 2026-08-31
Fix Huawei/HiSilicon Android ≤9 native playback crash via a fingerprint device-capability policy

sztomek:
## Description

Huawei/Honor devices on **Android 8/9 (SDK ≤ 28)** hard-crash a few seconds into playback (silent native crash, process dies and cold-restarts, no Java exception). Confirmed via a Play Console native tombstone:

```
libstagefright.so → android::ACodec::BaseState::onOMXFillBufferDone(...)
```

**Root cause:** when synced transcripts is enabled, the fingerprinting pipeline opens a **second `MediaCodec` audio decoder** (the eager full-file decode on downloads; the catch-up + on-demand chapter resolves on streams) concurrent with the playback decoder. On these devices the vendor OMX codec corrupts its output-buffer bookkeeping under concurrency + teardown and returns a bad buffer, faulting inside the platform media framework. `ACodec` only exists on Android ≤ 9 (Codec2 replaced it on 10+), which matches the affected slice, and it's a clean 8.11→8.15 regression (the second decoder shipped with fingerprinting). The Rust fingerprint library is ruled out — its panics are caught at the FFI boundary and it contains no `unsafe`.

**Fix:** a single device-capability policy, `FingerprintPolicy { PLATFORM, TAP_ONLY, DISABLED }`, resolved by `FingerprintDecodePolicy`:
- **Every non-affected device → `PLATFORM`**: the existing code path is untouched (one enum compare of overhead; the RC override is never even consulted for them).
- **Huawei/HiSilicon Android ≤ 28 → `TAP_ONLY`** (default): the second decoder is never opened (eager pass, catch-up, and on-demand resolves are gated off), while the passive PCM **tap** — which reuses the *playback* decoder and opens no codec — keeps building the transcript/chapter map in realtime. This makes the pipeline codec-equivalent to crash-free 8.11.
- **`DISABLED`**: the whole subsystem is off (kill-switch).

The tier is overridable per affected population via a Firebase Remote Config string (`fingerprint_policy_override`), so we can demote `TAP_ONLY → DISABLED` (or restore `PLATFORM`) without a release. Analytics reuse the existing free-form `markUnavailable(reason = "unsupported_device")` string, so **no EventHorizon schema change** is needed.

**Why safe for other devices:** `FingerprintDecodePolicy.resolve()` returns `PLATFORM` before reading the RC override for anything that isn't affected, so a global override value can never flip a Pixel/Samsung/Android-10+ device; a malformed override degrades to the affected default, never to `PLATFORM`-off. No shared playback-path files are edited (`ShiftyRenderersFactory` / `ShiftyAudioProcessorChain` / `DefaultAudioSink` / the playback decoder are untouched).

**User-facing impact on affected devices:** playback no longer crashes; transcript highlighting still works in realtime; the only loss is precise generated-chapter/transcript seeks that require an on-demand decode (they fall back to the chapter's own start time).

### Known limitation / follow-up
The device match is scoped to HiSilicon: manufacturer/brand contains `huawei`/`honor` **and** `Build.HARDWARE`/`BOARD` contains `kirin` or a `hi<digit>` codename (covers the Kirin 6xx `hi6250` and Kirin 9xx `hi36xx` families — Honor 9, MediaPad M5, Honor 6X/7X, P Smart, Mate 10 Lite, etc.), on SDK ≤ 28. This avoids over-capturing the many non-crashing Huawei Android-9 devices into feature-lite mode. The one gap: a **MediaTek**-based Huawei/Honor (e.g. Honor 9A, a secondary Android-8.1 report) is not matched — if its tombstone confirms the same `ACodec` fault, broaden by dropping the SoC narrowing (manufacturer + SDK ≤ 28 alone). Because the tier default and the RC override are tunable, this can be revisited without a rearchitecture.

We have **no physical affected device**; the decisive validation is a staged rollout watching crash-free-users on the affected slice, with RC → `DISABLED` as the instant fallback.

Fixes PCDROID-673 https://linear.app/a8c/issue/PCDROID-673/android-9-app-fully-crashes-5s-into-playback-on-815-regression-from

## Testing Instructions

1. `./gradlew :modules:services:utils:testDebugUnitTest` and `:modules:services:repositories:testDebugUnitTest` pass. New/updated tests:
   - `FingerprintDecodePolicyTest`: `PLATFORM` for non-affected devices and SDK ≥ 29; `TAP_ONLY` for Huawei/Honor HiSilicon (incl. `hi6250` Kirin 6xx) ≤ 28; the RC override is ignored on non-affected devices and parsed only for affected ones; empty/garbage overrides fall back to `TAP_ONLY`.
   - `GeneratedChapterSeekerTest`: under `TAP_ONLY` a dense-map hit resolves and a miss returns null — both **without** calling `resolvePlaybackTime` (no second decoder); under `DISABLED` it returns null without touching the timing manager.
   - `DriftFilterTest`: `resolvePlaybackTime` under a restricted policy returns `Unresolved(unsupported_device)` (with no dense anchors) rather than decoding.
2. On a normal device (unaffected), behavior is unchanged: synced transcripts and generated-chapter seeks work exactly as before.
3. To exercise the restricted path on any device, set the Remote Config `fingerprint_policy_override` to `TAP_ONLY` (or `DISABLED`) — note the override only takes effect on devices the policy already classifies as affected; for a non-affected test device you'd temporarily relax `isAffected` locally.

## Checklist
- [x] All strings that need to be localized are in `modules/services/localization/...` 
- [x] I have updated (or requested that someone edit) the Event Horizon schema to reflect any new or changed analytics

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/fingerprint/ChapterSeekResult.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/fingerprint/ChapterSeekResult.kt
@@ -14,5 +14,6 @@ sealed interface ChapterSeekResult {
         const val REASON_AUDIO_UNAVAILABLE = "audio_unavailable"
         const val REASON_NO_MATCH = "no_match"
         const val REASON_METERED_NETWORK = "metered_network"
+        const val REASON_UNSUPPORTED_DEVICE = "unsupported_device"
     }
 }
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/fingerprint/FingerprintTimingManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/fingerprint/FingerprintTimingManager.kt
@@ -20,6 +20,8 @@ import au.com.shiftyjelly.pocketcasts.utils.Network
 import au.com.shiftyjelly.pocketcasts.utils.Util
 import au.com.shiftyjelly.pocketcasts.utils.featureflag.Feature
 import au.com.shiftyjelly.pocketcasts.utils.featureflag.FeatureFlag
+import au.com.shiftyjelly.pocketcasts.utils.fingerprint.FingerprintDecodePolicy
+import au.com.shiftyjelly.pocketcasts.utils.fingerprint.FingerprintPolicy
 import com.automattic.eventhorizon.EventHorizon
 import com.automattic.eventhorizon.SyncedTranscriptsPreparationCompletedEvent
 import com.automattic.eventhorizon.SyncedTranscriptsPreparationFailedEvent
@@ -73,6 +75,7 @@ class FingerprintTimingManager @Inject constructor(
     private val settings: Settings,
     private val dataSourceFactory: Lazy<ExoPlayerDataSourceFactory>,
     private val pcmTap: FingerprintPcmTap,
+    private val decodePolicy: FingerprintDecodePolicy,
 ) {
 
     /** Who asked for preparation; decides whether streaming over a metered network is acceptable. */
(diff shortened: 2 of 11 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

geekygecko (approved):

PR #5820, opened 2026-09-01
Use the configured skip back interval for the Cast notification rewind button

sztomek:
## Description

When casting, the media notification's rewind button always showed 30 seconds, no matter what the user set for skip back. The Cast skip step was configured from the skip forward value, and the Cast SDK's rewind and forward actions both seek by that same step.

This change:
- Uses the configured skip back interval for the Cast skip step, so the rewind button reflects the user's setting.
- Adds a custom media intent receiver so rewind and forward seek by their own configured intervals instead of a shared one.

One Cast SDK limitation remains: the rewind and forward button icons can't show two different numbers, since the SDK supports a single skip step for both. The seek behaviour is now correct for both directions.

Fixes PCDROID-633 https://linear.app/a8c/issue/PCDROID-633/notification-player-skip-back-shows-wrong-interval-android-814-rc-3

## Testing Instructions

1. Set skip back to something other than 30 seconds (for example 10 seconds).
2. Cast an episode.
3. Confirm the rewind button shows the configured value and rewinds by that amount.
4. Confirm the forward button still skips forward by the configured forward value.

## Screenshots or Screencast

N/A (notification behaviour, verified via `dumpsys notification`).

## Checklist

No app UI changes (Cast notification, managed by the Cast SDK).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 8.21
 -----
+*   Bug Fixes
+    *   Fix the Cast notification rewind button showing 30 seconds regardless of the configured skip back interval
+        ([#5820](https://github.com/Automattic/pocket-casts-android/pull/5820))
 
 
 8.20
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 8.21
 -----
 *   Bug Fixes
-    *   Fix the Cast notification rewind button showing 30 seconds regardless of the configured skip back interval
+    *   Fix the skip forward and back intervals in the Cast notification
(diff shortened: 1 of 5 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>PocketCastsMediaIntentReceiver</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>

</p>

PR #5821, opened 2026-09-01
[TV] Fix Made for TV video playback

sztomek:
## Description

Two fixes for Made for TV video playback.

**Now Playing video did not show.** The video surface can connect before the player is ready. `SimplePlayer.setDisplay` used to drop the surface when that happened, so no video appeared. It now keeps the surface and attaches it once the player is prepared.

**Some Made for TV clips failed to play with a "Something went wrong" toast.** These are curated clips. They are missing from the podcast feed window. The cache server's find-by-episode endpoint also returns no episode for them. So the clip never reaches the database and there is nothing to play. The discover feed already carries a playable media url. It is the same url the tile preview uses. We now build an episode from that url and play it. This matches the Apple TV app.

## Why this is the right approach

The feed url is the only reliable source for these clips. Every uuid lookup fails for them by design: the database, the podcast refresh, and the cache find-by-episode all come back empty. Building the episode from feed data is simple. It adds no new endpoint and reuses a url we already have.

## Testing Instructions

1. Build and install the TV app on a device.
2. Open the **Made for TV** row on Home.
3. Focus a tile and let the preview play.
4. Press **SELECT**. The clip should open Now Playing and the video should play.
5. Try several clips, including ones that used to fail.

## Checklist

- [ ] with different themes

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SimplePlayer.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/SimplePlayer.kt
@@ -139,6 +139,7 @@ class SimplePlayer(
 
         player = null
         prepared = false
+        pendingSurface = null
 
         videoChangedListener?.videoNeedsReset()
     }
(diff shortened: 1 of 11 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

PR #5830, opened 2026-09-02
[DO NOT MERGE] Probe where and how the Android SLACK_WEBHOOK posts

mokagio:
> **Validation only — DO NOT MERGE.** This branch exists to make CI answer one question, and gets closed once it has. Branched off `main`; it is not stacked on anything.

Part of [AINFRA-3066](https://linear.app/a8c/issue/AINFRA-3066) / [AINFRA-3063](https://linear.app/a8c/issue/AINFRA-3063).

## Why CI has to answer this

A `SLACK_WEBHOOK` was just provisioned on the `pocket-casts-android` pipeline. Nothing in this repo has ever used it — the only Slack references in `.buildkite/` are Buildkite's own `notify: - slack: "#build-and-ship"` hooks, which go through the Buildkite Slack integration and not this secret. There is no fastlane Slack code at all.

So two things are unknown and unknowable locally, because the secret exists only on CI:

1. **Which channel** the webhook is bound to.
2. **Whether `username` / `channel` payload overrides are honoured** — legacy incoming webhooks honour them, modern Slack-app webhooks silently ignore them. This decides whether one webhook can serve both Pocket Casts channels or whether each needs its own.

## What to read

The **`:slack: [TEMP] Slack webhook probe`** job. It is **expected to pass**. Two probes run in it:

| Probe | What it proves |
|---|---|
| Real webhook, **no** `channel:` override | Where the webhook actually posts, and whether the `username: 'Pocket Casts Release Bot'` override took effect |
| Deliberately invalid webhook URL | `notify_slack` degrades to a log line instead of failing the job — the job stays green and the script greps for `Slack notification failed` |

The second assertion is the one that could go red: if the invalid URL were accepted, or the rescue never fired, the script exits 1.

One clearly-marked test message will appear in whichever channel the webhook points at:

> :test_tube: **Webhook wiring check — please ignore.** Confirming where `SLACK_WEBHOOK` posts, for AINFRA-3066.

## What CI cannot answer

Whether the resulting message *reads well* to the team, and which bot identity we want long-term — the existing iOS webhook posts as "pocket cats", so we may prefer to match that rather than introduce a second name. Human call, once we see the message.

## Note on the helpers

`send_slack_message` / `notify_slack` are written in the shape intended for the real change, modelled on `wpandroid`'s (the most recently updated Slack implementation in the roster, 2026-08-18), so the probe exercises the actual implementation rather than a stand-in. The `slack_webhook_probe` lane, the probe script, and the pipeline step are the throwaway parts.

PR #5831, opened 2026-09-02
Announce Pocket Casts Android rollouts on Slack from CI

mokagio:
## Description

The job that rolls out a Pocket Casts Android build now posts the Slack announcement that used to be a manual Releases V2 click.

Closes [AINFRA-3066](https://linear.app/a8c/issue/AINFRA-3066), part of [AINFRA-3063](https://linear.app/a8c/issue/AINFRA-3063).

Land this before the Releases V2 change that removes those clicks. Until that side sends `MILESTONE`, the message omits it.

`channel:` is required: `SLACK_WEBHOOK` is shared across products and posts to a test channel unless told otherwise.

A 1% production rollout announces a submission, not a percentage — that is how the scenario submits a build for Google to review.

## Testing Instructions

Wording is covered by `ruby fastlane/test/helpers_test.rb`, which this PR also adds as a CI step.

Slack wiring was checked on throwaway PR #5830 ([build 18722](https://buildkite.com/automattic/pocket-casts-android/builds/18722)): posting to `#pocket-casts-android` as Pocket Cats, and an invalid webhook leaving the job green.

End to end it only runs on a real rollout.

## Checklist

- [x] If this is a user-facing change, I have added an entry in CHANGELOG.md — n/a, CI only
- [x] Ensure the linter passes
- [x] All strings that need to be localized are in `strings.xml` — n/a
- [x] Any jetpack compose components I added or changed are covered by compose previews — n/a
- [x] I have updated the Event Horizon schema — n/a

### DIFF
--- a/.buildkite/commands/update-rollouts.sh
+++ b/.buildkite/commands/update-rollouts.sh
@@ -2,9 +2,20 @@
 
 set -euo pipefail
 
-: "${TRACK:?Missing value}"
-: "${ROLLOUT_PERCENT:?Missing value}"
-: "${RELEASE_VERSION:?Missing value}"
+REQUIRED_ENV_VARS=(TRACK ROLLOUT_PERCENT RELEASE_VERSION)
+
+missing=()
+for var in "${REQUIRED_ENV_VARS[@]}"; do
+  if [[ -z "${!var:-}" ]]; then
+    missing+=("$var")
+  fi
+done
+
+if [[ ${#missing[@]} -gt 0 ]]; then
+  echo "Error: missing or empty environment variables: ${missing[*]}" >&2
+  echo "They are passed by the ReleasesV2 scenario that triggers this pipeline." >&2
+  exit 1
+fi
 
 echo '--- :ruby: Setup Ruby Tools'
 install_gems
--- a/.buildkite/commands/update-rollouts.sh
+++ b/.buildkite/commands/update-rollouts.sh
@@ -24,4 +24,8 @@ echo '--- 🔐 Access Secrets'
 bundle exec fastlane run configure_apply
 
 echo '--- 🚀 Update Rollouts'
-bundle exec fastlane update_rollouts track:"$TRACK" percent:"$ROLLOUT_PERCENT" version:"$RELEASE_VERSION" milestone:"${MILESTONE:-}"
+bundle exec fastlane update_rollouts \
+  track:"$TRACK" \
+  percent:"$ROLLOUT_PERCENT" \
+  version:"$RELEASE_VERSION" \
+  milestone:"${MILESTONE:-}"
--- a/.buildkite/commands/update-rollouts.sh
+++ b/.buildkite/commands/update-rollouts.sh
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+: "${TRACK:?Missing value}"
+: "${ROLLOUT_PERCENT:?Missing value}"
+: "${RELEASE_VERSION:?Missing value}"
+
+echo '--- :ruby: Setup Ruby Tools'
+install_gems
+
+echo '--- 🔐 Access Secrets'
+bundle exec fastlane run configure_apply
+
+echo '--- 🚀 Update Rollouts'
+bundle exec fastlane update_rollouts track:"$TRACK" percent:"$ROLLOUT_PERCENT" version:"$RELEASE_VERSION" milestone:"${MILESTONE:-}"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -31,10 +31,11 @@ steps:
         artifact_paths:
           - "**/build/reports/lint-results*.*"
 
+  # Not on the `linter` queue: those agents only accept an allowlist of commands
+  # (danger, rubocop, swiftlint, validate_gradle_wrapper).
   - label: ":fastlane: fastlane Helper Tests"
     command: ruby fastlane/test/helpers_test.rb
-    agents:
-      queue: "linter"
+    plugins: [ $CI_TOOLKIT ]
 
   - label: 'Unit tests'
     command: ".buildkite/commands/run-unit-tests.sh"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -31,6 +31,11 @@ steps:
         artifact_paths:
           - "**/build/reports/lint-results*.*"
 
+  - label: ":fastlane: fastlane Helper Tests"
+    command: ruby fastlane/test/helpers_test.rb
+    agents:
+      queue: "linter"
+
   - label: 'Unit tests'
     command: ".buildkite/commands/run-unit-tests.sh"
     plugins: [ $CI_TOOLKIT ]
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -31,8 +31,6 @@ steps:
         artifact_paths:
           - "**/build/reports/lint-results*.*"
 
-  # Not on the `linter` queue: those agents only accept an allowlist of commands
-  # (danger, rubocop, swiftlint, validate_gradle_wrapper).
   - label: ":fastlane: fastlane Helper Tests"
     command: ruby fastlane/test/helpers_test.rb
     plugins: [ $CI_TOOLKIT ]
--- a/.buildkite/release-pipelines/update-rollouts.yml
+++ b/.buildkite/release-pipelines/update-rollouts.yml
@@ -3,7 +3,7 @@
 
 # This pipeline is meant to be triggered by the MC scenario in ReleasesV2
 
-# Expected variables passed to the pipeline by ReleasesV2: `TRACK`, `ROLLOUT_PERCENT`
+# Expected variables passed to the pipeline by ReleasesV2: `TRACK`, `ROLLOUT_PERCENT`, `RELEASE_VERSION`, `MILESTONE`
 
 steps:
   - label: "🚂 Update Rollouts"
@@ -15,7 +15,7 @@ steps:
       bundle exec fastlane run configure_apply
 
       echo '--- 🚀 Update Rollouts'
-      bundle exec fastlane update_rollouts track:"$TRACK" percent:"$ROLLOUT_PERCENT"
+      bundle exec fastlane update_rollouts track:"$TRACK" percent:"$ROLLOUT_PERCENT" version:"${RELEASE_VERSION}" milestone:"${MILESTONE}"
     plugins: [$CI_TOOLKIT]
     agents:
       queue: "mac-metal"
--- a/.buildkite/release-pipelines/update-rollouts.yml
+++ b/.buildkite/release-pipelines/update-rollouts.yml
@@ -15,7 +15,7 @@ steps:
       bundle exec fastlane run configure_apply
 
       echo '--- 🚀 Update Rollouts'
-      bundle exec fastlane update_rollouts track:"$TRACK" percent:"$ROLLOUT_PERCENT" version:"${RELEASE_VERSION}" milestone:"${MILESTONE}"
+      bundle exec fastlane update_rollouts track:"$TRACK" percent:"$ROLLOUT_PERCENT" version:"$RELEASE_VERSION" milestone:"$MILESTONE"
     plugins: [$CI_TOOLKIT]
     agents:
       queue: "mac-metal"
--- a/.buildkite/release-pipelines/update-rollouts.yml
+++ b/.buildkite/release-pipelines/update-rollouts.yml
@@ -7,15 +7,7 @@
 
 steps:
   - label: "🚂 Update Rollouts"
-    command: |
-      echo '--- :ruby: Setup Ruby Tools'
-      install_gems
-
-      echo '--- 🔐 Access Secrets'
-      bundle exec fastlane run configure_apply
-
-      echo '--- 🚀 Update Rollouts'
-      bundle exec fastlane update_rollouts track:"$TRACK" percent:"$ROLLOUT_PERCENT" version:"$RELEASE_VERSION" milestone:"$MILESTONE"
+    command: .buildkite/commands/update-rollouts.sh
     plugins: [$CI_TOOLKIT]
     agents:
       queue: "mac-metal"
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -1020,7 +1020,7 @@ platform :android do
   # usually run from `main`, whose `version.properties` has already moved on to the next
   # version, so the fallback is only right for a run from the release branch.
   def rollout_announcement(track:, percent:)
-    version = ENV.fetch('RELEASE_VERSION', nil).to_s
+    version = ENV.fetch('RELEASE_VERSION', nil).to_s.strip
     version = release_version_current if version.empty?
 
     # The scenario submits a build for Google to review by rolling production out to 1%, so that
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -341,11 +341,11 @@ platform :android do
   #
   # @param percent [Float] The rollout percentage, between 0 and 1
   # @param track [String] The Google Play track for which to update the rollout of. Must be either `beta` or `production`.
-  # @param version [String] (optional) The release version the announcement is about.
-  #                Typically provided by Releases V2. When omitted, uses the current project version, which on
-  #                `main` has already moved on to the next one.
+  # @param version [String] (optional) The release version the announcement is about. Provided by
+  #                Releases V2. When omitted, uses the current project version, which is only correct
+  #                on a release branch.
   # @param milestone [String] (optional) The milestone to name alongside the version.
-  #                  Typically provided by Releases V2.
+  #                  Provided by Releases V2, which is the only place it can come from.
   #
   lane :update_rollouts do |percent:, track:, version: release_version_current, milestone: nil|
     UI.user_error!('percent parameter must be between 0.0 and 1.0') if percent.to_f.negative? || percent.to_f > 1
@@ -1004,8 +1004,6 @@ platform :android do
   # Slack helpers
   #####################################################################################
 
-  # Sends a Slack message via the incoming webhook.
-  #
   # `channel` is not optional in practice: `SLACK_WEBHOOK` holds the webhook shared across
   # products, which posts to a test channel unless told otherwise.
   def send_slack_message(message:, channel: SLACK_CHANNEL)
@@ -1022,10 +1020,9 @@ platform :android do
     )
   end
 
-  # Posts to Slack without letting the caller fail over it. A rollout that reached Google Play
-  # must not be reported as failed because the announcement did not land. The failure is annotated
-  # and not just logged, because an announcement that quietly stops posting is the one nobody
-  # notices.
+  # Posts to Slack without letting the caller fail over it, reporting loudly and annotating the CI build
+  # (if any) instead. Rationale: A hard failure in the release process is costly to address. A missing
+  # Slack message is not worth halting the release process.
   #
   # Takes a block rather than a string so that building the message is covered too, not just
   # sending it.
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -1025,9 +1025,7 @@ platform :android do
 
     # The scenario submits a build for Google to review by rolling production out to 1%, so that
     # one rollout announces a submission instead of a percentage.
-    if track == 'production' && (percent - 0.01).abs < 0.001
-      return ":announcement: `#{version}` has been submitted to the Production track for Google to review."
-    end
+    return ":announcement: `#{version}` has been submitted to the Production track for Google to review." if track == 'production' && (percent - 0.01).abs < 0.001
 
     milestone = ENV.fetch('MILESTONE', nil).to_s.strip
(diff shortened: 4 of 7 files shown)


fastlane/Fastfile
```diff
@@ -978,6 +985,54 @@ platform :android do
...
+      return ":announcement: `#{version}` has been submitted to the Production track for Google to review."
+    end
+
+    milestone = ENV.fetch('MILESTONE', nil).to_s.strip
+    subject = ["`#{version}`", milestone].reject(&:empty?).join(' ')
+
+    ":announcement: #{subject} has started rolling out to #{(percent * 100).round}% of users."
+  end
+
+
```
dangermattic: <table data-meta="generated_by_pr-check">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/Automattic/pocket-casts-android/blob/a251a984c775e65943a854252447ab97108de140/fastlane/Fastfile#L1035"></span>Layout/EmptyLines: Extra blank line detected.</td>
    </tr>
  </tbody>
</table>

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

fastlane/Fastfile
```diff
@@ -370,6 +375,8 @@ platform :android do
       UI.important(message)
     end
     UI.user_error!('None of the 3 app variants were found in Google Play Console. We expected at least one') if not_found_variants.count == APPS.count
+
+    notify_slack(rollout_announcement(track: track, percent: percent.to_f))
```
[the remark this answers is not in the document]
mokagio: Fixed in ce329b959 — you're right, and it was reachable: `release_version_current` raises if `version.properties` can't be read, which is exactly the fallback path. `notify_slack` now takes a block, so building the message is inside the guard too.

*Posted by Claude (Opus 5) on behalf of @mokagio with approval.*

fastlane/Fastfile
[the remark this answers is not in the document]
mokagio: Fixed in ce329b959.

*Posted by Claude (Opus 5) on behalf of @mokagio with approval.*

fastlane/Fastfile
```diff
@@ -978,6 +985,54 @@ platform :android do
...
+  end
+
+  # Wording for a rollout update, matching what the release scenario used to post by hand.
+  #
+  # `RELEASE_VERSION` and `MILESTONE` are passed by the Releases V2 Buildkite action. Rollouts
+  # usually run from `main`, whose `version.properties` has already moved on to the next
+  # version, so the fallback is only right for a run from the release branch.
+  def rollout_announcement(track:, percent:)
+    version = ENV.fetch('RELEASE_VERSION', nil).to_s
+    version = release_version_current if version.empty?
```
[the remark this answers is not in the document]
mokagio: Declining both halves of this one, with the checks I ran.

**On the branch:** the Buildkite pipeline's `branch_configuration` is `main release/*`, and Releases V2 sets the rollout job's branch to `DEFAULT_BRANCH` (main) unless `publish_draft_github_release` is set — deliberately, per its own comment, "so that it works even after the release branch has been deleted". So rollouts really do usually run from `main`, and the comment stands. I couldn't find a `^release/` requirement in `update-rollouts.yml` or `should-skip-job.sh`; if you're looking at something I missed, point me at it and I'll recheck.

**On the beta version:** switching to `version_name_current` would change the message rather than correct it. Releases V2 posts `%version%` — the release version — for beta rollouts too. Real examples from #pocket-casts-android: `8.19 Intermediate Beta 1 has started rolling out to 100% of users.` and `8.20 Code Freeze has started rolling out to 20% of users.` — no rc suffix. Parity with what the channel sees today is the goal here, so `RELEASE_VERSION` is the right source. Announcing `8.20-rc-3` would be a separate, deliberate change worth its own discussion.

*Posted by Claude (Opus 5) on behalf of @mokagio with approval.*

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

.buildkite/release-pipelines/update-rollouts.yml
```diff
@@ -15,7 +15,7 @@ steps:
       bundle exec fastlane run configure_apply
 
       echo '--- 🚀 Update Rollouts'
-      bundle exec fastlane update_rollouts track:"$TRACK" percent:"$ROLLOUT_PERCENT"
+      bundle exec fastlane update_rollouts track:"$TRACK" percent:"$ROLLOUT_PERCENT" version:"${RELEASE_VERSION}" milestone:"${MILESTONE}"
```
mokagio:
```suggestion
      bundle exec fastlane update_rollouts track:"$TRACK" percent:"$ROLLOUT_PERCENT" version:"$RELEASE_VERSION" milestone:"$MILESTONE"
```

.buildkite/pipeline.yml
```diff
@@ -31,6 +31,12 @@ steps:
         artifact_paths:
           - "**/build/reports/lint-results*.*"
 
+  # Not on the `linter` queue: those agents only accept an allowlist of commands
+  # (danger, rubocop, swiftlint, validate_gradle_wrapper).
```
mokagio:
```suggestion
```

fastlane/lib/helpers.rb
```diff
@@ -0,0 +1,31 @@
...
+# @param version [String] The version the announcement is about.
+# @param milestone [String, nil] The milestone to name alongside the version, when there is one.
+# @param skipped_apps [Array<String>] Variants Google Play had no matching release for.
+# @return [String] The slack message body to use, typically in a call to the `slack()` fastlane action
+#
+def rollout_announcement(track:, percent:, version:, milestone: nil, skipped_apps: [])
+  version = version.to_s.strip
+  subject = ["`#{version}`", milestone.to_s.strip].reject(&:empty?).join(' ')
+
+  # Derived once so the announced percentage and the submission check cannot disagree.
```
mokagio:
```suggestion
```

fastlane/Fastfile
```diff
@@ -334,8 +341,13 @@ platform :android do
   #
   # @param percent [Float] The rollout percentage, between 0 and 1
   # @param track [String] The Google Play track for which to update the rollout of. Must be either `beta` or `production`.
+  # @param version [String] (optional) The release version the announcement is about. Provided by
+  #                Releases V2. When omitted, uses the current project version, which is only correct
+  #                on a release branch.
+  # @param milestone [String] (optional) The milestone to name alongside the version.
+  #                  Provided by Releases V2, which is the only place it can come from.
```
mokagio:
```suggestion
  # @param version [String] (optional) The release version the announcement is about. When
  #                omitted, uses the current project version, which is only correct on a release branch.
  # @param milestone [String] (optional) The milestone to name alongside the version.
  #                  Provided by Releases V2, which is the only place it can come from.
```

fastlane/Fastfile
```diff
@@ -978,6 +1000,40 @@ platform :android do
     UI.user_error!(error_message)
   end
 
+  #####################################################################################
+  # Slack helpers
+  #####################################################################################
+
+  # `channel` is not optional in practice: `SLACK_WEBHOOK` holds the webhook shared across
+  # products, which posts to a test channel unless told otherwise.
```
mokagio:
```suggestion
```

fastlane/Fastfile
```diff
@@ -334,8 +341,12 @@ platform :android do
   #
   # @param percent [Float] The rollout percentage, between 0 and 1
   # @param track [String] The Google Play track for which to update the rollout of. Must be either `beta` or `production`.
+  # @param version [String] (optional) The release version the announcement is about. When
+  #                omitted, uses the current project version, which is only correct on a release branch.
+  # @param milestone [String] (optional) The milestone to name alongside the version.
+  #                  Provided by Releases V2, which is the only place it can come from.
   #
-  lane :update_rollouts do |percent:, track:|
+  lane :update_rollouts do |percent:, track:, version: release_version_current, milestone: nil|
```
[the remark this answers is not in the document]
AliSoftware: I think that's a good point, and easy to fix, so worth addressing IMHO

.buildkite/release-pipelines/update-rollouts.yml
```diff
@@ -15,7 +15,7 @@ steps:
       bundle exec fastlane run configure_apply
 
       echo '--- 🚀 Update Rollouts'
-      bundle exec fastlane update_rollouts track:"$TRACK" percent:"$ROLLOUT_PERCENT"
+      bundle exec fastlane update_rollouts track:"$TRACK" percent:"$ROLLOUT_PERCENT" version:"$RELEASE_VERSION" milestone:"$MILESTONE"
```
AliSoftware: Should we error if `RELEASE_VERSION` is empty, i.e. use `${RELEASE_VERSION:?Empty value}`, to avoid potential issues if we accidentally add a new Task in ReleasesV2 that calls this but forgets to pass those env vars?

And same for `$TRACK` and `$ROLLOUT_PERCENT` while at it, I guess (But missing `$MILESTONE` is not fatal so probably ok).

```suggestion
      bundle exec fastlane update_rollouts track:"${TRACK:?Missing value" percent:"${ROLLOUT_PERCENT:?Missing value}" version:"${RELEASE_VERSION:?Missing value}" milestone:"$MILESTONE"
```

Note: the error message doesn't have to include the name of the env var we say are missing, because `bash` already prints it:

```bash
bash-3.2$ RELEASE_VERSION=''
bash-3.2$ echo "${RELEASE_VERSION:?Missing value}"
bash: RELEASE_VERSION: Missing value
```

mokagio: @AliSoftware I think that's a great idea. I asked Claude for [a validation PR](https://github.com/Automattic/pocket-casts-android/pull/5872) for the syntax as used in a Buildkite pipeline step command node and I'm afraid it showed it wouldn't [parse](https://buildkite.com/automattic/pocket-casts-android/builds/18906/canvas?sid=01a07f97-8b95-4bcf-9ccf-9adb6e57b198&tab=output):

Image

Time to extract this into a script!

.buildkite/commands/update-rollouts.sh
```diff
@@ -0,0 +1,27 @@
...
+fi
+
+echo '--- :ruby: Setup Ruby Tools'
+install_gems
+
+echo '--- 🔐 Access Secrets'
+bundle exec fastlane run configure_apply
+
+echo '--- 🚀 Update Rollouts'
+bundle exec fastlane update_rollouts track:"$TRACK" percent:"$ROLLOUT_PERCENT" version:"$RELEASE_VERSION" milestone:"${MILESTONE:-}"
```
mokagio:
```suggestion
bundle exec fastlane update_rollouts \
  track:"$TRACK" \
  percent:"$ROLLOUT_PERCENT" \
  version:"$RELEASE_VERSION" \
  milestone:"${MILESTONE:-}"
```

fastlane/env/user.env-example
```diff
@@ -5,3 +5,7 @@ GITHUB_TOKEN=<GitHub access token>
 #   Organizations: 'Automattic',
 #   REST Scopes: 'write_builds'
 BUILDKITE_TOKEN=<Buildkite Personal Access Token>
+
+# The Slack incoming webhook used to announce rollouts.
+# Shared across products, so every message must pass an explicit channel.
```
mokagio:
```suggestion
```

fastlane/lib/helpers.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+# The scenario submits a build for Google to review by rolling production out to 1%, so that one
+# rollout announces a submission instead of a percentage.
```
mokagio:
```suggestion
```

fastlane/lib/helpers.rb
```diff
@@ -0,0 +1,30 @@
...
+SUBMISSION_ROLLOUT_PERCENTAGE = 1
+
+# The rollout announcement.
```
mokagio:
```suggestion
# The rollout announcement message for Slack.
```

.buildkite/release-pipelines/update-rollouts.yml
AliSoftware: Oh shame, so despite this being a valid `bash` syntax, because this is interpolated by the `buildkite-agent pipeline upload` pre-processing step, that won't work because the `buildkite-agent`'s preprocessing implementation doesn't support this particular syntax, I see.

AliSoftware: Strange though: https://github.com/buildkite/interpolate 's own README suggests that `${parameter:?[word]}` is supposed to be supported syntax thought 🤔

But it's true that [the official doc of the `pipeline upload` command](https://buildkite.com/docs/agent/cli/reference/pipeline#environment-variable-substitution) doesn't mention that `${parameter:?[word]}` syntax as supported despite the Go lib supporting it 😕

.buildkite/commands/update-rollouts.sh
```diff
@@ -0,0 +1,27 @@
...
+REQUIRED_ENV_VARS=(TRACK ROLLOUT_PERCENT RELEASE_VERSION)
+
+missing=()
+for var in "${REQUIRED_ENV_VARS[@]}"; do
+  if [[ -z "${!var:-}" ]]; then
+    missing+=("$var")
+  fi
+done
+
+if [[ ${#missing[@]} -gt 0 ]]; then
```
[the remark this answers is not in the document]
mokagio: Got my Claude to verify this.  The bash 4.4 change covers *expanding* an empty array (`"${a[@]}"`, `"${a[*]}"`), not the `${#…}` length operator, which returns 0 safely on 3.2.

I ran it on the queue in question (`mac-metal`) via [#5874](https://github.com/Automattic/pocket-casts-android/pull/5874) ([build](https://buildkite.com/automattic/pocket-casts-android/builds/18912#01a07ff6-a959-4825-899c-e6ac3618fb25):

```
BASH_VERSION       = 3.2.57(1)-release
command -v bash    = /bin/bash
/bin/bash          = GNU bash, version 3.2.57(1)-release (arm64-apple-darwin24)
"${a[@]}" on an empty array is an error -> bash < 4.4 semantics
PASS  ${#empty[@]} = 0, guard skipped, shell still running
PASS  exited 0 -- the guard l
PASS  exited 0 under GNU bash, version 3.2.57(1)-release
PASS  stopped through the gua
```

The premise was right. `env bash` on these agents *does* resolve to system bash 3.2.57, and the negative control confirms the shell is genuinely pre-4.4.  The probe also ran the real `update-rollouts.sh` with `install_gems`/`bundle` stubbed: exits 0 on the happy path, and the guard still
fires on a missing `RELEASE_VERSION`.

I think we can keep the line as is. 

But this exploration makes me think we might want to ensure our `mac-metal` and `mac` queues are provisioned with modern Bash.

mokagio:
From the latest Claude review:

> Still open from earlier: whether the ReleasesV2 scenario already sends RELEASE_VERSION. The lane-level fallback is unreachable from CI because .buildkite/commands/update-rollouts.sh:5-17 still requires the variable, so its absence remains a rollout that never happens. (https://github.com/Automattic/pocket-casts-android/pull/5831#discussion_r3954947627)

Ah! Good point. There's a week still before the next release so I'd say we can trust the ReleasesV2 counterpart will land in time for it. (_Unless I just jinxed us?! Ahahahah_)

.buildkite/release-pipelines/update-rollouts.yml
```diff
@@ -3,19 +3,11 @@
...
-    command: |
-      echo '--- :ruby: Setup Ruby Tools'
-      install_gems
-
-      echo '--- 🔐 Access Secrets'
-      bundle exec fastlane run configure_apply
-
-      echo '--- 🚀 Update Rollouts'
-      bundle exec fastlane update_rollouts track:"$TRACK" percent:"$ROLLOUT_PERCENT"
+    command: .buildkite/commands/update-rollouts.sh
```
AliSoftware, line 10: I am not 100% sure this will work?

The `TRACK` / `ROLLOUT_PERCENT` / `RELEASE_VERSION` / `MILESTONE` env vars are  passed by ReleasesV2 when triggering the `pipeline upload`. So those are available at the time the `pipeline upload` does the pre-processing of the `.yml` to create/enqueue the jobs.

But at the time when those jobs will get assigned an agent and that agent will start running the jobs, the environment variables known to those jobs won't have those `TRACK` / `ROLLOUT_PERCENT` / `RELEASE_VERSION` / `MILESTONE` env vars at that stage anymore. So by the time the `update-rollouts.sh` script is run **at job run time** by the bash shell running inside the job, those env vars will not be set and that will fail the build.

This is why we had those `$TRACK` / `$ROLLOUT_PERCENT` etc in the `.yml` pipeline, so that the values would be replaced at the `pipeline upload` time, at the moment when those env vars are passed by ReleasesV2 to the API call triggering the build and the `buildkite-agent pipeline upload "${PIPELINE}"` call.

---

1️⃣  Solution 1: pass the env vars as parameters to your call to `update-rollouts.sh` in the `.yml`:

```yaml
command: .buildkite/commands/update-rollouts.sh "$TRACK" "$ROLLOUT_PERCENT" "$RELEASE_VERSION" "$MILESTONE"`
```
Then in the `update-rollouts.sh` script's implementation, read them back via `$1`/`$2`/`$3`/`$4`.

At this stage you could even directly use `TRACK=${1:?Empty or missing value}` syntax, because at that point that `.sh` will be interpreted by `bash` at job run time, not by the `pipeline upload` pre-processing.
So you could even remove your whole `REQUIRED_ENV_VARS=(TRACK ROLLOUT_PERCENT RELEASE_VERSION)` + `missing=()` and the `for` + `if` etc… to simplify back your `.sh` to this at the top of the file:
```bash
TRACK="${1:?Missing or empty value}"
ROLLOUT_PERCENT="${2:?Missing or empty value}"
RELEASE_VERSION="${3:?Missing or empty value}"
MILESSTONE="${4:-}"
```
Then go back to use them in your `bundle exec fastlane …` call.

2️⃣  Solution 2: propagate the env vars to the job by adding an `env:` attribute to the `command` step in the `.yml`:
```yaml
command: .buildkite/commands/update-rollouts.sh
env:
  TRACK: "$TRACK"
  ROLLOUT_PERCENT: "$ROLLOUT_PERCENT"
  RELEASE_VERSION: "$RELEASE_VERSION"
  MILESTONE: "$MILESTONE"
```
The `$xxx` values in this `env:` attribute will then be interpolated and replaced **at `pipeline upload` time**, so by the time the job actually gets an a agent and starts running the `update-rollouts.sh` script, the environment for that job will contain the resolved values (resolved at pipeline-upload time) associated with those 4 env vars, and when the `update-rollouts.sh` script will run, it will be able to use those env vars directly:
```bash
bundle exec fastlane update_rollouts \
  track:"${TRACK:?Missing or empty value}" \
  percent:"${ROLLOUT_PERCENT:?Missing or empty value}" \
  version:"${RELEASE_VERSION:?Missing or empty value}" \
  milestone:"${MILESTONE:-}"
```

_(We could also use different names for the env var passed to the job's environment, e.g. `ARG_TRACK: $TRACK` in the `.yml` then use `${ARG_TRACK:?Missing or empty value}` in the `.sh`, if that helps you better understand the distinction between what's provided at pipeline-upload time vs run time… but I think it'd be more consistent to keep the same names)_

3️⃣  Solution 3: Keep the initial call site in `update-rollouts.yml` and manage all that at the `Fastfile` level

 - Remove `update-rollouts.sh`
 - Revert to the way the `update-rollouts.yml` was calling `bundle exec fastlane update_rollouts track:"$TRACK" percent:"$ROLLOUT_PERCENT"` directly, allowing those env vars to be substituted at pipeline-upload time
 - Update the lane declaration in the `Fastfile` to not provide any default value for the parameters, then test for `.empty?`-ness on the Ruby side inside the lane.

---

In practice I think I prefer solution 3️⃣ , because it has the benefit to add the guard against empty values not just when the lane is invoked by the CI via a Buildkite pipeline, but also if the lane is called in any other context.

Also, _fastlane_ and Ruby already handle the error cases for when you fail to pass a parameter (e.g. if you have a lane with a `track:` parameter and try to call it without providing a `track:<value>`, it will catch the error and complain. So we might as well centralize the error handling on the fastlane side?

(e.g. replacing `release_version: current_release_version` parameter with default value with `release_version:` without a default value, then `UI.user_error!('Empty release version') if release_version.empty?` + `release_version ||= current_release_version`)

If you still prefer to handle it not at the `Fastfile` end but on the CI + pipeline end, and still want to split the command in a separate `update-rollouts.sh`, I think I'd prefer option 1️⃣  (also because that's the one we've been using in all other cases across repos when we need to handle such argument passing needs)

AliSoftware (changes requested):
Not 100% sure, but let's say 95% sure that this implementation won't work because the env vars in question will only be visible at pipeline-upload time not at job runtime.

Left some solution suggestions as alternative approaches that should work instead.

PR #5836, opened 2026-09-02
Fix multi-select clearing on rotation in podcast episode list

sztomek:
## Description

Fixes PCDROID-709 https://linear.app/a8c/issue/PCDROID-709

On the podcast screen, entering multi-select (long-press an episode) and rotating the device caused the app to leave multi-select mode and deselect every episode, forcing the user to rebuild the selection from scratch. This is especially painful when bulk-archiving a podcast's back catalogue.

### Root cause

`PodcastFragment` holds the multi-select state in `MultiSelectEpisodesHelper` / `MultiSelectBookmarksHelper`, which are constructor-injected into the retained `PodcastViewModel`, so their state (the `isMultiSelecting` flag and the `selectedSet`) already survives a configuration change.

The problem was `PodcastFragment.onPause()`, which unconditionally did:

```kotlin
viewModel.multiSelectEpisodesHelper.isMultiSelecting = false
viewModel.multiSelectBookmarksHelper.isMultiSelecting = false
```

`MultiSelectHelper.isMultiSelecting`'s setter clears `selectedSet` and empties the selected-list LiveData. Rotation destroys and recreates the activity, and `onPause()` fires during that teardown — wiping the selection even though the ViewModel that holds it is retained.

### The fix

Guard the `onPause()` clear with `!requireActivity().isChangingConfigurations`, so the selection is only cleared when the fragment is genuinely being left (navigation, backgrounding) and not during a configuration change such as rotation. This is the same idiom already used across the codebase (e.g. `PlayAllFragment`, `CreatePlaylistFragment`, `AddToPlaylistFragment`, playlists `SettingsFragment`).

On rotation the retained helpers keep both the `isMultiSelecting` flag and the selected set; `onViewCreated` re-runs `setupMultiSelect()`, and the existing `isMultiSelectingLive` observer restores the multi-select toolbar and re-renders the checkmarks. No new state plumbing was needed.

### Scope: which lists are affected

This bug is a shared pattern, but the fix is scoped to the podcast episode list as requested.

- **Podcast episode list (`PodcastFragment`) — fixed here.** Helper lives in the retained ViewModel; the only cause of loss was the `onPause()` clear.
- **Downloads / Starred / History (`ProfileEpisodeListFragment`), Files (`CloudFilesFragment`), Up Next (`UpNextFragment`), Filters/Playlists (`PlaylistFragment` via `PlaylistEpisodesAdapterFactory`)** all share the same class of bug, but they hold the helper as a **Fragment-scoped `@Inject` field** (re-injected fresh on recreation) **and** clear it in `onPause`. Fully fixing those needs the helper moved into a retained ViewModel per screen in addition to the same `onPause` guard — a larger, separate change not included here. See related PCDROID-686.

### Config change vs process death

This fix covers **rotation / configuration changes only**. On process death the `PodcastViewModel` (and therefore the helper's `selectedSet`) is destroyed, so the selection is not restored; that would require persisting the selection via `SavedStateHandle`, which is out of scope for this fix.

## Testing Instructions

Manual repro:
1. Open any followed podcast with several episodes (e.g. from Podcasts tab → tap a podcast) to reach its episode list.
2. Long-press an episode to enter multi-select mode; tap a few more episodes so several are selected (the multi-select toolbar shows the count).
3. Rotate the device portrait ↔ landscape.
4. Expected (with this fix): multi-select stays active and all previously selected episodes remain selected. Before the fix: multi-select exits and the selection is cleared.

## Screen recordings
| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/79a45e7d-64a3-4b67-a106-2b4479d1f8e4" /> | <video src="https://github.com/user-attachments/assets/ea600fac-6ed8-428e-bbc8-eb79c623db82" /> | 

## Checklist
- [x] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml` (no strings changed)
- [x] Any jetpack compose components I added or changed are covered by compose previews (no Compose changes)
- [x] I have updated (or requested that someone edit) the Event Horizon schema to reflect any new or changed analytics (no analytics changes)

- [ ] with different themes


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

geekygecko (approved):
Should this have a changelog entry as it's a user facing bug?

PR #5838, opened 2026-09-03
Upgrade Kotlin and KSP

geekygecko:
## Description
Updates the app to the latest Kotlin, KSP and AGP versions, with a few follow-on changes needed to keep the build green.

Kotlin goes from 2.3.21 to 2.4.10, KSP from 2.3.9 to 2.3.11, and AGP from 9.3.2 to 9.4.0. Two source changes were required because the project compiles with `allWarningsAsErrors`:

- The `-Xannotation-default-target=param-property` compiler argument is removed. It was a migration flag for the annotation-use-site-target change, and it is no longer needed on this Kotlin version.
- In `String.parseIsoDate()`, an explicit `toString()` inside a string template is replaced with `$this`, which the newer compiler flags as redundant.

### Removing the `kotlin-stdlib` resolution strategy

The Kotlin bump made every `lintAnalyze*` task fail:

```
Can't initialize detector com.android.tools.lint.checks.InferredThreadDetector.
```

Lint only reports an `InvocationTargetException`, so the underlying cause has to be dug out by constructing the detector directly against lint's resolved classpath:

```
KotlinReflectionInternalError: Function 'forEach' (JVM signature:
forEach(Ljava/lang/Iterable;Lkotlin/jvm/functions/Function1;)V)
not resolved in file class kotlin.collections.CollectionsKt: no members found
```

`kotlin-reflect` and `kotlin-stdlib` have to be on the exact same version. The root `build.gradle.kts` had an `allprojects` block forcing every `kotlin-stdlib*` artifact to the project's Kotlin version on *every* configuration. Lint resolves its own runtime through AGP's `androidLintTool` configuration, so the force applied there too, raising stdlib to 2.4.10 while lint's bundled `kotlin-reflect` stayed at 2.2.10. Harmless on Kotlin 2.3.21, fatal on 2.4.10.

That block has been removed rather than patched. It was added in #2791 as a Kotlin *ceiling*: the project was on Kotlin 1.9.24 and `aboutlibraries` was pulling stdlib 2.0 into the build. Every dependency now requests 2.4.10 or lower (`aboutlibraries` 15.1.1 included), so Gradle resolves to 2.4.10 on its own and the block no longer does anything for the app graph.

Written as `useVersion` it also does not cap tool classpaths, it raises them. It was doing the same thing to the KSP, Hilt and Kotlin compiler plugin classpaths, which were running stdlib 2.4.10 against `kotlin-reflect` 1.6.10 to 2.0.10. Lint is simply the first tool that did a reflective lookup close enough to the seam to notice. Removing the block puts those classpaths back on the versions their tools were built against.

Two notes:

- Upgrading AGP does not fix this on its own. Lint pins its Kotlin at 2.2.10 in 32.3.2, 32.4.0 and 32.5.0-alpha03 alike, and the failure reproduces on AGP 9.4.0.
- `kotlin-stdlib-jdk7`/`jdk8` now resolve to 1.9.24 rather than 2.4.10. Those artifacts have been empty stubs since Kotlin 1.8 folded their contents into `kotlin-stdlib` (one `module-info.class`, no code), so nothing changes in the built app.

## Testing Instructions

Smoke test the app: sign in, play an episode, browse Discover, and open a podcast's episode list.

Verified locally on AGP 9.4.0, all with `--rerun-tasks`:

- `./gradlew aggregatedLintRelease` and `./gradlew :app:assembleDebug` pass
- `./gradlew spotlessCheck` passes
- `./gradlew buildHealth` passes, with a report byte-identical to before the change

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

PR #5841, opened 2026-09-03
Apply the Sentry Gradle plugin to the tv module (on #5708)

mokagio:
## Description

Same change as #5840 — apply the Sentry Gradle plugin to `tv`, which is the only app module that never has, leaving the Android TV app's crashes unsymbolicated — but built on top of #5708 instead of `main`.

There the Sentry project slugs are inlined as the public identifiers they are, so `tv` follows suit with `projectName = "pocket-casts-tv"` and needs no `mobile-secrets` entry for the slug at all. The DSN stays a secret, as it does for every other module on that branch.

**Do not merge this and #5840.** They are the same change in the two worlds; whichever base lands first, close the other. This one exists so that the TV work isn't held hostage to #5708's review, and #5708 isn't held hostage to the TV beta.

## Gotchas

Stacked on `ainfra-2790-remove-sentryproperties-in-favor-of-public-values-and-env`, so the diff against `main` will show #5708's changes too until that merges. Review only the `tv/build.gradle.kts` commit.

`sentryTvDsn` still needs a `mobile-secrets` entry, and still needs #5682's `IS_TV_BUILD` to take effect at release time. Neither blocks this.

## Testing Instructions

`./gradlew :tv:tasks --all | grep -i sentry` lists the same Sentry tasks as `:wear:tasks --all`. `./gradlew spotlessCheck` passes.

### DIFF
--- a/.configure
+++ b/.configure
@@ -1,7 +1,7 @@
 {
   "project_name": "pocketcasts-android",
   "branch": "trunk",
-  "pinned_hash": "789d0d6ec3d5e23886ca25d767fb63452082412c",
+  "pinned_hash": "1e17440ffb52c59087b22337fccfcfcdb8b4aa8f",
   "files_to_copy": [
     {
       "file": "android/pocket-casts/secret.properties",
--- a/tv/build.gradle.kts
+++ b/tv/build.gradle.kts
@@ -2,8 +2,13 @@ plugins {
     alias(libs.plugins.android.application)
     alias(libs.plugins.ksp)
     alias(libs.plugins.hilt)
+    alias(libs.plugins.sentry)
     alias(libs.plugins.compose.compiler)
     alias(libs.plugins.google.services)
+}
+
+sentry {
+    projectName = "pocket-casts-tv"
 }
 
 android {


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-android/milestone/135">8.21</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

sztomek:
thanks for considering the tv app as well 🙏

AliSoftware (approved):

AliSoftware (commented):
As mentioned in https://github.com/Automattic/pocket-casts-android/pull/5708/changes#r3949800161, we need to add the `sentryTvDsn` to `secrets.properties` too.

The `.configure`'s `pinned_hash` in https://github.com/Automattic/pocket-casts-android/pull/5708 points to `mobile-secrets`' commit `789d0d6`, which does not yet include the `sentryTvDsn` value.

This PR should thus at least point `.configure`'s `pinned_hash` to `1e17440`, which includes the added `sentryTvDsn`, or to a commit hash that is later after that one (in particular, `1e17440` also contains `sentryTvProject` while https://github.com/Automattic/pocket-casts-android/pull/5708 has it inlined in the `.kts` already, and also contains comments above that `sentryTvProject` declaration (mentioning it could be inlined) that will now be obsolete, so we could probably add a commit in `mobile-secrets` that removes `sentryTvProject=` and those related comments from `secrets.properties` and point the `pinned_hash` to that new commit instead.

AliSoftware (changes requested):
We need to update `.configure`'s `pinned_hash` to `1e17440` and re-run `configure_apply` (so it updates `.configure-files/secrets.properties.enc` to the one containing the `sentryTvDsn` property) before merging this.

.configure
```diff
@@ -1,7 +1,7 @@
 {
   "project_name": "pocketcasts-android",
   "branch": "trunk",
-  "pinned_hash": "789d0d6ec3d5e23886ca25d767fb63452082412c",
+  "pinned_hash": "1e17440ffb52c59087b22337fccfcfcdb8b4aa8f",
```
mokagio, line 4: I don't understand why after rebasing to resolve the merge conflicts that arose after the latest changes in #5708 the commit now does not carry the `.enc` file update!

Yet when I run `configure_apply`, the resulting `secrets.property` contains the `sentryTv` data as exepected.

AliSoftware: Because the `.enc` file updated in #5708 is the encrypted version of the file at commit `789d0d6`. When you update the `pinned_hash`, you still need to run `bundle exec fastlane run configure_update` to update the `.enc` file that will be used by CI with the encrypted version of the file found in your local mobile-secrets at this new `pinned_hash` commit (See PCYsg-iIC-p2g)

> Yet when I run `configure_apply`, the resulting `secrets.property` contains the `sentryTv` data as exepected.

This local test from `configure_apply` is not enough, because when you run `configure_apply` on your **local** machine (i.e. in a machine that has access to the `.mobile-secrets` repo), it will directly copy the clear file from your local `~/.mobile-secrets` checkout at that commit, without doing any encryption/decryption. **The `.enc` file does not take part in the process this situation of running this locally**.

It's only when you are **on CI**—or on a machine that does not have `.mobile-secrets` repo checked out, and thus it can't copy the file in clear directly—that **it uses the `.enc` file instead** (and goes thru the decryption of it).

If the `.enc` file is not modified in this PR, all that means is that, even if you get the expected `secrets.properties` when running `configure_apply` locally, the CI will still decrypt to the old one (because it'd decrypt it from the `.enc` file, not from the mobile-secrets repo directly). That's what calling `configure_update` fixes, as its role is to update the `.enc` file with the encrypted version of the new content.

This is one of the many ugly and asymetric behaviors of `configure` that made the tool so confusing to use (and which made it easy to have a secret work locally but not on CI). Can't wait for us to switch to `a8c-secrets` to not having to deal with this whole confusing mess anymore…

AliSoftware: @mokagio Ok I checked and confirmed that the `secrets.properties.enc` is already up-to-date after all, probably updated to the right commit back from your previous PR before you did the rebase.

TL;DR: It's all good after all, `.enc` file is OK and already contains the following:

```
### Sentry
sentryAndroidDsn=…
sentryAutomotiveDsn=…1911937
sentryWearDsn=…520
# Note: This value could be inlined in clear in the code.
#       Here for added flexibility given the work is in flux.
# See:
# - https://github.com/Automattic/pocket-casts-android/pull/5708
# - https://github.com/Automattic/pocket-casts-android/pull/5840
# - https://github.com/Automattic/pocket-casts-android/pull/5841
sentryTvProject=pocket-casts-tv
sentryTvDsn=…
```

<details><summary>The way I checked it</summary>

To simulate the CI environment (i.e. when the `~/.mobile-secrets` repo was not present locally, so that `configure_apply` would use the `.enc` file as a source instead of the local file from mobile-secrets directly), I temporarily renamed `mv ~/.mobile-secrets ~/.mobile-secrets-bak`, ran `configure_apply` again, which told me that it was missing the decryption key—confirmed that it was now trying to decrypt the `.enc` file instead of using the local clear file directly. Then I exported the decryption key env var (the one that CI has and uses to decrypt the `.enc`) and tried again—which worked and decrypted correctly this time—which allowed me to properly confirm the proper content of the `.enc` file once decrypted was the expected one.
</details>

@mokagio I think we should mark this  one as "Ready for Review", then merge this one + https://github.com/Automattic/pocket-casts-android/pull/5708 into `trunk`.

AliSoftware (approved):
I've double-checked and now confirmed that the `secrets.properties.enc` encrypted file contains the expected properties now (the ones from `1e17440f`)

Let's ship this ASAP.

PR #5856, opened 2026-09-04
[TV] Fix mirrored/clipped logged-out banner artwork and single-line subtitle

sztomek:
## Description

Fixes POC-886 https://linear.app/a8c/issue/POC-886/inverted-image-in-the-logged-out-banner

1. **Mirrored phone mockup** — the shipped `tv_banner_create_account` artwork is horizontally flipped (the on-screen text reads backwards). Replaced it with a corrected asset derived from the original (mirroring the phone region back to the correct orientation) instead of the earlier `graphicsLayer { scaleX = -1f }` hack. Flipping the whole image in code shifts the phone to the left of the banner, away from the spec position; baking the correction into the asset keeps the phone anchored at the right edge where the spec places it.
2. **Cut at the top** — the artwork was `174dp` tall inside a `153dp` banner, so the phone was clipped. Reduced the Create-Account artwork height to `150dp` so the whole phone fits (status bar fully visible).
3. **Subtitle** — the Create-Account subtitle now uses a single line (was up to 2). Widened the Create-Account text area (`contentWidthFraction` 0.68 → 0.78) so the full phrase fits on one line instead of truncating with an ellipsis.

### Verified on device
Signed out on a Google TV Streamer and confirmed the Create-Account banner:
- the phone reads correctly (not mirrored: clock, "Good Times", "24 episodes", "Smart Rules", "Search", "Play all"),
- the phone is not clipped at the top (status bar fully visible) or bottom,
- the phone sits at the right edge of the banner, matching the spec composition,
- the subtitle "Your follows and progress exactly where you left them" fits fully on one line (no ellipsis).

## Testing Instructions

1. Sign out and open the TV Home.
2. The "Create account" banner's phone mockup should read correctly (not mirrored), the phone should not be clipped at the top, and the subtitle should fit on one line.

## Screenshots or Screencast

On-device render of the banner after the fix (Google TV Streamer, signed out):

create-account-banner-fixed
Screenshot_20260904_091402

## Checklist

- [x] with different themes


geekygecko (approved):
Looks good. 

Slightly separate from this change, but the background gradient on Android TV doesn't match Apple TV. 

**Android TV**

Image

**Apple TV**

Image

sztomek:
@geekygecko thanks, i've made changes to it:
Screenshot_20260907_144524

PR #5860, opened 2026-09-04
Network Discovery: persist the podcast's network list id

geekygecko:
## Description

The app now remembers which network a podcast belongs to, so a later change can turn the author name on the podcast page into a link to that network.

The podcast cache endpoint recently started returning a `network_list` object for podcasts that belong to a network. This PR decodes it, stores the network's list id against the podcast, and makes sure the value survives a refresh. There is no UI here; making the author line tappable is [PCDROID-761](https://linear.app/a8c/issue/PCDROID-761/network-discovery-link-the-podcast-page-to-its-network).

Only `list_id` is persisted, not the `source` URL the server sends alongside it. `source` is just `{SERVER_LIST_URL}/{list_id}.json`, so storing it would bake the staging or production host into the database and go stale across build variants.

The part most likely to go quietly wrong is a server write blanking the new column, so every path that writes a podcast row was checked:

- **Subscribe** goes through `PodcastResponse.toPodcast()` and a full-entity insert, so parsing alone covers it.
- **Refresh** hits the *same* endpoint but persists through `PodcastDao.updateRefresh`, a hand-written partial-column `UPDATE`. That is the one that would have dropped the value, so the column is added there and to the change detection in `PodcastRefresherImpl`.
- **Sync** mutates a podcast loaded from the database and writes it back whole, so the column round-trips untouched.
- **`MissingEpisodesSync`** inserts placeholder rows with `onConflict = IGNORE`, so it cannot blank an existing row.
- **`UpdateEpisodeTask`** and `EpisodeManagerImpl.downloadMissingPodcastEpisode` call the same response mapper but only ever persist episodes.

Because subscribe and refresh share one endpoint, the response is treated as the source of truth: if a podcast leaves its network, the next refresh clears the column. That matches how `funding_url` and `explicit` already behave.

Database version goes 136 → 137 with a one-column `ALTER TABLE`, and the exported schema is committed.

Fixes [PCDROID-755](https://linear.app/a8c/issue/PCDROID-755/network-discovery-persist-the-podcasts-network-list-id)

Screenshot 2026-09-05 at 9 25 54 am

## Testing Instructions

To check it by hand against the server:

1. Install a build over an existing install so the 136 → 137 migration runs on a populated database, and confirm the app starts and your podcast list is intact.
2. Subscribe to a podcast that belongs to a network, for example Analog(ue) on Relay.
3. Inspect the database (App Inspection → `podcasts`) and confirm `network_list_id` holds the network's uuid for that row.
4. Open the podcast page again to trigger a refresh, then re-check the row: the value should be unchanged.
5. Subscribe to a podcast that is not on a network and confirm its `network_list_id` is `NULL`.

## Checklist

### DIFF
--- a/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/AppDatabaseTest.kt
+++ b/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/AppDatabaseTest.kt
@@ -28,6 +28,7 @@ class AppDatabaseTest {
         private const val MIGRATION_DB = "migration-test-132-133"
         private const val MIGRATION_DB_133_134 = "migration-test-133-134"
         private const val MIGRATION_DB_135_136 = "migration-test-135-136"
+        private const val MIGRATION_DB_136_137 = "migration-test-136-137"
     }
 
     @Rule @JvmField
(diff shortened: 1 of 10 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

sztomek (approved):

PR #5868, opened 2026-09-07
Add safeguards to "Cache entire playing episode" for oversized/video files

sztomek:
## Description

Adds safeguards to the "Cache entire playing episode" feature (Settings → Advanced → Playback, on by default) so a single mis-published feed enclosure can no longer trigger a huge, silent background download.

**Root cause.** When an episode is streamed, `ExoPlayerDataSourceFactory.shouldUseCache()` gates a background `CacheWorker` that pulls the **entire** file into the ExoPlayer cache so scrubbing is instant. The predicate had no video exclusion and no size cap, and `CacheWorker` ran a media3 `CacheWriter` with an open-ended `DataSpec` (no length) — so it pulled however many bytes the resource had. A podcast feed pointed at a mis-published **44.8 GiB** raw `.mp4` export (`48,089,985,249` bytes); streaming it made `CacheWorker` try to pull all 48 GB in the background. A free-tier user burned ~23 GB of mobile data in ~30 min → a €750 carrier bill (Zendesk 11558980). Note `warnOnMeteredNetwork` defaults to `false`, so the cache job's network constraint is `CONNECTED` (cellular allowed) by default.

iOS already excludes video podcasts outright and checks reported size vs. free space before starting its parallel download. This brings Android to parity and adds one extra, metadata-independent backstop.

### Three layers of defence

1. **Exclude video + oversized episodes (reported-size pre-check).** Extracted the caching decision into a pure, testable `shouldCacheEntireEpisode(...)` that additionally rejects `episode.isVideo` (iOS `!videoPodcast()`) and episodes whose feed-reported `sizeInBytes` exceeds the ExoPlayer LRU cache size. The LRU cache can only retain `maxCacheSizeBytes` (Firebase default 500 MB), so pulling a larger file downloads the whole thing but immediately evicts most of it — pure waste. Unknown size (`0`) is allowed (matches iOS's residual behaviour); `maxCacheSizeBytes <= 0` is treated as "no size cap" so a transient `0` remote-config value can't disable the feature globally.

2. **Keep the fingerprint/transcript path in sync.** `FingerprintTimingManager` had a structurally identical copy of the predicate deciding whether to reuse the player's on-disk cache (`sharedCacheKey` → `followPlayerCache`). Left untouched it would diverge — the player would stop caching video/oversized episodes while the transcript pipeline still waited up to `FOLLOW_TIMEOUT_MS` (60 s) per region for cache writes that never arrive. It now calls the same shared `shouldCacheEntireEpisode(...)`.

3. **Metadata-independent hard byte-bound in `CacheWorker`.** The two guards above trust publisher metadata (`fileType`, `sizeInBytes`). A non-video enclosure that reports `length="0"` (or lies) but is actually enormous would still slip through. So `CacheWorker` now attaches a `CacheWriter.ProgressListener` that cancels the writer once the bytes pulled reach the cache size, landing in the existing catch as a `Result.failure()`. This bounds background network egress to ~500 MB regardless of what the feed claims. It is safe: `Result.failure()` never retries (no backoff is configured), and the fully-cached callback (`onCachingComplete`, which marks the episode 100 % buffered) only fires on `SUCCEEDED`, so a capped run never falsely marks the episode buffered. The request stays open-ended (`bytes=0-`), so there is no risk of a `416` from a bounded range on non-compliant CDNs — which is why this uses listener-cancel rather than `DataSpec.setLength`.

   **Scope of the byte-bound:** it caps egress **per worker run**. `CacheWriter` seeds its counter with bytes already present for the key, so re-runs of an already-capped episode are normally no-ops. The only residual is a slow "top-up drip": if other cache writes (1 MB next-episode prefetch, the fingerprint reader) evict part of a capped episode, a later player-prepare / `resetEpisodeCaching` re-enqueue pulls just the evicted delta back up to the cap. It is bounded (never a fresh full pull) and requires actively re-streaming a pathological enclosure while other content churns the shared cache — for the reported 48 GB file this is still a ~100× reduction. A persistent "already capped" marker to suppress the drip was considered and deliberately left out as disproportionate for this non-blocking edge case.

### Notes / trade-offs
- **Threshold = cache size** is stricter than iOS's free-space check: an episode only *modestly* larger than the cache (e.g. a 3 h+ high-bitrate show) loses the instant-scrub cache entirely rather than being partially cached. This is a conscious, documented choice — the cache can't hold the whole file anyway, and the byte-bound in layer 3 keeps the *unknown-size* case graceful.
- **Intentional "no cap" asymmetry:** in the predicate, `maxCacheSizeBytes <= 0` means "no size pre-check" (permissive, so we never disable the feature globally); in `CacheWorker`, a `<= 0` value falls back to a 500 MB hard cap. The worker must *always* have a bound — that's the whole point of layer 3.
- **Left out on purpose:** `PrefetchWorker` (the next-episode prefetch) is already hard-capped to 1 MB via `DataSpec.setLength`, so it can't reproduce this. No `UserEpisode` exclusion was added — the video + size guards already cover uploaded cloud files.

Fixes PCDROID-718

## Testing Instructions

This is a background/data-usage change with no UI surface, so there are no emulator screenshots.

1. Unit tests: `./gradlew :modules:services:repositories:testDebugUnitTest` — new `ShouldCacheEntireEpisodeTest` covers the guard truth table (video, downloaded/downloading, HLS, setting off, size > cap, size == cap boundary, unknown size, no-cap path) and `CacheWorkerRequestTest` covers the max-cache-bytes input-data plumbing.
2. Manual sanity (optional): with "Cache entire playing episode" on, stream a normal audio episode and confirm scrubbing is still instant (caching still runs). Stream a video podcast and confirm it plays via plain streaming with no background cache job.

The `CacheWorker` cap/cancel path itself isn't unit-tested — it needs media3/Robolectric infra, and this repo avoids player-state-machine tests. Its correctness (media3 1.10.1 `CacheWriter.cancel()` from within the progress listener terminates `cache()`) was verified by reading the media3 source and via an independent review.

## Screenshots or Screencast

N/A — no UI changes.

## Checklist
- [x] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml` (no new strings)
- [x] Any jetpack compose components I added or changed are covered by compose previews (none)
- [x] I have updated (or requested that someone edit) the Event Horizon schema to reflect any new or changed analytics (no analytics changes)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
         ([#5843](https://github.com/Automattic/pocket-casts-android/pull/5843))
     *   Closing the transcript search bar now keeps the transcript open at the same scroll position instead of exiting the transcript
         ([#5835](https://github.com/Automattic/pocket-casts-android/pull/5835))
+    *   Stop "Cache entire playing episode" from silently downloading oversized or video enclosures in the background
+        ([#5868](https://github.com/Automattic/pocket-casts-android/pull/5868))
 
 8.20
 -----
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/fingerprint/FingerprintTimingManager.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/fingerprint/FingerprintTimingManager.kt
@@ -14,6 +14,7 @@ import au.com.shiftyjelly.pocketcasts.preferences.Settings
 import au.com.shiftyjelly.pocketcasts.repositories.BuildConfig
 import au.com.shiftyjelly.pocketcasts.repositories.playback.ExoPlayerDataSourceFactory
 import au.com.shiftyjelly.pocketcasts.repositories.playback.PlaybackManager
+import au.com.shiftyjelly.pocketcasts.repositories.playback.shouldCacheEntireEpisode
 import au.com.shiftyjelly.pocketcasts.repositories.podcast.ChapterManager
 import au.com.shiftyjelly.pocketcasts.utils.AppPlatform
 import au.com.shiftyjelly.pocketcasts.utils.Network
@@ -222,7 +223,12 @@ class FingerprintTimingManager @Inject constructor(
         val audioSource = episode.downloadedFilePath ?: episode.downloadUrl
         // Reuse the player's on-disk cache (same UUID key) instead of a second download when it applies.
         val sharedCacheKey = episodeUuid.takeIf {
-            !episode.isDownloaded && !episode.isDownloading && !episode.isStreamUrlHls && settings.cacheEntirePlayingEpisode.value
+            shouldCacheEntireEpisode(
+                episode = episode,
+                isHlsStream = episode.isStreamUrlHls,
+                cacheEntirePlayingEpisodeEnabled = settings.cacheEntirePlayingEpisode.value,
+                maxCacheSizeBytes = settings.getExoPlayerCacheEntirePlayingEpisodeSizeInMB() * 1024 * 1024L,
+            )
         }
 
         scope.launch {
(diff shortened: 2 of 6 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

geekygecko (approved):
This seems like a great change to make!

PR #5869, opened 2026-09-07
Fail truncated media responses instead of completing the episode

sztomek:
## Description

Fixes PCDROID-714 https://linear.app/a8c/issue/PCDROID-714

Some hosts truncate an audio response mid-transfer and end it with a clean **HTTP/2 `END_STREAM`** while still advertising the full `Content-Length` (e.g. Rádio Câmara / *Feijoada Completa* sends `content-length: 231288061` over h2 then cuts the body at ~150 MB; Libsyn DAI-stitched *Lore*). Today the app treats that short body as a legitimate end of stream, so the episode is **marked played, its download is deleted, and Up Next advances — with no error shown**. The same episodes play fine on desktop/web and iOS.

### Root cause

The app never checks that the bytes it received match the server-declared `Content-Length`:

- **Media3 1.10.1** `OkHttpDataSource.readInternal` (and `DefaultHttpDataSource`) return `C.RESULT_END_OF_INPUT` the moment the underlying stream returns `-1`, with no `bytesRead == bytesToRead` check.
- **OkHttp 5.5.0** is asymmetric: `Http1ExchangeCodec.FixedLengthSource` throws `ProtocolException("unexpected end of stream")` on a short **HTTP/1.1** body, but `Http2Stream.FramingSource` returns `-1` on a clean `END_STREAM` short body with no length reconciliation — so truncation is loud on h1 (already handled) and **silent on h2**.

Downstream, a silent end-of-stream becomes `Player.STATE_ENDED` → `PlaybackManager.onCompletion` (mark played, remove from Up Next, auto-archive → delete download). The `EpisodeDownloader` path has the same gap: a single GET + `readAll`, guarded only against files < 150 KB.

### The fix

A new **network** interceptor (`ContentLengthValidatorInterceptor`) is added to the `@Player` and `@Downloads` OkHttp clients — the only two that transfer media bytes. It wraps the response body in an okio `ForwardingSource` and, when the stream reaches EOF (`read() == -1`) while the body is **grossly short** of the declared `Content-Length`, throws `ProtocolException`. This mirrors OkHttp's own HTTP/1.1 behavior and extends it to HTTP/2.

The thrown `IOException` flows through existing, tested error handling instead of a silent completion:
- **Streaming**: Media3 wraps it as `HttpDataSourceException`; ExoPlayer surfaces `onPlayerError` (retryable), never `STATE_ENDED`, so nothing is marked played.
- **Download**: `EpisodeDownloader` → `Result.ExceptionFailure` → retryable, capped at the existing max attempts.
- **Cache prefetch**: `CacheWorker` → `Result.failure()`; the "caching complete" buffer-to-end callback only runs on success, so no false completion.

**Why "grossly short" and not an exact match:** a functional server that *over-declares* its `Content-Length` but sends a complete, playable file would be indistinguishable from truncation on h2. To avoid ever breaking such a file (which would otherwise become permanently undownloadable), the body is only treated as truncated when it is short by **both** more than 512 KB **and** more than 10 % of the declared length. Real truncations here miss 34–95 %; benign length mismatches miss kilobytes.

Gating (decided once per response, never in the read hot path):
- feature flag on, `GET`, response `200`/`206`, and a positive `Content-Length`.
- excludes `HEAD` (used by `UpdateEpisodeDetailsTask` on the `@Downloads` client), `204`/`304`, redirects, and chunked/unknown-length responses.

It sits **below** `BridgeInterceptor`, so gzip'd responses are validated on the compressed stream (consistent length), and it only throws on real EOF — never on an early `close()` (seek / buffer-full / bounded Range reads), so seeking and partial reads are unaffected.

Behind a **`Feature.VALIDATE_CONTENT_LENGTH`** flag (default on, `hasFirebaseRemoteFlag = true`) as a remote kill switch.

### Scope / follow-ups (not in this PR)

- Covers **byte-truncation** (Rádio Câmara, confirmed via curl; the HTTP/2 short-body *Lore* variant with "no exception logged" in the reporter's log).
- A hypothetical "all declared bytes delivered but the MP3 decodes to ~82 s" variant would need an `onCompletion` low-completion guard — deferred, since a naive guard risks blocking legitimate completions on feeds with over-declared durations.
- Pinning the resolved DAI build URL across a session (same length, different content per request) is a separate, larger change — deferred.

**Please do not auto-close PCDROID-714** on merge; this covers the truncation cases above.

## Testing Instructions

This is a network-layer change with no UI. Behaviour is covered by unit tests:

1. `./gradlew :modules:services:servers:testDebugUnitTest --tests "*ContentLengthValidatorInterceptorTest"` — 11 cases: complete body passes; gross truncation throws (200 and 206); shortfall within the absolute (512 KB) or relative (≥ 90 % delivered) tolerance passes; unknown length / HEAD / non-2xx / disabled-flag not validated; early close before EOF does not throw.
2. Manual sanity (optional): stream and fully play a normal episode, and download one, on a real device — both complete normally (healthy servers deliver the full `Content-Length`, so nothing changes for them). A full device repro of the server-side truncation is impractical (geo/campaign-dependent), which is why the behaviour is unit-tested and verified against the Media3 / OkHttp sources.

## Checklist

- [ ] If a user-facing change, I added an entry in CHANGELOG.md
- [x] I considered whether it makes sense to add tests for my changes
- [x] All strings are already localized (no new strings)
- [x] No jetpack compose components added or changed
- [x] No new or changed analytics (Event Horizon schema unchanged)

### DIFF
--- a/modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/di/InterceptorModule.kt
+++ b/modules/services/servers/src/main/java/au/com/shiftyjelly/pocketcasts/servers/di/InterceptorModule.kt
@@ -7,6 +7,7 @@ import au.com.shiftyjelly.pocketcasts.servers.BuildConfig
 import au.com.shiftyjelly.pocketcasts.servers.CleanAndRetryInterceptor
 import au.com.shiftyjelly.pocketcasts.servers.OkHttpInterceptor
 import au.com.shiftyjelly.pocketcasts.servers.interceptors.BasicAuthInterceptor
+import au.com.shiftyjelly.pocketcasts.servers.interceptors.ContentLengthValidatorInterceptor
 import au.com.shiftyjelly.pocketcasts.servers.interceptors.InternationalizationInterceptor
 import au.com.shiftyjelly.pocketcasts.servers.sync.TokenHandler
 import au.com.shiftyjelly.pocketcasts.servers.sync.exception.RefreshTokenExpiredException
@@ -96,6 +97,8 @@ object InterceptorModule {
     ).toClientInterceptor() // Must be client interceptor. Otherwise calls cannot be retried.
 
     private val basicAuthInterceptor = BasicAuthInterceptor().toClientInterceptor()
+
+    private val contentLengthValidatorInterceptor = ContentLengthValidatorInterceptor().toNetworkInterceptor()
 
     @Provides
     @TokenInterceptor
@@ -231,6 +234,7 @@ object InterceptorModule {
             add(crashLoggingInterceptor.toClientInterceptor())
             add(basicAuthInterceptor)
             add(cleanAndRetryInterceptor)
+            add(contentLengthValidatorInterceptor)
 
             if (BuildConfig.DEBUG) {
                 val loggingInterceptor = HttpLoggingInterceptor().apply {
@@ -274,6 +278,7 @@ object InterceptorModule {
             add(crashLoggingInterceptor.toClientInterceptor())
             add(basicAuthInterceptor)
             add(cleanAndRetryInterceptor)
+            add(contentLengthValidatorInterceptor)
 
             if (BuildConfig.DEBUG) {
                 val loggingInterceptor = HttpLoggingInterceptor().apply {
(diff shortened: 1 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

PR #5872, opened 2026-09-08
[DO NOT MERGE] Validate Buildkite interpolation of env-var guards for #5831

mokagio:
**Validation only — do not merge.** Stacked on `ainfra-3066-pocket-casts-android-post-rollout-slack-announcements-from` (#5831). It will be closed once CI has answered.

### The question

[#5831 (comment)](https://github.com/Automattic/pocket-casts-android/pull/5831/files#r3949532121) suggests guarding the release env vars with `${RELEASE_VERSION:?Missing value}` in `update-rollouts.yml`, on the reasoning that bash will print the variable name. But a `command:` node is interpolated by `buildkite-agent pipeline upload` before bash ever sees it — the pipeline's upload step runs `buildkite-agent pipeline upload .buildkite/${PIPELINE:-pipeline.yml}` with interpolation on, which is also why `plugins: [$CI_TOOLKIT]` resolves at all.

So: does `${VAR:?msg}` reach bash, and if not, what does Buildkite's interpolator do with it?

### Why nothing else answers it

`update-rollouts.yml` only runs when ReleasesV2 triggers it with `PIPELINE` and the four env vars set, so no PR build exercises that upload path. Production evidence points one way — [build 13960](https://buildkite.com/automattic/pocket-casts-android/builds/13960)'s uploaded job command is `update_rollouts track:"production" percent:"0.10"`, from a YAML that reads `track:"$TRACK" percent:"$ROLLOUT_PERCENT"` — but that only shows plain `$VAR` being interpolated, not what the `:?` form does.

### What to read

One job, **expected to pass**: [:test_tube: Validate Buildkite interpolation](../../actions). It runs on the `upload` queue, the agent that actually interpolates the release pipelines. Every other step is deliberately stripped from `pipeline.yml` so this takes minutes, not a full Android build.

Each assertion is a `pipeline upload --dry-run` of a one-step probe pipeline:

| Form | Asserts |
|---|---|
| `${VAR:?msg}`, VAR **set** | upload **fails** — the interpolator reads `:` as a substring offset |
| `${VAR:?msg}`, VAR unset | upload **fails** the same way, so the form is simply invalid |
| `${VAR?msg}`, VAR set | upload succeeds, value substituted |
| `${VAR?msg}`, VAR unset | upload **fails** with `$VAR: msg` — a working guard, at upload time |
| `${VAR?msg}`, VAR **empty** | upload **succeeds** — the guard does *not* catch empty |
| `$${VAR:?msg}` | passes through literally, deferred to the job shell |
| `$VAR`, unset | silently empty — today's behaviour |

The step then evaluates `$${VAR:?msg}` for real, proving the escaped form survives the upload of *this* pipeline and fires in the job shell for both unset and empty.

The assertions can go red: flipping one expected pattern locally turns two into `FAIL` and exits 1.

### What this still cannot answer

Whether ReleasesV2 can pass an *empty* `RELEASE_VERSION` rather than omitting it. That decides between the two working forms, and only the ReleasesV2 scenario config can say.

---

## Answer — [build 18906](https://buildkite.com/automattic/pocket-casts-android/builds/18906), passed in 2s on `buildkite-agent 3.128.0` (`upload` queue)

All ten assertions ran. Verbatim:

```
--- :mag: Upload-time interpolation of a command node
PASS  colon form ${VAR:?msg}, VAR set: buildkite-agent: fatal: pipeline interpolation of "colon-set.yml" failed: interpolating command: Unable to parse offset: strconv.Atoi: parsing "?Missing value": invalid syntax
PASS  colon form ${VAR:?msg}, VAR unset: buildkite-agent: fatal: pipeline interpolation of "colon-unset.yml" failed: interpolating command: Unable to parse offset: strconv.Atoi: parsing "?Missing value": invalid syntax
PASS  bare form ${VAR?msg}, VAR set: "command": "echo probe:\"beta\"\n",
PASS  bare form ${VAR?msg}, VAR unset: buildkite-agent: fatal: pipeline interpolation of "bare-unset.yml" failed: interpolating command: $PROBE_UNSET: Missing value
PASS  bare form ${VAR?msg}, VAR empty -- guard does NOT fire: "command": "echo probe:\"\"\n",
PASS  escaped form $${VAR:?msg} -- deferred to the job shell: "command": "echo probe:\"${PROBE_UNSET:?Missing value}\"\n",
PASS  plain $VAR, unset -- silently empty (todays behaviour): "command": "echo probe:\"\"\n",
--- :mag: Escaped guard, evaluated by the job shell
VAR set:   [beta]
VAR unset: guard tripped -> /bin/bash: line 10: PROBE_UNSET: Missing value
VAR empty: guard tripped -> /bin/bash: line 16: PROBE_EMPTY: Missing value
```

So:

- `${VAR:?msg}` **breaks the upload unconditionally**. Buildkite reads `:` as the substring-offset operator; there is no `:?` in its grammar.
- `${VAR?msg}` is Buildkite's required-variable form and guards at **upload** time — but only against *unset*, not empty.
- `$${VAR:?msg}` reaches the job shell intact and catches unset *and* empty. The job log confirms it end to end: the step's own command block shows `${PROBE_UNSET:?Missing value}` after the real upload interpolated `$$` to `$`, and `/bin/bash` then tripped on it.

Still open: whether ReleasesV2 can pass an empty `RELEASE_VERSION` rather than omitting it. That is what decides between the two working forms.


mokagio:
Closing — it answered its question. Results are in the description; build: https://buildkite.com/automattic/pocket-casts-android/builds/18906

PR #5875, opened 2026-09-08
Refresh a podcast's private flag on refresh so sharing follows the server

sztomek:
## Description

`Podcast.isPrivate` was only written when a podcast row is first created from the server
(`PodcastInfo.toPodcast()` in `PodcastResponse.kt`, i.e. at follow time). The refresh path
`PodcastRefresherImpl.updatePodcastIfRequired()` compared and rewrote title, author, category,
description, next-episode estimate, frequency, refresh-availability, funding URL, explicit, web-feed
and network-list id — but never `isPrivate`. Because `Podcast.canShare = !isPrivate`, a podcast that
was private when the user followed it kept blocking sharing on that device forever, even after the
listing had been public server-side for months. The only workaround was to unfollow and re-follow
(a fresh `toPodcast()`), which loses per-episode progress. iOS already writes `podcast.isPrivate`
from `is_private` on every update (`ServerPodcastManager+Update.swift`).

This adds `isPrivate` to both the change-detection comparison and `PodcastDao.updateRefresh`, so the
flag tracks the server value in both directions (private → public and public → private), matching iOS.

No database migration is needed: the `is_private` column already exists (`MIGRATION_105_106`);
`updateRefresh` is a plain `@Query` UPDATE, so adding a column to its SET clause does not change the
schema.

**On null handling (deliberate):** `is_private` is deserialized as `Boolean?` and coerced to `false`
in `toPodcast()`, and the entity column is non-null, so "unknown" cannot be represented. I chose the
iOS-parity unconditional write rather than an "only overwrite when present" guard. The risk of a
null/absent `is_private` wrongly clearing a genuinely-private flag is bounded: follow and refresh hit
the same podcast-cache endpoint through the same `null → false` mapper, a non-2xx response yields a
null body so `toPodcast()` never runs, and `canShare` only gates the client share affordance — real
privacy is enforced server-side.

Fixes PCDROID-766 https://linear.app/a8c/issue/PCDROID-766/android-is-private-is-only-written-at-follow-time-and-never-refreshed

## Testing Instructions

1. Follow a podcast whose nodeweb listing is a private feed type (or set one to "manual private").
2. In nodeweb, set the listing back to public and Update Cache / force upload to S3.
3. On the device, open the podcast, use the three-dot menu → Refresh Episode List (and/or let a
   background refresh run).
4. Tap the share icon on the podcast page.

Expected: after the refresh the app picks up `is_private: false` and the share sheet opens (previously
it showed the toast "Sharing is not available for private podcasts" until unfollow/re-follow).

Automated coverage: `PodcastRefresherImplTest` gains a private → public case (the exact ticket repro)
and a public → private case. Run:
`./gradlew :modules:services:repositories:testDebugUnitTest --tests "*PodcastRefresherImplTest"`.

## Checklist

- [x] If user-facing change, I added entry in CHANGELOG.md
- [x] Ensure linter passes (`./gradlew spotlessApply` was run)
- [x] I considered whether it makes sense to add tests to my changes
- [ ] All strings that need to be localized are in `modules/services/localization/.../strings.xml` — no new strings
- [ ] Any jetpack compose components I added or changed are covered by compose previews — no Compose changes
- [ ] I updated the Event Horizon schema — no analytics changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
         ([#5835](https://github.com/Automattic/pocket-casts-android/pull/5835))
     *   Stop "Cache entire playing episode" from silently downloading oversized or video enclosures in the background
         ([#5868](https://github.com/Automattic/pocket-casts-android/pull/5868))
+    *   Refresh a podcast's private flag so sharing becomes available again once a podcast is made public
+        ([#5875](https://github.com/Automattic/pocket-casts-android/pull/5875))
 
 8.20
 -----
--- a/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/PodcastDaoTest.kt
+++ b/app/src/androidTest/java/au/com/shiftyjelly/pocketcasts/models/db/PodcastDaoTest.kt
@@ -230,5 +230,6 @@ class PodcastDaoTest {
         explicit = null,
         webFeed = false,
         networkListId = networkListId,
+        isPrivate = false,
     )
 }
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/PodcastDao.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/PodcastDao.kt
@@ -292,8 +292,8 @@ abstract class PodcastDao {
     @Update
     abstract suspend fun updateSuspend(podcast: Podcast)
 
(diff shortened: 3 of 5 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

PR #5876, opened 2026-09-08
Make Wear OS login failures diagnosable (PCDROID-763)

sztomek:
## Description

Fixes PCDROID-763 https://linear.app/a8c/issue/PCDROID-763/wear-os-instant-loginfailed-on-both-login-routes-galaxy-watch-ultra-2

**This is a diagnostics-only change. It does not — and cannot yet — fix the reported failure.** The root cause is unconfirmable from the outside because the failure reason is swallowed at every layer, and this reporter can't send us watch logs (in-app export is gated behind the signed-in state they can't reach, PCDROID-476). So instead of shipping a blind behavioral change to a login path every Wear user depends on, this PR makes the *next* occurrence legible — via readable logs and a disambiguated on-screen error — so we can actually pin the cause.

What was invisible before this PR:

1. **The identity in phone→watch logs was R8-obscured.** `LoginIdentity` is a `sealed class` of `object`s with no `toString`, so `WatchSync` logged `identity: tq6@6084f14` in release. Now it logs `loginIdentity.key` (`Google` / `PocketCasts` / `QrCode`).
2. **The actual login exception was dropped on the floor.** `SyncManagerImpl.handleLogin` caught the exception, sent it to `Timber.e(…)`, then replaced it with a generic string. But on a **release watch Timber is never planted** (and `TimberDebugTree` doesn't forward to `LogBuffer` anyway) — so the real exception (HTTP status, `IOException`, `SecurityException`, TLS, …) was lost even from the exportable log. It now goes to `LogBuffer.e(…, ex, …)`, recovering the signal across all release platforms.
3. **The phone-sync failure reason was unlogged.** `WatchSync.processAuthDataChange` logged only `"Login result: Failed"` (itself R8-degraded). It now logs the server `message` + `messageId`.
4. **The Google-route toast was ambiguous.** `State.Failed.GoogleLoginUnavailable` (pre-picker: Play services / no credential) and `State.Failed.Other` (post-picker: server rejected the token) both showed the same "Unable to launch Google sign in." A distinct toast now covers the server-side case, so the reporter's screenshot alone tells us client-side vs server-side.

No login success/failure behavior changes. The only user-visible delta is a more accurate toast on the post-picker Google case; the dominant pre-picker path keeps its exact current string.

## Testing Instructions

This is not reproducible in-house (no Galaxy Watch Ultra 2). Verification is of the diagnostics themselves:

1. Unit tests (green): `:modules:features:account` (36), `:modules:services:repositories` (963), `:wear` (61) — 1060 total, 0 failures. `WatchSyncTest` still asserts the `Failed` result reaches the callback (the `onResult` invariant is preserved).
2. Manual (optional, closest to a real repro): boot a **Wear OS API 36 emulator** (> the API-34 gate) and tap **Log in with Google** with no usable Google credential — this exercises the modern Credential Manager path the Ultra 2 takes. Observe which toast appears: the unchanged "Unable to launch Google sign in." (`GoogleLoginUnavailable`, pre-picker) vs. the new "Couldn't sign in with Google. Please try again." (`Other`, server-side).
3. To confirm the R8 log fix survives obfuscation, build a `prototype` wear APK, trigger a phone→watch sync, and grep the on-device log for `identity: Google` (not `identity: tq6@…`).

## Screenshots or Screencast

N/A — no on-device UI layout changes (a toast string differs on one Google error path). No repro hardware available.

## Checklist

### DIFF
--- a/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/watchsync/WatchSync.kt
+++ b/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/watchsync/WatchSync.kt
@@ -90,7 +90,10 @@ constructor(
                     loginIdentity = data.loginIdentity,
                     signInSource = SignInSource.WatchPhoneSync,
                 )
-                LogBuffer.i(TAG, "Login result: ${result::class.simpleName}")
+                when (result) {
+                    is LoginResult.Success -> LogBuffer.i(TAG, "Login from phone succeeded")
+                    is LoginResult.Failed -> LogBuffer.e(TAG, "Login from phone failed: ${result.message} (id: ${result.messageId})")
+                }
                 onResult(result)
             } else {
                 LogBuffer.i(TAG, "Already logged in, skipping login")
--- a/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/watchsync/WatchSync.kt
+++ b/modules/features/account/src/main/java/au/com/shiftyjelly/pocketcasts/account/watchsync/WatchSync.kt
@@ -51,7 +51,7 @@ constructor(
                 if (watchSyncAuthData == null) {
                     LogBuffer.i(TAG, "Removing auth token from Wear: Phone not logged in to Pocket Casts")
                 } else {
-                    LogBuffer.i(TAG, "Sending auth token to Wear Data Layer (identity: ${watchSyncAuthData.loginIdentity})")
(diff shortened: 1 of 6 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

PR #5877, opened 2026-09-08
Fix the Playback Effects icon disappearing when a podcast has no colours yet

### LINKED ISSUE #5529: Playback Effects icon missing [beta 8.15-rc-2 (9437)]
### Description

Playback Effects icon missing from Now Playing screen 

Go to Now Playing screen

Missing from screen:
Image

Edit:
Image

Still missing:
Image

Pixel 8, android 16, beta 8.15-rc-2 (9437)

sztomek:
## Description

On the Now Playing screen, the Playback Effects icon was sometimes missing from the actions shelf. Its slot was still there — it just showed nothing, while all the other icons were fine.

The effects icon is drawn using the podcast's accent colour (the same colour the player uses for its highlights). When playback effects are turned on, the icon is tinted with that colour. The catch is that a podcast's colours are filled in later — after they're fetched from the server or worked out from the artwork. Until then the colour value is `0`, which is fully transparent, so the icon was being painted with an invisible colour and vanished. Every other shelf icon uses a fixed theme colour, so they stayed visible. This is also why it came and went on its own: once the podcast's colours were populated, the icon reappeared.

The rest of the app already handles this. `getTintColor()` falls back to a visible default colour when a podcast has no colours yet, but the player was using `getPlayerTintColor()`, which returned the raw value with no fallback.

The fix is to build the player tint from the same guarded `getTintColor()`. For podcasts that already have colours nothing changes; for podcasts without colours yet the accent now falls back to a visible colour instead of being transparent. The now-unused `getPlayerTintColor()` is removed.

Fixes #5529
Fixes PCDROID-652 https://linear.app/a8c/issue/PCDROID-652/playback-effects-icon-missing-beta-815-rc-2-9437

## Testing Instructions

This is hard to trigger by hand because it only happens in the short window before a podcast's colours have been fetched, with playback effects enabled. The behaviour is covered by a unit test:

1. `./gradlew :modules:services:compose:testDebugUnitTest`
2. `PodcastColorsTest` checks that a podcast with no extracted colours produces a visible (opaque) player tint, and that a podcast with a real colour keeps it.

## Checklist

- [ ] with different themes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
         ([#5835](https://github.com/Automattic/pocket-casts-android/pull/5835))
     *   Stop "Cache entire playing episode" from silently downloading oversized or video enclosures in the background
         ([#5868](https://github.com/Automattic/pocket-casts-android/pull/5868))
+    *   Keep the Playback Effects icon visible on the Now Playing screen when a podcast's colours haven't loaded yet
+        ([#5877](https://github.com/Automattic/pocket-casts-android/pull/5877))
 
 8.20
 -----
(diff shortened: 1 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

geekygecko (approved):
Great fix!

PR #5879, opened 2026-09-08
Fix played/archived status not syncing to episodes missing from the local library

sztomek:
## Description

Incremental sync silently dropped server-side episode changes — played status, play position, archived, and starred — for any episode that was **not already in the local database**, and then never received them again. The sync server sends each change only once (an incremental delta by `modified_at`), so once a device's `lastModified` advanced past the change it was gone. The result: episodes stayed unplayed / stuck in **New Releases** on one device even after being completed on another, and only a full re-sync (log out / log back in) recovered them.

`EpisodeSync.processIncrementalResponse` applied server episode records only to episodes returned by `episodeManager.findByUuids(...)` and dropped the rest. This change makes it fetch the missing ones — for episodes whose podcast is subscribed locally — via the existing `/user/podcast/episodes` endpoint (the same call the manual-playlist `MissingEpisodesSync` uses, which returns the authoritative per-user state), insert them, and include them in the normal apply pass.

This mirrors iOS, whose `SyncTask+ServerChanges.importEpisode` already fetches missing episodes (`addMissingEpisode`) before applying sync changes — which is why the bug was Android-only. The behaviour has existed since the original JSON sync (the Protobuf rewrite preserved it), so this is a long-standing fix rather than a regression.

Fixes PCDROID-521 https://linear.app/a8c/issue/PCDROID-521/android-last-played-episode-doesnt-sync-completed-status-to-android

## Testing Instructions

The end-to-end path needs two synced clients, so the core logic is covered by new unit tests (`EpisodeSyncTest`). To verify manually:

1. On device A (or iOS / web), complete an episode of a subscribed podcast that device B does **not** yet have in its local library.
2. On device B (Android, this build), refresh / reopen the app to trigger an incremental sync.
3. The episode now appears as **played** (and archived, if it was archived on device A) instead of remaining unplayed in New Releases — without needing to log out and back in.

## Checklist

### DIFF
--- a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/data/EpisodeSync.kt
+++ b/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/data/EpisodeSync.kt
@@ -7,6 +7,8 @@ import au.com.shiftyjelly.pocketcasts.preferences.Settings
 import au.com.shiftyjelly.pocketcasts.repositories.playback.PlaybackManager
 import au.com.shiftyjelly.pocketcasts.repositories.podcast.EpisodeManager
 import au.com.shiftyjelly.pocketcasts.repositories.podcast.PodcastManager
+import au.com.shiftyjelly.pocketcasts.repositories.sync.SyncManager
+import au.com.shiftyjelly.pocketcasts.utils.log.LogBuffer
 import com.google.protobuf.boolValue
 import com.google.protobuf.int32Value
 import com.google.protobuf.int64Value
@@ -16,6 +18,7 @@ import com.pocketcasts.service.api.durationOrNull
 import com.pocketcasts.service.api.isDeletedOrNull
 import com.pocketcasts.service.api.playedUpToOrNull
 import com.pocketcasts.service.api.playingStatusOrNull
+import com.pocketcasts.service.api.podcastsEpisodesRequest
 import com.pocketcasts.service.api.record
 import com.pocketcasts.service.api.starredModifiedOrNull
 import com.pocketcasts.service.api.starredOrNull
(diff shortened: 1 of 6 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/sync/data/EpisodeSync.kt
```diff
@@ -91,6 +97,39 @@ internal class EpisodeSync(
         episodeManager.updateAllSyncFields(localEpisodes)
     }
 
+    private suspend fun fetchMissingEpisodes(
+        serverEpisodesMap: Map<String, SyncUserEpisode>,
+        presentUuids: Set<String>,
+    ): List<PodcastEpisode> {
```
geekygecko, line 103: I’m not sure about this approach. There’s a possible case where the episode no longer exists in the podcast database and we only have the user’s sync data for it. In that situation, I don’t think we’d want to add the episode back into the local database.

I’m wondering if, when we detect missing episodes, we should first refresh the podcast JSON if required to make sure we have the latest data, similar to what we already do when the user visits the podcast page.

CookieyedCodes: Hmmmm Phil users do need something for orphan episodes & a way to relink episodes that they have as the uuid can at least linked with upnext at minimum, it will help with cross device transfer & help with archival, i hate lost media & podcasts are very prone for it maximising it, I'm not gonna comment on is this the best way of doing it as you guys are the experts but PC dose to be better on this front.

Community discussed it here a few weeks ago but I've seen it as a regular complaint over the past few years on other platforms too
https://pocketcastsbeta.slack.com/archives/C0GKY84CR/p1787067058037999?thread_ts=1787067058.037999&cid=C0GKY84CR

PR #5893, opened 2026-09-09
Fix prototype download links and build details

iangmaia:
Related to AINFRA-2488

## Description

Fixes prototype build comments so Wear, Automotive, and TV testers get the correct download links and build details.

Wear, Automotive, and TV prototype comments were showing the mobile app's QR code and build details. Each now shows its own download link, version, and application ID.

Manually requested prototype builds also run when only automation files change.

## Testing Instructions

1. Run the optional **Prototype Builds** job.
2. Check each comment's QR code opens the correct app and its build details match the uploaded artifact. Mobile should still use Firebase App Distribution.

Verified in the [CI prototype run](https://buildkite.com/automattic/pocket-casts-android/builds/18956/list?sid=01a08724-1977-45d8-b231-601631847956&tab=output).

Resulting comment:
- [Before](https://github.com/Automattic/pocket-casts-android/pull/5682#issuecomment-5588367126) 
- [After](https://github.com/Automattic/pocket-casts-android/pull/5893#issuecomment-5605808490)

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

fastlane/Fastfile
```diff
@@ -693,6 +696,9 @@ platform :android do
     # Build wear/automotive/TV and upload to S3 (PRs only)
     next unless pr_number
 
+    # The comment action otherwise reuses the mobile build's Firebase URL and metadata for S3 builds.
+    lane_context.delete(SharedValues::FIREBASE_APP_DISTRO_RELEASE)
```
AliSoftware, line 700: Ooof, nice catch, that was subtle!

fastlane/Fastfile
```diff
@@ -757,6 +763,19 @@ platform :android do
     { app: '📱 Mobile', wear: '⌚ Wear', automotive: '🚗 Automotive', tv: '📺 TV' }.fetch(key, '❔ Unknown')
   end
 
+  def apk_metadata_for_comment(apk_path:)
+    metadata_path = File.join(File.dirname(apk_path), 'output-metadata.json')
```
AliSoftware, line 767: It took me a good minute searching through this `Fastfile` and any other part of our source code and gradle config/properties files to find where/what was the code generating this `output-metadata.json` file.

Turns out that this seems to be was generated by Gradle itself by default automatically. TIL.
Gradle seems to generate those in the `{module}/build/outputs/apk/debugProd/` folder, next to the `.apk`, hence the `File.dirname(apk_path)` to construct the path to that `.json` (and the fact that they are automatically already uploaded as Buildkite artifacts, given our `artifact_paths: ["**/build/outputs/apk/**/*"]` in the `pipeline.yml` for this job)

fastlane/Fastfile
AliSoftware: Btw, note for any other reviewer:
The PR description says this:

> Wear, Automotive, and TV prototype comments were showing the mobile app's QR code and build details.

But this confused me at first because when I looked at the Direct Download link in old PRs, it all looked correct to me… but this was a red herring that almost caught me in missing the point.

Before this fix, the **Direct Download** link shown in the Prototype PR comment was correct. It pointed to the right `.apk` files in `cdn.a8c-ci.services` / S3. See [this comment from the PR this one is stacked on top of, for example](https://github.com/Automattic/pocket-casts-android/pull/5682#issuecomment-5588367126)

Image

But what was indeed wrong though, was:
 - The `Build Number` and `Version` rows were showing the values of the app module
 - Those tables still showed a row for "Installation URL" (with the installation link to the mobile app in Firebase, so wrong link)—in addition to the "Direct Download" pointing to the right link.
 - The QR codes indeed were pointing to that "Installation URL" (Firebase URL), not to the "Direct Download" URL

Me initially comparing the "Direct Download" link before/after and seeing it has always pointed to the correct link in S3 made me question for a second if you hallucinated that bug about wrong link and metadata… but now I see the subtlety that this was about "Installation URL" being wrong, not "Direct Download"

AliSoftware (approved):
Checked that:
 - The QR codes in the Automotive+Wear+TV PR comment are now all pointing to the `.apk`s in S3 only and not to Firebase
 - There is no "Installation URL" row anymore (which was pointing to Firebase URL). Only "Direct Download" row with the right link to S3 CDN.
 - The `Build Number` and `Version` metadata are now correct, not the ones from the app.

:shipit:

PR #5905, opened 2026-09-10
[TV] Scroll the search field away with results and stop clipping the last row

sztomek:
## Description

On the TV Search screen, once you typed a query the search field and filter tabs stayed pinned while you scrolled down through the results, and the bottom row of each tab got cut off. With an empty query the header slid away fine, so the two cases were inconsistent.

Now the whole search header — the field, recent searches, and filter tabs — slides away as you move down into the results, so they fill the screen and nothing is clipped. Pressing up brings the filter tabs and the field back.

The header now collapses based on whether the results have focus, rather than on each list's scroll position (which never triggered for the podcast/folder grids or for short result sets).

Fixes POC-902 https://linear.app/a8c/issue/POC-902/search-screen-scroll-enhancements

## Testing Instructions

1. Open Search and type a query, e.g. "news".
2. Move down into the results — the field and filter tabs slide away and the last row is fully visible.
3. Press up from the first result — focus goes back to the filter tabs, then to the search field.
4. Press back — the list scrolls to the top and the header returns.

## Screenshots or Screencast
[search-focus.webm](https://github.com/user-attachments/assets/d954150e-51d5-47a9-9790-960353c8dcfb)

## Checklist

- [ ] with different themes

### DIFF
--- a/tv/src/main/java/au/com/shiftyjelly/pocketcasts/search/TvSearchScreen.kt
+++ b/tv/src/main/java/au/com/shiftyjelly/pocketcasts/search/TvSearchScreen.kt
@@ -40,6 +40,7 @@ import androidx.compose.runtime.snapshotFlow
 import androidx.compose.runtime.withFrameNanos
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.focus.FocusDirection
 import androidx.compose.ui.focus.FocusRequester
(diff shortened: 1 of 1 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

PR #5914, opened 2026-09-11
Migrate auto add to Up Next podcasts from RxJava to Flow

geekygecko:
## Description

No behaviour change: the Auto Add to Up Next settings screen now gets its data through Kotlin Flow instead of RxJava.

One step in the incremental RxJava removal. The screen was the only consumer of a Room query that was declared as a `Flowable`, exported as a `Flowable` on `PodcastManager`, and then combined in the ViewModel with two settings values that were *already* Flows but had to be pushed back through `asObservable().toFlowable(LATEST)` to join the Rx chain. Dropping the Rx end lets the whole pipeline stay in Flow and deletes that round-trip.

- `PodcastDao.findAutoAddToUpNextPodcastsRxFlowable()` becomes `findAutoAddToUpNextPodcastsFlow()`, same `@Transaction` and SQL.
- `PodcastManager.autoAddToUpNextPodcastsRxFlowable()` becomes `autoAddToUpNextPodcastsFlow()`. One caller, no test stubs, and `PodcastManagerImpl` is the only implementation.
- `AutoAddSettingsViewModel` swaps `combineLatest` for `combine` and `toLiveData()` for `asLiveData()`, and is now Rx-free. `autoAddPodcasts` stays a `LiveData`, so the XML fragment is untouched.

One difference worth a reviewer's eye: `asLiveData()` keeps collecting for its default 5s timeout after going inactive and publishes via `setValue` rather than `postValue`, so bursty emissions are no longer coalesced. The observer just does `submitList` behind a `DiffUtil` callback, so extra invocations are invisible.

## Testing Instructions

With a few podcasts subscribed, go to **Profile → Settings → Auto add to Up Next**.

1. Tap **Choose podcasts**, select two, go back. The list should update without leaving the screen, and the count on the **Choose podcasts** row should match.
2. Change a podcast's position, the **Auto add limit**, and **When auto add limit reached**. Each subtitle should update immediately.
3. Rotate the device. The list and all three subtitles should survive.
4. Deselect everything. The list should empty out.

## Checklist

### DIFF
--- a/modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/viewmodel/AutoAddSettingsViewModel.kt
+++ b/modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/viewmodel/AutoAddSettingsViewModel.kt
@@ -1,7 +1,7 @@
 package au.com.shiftyjelly.pocketcasts.settings.viewmodel
 
 import androidx.lifecycle.ViewModel
-import androidx.lifecycle.toLiveData
+import androidx.lifecycle.asLiveData
 import androidx.lifecycle.viewModelScope
 import au.com.shiftyjelly.pocketcasts.models.entity.Podcast
 import au.com.shiftyjelly.pocketcasts.preferences.Settings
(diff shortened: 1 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

PR #5917, opened 2026-09-12
Migrate the two Settings observables from RxJava to StateFlow

geekygecko:
## Description

Internal refactor with no user-visible change: the two app preferences that broadcast their updates to the UI (the multi-select action order, and the podcast sort order used when adding podcasts to a folder) now use Kotlin coroutines instead of RxJava.

These were the last two RxJava members of the `Settings` interface, and every consumer immediately bridged back out of Rx, so the round trip was pure overhead:

| Before | After |
| --- | --- |
| `selectPodcastSortTypeObservable: Observable<PodcastsSortType>` | `selectPodcastSortTypeFlow: StateFlow<PodcastsSortType>` |
| `multiSelectItemsObservable: Observable<List<String>>` | `multiSelectItemsFlow: StateFlow<List<String>>` |

`BehaviorRelay` becomes `MutableStateFlow` in `SettingsImpl`, and the three consumers drop their `toLiveData()` / `asFlow()` bridges. **`modules/services/preferences` and `modules/services/views` are now Rx-free modules.** Both still declare Rx in their build files, because those `api(...)` entries leak `io.reactivex` transitively to feature modules — removing them is separate work.

One deliberate difference: `StateFlow` deduplicates by equality where `BehaviorRelay` did not, so persisting an unchanged value no longer re-emits. Both consumers are idempotent, and `MultiSelectFragment` already updates its adapter locally before persisting, so nothing depended on that re-emission.

## Testing Instructions

A pure refactor, so there is nothing new to look at. What could plausibly have broken is the multi-select toolbar and the folder sort order still reacting when their preference changes.

1. Long-press an episode to enter multi-select. The toolbar shows the expected actions, and they enable/disable as you change the selection.
2. From the multi-select overflow, choose "Rearrange actions" and drag actions between the two sections. The new order applies to the toolbar immediately and survives an app restart. Dropping an action back where it started changes nothing.
3. Podcasts tab → edit a folder → "Add podcasts". Changing the sort order re-sorts the list immediately, keeps your current selection, and survives rotation.
4. If you can, upgrade over an older install rather than installing fresh, and check the multi-select toolbar still shows the right actions on first launch (a background migration rewrites the stored action ids).

## Checklist

### DIFF
--- a/app/src/test/java/au/com/shiftyjelly/pocketcasts/podcasts/view/folders/FolderEditViewModelTest.kt
+++ b/app/src/test/java/au/com/shiftyjelly/pocketcasts/podcasts/view/folders/FolderEditViewModelTest.kt
@@ -14,7 +14,6 @@ import com.automattic.eventhorizon.EventHorizon
 import com.automattic.eventhorizon.FolderCreateColorShownEvent
 import com.automattic.eventhorizon.FolderSavedEvent
 import io.reactivex.Flowable
-import io.reactivex.Observable
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.flowOf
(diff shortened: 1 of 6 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-android/milestone/135">8.21</a>. This milestone is due in less than 2 days.<br />
    </tr>
  </tbody>
</table>

</p>

PR #5919, opened 2026-09-12
Migrate the downloaded episodes stream from RxJava to Flow

geekygecko:
## Description

The list of downloaded episodes is now delivered to the app as a Kotlin Flow rather than an RxJava stream. Nothing about the app's behaviour changes: the Downloads screen in Android Auto, the storage and clean-up settings screens, the storage-move routine and the auto-play picker all read the same data as before.

Part of the RxJava removal effort. The backlog entry is `EpisodeManager` / `EpisodeDao` `findDownloadedEpisodesRxFlowable`, which is one of the remaining Rx signatures on the `EpisodeManager` interface.

`EpisodeDao.findDownloadedEpisodesRxFlowable` becomes `findDownloadedEpisodesFlow` and returns `Flow` instead of `Flowable`. Room supports both natively, so the query itself is untouched. That propagates through `EpisodeManager` and `EpisodeManagerImpl` (the only implementor) and out to all seven call sites. `AutoPlaySelector`, `StorageSettingsViewModel` and `ManualCleanupViewModel` end up with no RxJava references left at all.

The most substantial call site is `ManualCleanupViewModel`, where the rxkotlin `Flowable.combineLatest` becomes `combine`. That forced its "include starred" `BehaviorRelay` to change too, and it became a `MutableStateFlow` rather than gaining a new `asFlow()` bridge.

## Testing Instructions

The risk is that one of these screens silently stops updating, so each step is worth watching rather than just loading once.

1. Download a few episodes, including at least one starred episode and a mix of unplayed, in-progress and played ones.
2. Go to Profile → Settings → Storage. Confirm the "Downloaded files" size is populated and is not zero.
3. While that screen is open, delete a download from elsewhere (or download another episode). Confirm the size updates live without leaving the screen.
4. Go to Profile → Settings → Storage → Manage Downloads. Confirm the Unplayed / In progress / Played rows show the right episode counts and sizes.
5. Toggle "Include starred". Confirm the starred episode appears in the rows when it is on and disappears when it is off, and that the total size and the enabled state of the clean-up button follow.
6. Tick one or more categories and run the clean-up. Confirm the right episodes are removed and the rows refresh.
7. Rotate the Manage Downloads and Storage screens and confirm the numbers survive.
8. Connect to Android Auto (or the Desktop Head Unit) and open Pocket Casts → Downloads. Confirm the downloaded episodes are listed and playable. This is the call that stopped blocking a thread, so also back out of the node quickly while it loads and confirm nothing hangs or crashes.
9. With auto-play enabled and its source set to Downloads, let an episode finish and confirm the next downloaded episode plays.
10. In Profile → Settings → Storage, change the storage location to move files to another volume. Confirm downloaded episodes are moved and still play afterwards.
11. Check the home screen / profile downloaded-size figure still reflects reality after downloading and deleting.

## Checklist

### DIFF
--- a/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivityViewModel.kt
+++ b/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivityViewModel.kt
@@ -39,7 +39,6 @@ import kotlinx.coroutines.flow.stateIn
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.reactive.asFlow
-import kotlinx.coroutines.reactive.collect
 import timber.log.Timber
 import au.com.shiftyjelly.pocketcasts.localization.R as LR
 
@@ -92,7 +91,7 @@ class MainActivityViewModel
         }
 
         viewModelScope.launch {
-            episodeManager.findDownloadedEpisodesRxFlowable()
+            episodeManager.findDownloadedEpisodesFlow()
                 .collect { result ->
                     _downloadedEpisodeState.update { state -> state.copy(downloadedEpisodes = result.sumOf { it.sizeInBytes }) }
                 }
--- a/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivityViewModel.kt
+++ b/app/src/main/java/au/com/shiftyjelly/pocketcasts/ui/MainActivityViewModel.kt
@@ -59,9 +59,6 @@ class MainActivityViewModel
     private val _state = MutableStateFlow(State())
     val state = _state.asStateFlow()
 
-    private val _downloadedEpisodeState = MutableStateFlow(DownloadedEpisodesState())
-    val downloadedEpisodeState = _downloadedEpisodeState.asStateFlow()
-
     private val _snackbarMessage = MutableSharedFlow<Int>()
     val snackbarMessage = _snackbarMessage.asSharedFlow()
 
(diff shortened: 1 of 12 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

geekygecko:
Thanks — agreed on the analysis, and this matches what both pre-PR reviews found independently. Declining the fix here, deliberately.

`ManualCleanupPage.kt:49` is a pre-existing UI state-duplication bug: `includeStarredSwitchState` is a second copy of `ManualCleanupViewModel.switchState`, and after a config change the ViewModel keeps filtering with starred included while the switch renders off. That is true on `main` today and this PR does not change the visible outcome, which you and I both traced to the same place: the `onStarredSwitchClicked(true)` that `MutableStateFlow` now dedupes would have recomputed a byte-identical `State` under `BehaviorRelay`, because the transform is a pure function of the episode list, the switch value and the current checked flags.

This PR is one entry off the RxJava removal backlog and the whole workflow depends on each of those staying reviewable on its own. Hoisting the switch state into `ManualCleanupViewModel.State` and deleting the local `remember` is a real user-visible fix to a real rotation bug, which is exactly why it deserves its own PR and its own testing rather than riding along in a refactor whose stated bar is zero behaviour change.

Noted in the backlog alongside the other two things this PR surfaced and left alone: `EpisodeManager.findDownloadingEpisodesRxFlowable` being entirely dead (3 removable Rx signatures across 3 files), and `StorageSettingsViewModel.start()` stacking an uncancelled collector per view recreation.

PR #5925, opened 2026-09-12
Migrate the cloud files sorted query from RxJava to Flow

geekygecko:
## Description
Internal refactor with no user-visible change: the Files list keeps working exactly as it does today, it just gets its data through Kotlin Flow instead of RxJava.

Part of the RxJava removal effort — the `UserEpisodeDao` / cloud files sorted query entry from the backlog.

The six sorted cloud-files queries in `UserEpisodeDao` returned `Flowable`, and both of the things that consumed them immediately converted the result with `.asFlow()`. Returning a native Room `Flow` deletes those two bridges rather than adding any. `UserEpisodeManager.userEpisodesSortedRxFlowable` becomes `userEpisodesSortedFlow`; its only two callers are `CloudFilesManager` and wear's `FilesViewModel`, both already Flow consumers, so `sortedCloudFiles` keeps its existing type and nothing downstream changes.

`removeCloudStatusFromFiles` also moves to the pre-existing suspend twin `findUserEpisodesDesc()`. It was the last caller keeping the Rx `Desc` Flowable alive, and it was only ever taking a one-shot snapshot (`.firstElement().awaitSingleOrNull()`) anyway.

Semantics checked by hand:
- Identical SQL on all six queries, so the sort-order mapping is unchanged.
- Room `Flow` and Room's Rx `Flowable` adapter are both backed by the same invalidation flow, so initial emission and re-query on table change are preserved.
- The `filterNot(MISSING)` filter is unchanged; `.map` is now `kotlinx.coroutines.flow.map`.
- `.firstElement().awaitSingleOrNull()?` → non-null `List`: Room never emits null for a list query, and the call stays inside `withContext(Dispatchers.IO)`.
- **One deliberate difference worth a reviewer's eye:** that `filterNot` used to run on Room's query executor and now runs in the collector's context (main thread for both consumers). The list is the user's own uploaded files, so this is a trivially small amount of work; I left it rather than add a `flowOn` buffer hop the Rx version never had. Happy to add `.flowOn(Dispatchers.Default)` if you'd rather.

Rx bridges deliberately left in place: `findDownloadingUserEpisodesRxFlowable`, `findEpisodeRxFlowable`, `findEpisodeByUuidRxMaybe`, `insertRxCompletable`, `updateServerStatusRxCompletable` and `updateUploadErrorRxCompetable` all stay. They are blocked on public `UserEpisodeManager` signatures consumed from `PlaybackManager`, `UpNextSync` and `UploadEpisodeTask`, which are out of scope here.

## Testing Instructions
1. Sign in to an account with Pocket Casts Plus and open **Profile → Files**.
2. Confirm the uploaded files list appears and is correct.
3. Change the sort order (Newest to oldest, Oldest to newest, A–Z, Z–A, Shortest to longest, Longest to shortest) and confirm the list reorders correctly for every option.
4. Upload a new file and confirm it appears in the list without needing to leave and re-enter the screen.
5. Delete a file and confirm it disappears from the list immediately.
6. Confirm files with a "missing" server status are still hidden from the list.
7. On a Wear OS device or emulator, open the **Files** screen and confirm the list loads and is sorted per the phone's sort setting.
8. Go to **Profile → Settings → Files** (or cancel your subscription flow) and trigger "remove cloud status from files"; confirm cloud-only files are deleted and downloaded files revert to local.
9. With Up Next empty and auto-play configured to continue from Files, confirm playback continues into a cloud file.

## Checklist

### DIFF
--- a/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/UserEpisodeDao.kt
+++ b/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/UserEpisodeDao.kt
@@ -46,25 +46,25 @@ abstract class UserEpisodeDao {
     abstract suspend fun findAllUuids(): List<String>
 
     @Query("SELECT * FROM user_episodes ORDER BY added_date DESC")
-    abstract fun findUserEpisodesDescRxFlowable(): Flowable<List<UserEpisode>>
+    abstract fun findUserEpisodesDescFlow(): Flow<List<UserEpisode>>
 
     @Query("SELECT * FROM user_episodes ORDER BY added_date DESC")
     abstract suspend fun findUserEpisodesDesc(): List<UserEpisode>
 
     @Query("SELECT * FROM user_episodes ORDER BY added_date ASC")
-    abstract fun findUserEpisodesAscRxFlowable(): Flowable<List<UserEpisode>>
+    abstract fun findUserEpisodesAscFlow(): Flow<List<UserEpisode>>
 
     @Query("SELECT * FROM user_episodes ORDER BY title ASC")
-    abstract fun findUserEpisodesTitleAscRxFlowable(): Flowable<List<UserEpisode>>
+    abstract fun findUserEpisodesTitleAscFlow(): Flow<List<UserEpisode>>
 
     @Query("SELECT * FROM user_episodes ORDER BY title DESC")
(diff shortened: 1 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

PR #5920, opened 2026-09-12
Remove the dead findDownloadingEpisodesRxFlowable chain

geekygecko:
This removes three unused RxJava methods that nothing in the app called. There is no behaviour change.

Part of the RxJava removal effort (backlog entries `EpisodeDao.kt`, `EpisodeManager.kt` and `UserEpisodeManager.kt`). `EpisodeManager.findDownloadingEpisodesRxFlowable` had zero consumers, which made `EpisodeDao.findDownloadingEpisodesRxFlowable` and `UserEpisodeManager.downloadUserEpisodesRxFlowable` dead as well, so all three go in one pure-deletion diff: 15 lines removed, nothing added.

- No Rx semantics needed preserving: nothing subscribed to these streams, so the `mergeWith` in the deleted override had no observer and no Room invalidation was registered.
- No bridges added or removed, and no imports orphaned: `io.reactivex.Flowable` is still used by other signatures in all four files.
- The in-app downloads list already uses the Flow path (`findDownloadingEpisodesIncludingFailedFlow`), which is untouched.

Two follow-ups found during review and deliberately left out to keep this diff purely the dead-code removal:
- `UserEpisodeDao.findDownloadingUserEpisodesRxFlowable` (`UserEpisodeDao.kt:70`) becomes dead with this change and can be deleted next.
- `UserEpisodeManager.kt:38` imports `io.reactivex.Single` without using it, unchanged from `main`.

Note for whoever merges second: draft PR #5919 touches `EpisodeDao.kt`, `EpisodeManager.kt` and `EpisodeManagerImpl.kt` in different hunks, so a trivial conflict is possible.

## Testing Instructions

Nothing should change, so this is a regression check that downloads still work.

1. Open the app, go to a podcast and download two or three episodes.
2. Open Profile > Downloads and confirm the episodes appear and their progress bars advance while downloading.
3. Go to Profile > Files, upload a file, and download it. Confirm it also shows progress and completes.
4. Pause and resume a download, and cancel one, confirming the list updates each time.
5. Background the app during a download and return to it: the list should still be accurate.

## Checklist
- [x] If this is a user-facing change, I have added an entry in CHANGELOG.md
- [x] Ensure the linter passes (`./gradlew spotlessApply` to automatically apply formatting/linting)
- [x] I have considered whether it makes sense to add tests for my changes
- [x] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml`
- [x] Any jetpack compose components I added or changed are covered by compose previews
[...]

PR #5934, opened 2026-09-13
Remove the dead unsubscribed-podcast Rx DAO queries

geekygecko:
## Description
This removes three database queries that nothing in the app uses any more. There is no behaviour change.

`EpisodeDao.findEpisodesRxFlowable`, `EpisodeDao.countRxFlowable` and `PodcastDao.findUnsubscribedUuidRxFlowable` only fed the two `EpisodeManager` Rx pipelines deleted in #5652, and nothing has called them since. Deleting DAO query methods has no schema or migration impact.

Worth checking when merging: #5920 deletes the only other `Flowable` in `EpisodeDao`, so whichever of the two lands second needs to drop the now-unused `io.reactivex.Flowable` import or ktlint will fail.

## Testing Instructions
Nothing calls the deleted code, so there is no path through the app to exercise. A green CI build (all app modules compile, unit tests pass) is the check.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

PR #5937, opened 2026-09-14
Replace the Rx emitter in ProgressRequestBody with a progress callback

geekygecko:
## Description
The code that reports upload progress for user files no longer depends on RxJava, with no user-visible change.

`ProgressRequestBody.create` took a `FlowableEmitter<Float>` only to call `onNext` on it. It now takes a plain `(Float) -> Unit` callback, and `SyncServiceManager.uploadToServer` passes `emitter::onNext`, so its public `Flowable<Float>` stays the same. A small unit test covers the progress callback.

## Testing Instructions
1. Signed in with a Plus account, open Profile → Files and add an audio file from the device.
2. Upload it to the cloud. The upload progress advances to 100% and the file shows as uploaded.

## Screenshots or Screencast

https://github.com/user-attachments/assets/46742ec4-d6c6-419b-b539-19843f6b15d2

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

geekygecko:
Re the nit on `ProgressRequestBody.kt:62` (folding `Listener` into a `(Float) -> Unit` constructor parameter): I checked, and `Listener` and the `Listener`-taking constructor are only used inside this file, so the cleanup would be safe. I'm leaving it out on purpose. This PR only removes the Rx type, and reshaping the class's public constructor is a separate tidy-up that would make a zero-behaviour-change diff harder to check. Leaving as-is.

PR #5955, opened 2026-09-15
Remove EpisodeDao.findByUuidRxMaybe in favour of its suspend twin

geekygecko:
## Description
Looking up a single episode by id now goes through a Kotlin coroutine query instead of an RxJava database query, with no user-visible change.

`EpisodeDao.findByUuidRxMaybe()` had one caller, `EpisodeManagerImpl.findByUuidRxMaybe()`. That method keeps its public `Maybe` but now wraps the existing suspend `findByUuid()`, which runs the same query, in `rxMaybe(ioDispatcher) { }`, the same bridge `downloadMissingEpisodeRxMaybe` uses. The Room Rx query is deleted.

Worth checking: the result is now delivered on the IO dispatcher rather than Room's query executor. Both are background threads, and every caller either switches to main itself or does not care which thread it lands on.

## Testing Instructions
1. Open an episode from a podcast page. The episode screen loads its details and show notes.
2. Open a shared episode link for a podcast you don't follow. The episode screen loads.
3. In Discover, play a featured episode. It starts playing.
4. Play an episode and an uploaded file in turn, and open the full player and Up Next each time. Title, artwork and playback controls show the right episode, including in the media notification.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

PR #5978, opened 2026-09-16
Remove the Advertising ID permission from all apps

geekygecko:
## Description

Stops all four Pocket Casts apps from asking Android for permission to read the device's advertising ID, so none of them can be handed one.

We do not use the advertising ID anywhere: there is no third-party ad SDK in the project, and nothing in our code reads it. The permission only arrives transitively, from the Play Services measurement libraries that Firebase Analytics depends on. The mobile app has excluded it since 2022 (1dcfb3b1fb), and Automotive excluded the whole set, but the TV app never did, so shipped TV builds were declaring `com.google.android.gms.permission.AD_ID` even though we answer "No" to the advertising ID question in the Play Console. That mismatch between the declaration and the manifest is what this fixes.

Each app manifest now marks the relevant permissions with `tools:node="remove"` so the merger drops the libraries' declarations:

- `tv` gained removals for `com.google.android.gms.permission.AD_ID`, `ACCESS_ADSERVICES_AD_ID` and `ACCESS_ADSERVICES_ATTRIBUTION`. It previously had none.
- `app` and `wear` already removed the GMS permission and gained the two AdServices ones, matching what `automotive` was already doing.
- `automotive` is unchanged, it already removed all three.

The two `ACCESS_ADSERVICES_*` permissions are the Privacy Sandbox equivalents rather than the permission the Play Console asks about, but they are removed here too so the four manifests are consistent and nothing is left declaring an ad identifier we do not read.

No behaviour changes. Firebase Analytics identifies installs by its own app-instance ID, not the advertising ID, so active user counts and event reporting are unaffected. What the advertising ID would enable, and which we do not use, is Google Ads attribution, remarketing audience export, and demographics and interests reporting.

## Testing Instructions

This is a manifest-only change, so the verification is on the merged manifests rather than in the running app.

1. Check out this branch and run:
   `./gradlew :app:processReleaseMainManifest :automotive:processReleaseMainManifest :tv:processReleaseMainManifest :wear:processReleaseMainManifest`
2. For each of `app`, `automotive`, `tv` and `wear`, grep the merged output:
   `grep -n "ADSERVICES\|permission.AD_ID" <module>/build/intermediates/merged_manifest/release/processReleaseMainManifest/AndroidManifest.xml`
3. Confirm every one of the four returns no matches.
4. For contrast, repeat steps 1 to 3 on `main` and confirm `tv` reports all three permissions, while `app` and `wear` each report the two `ACCESS_ADSERVICES_*` ones.
5. Install a build on a device or emulator and confirm the app starts, plays an episode, and that Firebase events still arrive as before.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

app/src/main/AndroidManifest.xml
```diff
@@ -27,6 +27,8 @@
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE_MEDIA_PLAYBACK" />
 
     <uses-permission android:name="com.google.android.gms.permission.AD_ID" tools:node="remove"/>
+    <uses-permission android:name="android.permission.ACCESS_ADSERVICES_AD_ID" tools:node="remove"/>
+    <uses-permission android:name="android.permission.ACCESS_ADSERVICES_ATTRIBUTION" tools:node="remove"/>
```
geekygecko, line 31: I have included the new versions of this permission that the Automotive app was already using.

> android.permission.ACCESS_ADSERVICES_AD_ID is the new framework version designed to coexist with and eventually succeed the classic com.google.android.gms.permission.AD_ID permission. 
> This transition is part of Google’s long-term Privacy Sandbox on Android initiative. While the old permission is managed by Google Play Services, the new permission utilizes the native Android system framework to provide access to the user's advertising ID in a more secure and privacy-focused manner.

sztomek (approved):

PR #5981, opened 2026-09-16
Fix full screen player transcript button not working if playback hasn't started

geekygecko:
## Description

The transcript button on the full screen player now becomes available as soon as an episode is loaded, instead of staying greyed out until you start playing the episode or open its episode details page.

Whether a transcript exists is read from a local database table, and that table is only written when an episode's show notes are fetched and processed. Until now nothing fetched show notes just because an episode became the current one: it happened when the playback engine loaded an episode to play, when the show notes tab or the episode details page was opened, or when the episode was downloaded. So on a fresh start the table was empty for the current episode, the availability flag was genuinely `false`, and the button stayed disabled until one of those other things happened to run.

`PlaybackManager.setup()` now also collects the Up Next current episode and preloads its show notes, next to the two collectors already watching that flow for prefetching and HLS availability. It de-duplicates on episode UUID so unrelated changes (starred, download status) don't refetch, and `loadShowNotes` returns a state rather than throwing, so a network failure can't kill the collector.

This adds one show notes request per current-episode change, including one at startup where there previously was none. It goes through the same cached endpoint that playback already uses, so repeats are cheap.

Fixes https://linear.app/a8c/issue/PCDROID-601/transcript-button-remains-disabled-on-now-playing-screen-until-episode

## Testing Instructions

1. Load an episode with a transcript into your player.
2. Sync to your account.
3. Reinstall from refresh and login
4. Once the initial sync is complete, open the full screen player
5. Tap the more button in the shelf
6. ✅ Verify the transcript button is enabled (before it would be greyed out)

## Screenshots or Screencast

https://github.com/user-attachments/assets/ba5b5956-5639-4440-8de3-d626765c771c

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

geekygecko:
Fixed in another PR https://github.com/Automattic/pocket-casts-android/pull/6034

PR #5956, opened 2026-09-15
Make account deletion a suspend call instead of an Rx Single

geekygecko:
## Description
Deleting your account from the profile screen now runs through Kotlin coroutines instead of RxJava, with no user-visible change.

## Testing Instructions
Account deletion is permanent, so use a throwaway account you create for this.

1. Sign in with the throwaway account and turn on airplane mode. Go to Profile, tap your account, choose Delete account and confirm both dialogs. The "delete account failed" alert appears and you stay signed in.
2. Turn airplane mode off and delete the account again. You are signed out, and signing back in with that account fails.

## Screencast

https://github.com/user-attachments/assets/88f084fd-f9f5-4b7b-bf99-b2bbac02d95f

## Checklist
- [ ] If this is a user-facing change, I have added an entry in CHANGELOG.md
- [ ] Ensure the linter passes (`./gradlew spotlessApply` to automatically apply formatting/linting)
- [ ] I have considered whether it makes sense to add tests for my changes
- [ ] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml`
- [ ] Any jetpack compose components I added or changed are covered by compose previews
- [ ] I have updated (or requested that someone edit) [the Event Horizon schema](https://github.com/Automattic/EventHorizonSchemas/blob/trunk/schema/pocket-casts.yml) to reflect any new or changed analytics.

PR #6003, opened 2026-09-17
Move the file usage endpoint off RxJava

geekygecko:
## Description
The cloud storage usage figure shown after a file is removed from the cloud is now fetched with Kotlin coroutines instead of RxJava, with no user-visible change.

The `/files/usage` endpoint was the last RxJava `Single` in the file delete path: the Retrofit endpoint, the `SyncServiceManager` wrapper and `SyncManager.getFileUsageRxSingle()` are now `suspend`, and the only caller drops its `.await()` bridge. This follows the same shape as #5956, #5967 and #5969.

Worth checking: the suspend token helper differs from the Rx one in a single, barely reachable case. Signed in with no access token only happens if the account is removed mid-call, and there the call still fails, now with `Failed to get access token` instead of `Failed to get token` after one no-op token invalidation. Every other endpoint on the suspend helper already behaves this way.

## Testing Instructions
1. Sign in with a Plus account and open Profile → Files. The storage usage bar shows your current usage.
2. Upload a file, then remove it from the cloud. The file returns to local-only and the usage figure drops without leaving the screen.
3. Turn off the network and remove another file from the cloud. The failure is logged and the app does not crash; the usage figure is left alone.

## Screencast

https://github.com/user-attachments/assets/30d3ede6-c678-499b-89e4-e7d35d672393

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

geekygecko:
Good catch, and you are right — I checked `SyncAccountManagerImpl.getAccessToken()`: it returns `null` only when `getAccount()` is null, and `isLoggedIn()` is that same check, so the logged-in-with-no-token state is only reachable if the account is removed between the two calls. In that race the call still fails either way; the only real difference is the exception message (`Failed to get access token` instead of `Failed to get token`) and one no-op `invalidateAccessToken()`. I have reworded the description accordingly.

I have kept the test, since it pins the `getCacheTokenOrLogin` contract the same way the existing `signed playback url is retried with a refreshed token when unauthorized` case does, rather than claiming a production recovery path.

PR #6004, opened 2026-09-17
Read Discover subscription state from podcastSubscriptionsFlow

geekygecko:
## Description
Discover rows now read the followed-podcast list through Kotlin Flow instead of RxJava, with no user-visible change.

`DiscoverViewModel.addSubscriptionStateToPodcasts` merged a one-shot `Single` fetch of the subscribed uuids with a `Flowable` of later changes. `podcastSubscriptionsFlow()` already does both halves, so the merge collapses onto that Flow, bridged back with `asFlowable()` because the surrounding chain is still Rx. That left `getSubscribedPodcastUuidsRxSingle()` and `podcastSubscriptionsRxFlowable()` without callers, so both are deleted from `PodcastManager` (public Rx signatures 6 → 4).

## Testing Instructions
1. Open Discover. Every row shows the correct followed state for podcasts you already follow.
2. Follow and unfollow a podcast from a Discover row. The button updates immediately, and other rows containing that podcast follow suit.
3. Discover → See all on a list, follow/unfollow there, then go back. State is consistent in both directions.
4. Follow a podcast from its podcast page, then return to Discover. The row reflects it without an app restart.
5. Open a shared podcast list and the podcast page recommendations (other consumers of the same Flow) and confirm followed state is still correct.

## Screencast

https://github.com/user-attachments/assets/9e32b624-1540-4277-be86-bbedd4ddf548

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

PR #6012, opened 2026-09-18
Improve search result chips

geekygecko:
## Description
The filter chips at the top of the search results (such as the network chip) now scroll edge to edge without a fade, so chips further along the row are easier to spot.

Previously the chip row faded out on the right and was inset by 16dp on both sides, which made the later chips (the network chip in particular) look cut off and easy to miss. This change matches the iOS app.

The row now uses a plain `LazyRow` instead of `FadedLazyRow`, and the horizontal 16dp padding moved from the modifier to `contentPadding`. The first chip still lines up with the screen content, but the chips can now scroll all the way to the screen edges.

## Testing Instructions
1. Open the Search tab and search for something that returns results for all filter types (e.g. "news").
2. Check the filter chip row above the results: there is no fade on the right edge and the first chip is aligned with the content below it.
3. Scroll the chip row horizontally and confirm the chips scroll to the screen edges and the last chip (network) is fully visible when scrolled to the end.
4. Tap each chip and confirm the results filter as before.

## Screenshot and Screencast 

**Before**

Screenshot_20260918_123033

**After**

https://github.com/user-attachments/assets/1a757c5a-fdc0-400b-8039-f8463cfee36e

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

PR #6017, opened 2026-09-18
Keep the app open when changing the app icon

sztomek:
## Description

Changing the app icon closed Pocket Casts and dumped the user on the home screen, without them ever getting to read the "Icon updated" confirmation.

`SettingsAppearanceViewModel.updateGlobalIcon` flipped the launcher alias immediately and the fragment then showed the dialog. Because the running task had been launched through the alias that was being disabled, the system tore the task down: the dialog was created, leaked its window, and disappeared with the app.

```
E WindowManager: android.view.WindowLeaked: Activity …ui.MainActivity has leaked window
    …[MainActivity_1] that was originally added here
    at androidx.appcompat.app.AlertDialog$Builder.show(AlertDialog.java:1007)
    at …AppearanceSettingsFragment.onViewCreated$lambda$2$1(AppearanceSettingsFragment.kt:108)
```

This is pre-existing — it reproduces on `main` before #5609 for anyone already using a custom icon. #5609 removed the one case that was exempt (users on the default icon had a task rooted at `MainActivity`, which is never disabled), so it now applies to every icon change.

Selecting an icon now only stores the choice. The alias is applied when the app goes to background, so the icon is already updated by the time the user looks at their launcher.

The eager `enableSelectedAlias` in `setupApp()` had to go with it. It was harmless before, because the alias was always applied the moment the user picked it, so the startup call never had anything to do. With the change deferred, a selection that never reaches `onStop` (crash, force stop) leaves a pending change that the *next* cold start would apply from `Application.onCreate` — disabling the very alias the launch had just come through and killing the task on sight. Verified on device: the app showed its splash and dropped straight back to the launcher. Alias updates now only happen when no foreground task depends on them: on background, and from `AppIconUpdateReceiver` on `MY_PACKAGE_REPLACED`, which is the path #5609 relies on for migrating existing installs.

`enableSelectedAlias` also skips work when the requested icon is already the enabled one. Without that, every backgrounding would re-issue the alias update — one batched call on API 33+, but 19 individual `setComponentEnabledSetting` binder calls on the main thread below it, and `minSdk` is 24. It also removes the duplicate batch the update receiver used to issue right after startup.

## Testing Instructions

1. Open **Profile → Settings → Appearance**.
2. Pick a different **App Icon**.
3. Confirm the **Icon updated** dialog appears and stays, and that the app does **not** close.
4. Tap **OK**, then leave the app with the Home button.
5. Confirm the launcher now shows the icon you picked, and that there is exactly one Pocket Casts entry in the all-apps screen.
6. Repeat with the **Default** icon to check it round-trips.

Ground truth for the launcher entries, since launchers cache aggressively:

```
adb shell pm query-activities -a android.intent.action.MAIN \
  -c android.intent.category.LAUNCHER --brief | grep pocketcasts
```

Verified on a Pixel 6 (Android 16) with the `debug` build. Before the change, selecting an icon bounced straight to the home screen and logged `WindowLeaked`. After it, the dialog stays, `logcat` has no `WindowLeaked`, the alias stays put while the app is foregrounded, and flips on Home — checked across Round Light → Round Dark → Indigo → Rosé, each time leaving exactly one enabled launcher alias.

Also covered: pick an icon, force stop the app without backgrounding it, then launch from the launcher. Before this last change the app closed immediately; now it opens normally and the pending icon applies the next time you leave it.

Two things worth knowing, both inherent to the activity-alias approach rather than introduced here:

- Disabling the previously enabled alias still ends the task, so returning from recents after an icon change starts Pocket Casts fresh instead of resuming. It now happens after the user has left the app rather than in front of them.
- `ProcessLifecycleOwner` treats a screen lock as leaving the app, so locking the phone shortly after changing the icon will apply it and end the task then. Because the alias update is skipped when the icon has not actually changed, this only applies in the window right after a change.

## Checklist
- [x] All strings that need to be localized are in `modules/services/localization/src/main/res/values/strings.xml` — no string changes
- [ ] Any jetpack compose components I added or changed are covered by compose previews — n/a, no Compose changes
- [ ] I have updated (or requested that someone edit) [the Event Horizon schema](https://github.com/Automattic/EventHorizonSchemas/blob/trunk/schema/pocket-casts.yml) to reflect any new or changed analytics. — n/a, `SettingsAppearanceAppIconChangedEvent` still fires on selection

n/a — no layout or screen changes; the only visible difference is that an existing dialog now stays on screen.
- [ ] with different themes


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

geekygecko (approved):
Nice change. It worked well in my testing on both a Google Pixel 9 and Samsung Galaxy S25.

PR #6019, opened 2026-09-18
[TV] Update Connect to TV screen on mobile

sztomek:
## Description

When you scan the QR code shown by the Pocket Casts TV app, the phone opens the device approval sheet. It was device-agnostic — an Android robot next to the Pocket Casts roundel, and copy that said "Connect your device". This names the TV instead, and reworks the sheet to the design in Ftk3KwnfqaK4g57yCN63p0-fi-2785_14076.

What changed:

*   The right-hand badge is now the Google TV mark (new `ic_google_tv` vector), and the badge pair matches the spec — 80dp circles, the darker blue gradient, and the connector dots in `primaryUi05`.
*   Copy names Google TV: "Connect to Google TV", "Google TV connected", "Your Google TV is now connected to your account.", and the expired message points back at "your TV".
*   The sheet is now full height, with the badges/title/account block centred and **Connect** pinned to the bottom, instead of a wrap-content sheet with the button inline.
*   New "Not you? Switch account" link under the account card. It asks for confirmation, then signs out and clears local data, and opens the login flow once the sign-out has actually finished. Clearing data matters here: without it the previous account's podcasts, filters and history stay on the device and sync up into whichever account signs in next.
*   Removed `ic_android_robot`, which had no remaining references after the badge swap.

Figma: Ftk3KwnfqaK4g57yCN63p0-fi-2783_9020
Fixes: POC-889 https://linear.app/a8c/issue/POC-889/change-connect-to-apple-tv

## Testing Instructions

1.  Sign in on the phone.
2.  Fire the pairing deep link (a raw QR scan is not auto-verified for the `.debug` package):
    `adb shell am start -n au.com.shiftyjelly.pocketcasts.debug/au.com.shiftyjelly.pocketcasts.ui.MainActivity -a android.intent.action.VIEW -d "https://pocketcasts.com/pair?user_code=ABCD12"`
3.  Check the sheet against the Figma: Google TV badge, "Connect to Google TV", your account card, the pairing code, the switch-account link, and Connect at the bottom.
4.  Tap **Not you? Switch account** — confirm the warning appears, and that accepting signs you out and lands you in the login flow.
5.  Sign back in, reopen the link and tap **Connect**. The TV should pair and the sheet should read "Google TV connected".
6.  With a stale or already-used code, confirm the expired copy reads "Please go back to your TV to generate a new one."

## Screenshots or Screencast
Screenshot_20260918_211200

## Checklist
 
- [ ] with different themes


modules/services/localization/src/main/res/values/strings.xml
```diff
@@ -2985,15 +2985,19 @@
     <string name="tv_onboarding_welcome_back">Good to see you</string>
     <string name="tv_onboarding_syncing_subtitle">Loading your podcasts now.</string>
     <string name="tv_create_account_title">Create your free account</string>
-    <string name="device_approve_title">Connect your device</string>
+    <string name="device_approve_title">Connect to Google TV</string>
```
geekygecko, line 2988: Isn't this used QR code approval page shared with Wear OS now as well?

sztomek: Good catch, and you're exactly right — `deviceAuthFlow` has two callers now, `TvDeviceAuth.kt` and Wear's `LoginWithCodeViewModel`, and both QR codes land on this same sheet via `PairDeviceDeepLink`. Naming Google TV here would have been wrong for anyone pairing a watch.

I went looking for a way to tell the two apart and there isn't one today:

* `DeviceApproveRequest` is just `user_code` + `deny`, and `/device/approve` returns an empty body (iOS's `Api_DeviceApproveResponse` has no fields at all).
* The deep link only ever carries `user_code`.
* I checked against staging to be sure — `POST /device/authorize` with `{"scope":"tv"}` and `{"scope":"watch"}` return an identical `verification_uri_complete` of `.../pair?user_code=XXXXXX`. No device hint either way.

So I've reverted the copy and the icon back to the generic versions and dropped the now-unused `ic_google_tv`. The rest of the PR (the layout rework and the switch-account link) is unaffected.

The server does know which it is — it validates the scope per device type and rejects `mobile` — so the fix is a small backend one: append the scope to `verification_uri_complete` (`.../pair?user_code=WR6N3U&scope=tv`). That's one field, no new endpoint, and it covers every approver (us, iOS, web player) and every requester (our TV, Wear, and Apple TV) at once, with absent/unknown falling back to the generic copy so TV builds already in the wild keep working. It has to be on that URL rather than in the approve response, since the copy is on screen before anyone taps Connect. I'll raise it with the backend folks and bring the Google TV branding back once we can actually tell the devices apart.

geekygecko (approved):

